PR #1, opened 2015-04-19
GROOVY-7159: STC claims non-existent Diamond in Anonymous Inner Class wi...

shils:
...th Generics. See https://issues.apache.org/jira/browse/GROOVY-7159


melix:
Merged, thanks!

PR #3, opened 2015-04-20
Bug Fix for GROOVY-7259

antoaravinth:
Bug fix for the issue:

https://jira.codehaus.org/browse/GROOVY-7259?jql=project%20%3D%20GROOVY%20AND%20resolution%20%3D%20Unresolved%20AND%20priority%20%3D%20Trivial%20ORDER%20BY%20key%20DESC

And my first PR in incubator-groovy :)


PascalSchumacher:
Hi antoaravinth,

It looks very good, but I guess there should be a check if the setter and/or getter is private. Here are some quick sketches of tests (names etc. are lacking):

    public void testProperties_public_get_private_set() throws Exception {
        testPropertiesFromGetterSetter("GeneratePropertyFromGetSet", "",
            "<a href=\"#_public_get_private_set\">_public_get_private_set</a>", false);
    }
    
    public void testProperties_private_get_public_set() throws Exception {
        testPropertiesFromGetterSetter("GeneratePropertyFromGetSet", "",
            "<a href=\"#_private_get_public_set\">_private_get_public_set</a>", false);
    }
    
    public void testProperties_private_get_private_set() throws Exception {
        testPropertiesFromGetterSetter("GeneratePropertyFromGetSet", "",
            "<a href=\"#_private_get_private_set\">_private_get_private_set</a>", false);
    }

and methods added to GeneratePropertyFromGetSet:

    private void set_public_get_private_set(String public_get_private_set) {
    }
   
    public String get_public_get_private_set() {
        return "";
    }

    public void set_private_get_public_set(String private_get_public_set) {
    }
   
    private String get_private_get_public_set() {
        return "";
    }
    
    private void set_private_get_private_set(String private_get_private_set) {
    }
    
    private String get_private_get_private_set() {
    }

All these test fail.

paulk-asert:
You might want to use java.beans.Introspector.decapitalize which handles some special cases like getZ and getURL. Also we probably need to handle primitive boolean methods of the form isPropName().

paulk-asert:
It is probably also worth checking for no parameters for getters (we don't normally consider indexed properties) and likewise exactly one parameter only for setters. I am a little fuzzy also whether we should include read-only or write-only properties (i.e. only a getter or only a setter) - I'll have to think about that one a bit more.

paulk-asert:
Also, the setters could be checked for void return type. And if we do eventually support read-only getters, we need to rule out some things Groovy adds like getMetaClass and getDeclaringClass and we probably don't want getClass from Java either.

antoaravinth:
@PascalSchumacher @paulk-asert Sure thanks for the comments and ideas. Will take up this one and update the PR soon :)

antoaravinth:
@PascalSchumacher @paulk-asert Done! Kindly have a look at it and let me know, if any more changes is needed from my side!

PascalSchumacher:
@antoaravinth thanks for the update! :)

It looks like your pull request is still missing these requirements from Paul:

"It is probably also worth checking for no parameters for getters (we don't normally consider indexed properties) and likewise exactly one parameter only for setters."

"Also, the setters could be checked for void return type."

antoaravinth:
@PascalSchumacher Sorry didn't check it, actually let me know my understanding is correct:

1. Check getters doesn't have any params.
2. Check set has always void as return type. 

These two changes needs to be added right?

PascalSchumacher:
@antoaravinth Yes, your understanding is correct.

antoaravinth:
@PascalSchumacher Now its done! :) Let me know if anything is needed from my end.

PascalSchumacher:
Looks fine to me, but I never worked on groovydoc. @paulk-asert what do you think?

paulk-asert:
I think it is very close. There are little things which I'd probably tweak, e.g.:
setOrGet.contains("is") would match "history", so startsWith would be better.

antoaravinth:
@paulk-asert : Any update on this?

PascalSchumacher:
@antoaravinth I can do the final clean up myself, but could you please squash the commits? My git skills are not sufficient to squash or even cherry-pick your changes. Thanks!

antoaravinth:
@PascalSchumacher Sure. Will make the changes of what Paulk as suggested.

PascalSchumacher:
Merged: c1516a8180a06c88be2fab18bea4d06062591790 Thanks!

PR #4, opened 2015-04-21
[WIP] Event feature proposal

AugierLe42e:
This is a proposition to easier event programming in Groovy. The feature is inspired from [C#'s Event Handler](https://msdn.microsoft.com/en-us/library/ms366768.aspx) feature syntax and aims to be very similar to Qt's signals and slots feature to exchange messages between objects in an event-driven programming style. This is an idea I've had while working on [an android application written in Groovy](https://github.com/AugierLe42e/Pheobius)

Please note this is a very early work and I'm completely new in contributing to Groovy so please forgive me if I make big mistakes. I'm opening it very soon to have an early feedback.

The current system works like this :

```
class ObjectWithEvent{
    // Declare the event
    public Event somethingChanged = new Event()

    public void makeSomething()
    // Code going on and then, something happens, you throw the event
    somethingChanged(newValue)
}

class ObjectListening{
    //
    private ObjectWithEvent objectWithEvent
    public ObjectListening(){
         // Subscribe to the event
         objectWithEvent.somethingChanged += this.&onSomethingChanged
    }

    private void onSomethingChanged(int newValue){
        // Code here
    }
}
```

The main concern I have with the current implementation is the everything or nothing apsect. Either you make the event public but you are able to fire it from everywhere, either you make it private and you are not able to subscribe to it from outside. Maybe you can help me with that.

The other concern I have is the reaction it could have with the `@CompileStatic` anotation for the type of the Closure.

Please tell me if this work interests you.
Thank you.

### DIFF
--- a/src/main/groovy/util/Event.groovy
+++ b/src/main/groovy/util/Event.groovy
@@ -1,0 +1,35 @@
+package groovy.util
+
+/**
+ * This clas provides a simple event management for Groovy.
+ *
+ * You can add a callback to be executed when an event spawns by simply using the << operator and remove it using the >> operator
+ */
+public class Event
+{
+	private Set<Closure> subscribers = []
+
+	/**
+	 * Adds a new callback to be executed when a event spawns
+	 * @param callback Callback to execute
+	 */
+	void leftShift(Closure callback){
+		this.subscribers.add(callback)
+	}
+
+	/**
+	 * Removes a callcack registered to the event
+	 * @param callback Callback to be removed
+	 */
+	void rightShift(Closure callback){
+		this.subscribers.remove(callback)
+	}
+
+	/**
+	 * Fires the event
+	 * @param args Arguments to be passed to the callback
+	 */
+	void call(Object... args){
+		subscribers*.call(args)
+	}
+}
--- a/src/test/groovy/util/EventTest.groovy
+++ b/src/test/groovy/util/EventTest.groovy
@@ -1,0 +1,53 @@
+package groovy.util
+
+import junit.framework.TestCase
+
+
+class EventTest extends TestCase
+{
+	Event testEvent
+	boolean passed1
+	boolean passed2
+	Closure onEvent1 = { passed1 = true }
+	Closure onEvent2 = { passed2 = true }
+
+	@Override
+	public void setUp() throws Exception
+	{
+		testEvent = new Event()
+		passed1 = false
+		passed2 = false
+	}
+
+	public void testShouldHaveCalledClosure1() throws Exception
+	{
+		assertEquals(passed1, false)
+		testEvent << onEvent1
+		testEvent()
+		assertEquals(passed1, true)
+	}
+
+	public void testShouldHaveCalledBothClosures() throws Exception
+	{
+		assertEquals(passed1, false)
+		assertEquals(passed2, false)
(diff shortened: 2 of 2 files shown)


PascalSchumacher:
Hi AugierLe42e,

thanks for your contribution. :)

New features should be discussed on the developer mailing list. See: http://groovy-lang.org/mailing-lists.html on how to subscribe.

Thanks!

AugierLe42e:
Sure ! Of course ! Thank you, I'm writing it.

src/main/groovy/util/Event.groovy
```diff
@@ -0,0 +1,21 @@
+package groovy.util
+
+import groovy.transform.InheritConstructors;
+
+@InheritConstructors
+public class Event extends Closure
+{
+	private List<Closure> subscribers = []
```
bsideup: Why not "Set"? It should not add listener twice in case of double call

src/main/groovy/util/Event.groovy
```diff
@@ -0,0 +1,21 @@
...
+	public Event plus(Closure closure)
+	{
+		this.subscribers.add(closure)
+		return this
+	}
+
+	@Override
+	Object call(Object... args)
+	{
+		subscribers.each{ it.call(args) }
```
bsideup: can be replaced with: `subscribers*.call(args)`

src/main/groovy/util/Event.groovy
AugierLe42e: Seems reasonnable. As I said, I'm still very new in Groovy. This is the kind of differences I'm not aware of yet ;)

bsideup: it's not Groovy, it's common for every language ; )

AugierLe42e: Well let's say I'm very new in every language ! :p

paulk-asert:
I am not sure whether you have received all the answers you need to progress or whether you have some new ideas but if you want to leave this PR open, you might like to fix up the compilation error (one too many closing braces).

AugierLe42e:
Oh ! Right ! Well, yes, I want to to finish this PR. What do you need to merge it ? Unit tests ?
I suppose I'll need:

 * add a `Closure` with `leftShift` and verify it is in the list,
 * call without any argument and check the `Closure` has been called,
 * call with multiple arguments and check the `Closure` has been called with the correct argument.

Do you need any thing else ?

Sorry I've not been active on this recently. I'm busy with the exams comming and my Android project to finish.

AugierLe42e:
Where can I see the compilation errors ?

PascalSchumacher:
http://ci.groovy-lang.org/viewLog.html?buildId=22697&tab=buildResultsDiv&buildTypeId=Groovy_Jdk6Build (Login as Guest)

AugierLe42e:
Ok, this look finished to me. Ready for review.

AugierLe42e:
Actually, I do have a concern about this PR. I'd like to find a way to constraint parameters of the closure from what the user would pass to the constructor. Something like this:

    Event event = new Event(variable1: int, variable2: String)

so variables would be directly available when declaring an anonymous closure:

    event << { 
        int var = variable1
        variable2.each{ print it }
    }

Would it be possible ?

PR #5, opened 2015-04-22
GROOVY-7378: Support JAVA_OPTS containing quoted arguments on *nix

jwadamson:
Apply the previously provided patch for GROOVY-7378.

### DIFF
--- a/src/bin/startGroovy
+++ b/src/bin/startGroovy
@@ -261,18 +261,18 @@ startGroovy ( ) {
     if $useprofiler ; then
         runProfiler
     else
-        exec "$JAVACMD" $JAVA_OPTS \
-            -classpath "$STARTER_CLASSPATH" \
-            -Dscript.name="$SCRIPT_PATH" \
-            -Dprogram.name="$PROGNAME" \
-            -Dgroovy.starter.conf="$GROOVY_CONF" \
-            -Dgroovy.home="$GROOVY_HOME" \
-            -Dtools.jar="$TOOLS_JAR" \
+        eval exec "\"\$JAVACMD\"" $JAVA_OPTS \
+            -classpath "\"\$STARTER_CLASSPATH\"" \
+            -Dscript.name="\"\$SCRIPT_PATH\"" \
+            -Dprogram.name="\"\$PROGNAME\"" \
+            -Dgroovy.starter.conf="\"\$GROOVY_CONF\"" \
+            -Dgroovy.home="\"\$GROOVY_HOME\"" \
+            -Dtools.jar="\"\$TOOLS_JAR\"" \
             $STARTER_MAIN_CLASS \
             --main $CLASS \
-            --conf "$GROOVY_CONF" \
-            --classpath "$CP" \
-            "$@"
+            --conf "\"\$GROOVY_CONF\"" \
+            --classpath "\"\$CP\"" \
+            "\"\$@\""
     fi
 }
 


src/bin/startGroovy
```diff
@@ -261,18 +261,18 @@ startGroovy ( ) {
...
-            -Dgroovy.starter.conf="$GROOVY_CONF" \
-            -Dgroovy.home="$GROOVY_HOME" \
-            -Dtools.jar="$TOOLS_JAR" \
+        eval exec "\"\$JAVACMD\"" $JAVA_OPTS \
+            -classpath "\"\$STARTER_CLASSPATH\"" \
+            -Dscript.name="\"\$SCRIPT_PATH\"" \
+            -Dprogram.name="\"\$PROGNAME\"" \
+            -Dgroovy.starter.conf="\"\$GROOVY_CONF\"" \
+            -Dgroovy.home="\"\$GROOVY_HOME\"" \
+            -Dtools.jar="\"\$TOOLS_JAR\"" \
```
jwagenleitner: Looks like there is an extra backslash before the environment variable that will cause it to not interpolate.  For example, think change should just be:

`"\"$VAR\""` (no slash before `$`)

For reference I checked [Tomcat's catalina.sh] (https://github.com/apache/tomcat/blob/86b1281a190ec2f17fd1ac5ae3d3889166dcd360/bin/catalina.sh#L325-L339) to see how they handled it.

russel:
eval exec strikes me as a code smell. There must be a better way of dealing with the quoting problem here.

jwadamson:
I agree catalina.sh uses `eval exec "\"$X\""`
Not sure where I got the extra \$ of `eval exec "\"\$X\"`.

A little testing with local shellscripts seems to indicate it doesn't matter (which means I am probably missing something).

    #!/bin/bash
    CMD="./testrun.sh"
    OPTS="'foo bar' \"bar baz\" \"\\\"quoted argument\\\"\""
    ARG="x y"
    (
      echo test1
      eval exec "\"\$CMD\"" "\"\$ARG\"" $OPTS
    )
    echo 
    (
      echo test2
      eval exec "\"$CMD\"" "\"$ARG\"" $OPTS
    )

with my testrun.sh being a simple loop that prints out the number of arguments and then one argument per line

    #!/bin/bash
    echo '$#='"$#"
    i=0
    for VAR in "$@"
    do
    	echo "$i:$VAR"
    	i=$((i + 1))
    done

I can not find a scenario that causes them to have different outputs.
For sake of consistency with catalina, I will probalby remove the extra \$ sequences.

jwadamson:
K, stand corrected. When I use an `ARG="x' \" y"`, the second test `eval exec "\"$ARG\""` breaks with a parser error.

I think the escape of the $ in "\"\$X\"" is needed so that the value from X is not expanded until after the last pass of interpolation. i.e. the eval pass instead of being substituted in during the outer bash execution and then getting its value eval'd.

BTW tomcat will fail if any of those $X variables contain a double-quote character e.g.

    ~/Developer/opt/tomcat"dir$ ./bin/catalina.sh run
    Using CATALINA_BASE:   /home/jwa/Developer/opt/tomcat"dir
    Using CATALINA_HOME:   /home/jwa/Developer/opt/tomcat"dir
    Using CATALINA_TMPDIR: /home/jwa/Developer/opt/tomcat"dir/temp
    Using JRE_HOME:        /usr/lib/jvm/java-7-oracle
    Using CLASSPATH:           /XXX/Developer/opt/tomcat"dir/bin/bootstrap.jar:/XXX/Developer/opt/tomcat"dir/bin/tomcat-juli.jar
./bin/catalina.sh: 1: eval: Syntax error: Unterminated quoted string

jwadamson:
BTW after applying my fix to catalina.sh, tomcat Boostrap.java forbids double-quotes in classpath elements arbitrarily(?). But I can pass a CATALINA_TMPDIR that uses a double-quote literal char

    $ CATALINA_TMPDIR=/tmp/catalina\"dir ./bin/catalina.sh run 
    Using CATALINA_BASE:   /home/XXX/Developer/projects/apache/tomcat80/output/build
    Using CATALINA_HOME:   /home/XXX/Developer/projects/apache/tomcat80/output/build
    Using CATALINA_TMPDIR: /tmp/catalina"dir
    Using JRE_HOME:        /usr/lib/jvm/java-7-oracle
    Using CLASSPATH:           /home/XXX/Developer/projects/apache/tomcat80/output/build/bin/bootstrap.jar:/home/XXX/Developer/projects/apache/tomcat80/output/build/bin/tomcat-juli.jar
    ./bin/catalina.sh: 1: eval: Syntax error: Unterminated quoted string

vs a functional tomcat

    $ CATALINA_TMPDIR=/tmp/catalina\"dir ./bin/catalina.sh run 
    Using CATALINA_BASE:   /home/XXX/Developer/projects/apache/tomcat80/output/build
    Using CATALINA_HOME:   /home/XXX/Developer/projects/apache/tomcat80/output/build
    Using CATALINA_TMPDIR: /tmp/catalina"dir
    Using JRE_HOME:        /usr/lib/jvm/java-7-oracle
    Using CLASSPATH:       /home/XXX/Developer/projects/apache/tomcat80/output/build/bin/bootstrap.jar:/home/XXX/Developer/projects/apache/tomcat80/output/build/bin/tomcat-juli.jar
    15-Jan-2016 16:11:11.726 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log     Server version:    Apache Tomcat/8.0.31-dev
    15-Jan-2016 16:11:11.727 INFO [main] org.apache.catalina.startup.VersionLoggerListener.log Server built:          Jan 15 2016 21:07:24 UTC

jwagenleitner:
You're right, sorry about that, I wasn't testing with `eval` so that was my mistake.

jwadamson:
@russel I agree it is messy looking and smells bad. Unfortunately when I last looked into this and compared with tomcat and ant, it seems like a sort of necessary evil.
Ant's structure looks slightly better where they assign the string to a var then eval that var. But they actually have the same flaw if any of the quoted items contain double-quote (JAVACMD, LOCALCLASSPATH, ANT_HOME etc)

    ant_exec_command="exec \"$JAVACMD\" $ANT_OPTS ...."
    eval $ant_exec_command

Again escaping the $ to defer the expansion until eval resolves that problem. The ant pattern seems more readable.

PascalSchumacher:
Thanks! :+1:

PR #7, opened 2015-04-28
Fix 'call' in Closures/Memoization

DarthPumpkin:
Typo: *call* instead of *can*


PascalSchumacher:
Merged: https://github.com/apache/incubator-groovy/commit/40dd9b30879fe057626bedec64a824fde114e8d8

Thanks :+1: 

Could you please close this pull request? I can not do it using the github gui because this is a read only mirror.

Thanks!

PR #9, opened 2015-05-10
Add includeSuper property to @Canonical annotation

andyjduncan:
Hi,

This is something that came up in discussion at a recent London Groovy and Grails User Group meetup.  It's a bit annoying having to use the three separate AST transformations that make up @Canonical to include the super class in the transformation(s).  I had a look at extending the @Canonical implementation to support this and this is what I came up with.

This adds a property called includeSuper to @Canonical which has has the same effect as setting @EqualsAndHashCode.callSuper, @ToString.includeSuper and @TupleConstructor.includeSuperProperties to true.  I've tried to keep this as low impact as possible so it shouldn't do anything unexpected.

Is there anything I've overlooked here?

Andy Duncan

### DIFF
--- a/src/main/groovy/transform/Canonical.java
+++ b/src/main/groovy/transform/Canonical.java
@@ -129,4 +129,15 @@ public @interface Canonical {
      * the value of this attribute can be overridden within the more specific annotation.
      */
     String[] includes() default {};
+
+    /**
+     * Whether to include super in all transformations.
+     * Acts as if {@code EqualsAndHashCode.callSuper}, {@code ToString.includeSuper} and
+     * {@code TupleConstructor.includeSuperProperties} were all set to true.
+     *
+     * If the {@code @Canonical} behavior is customised by using it in conjunction with one of the more specific
+     * related annotations (i.e. {@code @ToString}, {@code @EqualsAndHashCode} or {@code @TupleConstructor}), then
+     * the value of this attribute can be overridden within the more specific annotation.
+     */
+    boolean includeSuper() default false;
 }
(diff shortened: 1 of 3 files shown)


paulk-asert:
I have just created a different but related PR (https://github.com/apache/incubator-groovy/pull/21). It turns @Canonical into a meta-annotation (aka alias/aggregate/collector). It doesn't provide exactly what you're after but comes very close, e.g.:

    import groovy.transform.*
    @Canonical
    class Foo {
        int a
    }
    @Canonical(callSuper=true, includeSuperProperties=true, includeNames=true)
    class Bar extends Foo {
        int b
    }
    @Canonical(callSuper=true, includeSuper=true)
    @TupleConstructor(includeSuperProperties=true)
    class Baz extends Foo {
        int b
    }
    def (b1, b2, b3) = [new Bar(a:5, b:20), new Bar(10, 20), new Baz(15, 20)]
    assert [b1, b2, b3].toString() == '[Bar(b:20, a:5), Bar(b:20, a:10), Baz(20, Foo(15))]'
    assert b1 != b2
    assert b1.hashCode() != b2.hashCode()

More importantly, the plumbing done to enable the above PR would prohibit the proposals you have here but would allow you to define your own @Canonical2 or whatever that had exactly the semantics that you would like without any programming per se (just define an alternative alias). Let me know if you have any feedback.

paulk-asert:
Thanks for your efforts in creating the PR. I have merged #21, so I'll close this one - but if you feel you need further changes, feel free to create another PR based on the current master. Thanks again.

andyjduncan:
Hi Paul, thanks for looking at this.  I'll take a look at using a meta-annotation for this and check in on the mailing list if I have any problems.  Cheers.

PR #10, opened 2015-05-10
GROOVY-7420 Add a failing test

fpavageau:


paulk-asert:
A trick we use to be able to add such tests into the build is to use `notYetImplemented` which is an available method in your test class, e.g. add the following as the first line of your test:

    if (notYetImplemented()) return

Also, you might like to update the comment header from one of the other recent Groovy source files - we are getting ready to release under Apache guidelines and have tweaked the wording slightly.

fpavageau:
Thanks Paul, I've updated the license in the header and added the `notYetImplemented()` call to the (otherwise failing) method.

PascalSchumacher:
@paulk-asert I guess it's functionally equivalent, but even even easier/more obvious to annotate the test method with @NotYetImplemented (http://docs.groovy-lang.org/latest/html/gapi/groovy/transform/NotYetImplemented.html)?

fpavageau:
I like that, I've pushed the modified version.

paulk-asert:
Yep, that's good too. I tend to always use the annotation version with my JUnit 4 tests but often use the old version with my JUnit 3 style tests - but maybe that's just out of habit. Using the annotation is good in that case too.

PascalSchumacher:
Thanks!

PR #11, opened 2015-05-11
Upgrade to commons-cli 1.3: from OptionBuilder to Option.builder()

jacopoc:


PascalSchumacher:
Looks good as far as I can tell. :+1:

PascalSchumacher:
@paulk-asert What do you think? Is this ready to merge?

jacopoc:
I guess it is not a question for me, but I am not sure, so just in case: in my opinion it should be merged but I am available to do any tests you think are important.

paulk-asert:
Looks good from what I can see. Given that we don't have much "integration test" coverage for using our tools from the commandline, I just wanted to be a little cautious and manually double check all the changes. Also, I wondered whether the @SuppressWarnings("static-access") is related to the "synchronized" or not. If yes, we should be able to delete it. I might have time for those checks next week but happy for someone else to quickly cast their eyes over it and check that nothing changed and if that's the case, we should be good to go.

paulk-asert:
Applied, thanks!

PR #15, opened 2015-05-13
Access to Method parameter names at runtime

nickgrealy:
Given that JDK8 now supports [accessing parameter names at runtime](http://openjdk.java.net/jeps/118), I've incorporated this behaviour
into the Groovy code base (configurable).

Configuration is similar to the javac ```-parameters``` parameter...

It can be enabled by passing ```-pa``` or ```--parameters``` to the ```groovy``` or ```groovyc``` executables, or alternatively, you can set the ```-Dparameters``` system property.

Happy to have feedback on changes/suggestions. Thanks to @blackdrag for showing me where to get started!

Reference: [mail-archives](http://mail-archives.apache.org/mod_mbox/incubator-groovy-dev/201505.mbox/%3CCAK%2BHH6c1Z0FwWPY709uVccWe919viTw%3DRMS5%2BeN9P%2B16Qp7g8Q%40mail.gmail.com%3E)

JIRA: [GROOVY-7423](https://issues.apache.org/jira/browse/GROOVY-7423)

### DIFF
--- a/src/test/org/codehaus/groovy/control/customizers/ASTTransformationCustomizerTest.groovy
+++ b/src/test/org/codehaus/groovy/control/customizers/ASTTransformationCustomizerTest.groovy
@@ -192,7 +192,7 @@ class ASTTransformationCustomizerTest extends GroovyTestCase {
 
 boolean interrupted = false
 try {
-    50.times {
+    100.times {
         Thread.sleep(100)
     }
 } catch (TimeoutException e) {


src/main/org/codehaus/groovy/classgen/AsmClassGenerator.java
```diff
@@ -375,6 +375,13 @@ protected void visitConstructorOrMethod(MethodNode node, boolean isConstructor)
             visitParameterAnnotations(parameters[i], i, mv);
         }
 
+        // Add parameter names to the MethodVisitor (jdk8+ only)
+        if (getCompileUnit().getConfig().getParameters()){
```
roranade: A very minor nit - there should be a space between `...getParameters())` and `{`

nickgrealy:
HI @roranade , good catch, thanks! I've made the change and auto-formatted that line.

Just on the formatting...

Re: ```groovy.ui.GroovyMain``` - I've tried to make the formatting similar to what's already there. Intellij doesn't like the code current formatting (in ```GroovyMain```), so I've left it as is (3 space tab indents, strange indenting in the ```#buildOptions()``` code).

src/main/groovy/ui/GroovyMain.java
```diff
@@ -305,6 +310,7 @@ private static boolean process(CommandLine line) throws ParseException, IOExcept
         main.isScriptFile = !line.hasOption('e');
         main.debug = line.hasOption('d');
         main.conf.setDebug(main.debug);
+        main.conf.setParameters(line.hasOption("pa") || System.getProperty("parameters") != null);
```
paulk-asert: We wouldn't normally check system properties here but rather in CompilerConfiguration#configure. We'd also normally prefix properties e.g. like "groovy.parameters" unless Java is already using something without a java prefix in which case we might use directly what Java uses.

paulk-asert:
Very nice! It might also be nice to enable the functionality via the ant tasks:

    org.codehaus.groovy.ant.Groovy
    org.codehaus.groovy.ant.Groovyc

Have a look at how the stacktrace flag is handled (and possibly how the verbose flag is passed on for joint options in groovyc - might be relevant to do this too - I haven't thought through the joint scenario fully yet).

The only other comment I have is that the patch currently doesn't handle some tricky cases where setting the access might be appropriate, e.g. first parameter "this$0" for an inner class should probably be marked synthetic/implicit and possibly a couple of spots within enums too. But so long as not handling such cases doesn't cause a problem, we can probably endeavor to tweak the behavior to be smarter down the track. Maybe add a TODO comment in your patch for now, e.g. "// TODO handle synthetic/implicit?".

paulk-asert:
Forgot to mention the relevant place for doco changes (which would be nice too):
src/spec/doc/tools-groovyc.adoc (two places, groovyc commandline and groovyc ant task)
src/spec/doc/groovy-ant-task.adoc

nickgrealy:
Thanks @paulk-asert , I didn't even think of ant tasks or the doco. I'll make those changes later today.

**RE: Access Settings** - I didn't understand what the difference was between synthetic/implicit(/???), so I just left them as ```0``` for now. I could just add a ```todo```, but I don't mind doing it properly either. I'll reverse engineer/asm some JDK8 classes, so see how the value changes for enums and inner classes. If you can point me to any documentation, that'd be a great help too!

paulk-asert:
Don't know whether this will help:
http://docs.oracle.com/javase/tutorial/displayCode.html?code=http://docs.oracle.com/javase/tutorial/reflect/member/example/MethodParameterSpy.java

See the Java tutorial for details:
http://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html

But as you may or may not know, Groovy's enums are a bit different from Java's enums under the covers, so there might not be an exact one-to-one correspondence.

paulk-asert:
For bonus marks, you'd also update the doco on the groovy commandline itself but we haven't migrated that across to the new documentation yet.

nickgrealy:
Thanks @paulk-asert , that's perfect. Reading up on http://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html#implcit_and_synthetic now.

Will update groovysh (```org.codehaus.groovy.tools.shell.Main```), with documentation later.

---

Just some other thoughts (I'll leave this for another day):

1) Does gradle's [groovy plugin](http://gradle.org/docs/current/userguide/groovy_plugin.html) support compiler args? (I could only enable ```-parameters``` by using a [configurationScript](http://gradle.org/docs/current/dsl/org.gradle.api.tasks.compile.GroovyCompileOptions.html#org.gradle.api.tasks.compile.GroovyCompileOptions:configurationScript)).

    compileGroovy {
        groovyOptions.compilerArgs << '-parameters'
    }

2) Update [Groovy Eclipse Maven plugin](https://github.com/groovy/groovy-eclipse/) to support ```-parameters``` arg?
3) Intellij > 'Groovy Compiler' settings (<kbd>Ctrl</kbd>+<kbd>Alt</kbd>+<kbd>S</kbd>) doesn't support compiler args?

paulk-asert:
OK, now a placeholder doco for groovy commandline:

    src/spec/doc/tools-groovy.adoc

We may move that around yet.

Yep, groovysh and groovyConsole would be lower priority but should be handled eventually. And yes, we'd need to percolate it out to other tools eventually. Don't know a better Gradle option off the top of my head.

nickgrealy:
**Notes**

I tried using the ASMifier, to decompile all (as many) variations of parameter accessors (e.g. inner classes, implicit constructors, enums) as I could. The only time I got a value other than ```0```, was for enums, which had ```ACC_SYNTHETIC``` on their method parameters. Saying that, the parameters still work using ```0``` for an enum.

e.g. (groovyConsole)

    class Foobar {
        enum Alpha { AAA, BBB}
        def run(String firstParam, int secondParam){}
    }
    
    println Foobar.methods.find{it.name=='run'}.parameters.collect{it.name}
    println Foobar.Alpha.methods.find{it.name=='valueOf'}.parameters.collect{it.name}

Also, groovyConsole's argument handling was "manual", so I updated it to use the ```CliBuilder```.

**Done**

- Add to groovy
- Add to groovyc
- Add to groovysh
- Add to groovyConsole
- Change system property config
- Comment TODO on ASM accessor

**Backlog**

- Add to ant tasks
- Update documentation
- Test

PascalSchumacher:
@nickgrealy Thanks for the great work so far. :)

Could you also rebase to latest master, because the merge of https://github.com/apache/incubator-groovy/pull/11 caused some conflicts with your pull request? Thanks!

nickgrealy:
@PascalSchumacher , sure thing.

paulk-asert:
And a squash might be good too.

nickgrealy:
Roger wilco.

paulk-asert:
Did you get a chance to squash? Are you finished for now wrt the backlog you mentioned?

nickgrealy:
So, quick update. Finally managed to make some time to rebase and squash the code.

Since then, I implemented the ant groovy tasks (just testing atm) and added everything to the documentation.

paulk-asert:
Excellent work! The rebase/squash seems a little weird? Squashed down to 112 commits? Thanks for spotting the doco issues - yep, I will fix up manually in the 2_4_X branch. I'll check on where groovy.parameters doco should go.

nickgrealy:
Yeh, it was my first rebase/squash... so I'm not sure it did what it was supposed to(??).

Ant tasks look good, except for one (existing?) issue that I'm looking at. [Joint compilation](http://www.groovy-lang.org/groovyc.html#ThegroovycAntTask-JointCompilation) (i.e. integrated java compilation) is NOT using/passing on the `-parameters` flag (for Java classes only).

```xml
<groovyc srcdir="src" destdir="out" parameters="true">
	<javac source="1.8" target="1.8" debug="on" srcdir="src" destdir="out">
		<compilerarg line="-parameters"/>
	</javac>
</groovyc>
```

paulk-asert:
you did "git rebase master"?

nickgrealy:
After fetching `upstream/master` (i.e. -> `apache/incubator-groovy`), I used `git rebase -i HEAD~8`, where 8 was my last 8 commits (obviously this isn't correct). I'll give `git rebase master` a go this evening.

paulk-asert:
hmmm, I couldn't see the problem you mentioned in tools-groovyc.adoc, perhaps something weird happened with your rebase attempt? Unless I am missing something.

nickgrealy:
@paulk-asert , take a look at the latest docs for the `master` branch, on the CI server.
http://ci.groovy-lang.org/repository/download/Groovy_Jdk7Build/23592:id/target/distributions/apache-groovy-docs-2.5.0-SNAPSHOT.zip%21/groovy-2.5.0-SNAPSHOT/html/documentation/index.html#section-groovyc-compiler

Notice how the groovyc table, is truncated after the `--basescript` parameter? There should be a few more parameters, which aren't showing up because of the `--indy` parameter.

paulk-asert:
Ah, my bad - was looking in tools-groovy.adoc not tools-groovyc.adoc. I see the problem now. Thanks.

paulk-asert:
fixed and you should get that change if you rebase again (sorry)

paulk-asert:
If the rebase is too hard, you could create a diff of just your changes, close this PR, create a fresh branch from master and apply the diff and make that into a PR.

nickgrealy:
Awesome, thanks for fixing the docs. Not a problem, I've almost nailed down the ant java compiler thing. Will have another go at rebasing -> squashing after that. Will stick with this pull request for now. :)

nickgrealy:
Hi @paulk-asert , I've done a `git fetch upstream` and a `git rebase -i upstream/master`, which has (hopefully) squashed my changes. Please have a look when you get a minute.

paulk-asert:
It still looks a little weird to me. I would be expecting just a small number of commits - possibly just one if you squashed completely. I'll see if anyone else has some git advise on how to correct. Me, I'd do old school close and re-open at this point.

sagarsane:
I have seen this happen before when I squash the commits on my local branch and push my local branch to remote repository without `force pushing` it.

There are a few ways this can be corrected. One of them is to --

`git reset --hard <shaOfLastest upstream/master commit>`
`git cherry-pick <squashed commit's sha>` -- (Assuming there is only 1 final squashed commit)
`git push -f origin master` -- A force push to forked master branch

Also, as Paul said, closing this branch/PR and opening a new branch and cherry-picking the commits on the new branch may probably be cleaner (albeit that would not be much different from `git reset --hard` in this case)

nickgrealy:
Thanks, I'll give the cherry picking a go, and fall back to the diff if I'm still having troubles. 

I'm a bit of a newb to git, so just for next time, should I be working in a branch, within my fork? I was simply working off my master...

sagarsane:
IMO, unless you are working on multiple features at once, having just master is probably fine but in general, yes I do think its better to create a branch to work on.

nickgrealy:
Continued in https://github.com/apache/groovy/pull/38

PR #19, opened 2015-05-19
GROOVY-7338: URL getText(requestProperties) Map now accepts a GString

eginez:
What do you guys think about this change for http://jira.codehaus.org/browse/GROOVY-7338, is in a similar location as my previous PR


melix:
+1

erichelgeson:
:+1: Thanks!

PR #22, opened 2015-05-22
    GROOVY-7227  AST transformations relating to properties/fields should validate their includes/excludes lists

paulk-asert:
Ported (copied and reworked for recent changes) from the old repo:
https://github.com/groovy/groovy-core/pull/569


paulk-asert:
Thanks for your work John.

PR #23, opened 2015-05-23
GROOVY-7416: Problem With CompileStatic And Interface Methods In Abst…

paulk-asert:
…ract Subclass

### DIFF
--- a/src/test/groovy/transform/stc/BugsSTCTest.groovy
+++ b/src/test/groovy/transform/stc/BugsSTCTest.groovy
@@ -650,4 +650,42 @@ Printer
             List tmp = (List) m.get("some_key_here")     // <---  actually groovy crashes here!!
         '''
     }
+
+    // GROOVY-7416
+    void testMethodsFromInterfacesOfSuperClassesShouldBeVisible() {
+        assertScript '''
+            import groovy.transform.CompileStatic
+
+            interface SomeInterface {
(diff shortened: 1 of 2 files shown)


blackdrag:
I have the feeling that this code will still ignore methods coming from an interface that has been extended, especially if there are multiple such interfaces.  

    interface SomeInterface {void foo()}
    interface AnotherInterface extends SomeInterface;
    abstract class Parent implements AnotherInterface;
    abstract abstract class Child extends Parent{ void bar(){foo()}}

paulk-asert:
It seems to be okay. I added another testcase but perhaps I didn't quite catch the case you were after?

src/test/groovy/transform/stc/BugsSTCTest.groovy
```diff
@@ -650,4 +650,42 @@ Printer
...
+        assertScript '''
+            import groovy.transform.CompileStatic
+
+            interface SomeInterface {
+                void someInterfaceMethod()
+            }
+
+            abstract class AbstractSuperClass implements SomeInterface {}
+
+            @CompileStatic
```
shils: Are the CompileStatic annotations here and in the following test necessary? BugsSTCTest extends StaticTypeCheckingTestCase.

paulk-asert:
They can be removed.

paulk-asert:
merged

PR #25, opened 2015-05-27
ResourceGroovyMethods.splitEachLine closure params should be List<String> 

houbie:
ResourceGroovyMethods.splitEachLine closure params should be List<String> iso String[]


paulk-asert:
Thanks for spotting the anomaly with RGM#splitEachLine. I created an issue here:
https://issues.apache.org/jira/browse/GROOVY-7440
And ended up with a slightly different solution:
https://github.com/apache/incubator-groovy/commit/5b521788

PR #26, opened 2015-05-27
Fix IDE integration instructions in README.

pledbrook:
The README stated that you should run

    gradlew jarAll idea

before loading the project into IntelliJ IDEA. It seems, though, that the `jarAll`
task breaks the IDE integration and you can't run tests. Running

    gradlew jar idea

seems to work fine, so I've updated the README to reflect this.


PascalSchumacher:
Thanks! :+1:

PR #29, opened 2015-05-30
Append the power assignment operator to the docs

EPadronU:
The power assignment arithmetic operator was appended to the docs under the `Assignment arithmetic operators` section. Additionally, the corresponding test was included inside the `/src/spec/test/OperatorsTest.groovy` file.


PascalSchumacher:
Thanks! :+1:

PR #46, opened 2015-06-26
GROOVY-7476 - Fix JSON parsing error when escape character is read at the end of the buffer

gvaneyck:


PascalSchumacher:
Thanks! :+1:

PR #50, opened 2015-07-02
GROOVY-7485: BUG! exception in phase 'semantic analysis'... when tryi…

paulk-asert:
…ng to call instance method as default for a constructor parameter


paulk-asert:
merged

PR #51, opened 2015-07-02
groovysh/help: make it clear, that `-e` evaluates code

christoph-frick:
Also fixes a typo. Only Chuck Norris would evaluate the fist option... ;)


PascalSchumacher:
Thanks! :)

PR #55, opened 2015-07-06
GROOVY-7474: Difference between junit3 and junit 4 shouldFail {...} not described in docs

and-dmitry:


paulk-asert:
added - thanks!

PR #56, opened 2015-07-07
escaped less-than and greater-than signs in Javadoc

keeganwitt:


PascalSchumacher:
Thanks! :)

keeganwitt:
My pleasure.  Lemme know if you'd like me to fix the non-public ones too (like `MetaClassImpl.filterPropertyMethod`).

PR #58, opened 2015-07-09
GROOVY-7495: Diamond inheritance of interfaces makes method return ty…

paulk-asert:
…pe incompatible


paulk-asert:
Jochen, I think our fixes are logically equivalent for the cases we have been given. For trickier cases, I don't think they are quite the same but I'm not fully confident which would be the best way to go - so happy to go with yours and if something trickier ever arises we can reassess options at that point. Such trickier cases would certainly be questionable style in any case.

PR #59, opened 2015-07-10
GROOVY-7486: In groovy.util.Node.depthFirst please Provide a way to S…

paulk-asert:
…pecify pre-order, post-order Ordering

### DIFF
--- a/src/main/groovy/util/Node.java
+++ b/src/main/groovy/util/Node.java
@@ -23,6 +23,7 @@ import groovy.lang.DelegatingMetaClass;
 import groovy.lang.GroovySystem;
 import groovy.lang.MetaClass;
 import groovy.xml.QName;
+import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 import org.codehaus.groovy.runtime.InvokerHelper;
 
 import java.io.PrintWriter;
@@ -33,6 +34,7 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Stack;
 
 /**
  * Represents an arbitrary tree node which can be used for structured metadata or any arbitrary XML-like tree.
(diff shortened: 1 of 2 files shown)


src/test/groovy/tree/NavigationTest.groovy
```diff
@@ -57,11 +65,25 @@ class NavigationTest extends GroovyTestCase {
...
+
+    private static createTreeFromXmlParser() {
+        def xml = '''
+        <root>
+          <parent1>
+            <childa/>
+            <childb/>
+          </parent1>
+          <parent2>
+            <childa/>
```
keeganwitt: I think it'd improve this test to have all nodes uniquely named.

paulk-asert: Good idea, I'll make that change when I get a chance.

src/test/groovy/tree/NavigationTest.groovy
```diff
@@ -22,29 +22,37 @@ package groovy.tree
...
+    }
+
+    void testPrePostOrder() {
+        def root = createTreeFromXmlParser()
+        def combos = [[false, true], ['depthFirst', 'breadthFirst']].combinations()
+        def actual = combos.collect{ preorder, type ->
+            root."$type"(preorder)*.name()
+        }*.toString().join('\n')
+        def expected = '''
+            [childa, childb, parent1, childa, grandchild, childb, childc, parent2, root]
```
keeganwitt: Might improve readability to use capital letters or numbers on children.

src/main/groovy/util/Node.java
```diff
@@ -537,26 +539,40 @@ private NodeList getByName(String name) {
...
      */
     public List depthFirst() {
+        return depthFirst(true);
+    }
+
+    /**
+     * Provides a collection of all the nodes in the tree
+     * using a depth-first traversal.
+     *
+     * @param preorder if false, a postorder depth-first traversal will be performed
```
keeganwitt: Solution doesn't provide In-order traversal, which was also requested in 7486.

src/test/groovy/tree/NavigationTest.groovy
keeganwitt: Minor point, but might improve test readability by changing from a multiline string to a concatenated string with trailing comments labeling the combination.

src/main/groovy/util/Node.java
paulk-asert: As mentioned in the Jira, I am not aware of any widely used inorder traversal algorithms for non-binary trees. I have followed some of the computer science literature/research in that area but not all. Do you have references to such an algorithm?

src/test/groovy/tree/NavigationTest.groovy
```diff
@@ -22,29 +22,37 @@ package groovy.tree
...
+            [root, parent1, childa, childb, parent2, childa, childb, grandchild, childc]
+            [grandchild, childa, childb, childc, childa, childb, parent1, parent2, root]
```
keeganwitt: I've never heard of pre/post order BFS.  Is that a thing?  I'd kinda have expected this method to reverse the order of each level (i.e. [root, parent2, parent1, childc, childb, childa, childb, childa, grandchild]).

src/main/groovy/util/Node.java
keeganwitt: That's true, but one could convert the tree to a BST before performing the traversal.  Pretty expensive, but could be done.

paulk-asert: I am not sure I understand where you are going with the BST idea but I have tweaked the tests a little as per your earlier comments.

keeganwitt: You'd traverse the tree (traversal type doesn't matter), adding each node to a TreeSet, then return a List of the TreeSet.  Kinda funky, I'm not sure it's a good idea.  It's just the only way I could think of to offer in-order traversal for non-binary trees.

paulk-asert:
merged

PR #60, opened 2015-07-12
GROOVY-7501: omit package name from builder class name

Controlix:


paulk-asert:
looks good - any chance you can add a test case which currently fails and is fixed after the change?

paulk-asert:
Test added and PR merged, thanks!

PR #62, opened 2015-07-17
Update json-userguide.adoc

anshbansal:
Seems a grammar mistake.


anshbansal:
Also at the end it says "For more details on builders, have a look at the builders chapter". But there is no link. And it is not obvious that it is sub part of DSL chapter.

PascalSchumacher:
Thanks! :+1:

PR #64, opened 2015-07-18
Update core-program-structure.adoc

anshbansal:
- Added why the packages are imported by defaults and also removed the 6 part as two classes are also imported.
- Changed the star import content as that was rendered as bold on the page.
- is instead of it

There are 2 other things that may need some clarification but I am not really sure
>dynamic than Java in that it allows you to define methods with the same name as an imported method as long as you have different types

This needs an example. I am not sure what this means so couldn't add.

>How can we fix it in one place, outside of the original class, without changing all the code that’s using it? Groovy has an elegant solution to this problem.

Is this example valid? I mean the imports will need to be changed. Correct?

### DIFF
--- a/src/spec/doc/core-program-structure.adoc
+++ b/src/spec/doc/core-program-structure.adoc
@@ -56,7 +56,9 @@ Default imports are the imports that Groovy language provides by default. For ex
 include::{projectdir}/src/spec/test/PackageTest.groovy[tags=default_import,indent=0]
 ----
 
-The same code in Java needs an import statement to `Date` class like this: ++import java.util.Date++. Groovy by default imports these classes for you. There are six packages that groovy imports for you, they are:
+The same code in Java needs an import statement to `Date` class like this: ++import java.util.Date++. Groovy by default imports these classes for you. 
+
+The below imports are added by groovy for you:
 
 [source,groovy]
 ----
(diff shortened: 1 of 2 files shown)


anshbansal:
Added spaces as the imported method was not readable in the docs

PascalSchumacher:
> dynamic than Java in that it allows you to define methods with the same name as an imported method as long as you have different types

compile error in java:

```java
import static java.lang.String.format;

public class Dummy {
		System.out.println(format("String")); // compile error
		System.out.println(new Dummy().format(Integer.valueOf(1)));
	}
	
	public String format(Integer i) {
		return i.toString();
	}
}
```

valid groovy:
```groovy
import static java.lang.String.format

class Dummy {
	static void main(String[] args) {
		println(format("String"))
		println(new Dummy().format(Integer.valueOf(1)))
	}

	String format(Integer i) {
		return i.toString()
	}
}
```

> How can we fix it in one place, outside of the original class, without changing all the code that’s using it? Groovy has an elegant solution to this problem.
> Is this example valid? I mean the imports will need to be changed. Correct?

I agree the example is not very helpful. As you said the imports will need to be changed when the class is not defined in the same file. It would be better to change the section with an import alias e.g using `java.util.Date` and `java.sql.Date as SQLDate` or something similar.

PascalSchumacher:
Thanks! :+1: 

If you like to do the other two improvements create another pull request. Otherwise I will do the improvements in a few weeks.

anshbansal:
For improving import aliasing you can take the material from http://www.groovy-lang.org/style-guide.html.

PR #66, opened 2015-07-18
Update GrapeIvy.groovy: set ivyInstance to IvyContext, for speed up

qxo:
When custom grape.config is present(-Dgrape.config=...), there are no need load default ivysettings!

if you want see the diff,run with -Dgrape.config=... and -Divy.message.logger.level=4

groovy-grape-custom-config-no-need-load-default-settings

https://github.com/groovy/groovy-core/pull/665

### DIFF
--- a/src/main/groovy/grape/GrapeIvy.groovy
+++ b/src/main/groovy/grape/GrapeIvy.groovy
@@ -109,6 +109,7 @@ class GrapeIvy implements GrapeEngine {
 
         settings.setVariable("ivy.default.configuration.m2compatible", "true")
         ivyInstance = Ivy.newInstance(settings)
+        org.apache.ivy.core.IvyContext.getContext().setIvy(ivyInstance);
         resolvedDependencies = []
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
@paulk-asert As you are the grape expert, what's your opinion?

paulk-asert:
hoping to get time to look soon

PascalSchumacher:
@paulk-asert All tests pass, I guess this is safe to merge into master?

paulk-asert:
We won't have any tests covering this. I suspect it is safe to merge - just doing some checking. I think it applies all the time, not just when a custom config is specified via `-Dgrape.config`, but I think that might be OK anyway, since I don't think we ever fall back to that config, we use the one in `defaultGrapeConfig.xml`.

paulk-asert:
I did further testing and everything seems to be working well. Thanks for spotting this! I will close this once I create a Jira ticket.

PR #67, opened 2015-07-20
document ivy.message.logger.level

keeganwitt:
I documented the `document ivy.message.logger.level` argument.  I also mentioned the log arguments can be passed via Groovy's invocation or `JAVA_ARGS`.


PascalSchumacher:
Thanks! :+1:

PR #70, opened 2015-07-23
GROOVY-7518: hashCode() throws NPE when using @CompileStatic with @Eq…

paulk-asert:
…ualsAndHashCode

### DIFF
--- a/src/main/org/codehaus/groovy/util/HashCodeHelper.java
+++ b/src/main/org/codehaus/groovy/util/HashCodeHelper.java
@@ -41,7 +41,15 @@ public class HashCodeHelper {
         return shift(current) + (int) var;
     }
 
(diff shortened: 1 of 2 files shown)


blackdrag:
while you did add method for the wrapper types, I do miss the null-check, like it is for example in the null taking method. Plus... well this is an extra, but I would suggest reusing the primitive typed methods from the wrapper typed methods to not have the logic (even though it is small) twice.

paulk-asert:
The null check didn't seem to be needed for the test included in the PR?

blackdrag:
If the Object taking method is used, it means type information is lost somewhere or not used for this combination with CompileStatic. That's bound to be another bug somewhere, which would surface NPEs once fixed.

PR #76, opened 2015-07-29
GROOVY-7509 Problem With Stub Generator And Static Import Aliases

shils:


src/test/org/codehaus/groovy/tools/stubgenerator/ImportStaticAliasTest.groovy
```diff
@@ -0,0 +1,63 @@
...
+    Map<String, String> provideSources() {
+        [
+                'Constants.java': '''
+                    package test;
+                    public class Constants {
+                        public static final String C1 = "c1";
+                    }
+                ''',
+
+                'MyAnnotation.java': '''
```
keeganwitt: Could we simplify the test a little and just refer to the constant directly rather than through an annotation?

src/test/org/codehaus/groovy/tools/stubgenerator/ImportStaticAliasTest.groovy
```diff
@@ -0,0 +1,63 @@
...
+                    package test;
+                    public class SomeJavaClass {
+                        Test test;
+                    }
+                '''
+        ]
+    }
+
+    void verifyStubs() {
+
```
keeganwitt: I think you forgot to add your assertions.

src/main/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
```diff
@@ -925,7 +931,8 @@ private void printImports(PrintWriter out, ClassNode classNode) {
         imports.addAll(Arrays.asList(ResolveVisitor.DEFAULT_IMPORTS));
 
         for (Map.Entry<String, ImportNode> entry : moduleNode.getStaticImports().entrySet()) {
-            imports.add("static "+entry.getValue().getType().getName()+"."+entry.getKey());
+            if (entry.getKey().equals(entry.getValue().getFieldName()))
```
keeganwitt: Could this just be
```java
imports.add("static " + entry.getValue().getType().getName() + "." + entry.getValue().getFieldName());
```
without the conditional?

shils: I'm using a strategy similar to the one used for regular imports - excluding aliased static imports, and replacing their occurrences with their fully qualified names.

src/test/org/codehaus/groovy/tools/stubgenerator/ImportStaticAliasTest.groovy
shils: I'm just testing for successful compilation here - this test is a variant of the one in Groovy7366Bug.groovy.

src/main/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
keeganwitt: I'm rethinking my comment now.  I thought it'd be better to include the non-aliased name, but that could cause a problem if it was aliased to work around a name conflict.  I'm thinking now what you did is the right thing.

src/test/org/codehaus/groovy/tools/stubgenerator/ImportStaticAliasTest.groovy
keeganwitt: I know you just copied what was done previously, but I think we can do better.  If we exclude aliased static imports (which I'm now thinking is right to do), then I'd like us to use the assertions I made in Jira comment, but negate them.  As the test is now, it passes both with and without your changes.  That's not very useful.

shils: If I remember correctly, the test shouldn't pass without my changes. I can't verify right now (don't have access to a computer at the moment), but that's something I specifically tested for. In any case, making those assertions explicit might clear things up so I'll make those changes when I can.

keeganwitt: OK.  If it does, let me know.  Maybe something's goofed up in my environment.

shils: I just ran the test on Master and it failed.

keeganwitt: You're right.  I'm not sure what test report I was reading.  Maybe I had tested without your annotation or something.  Sorry for the false alarm.

src/test/org/codehaus/groovy/tools/stubgenerator/ImportStaticAliasTest.groovy
shils: Never addressed this - the constant is referred to through an annotation so that its usage appears in stubs.

melix:
The PR looks good to me!

keeganwitt:
Thanks!

shils:
Thank you both for reviewing!

PR #77, opened 2015-07-30
GROOVY-7529: Fix up transforms (apart from TupleConstructor) which ar…

paulk-asert:
…e affected by empty includes default

### DIFF
--- a/src/spec/doc/core-metaprogramming.adoc
+++ b/src/spec/doc/core-metaprogramming.adoc
@@ -784,7 +784,7 @@ include::{projectdir}/src/spec/test/CodeGenerationASTTransformsTest.groovy[tags=
 ----
 include::{projectdir}/src/spec/test/CodeGenerationASTTransformsTest.groovy[tags=tostring_example_excludes,indent=0]
 ----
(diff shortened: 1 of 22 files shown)


src/spec/doc/core-metaprogramming.adoc
```diff
@@ -966,6 +966,25 @@ include::{projectdir}/src/spec/test/CodeGenerationASTTransformsTest.groovy[tags=
 ----
 |=======================================================================
 
+[[xform-MapConstructor]]
+===== @groovy.transform.MapConstructor
+
+The `@MapConstructor` annotation aims at eliminating boilerplate code by generating a map constructor for you. A map
```
keeganwitt: Should this also have an attribute table like the other transformations?

paulk-asert: Yes, MapConstructor was missing from the documentation altogether. I added at least a placeholder but I'll probably try to add the doco in a separate commit and then rework this PR accordingly.

paulk-asert:
merged

PR #81, opened 2015-08-07
Change in core differences with Java - Different to extra, more precise

anshbansal:
These are extra keywords not different.


keeganwitt:
Please edit the PR description and commit message to something more descriptive.  Maybe something like "rename 'Different keywords' section to 'Extra keywords'".

keeganwitt:
Thanks!

PR #82, opened 2015-08-09
Added that default must be last in switch case

anshbansal:
Taken from http://www.groovy-lang.org/semantics.html


PascalSchumacher:
Thanks!

PR #87, opened 2015-08-17
Correct spelling of "Windowz" to "Windows".

adjohnson916:
Unless for some odd reason, this was intentional?

### DIFF
--- a/src/bin/startGroovy.bat
+++ b/src/bin/startGroovy.bat
@@ -20,7 +20,7 @@
 @if "%DEBUG%" == "" @echo off
 @rem ##########################################################################
 @rem                                                                         ##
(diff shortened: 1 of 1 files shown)


paulk-asert:
That is very common slang for Windows to hint that some people think it slow and sleepy but you are correct that we might as well fix it up. However, we wouldn't fix {{gradlew.bat}} since that is a generated file from gradle and the change will only be erased next time we upgrade gradle. You could raise a similar report to them if you felt so inclined and if changed would eventually find it's way back into a future version of Groovy. Could you remove that file from your PR? Thanks.

adjohnson916:
@paulk-asert Sure, I've reverted the changes to `gradlew.bat`, and force pushed a rebased commit.

paulk-asert:
merged, thanks

PR #91, opened 2015-08-20
GROOVY-7551: groovy-all has a dependency on com.googlecode:openbeans …

paulk-asert:
…that is not available on Maven Central


paulk-asert:
merged

PR #92, opened 2015-08-20
Completion fixes

tkruse:
This PR contains multiple minor fixes for tab completion in groovysh. I detected those while working on upgrading to jline 2.13. The main fix is for completion of files or folders with blanks or hyphens in the name. The previous implementation tried to juggle with opening and closing hyphens, but this caused several bugs.

Aside from this, I changed a bit the handling of blanks, such that completion of a command without arguments (like ```:exit```) is not completed with a blank at the end, suggesting more arguments are expected/allowed.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -218,7 +218,7 @@ The following class within this product:
     org.codehaus.groovy.tools.shell.completion.FileNameCompleter
 
 was derived from JLine 2.12, and the following patch:
-https://github.com/jline/jline2/issues/90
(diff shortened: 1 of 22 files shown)


tkruse:
I detected a small bug for completion inside an argument, maybe wait before merging...

tkruse:
I think the bug is in jline 2.13 and so this only concerns #93, so this PR here can be reviewed and merged

PascalSchumacher:
Thanks! :+1:

PR #93, opened 2015-08-20
upgrade to Jline2 2.13

tkruse:
This PR includes PR #92 (and depends on the changes therein), so you will see duplicate commits until the other PR is merged and this one is rebased.

Several classes that patched jline2 could be removed. The only Classes still lingering (pending future JLine releases) are the FileNameCompleter, which still has issues in JLine 2.13, and StricterArgumentCompleter, which also still has issues in JLine 2.13.


tkruse:
There may be several regressions in JLine 2.13, I found this one:
https://github.com/jline/jline2/issues/208

I think it might be wise to wait for a bit, since 2.13 seems to have done some risky changes (JLine does not have that many unit tests).

paulk-asert:
just confirming, you are saying #92 is good to go from your side but hold off on this one?

tkruse:
#92 is good, #93 should wait a couple of weeks at least.

paulk-asert:
I know you said hold off but the CI server doesn't like the current PR:
http://ci.groovy-lang.org/viewLog.html?buildId=26783&tab=buildResultsDiv&buildTypeId=Groovy_Jdk6Build
You've deleted JAnsiHelper but it is still references in a test?

tkruse:
Sorry. Rebased and using static helper method instead of JaniHelper.

PascalSchumacher:
@tkruse JLine is dead (as usual), so no idea when 2.13.1 will be released. :(. Do you think we should merge this anyway?

tkruse:
I think jline/jline2#208 is a bit annoying, but not terribly so. I guess there are no worse regressions in 2.13, given a glance at the bug reports.
On the other hand moving to Jline2.13 does not improve the behavior, just the codebase, I think.

So as a vote I'd say -0.

PascalSchumacher:
Well there are also some bugfixes in jline 2.13.

Getting rid of most of the patched classes is very nice.

And I do not like to keep around so many unmerged pull requests.

PascalSchumacher:
Thanks Thibault! :+1:

PR #95, opened 2015-08-21
Improve StreamingJsonBuilder support for @CompileStatic

graemerocher:
I made some non-breaking API changes to `StreamingJsonBuilder` to better support `@CompileStatic` by allowing calls through the `call` method. 

Ideally it would be nice to have some native way to extend static compilation so that dynamic calls could be converted to static calls, but that is for another pull request.


melix:
Merged into master and 2.4.x, thanks!

PR #96, opened 2015-08-21
Refactor InvokerHelper formatting methods

tkruse:
These are some refactorings and minor enhancements of the formatting methods in InvokerHelper.

The formatting of maps, arrays, collections and ranges should be more consistent, allowing the same parameters (verbose, maxSize, safe) everywhere. The 'safe' parameter gets used in many more contexts (though default is still false).

I hope that existing behavior and API was not changed (too much). I believe I did not change any existing *public* method signature, though I added a few with more options.

The behavior should be the same, except for the case when an Object.toString() method throws a checked exception, then this will be rethrown wrapped in a GroovyRuntimeException, when a method is invoked with 'safe=true', which I believe is done by PowerAsserts.

Some changes might need further discussions. One thing annoying me still is that after this PR, the methods
```toArrayString(Object[] collection, boolean verbose, int maxSize, boolean safe)``` and ```formatCollection(Collection collection, boolean verbose, int maxSize, boolean safe)``` are syntactically equal, not sure if there is a nice way not to replicate code, yet also not transforming arrays to collections or vice versa.

The preAllocation of the StringBuffers might be too big, maybe a value like 3 or 4 might be nicer. It's always a heuristic value, maybe premature optimization.

I hope next to change the default printing in PowerAsserts and Groovysh to use the "verbose" flag.

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/InvokerHelper.java
+++ b/src/main/org/codehaus/groovy/runtime/InvokerHelper.java
@@ -62,6 +62,9 @@ public class InvokerHelper {
     protected static final Object[] EMPTY_ARGUMENTS = EMPTY_ARGS;
     protected static final Class[] EMPTY_TYPES = {};
 
+    // heuristic size to pre-alocate stringbuffers for collections of items
+    private static final int ITEM_ALLOCATE_SIZE = 5;
+
     public static final MetaClassRegistry metaRegistry = GroovySystem.getMetaClassRegistry();
 
     public static void removeClass(Class clazz) {
@@ -645,7 +648,8 @@ public class InvokerHelper {
         if (map.isEmpty()) {
             return "[:]";
         }
-        StringBuilder buffer = new StringBuilder("[");
+        StringBuilder buffer = new StringBuilder(ITEM_ALLOCATE_SIZE * map.size() * 2);
+        buffer.append('[');
         boolean first = true;
         for (Object o : map.entrySet()) {
             if (first) {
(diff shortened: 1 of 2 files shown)


tkruse:
rebased and added another little fix to escaping (with test)

blackdrag:
on first look looks good to me... btw, very nice breaking up of commits and commit messages!

PascalSchumacher:
@tkruse: Could you please add a Apache license header to the test? Thanks!

tkruse:
Header added to new file. Note that there is some discussion on https://issuesPRE.apache.org/jira/browse/GROOVY-7564, so I guess this should not be merged until 7564 is okay with Paul, though I can make 7564 a separate PR if necessary. It is just one commit, independent of the rest.

Maybe a third opinion on this could help.

tkruse:
After reconsidering, reduced this PR to just providing strict refactorings, meaning (hopefully) no change of output.
The more controversial changes I have moved to https://github.com/apache/incubator-groovy/pull/111

tkruse:
Added several more tests in a commit before the changes to show API and behavior remain the same (tests should run before and after the refactorings)

paulk-asert:
compileJava is failing:
http://ci.groovy-lang.org/viewLog.html?buildId=26797&tab=buildResultsDiv&buildTypeId=Groovy_Jdk6Build

tkruse:
all PRs good now, I think. There were minor problems stemming from the reordering of commits to separate commits that refactor from commits that change behavior. Sorry that I don't retest locally after every rebase or commit change.

tkruse:
bump

paulk-asert:
Merged (finally!!), thanks!

PR #98, opened 2015-08-24
Document numeric conversions

keeganwitt:


PascalSchumacher:
Nice :), good to merge I guess.

keeganwitt:
Main reason I hadn't was I was trying to decide whether to wait for [GROOVY-7557](https://issues.apache.org/jira/browse/GROOVY-7557) to simplify things.

PascalSchumacher:
I did not follow the discussion on GROOVY-7557, but if it is implemented I think it will only be for 2.5/3.0. Therefore documenting the current behavior is a good idea imho. Of course it's up to you to decide when to merged the pull request. :)

keeganwitt:
Yea, you're probably right.  Some of what I proposed are breaking changes.  Just such a stupidly big table...  The way 7557 could make it smaller is if the behavior were consistent across the primitive and wrapper types, the table could be 1/2 the width with a note below it saying something like "the primitives have the same behavior as the wrapper types in the table".  But as it is today, sometimes you can convert using the wrapper, sometimes you can't.

keeganwitt:
Argh...don't know why I was having such trouble merging.  Created PR #108 to replace this.

PR #99, opened 2015-08-25
Minor fixes

tkruse:
some small issues I saw while doing other things. See commit messages for details


tkruse:
seems this breaks test, will repoen after fixing...

tkruse:
I had created the kind of bug the additional javadoc warns against. Fixed now.

PR #100, opened 2015-08-26
Groovysh: Fix imports not working at all in interpreter mode

tkruse:
Hi,

this is a bit embarassing. The interpreter mode introduced with Groovy 2.4 does not handle imports correctly. With the bug, it becomes impossible to import anything, as long as the shell is in interpreter mode. The fix is simple, since the interpreter mode merely missed to used the 'import' keyword, that I had stripped away for brevity from the list of imports.

I added a test which breaks if imports fail.

### DIFF
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/Groovysh.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/Groovysh.groovy
@@ -245,7 +245,7 @@ class Groovysh extends Shell {
         String variableBlocks = null
         // To make groovysh behave more like an interpreter, we need to retrieve all bound
         // vars at the end of script execution, and then update them into the groovysh Binding context.
-        Set<String> boundVars = ScriptVariableAnalyzer.getBoundVars(importsSpec + Parser.NEWLINE + current.join(Parser.NEWLINE))
+        Set<String> boundVars = ScriptVariableAnalyzer.getBoundVars(importsSpec + Parser.NEWLINE + current.join(Parser.NEWLINE), interp.classLoader)
         if (boundVars) {
             variableBlocks = "$COLLECTED_BOUND_VARS_MAP_VARNAME = new HashMap();"
             boundVars.each({ String varname ->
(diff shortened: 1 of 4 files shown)


tkruse:
Also added another fix with test, where it is impossible to use a defined class in interpreter mode:
```
groovy:000> class Me { String name }
===> true
groovy:000> m = new Me()
script14405523893121089774169.groovy: 2: unable to resolve class Me 
```
The fix is to use the shell's classloader as parent classloader to the Scriptanalyzer, so that the classes can be loaded. This was broken also in all of 2.4

tkruse:
BTW: Do you want me to create JIRA issues for all my current PRs?

PascalSchumacher:
@tkruse Thanks for all you PRs. :) Yes, I would be very nice if you would create JIRA issues for them.

tkruse:
done by now. Though not for refactorings.

PascalSchumacher:
Thanks! :+1:

ssudame:
It is still seen in 2.4.7

tkruse:
Thanks for reporting. Sadly I myself have little occasion to use groovysh.

I confirm the following still does not work:

```
groovy:000> :set interpreterMode true
groovy:000> import java.rmi.Remote
java.rmi.Remote
===> java.rmi.Remote
groovy:000> Remote r
DEBUG [org.codehaus.groovy.tools.shell.InteractiveShellRunner] Read line: Remote r
DEBUG [org.codehaus.groovy.tools.shell.RigidParser] Parsing: import java.rmi.Remote;
Remote r
DEBUG [org.codehaus.groovy.tools.shell.RigidParser] Parse complete
DEBUG [org.codehaus.groovy.tools.shell.Groovysh] Evaluating buffer...
ERROR org.codehaus.groovy.control.MultipleCompilationErrorsException:
script1474514709372448239816.groovy: 1: unable to resolve class Remote 
 @ line 1, column 8.
   Remote r
          ^
```
However, the cause remains unclear, the create buffer for evaluation seems correct and complete:
```
DEBUG [org.codehaus.groovy.tools.shell.RigidParser] Parsing: import java.rmi.Remote;
Remote r
```
And a unit test should have caught this problem...

tkruse:
Ah, well, the fixing commits are on the master branch, but not in 2.4.7. So nothing I can do. @PascalSchumacher

PR #103, opened 2015-08-26
Add withAutoClosable method

alien11689:


PascalSchumacher:
Hi Dominik,

thanks for the pull request.

Groovy still supports Java 6, so it can not be merged as it is: http://ci.groovy-lang.org/viewLog.html?buildId=26394&tab=buildResultsDiv&buildTypeId=Groovy_Jdk6Build

The groovy-nio subproject supports java 7. I do not know if it makes sense to move the methods to `org.codehaus.groovy.runtime.NioGroovyMethods`, what do you think?

alien11689:
Hi Peter,
I have moved this method to NioGroovyMethods,

PascalSchumacher:
Thanks,

I created https://issues.apache.org/jira/browse/GROOVY-7572 to track this.

tkruse:
Since Pascal asked for a review, I would say: It would be nice to have regression tests for: 
- invoking closeWithWarning() with closeable closes closeable
- invoking closeWithWarning() with null does not fail
- invoking closeWithWarning() with closeable that throws Exception on close() does not throw Exception
- invoking withAutoCloseable() closes even when action throws exception
- invoking withAutoCloseable() closes even when action throws exception e1, and e1 is thrown even when close() throws another exception e2.


The code could have a brief comment explaining the strategy of wrapping the call to close in case the call to self action fails, so that the first Exception is not swallowed by the second.

Also withAutoCloseable is vulnerable to invoking with null, will throw NPE. That might be okay, I just thought I mention it.

PascalSchumacher:
Thanks! :+1:

PR #104, opened 2015-08-28
GROOVY-6514: GrapeIvy Classloader selection change, Groovysh :grab and rescan commands

tkruse:
This PR allows to grab Grapes via a new ```:grab``` command. Classes from the new jar are available ia the import command completion. 

To allow the completion to work, I had to change the selection of the classLoader in GrapeIvy. I do not know if that will break anything. Grabbing without completion is already possible in groovysh in a cumbersome way using the grapes code API. Merely adding a command to hide that would also be nice, but the real benefit would be to have import completion, as manually typing import commands is very tiring.

### DIFF
--- a/src/main/groovy/grape/GrapeIvy.groovy
+++ b/src/main/groovy/grape/GrapeIvy.groovy
@@ -185,14 +185,20 @@ class GrapeIvy implements GrapeEngine {
                 throw new RuntimeException("No suitable ClassLoader found for grab")
             }
         }
+        // use the highest available valid loader in hierarchy
+        // (assuming they are all connected), so that
+        // e.g. Groovysh can pick up new classes for import tab completion
+        while (isValidTargetClassLoaderClass(loader.getParent()?.class)) {
+            loader = loader.parent
+        }
         return loader
     }
 
-    private boolean isValidTargetClassLoader(loader) {
+    private static boolean isValidTargetClassLoader(ClassLoader loader) {
         return isValidTargetClassLoaderClass(loader?.class)
     }
 
-    private boolean isValidTargetClassLoaderClass(Class loaderClass) {
+    private static boolean isValidTargetClassLoaderClass(Class loaderClass) {
         return (loaderClass != null) &&
             (
              (loaderClass.name == 'groovy.lang.GroovyClassLoader') ||
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
@@ -1,0 +1,83 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.codehaus.groovy.tools.shell.commands
+
+import jline.console.completer.Completer
+import org.codehaus.groovy.tools.shell.CommandSupport
+import org.codehaus.groovy.tools.shell.Groovysh
+
+/**
+ * The 'grab' command.
+ *
+ * @author Thibault Kruse
+ */
+class GrabCommand
+    extends CommandSupport
+{
+    public static final String COMMAND_NAME = ':grab'
+
+    GrabCommand(final Groovysh shell) {
+        super(shell, COMMAND_NAME, ':g')
+    }
+
+    @Override
+    protected List<Completer> createCompleters() {
+        return []
+    }
+
+    @Override
+    Object execute(final List<String> args) {
+        assert args != null
+
+        if (args.size() == 0) {
+            fail("Command '$COMMAND_NAME' requires at least one argument") // TODO: i18n
+        }
+        String specString = buildSpec(args)
+
+        log.debug("Attempting to grab: \"$specString\"")
+        shell.execute("groovy.grape.Grape.grab($specString)")
+        shell.packageHelper.reset()
+    }
+
+    protected String buildSpec(List<String> args) {
+        Map<String, String> spec = [:]
+        List<String> splitArgs
+        if (args.size() == 1 && args[0].contains(':')) {
+            splitArgs = args[0].split(':')
+        } else {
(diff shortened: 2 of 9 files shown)


PascalSchumacher:
@paulk-asert What do you think of the changes to grape?

Dispader:
...interesting.  The `Grape` modification might not be necessary.

It seems the effect we want after the `GrabCommand` gets run is a modification of the `Completer` of the `ImportCommand`.  I'm eyeing up the source of `ImportCompleter` in there... and it... seems to pass `shell` by reference.

Since we get the same `shell` when we construct `GrabCommand`, I'm going to see if we can't send a message up the chain.

Dispader:
@tkruse — can you take a look at my comment on 9244919 and see if it makes sense?

It seems we can specify a `classloader` to Grab without patching it.  That said— I've also been trying to modify `PackageHelper`, but it doesn't seem to have much luck getting `ImportCommand`'s completer to pick up any changes.

PascalSchumacher:
I would have merged this long ago if this did not modify Grape. If this modification could be removed I would merge immediately.

Dispader:
Yes— I _really_ think we can do this.  I've managed to get the Grab command to insert the loaded dependency in a `classloader` accessible to the `ImportCommand` completion scanner, and I've managed to get the scanner to reload its import map by sending messages.

I'm tidying up a script version of this code, and I'll post a link here.

What I _haven't_ been able to do is figure out how to test a new `Command` that I add to the `groovysh` tree: I can _build_ it, but I'm too dense to figure out how to execute the built shell. @tkruse 's work looks really good here— so it might be a matter of modifying the good work he's done.

Flipside, if someone could point me in the direction of some help testing a GroovyShell built from the core tree— I'd feel more comfortable creating one that's more appropriate to include.

Dispader:
I finished up a [small project](https://github.com/Dispader/groovysh-command-grab) with a simple, script-style [GrabCommand](https://github.com/Dispader/groovysh-command-grab/blob/master/src/main/groovy/GrabCommand.groovy).

The critical bit for the `classloader` is (currently) [line 40](https://github.com/Dispader/groovysh-command-grab/blob/master/src/main/groovy/GrabCommand.groovy#L40) in the method that does the grab

```Groovy
    private void grab(String dependency) {
        Map<String, Object> dependencyMap = org.codehaus.groovy.tools.GrapeUtil.getIvyParts(dependency)
        groovy.grape.GrapeIvy grapeIvy = groovy.grape.Grape.instance
        grapeIvy.grab([classLoader: shell.interp.classLoader.parent,refObject: shell.interp], dependencyMap)
    }
```

There are still a few problems with it— notably that it toggles import completion "on", even if it's previously been set "off" intentionally.

```Groovy
    private void reloadImportCompletion() {
        shell.packageHelper.preferenceChange(ignorePackageCompletion(true))
        shell.packageHelper.preferenceChange(ignorePackageCompletion(false))
    }
```

Anyhow, hope this helps.  I'd be happy to try to make it a Pullable command, if someone can help me understand how to execute a build shell.  But I'm wagering a more seasoned Groovy contributor could get this done faster than myself.

Cheers.

PascalSchumacher:
Hi Jake,

thanks for your contribution. :)

>What I haven't been able to do is figure out how to test a new Command that I add to the groovysh tree: I can build it, but I'm too dense to figure out how to execute the built shell. @tkruse 's work looks really good here— so it might be a matter of modifying the good work he's done.

I'm not sure if I understand you correctly, but you register a command by adding it to  `DefaultCommandsRegistrar` see the pull request for an example. 

Not sure if my memory is correct but then you should be able to run "org.codehaus.groovy.tools.shell.Main" from your IDE.

You can also run `./gradlew distBin` to build the distribution then unzip the zip under `target/distributions` and run the shell from there.

 >But I'm wagering a more seasoned Groovy contributor could get this done faster than myself.

Thibault is the most knowledgeable. The initial creator of groovysh does not contribute anymore and the rest of the team only did minimal fixes/enhancements.

PascalSchumacher:
@Dispader It would be very nice if you could clone the repo of this pull request https://github.com/tkruse/incubator-groovy/tree/groovysh-grab and create a pull request with your updates.

tkruse:
@Dispader thanks for following up on this. I have been away for this weekend, and now I'm down with a cold, so I cannot think too straight.
Regarding how to start a shell with modifications, my workflow has been very messy so far:

1. I run gradle dist to create a distribution, and unzip it to some location (to have dependencies to put on the classpath).
2. I delete groovy-groovysh-$VERSION.jar from that location
3. Then I remove all groovy-based .class files left in the source dir that could be/become stale:
`rm -rf subprojects/groovy-groovysh/target/classes/.../shell/*.class`
4. Finally I run a script that starts the Main:

```
export CLASSPATH=\
subprojects/groovy-groovysh/target/classes/main:\
subprojects/groovy-groovysh/target/resources/main:\
target/classes/main:\
target/resources/main:\
.../commons-cli-1.3.1.jar:\
.../asm-5.0.3.jar:\
.../jline-2.12.jar:\
.../antlr-2.7.7.jar:\

java -classpath $CLASSPATH  org.codehaus.groovy.tools.shell.Main $@
```

This is the only way I found to be sure to run the groovy classes I am modifying against the latest java classes and latest groovysh groovy sources. Probably there is a better way, but once I got this to work, I did not bother to find that better way. Obviously it is too messy to document as recommendation.

Starting in the IDE should also work, but I usually avoid it, as I am suspicious of differences between a real command line and the IDEs command-lines.

Dispader:
@tkruse That is exactly* the information I was looking for.  Thank you so much, sir!  (And I hope you're feeling better soon— blech.)

@PascalSchumacher And thank you, sir!  I've been having problems importing the whole Groovy project into Eclipse with BuildShip— I've got it hobbling along enough to navigate the code, but I'm getting things such as

```
A cycle was detected in the build path of project 'groovy'. The cycle consists of projects {groovy, groovy-ant, groovy-groovydoc, groovy-templates, groovy-xml, groovy-test, groovy-bsf, groovy-console, groovy-swing, groovy-docgenerator, groovy-groovysh, groovy-jmx, groovy-json, groovy-jsr223, groovy-macro, groovy-nio, groovy-servlet, groovy-sql, groovy-testng}
```

...but I think my major problems were build fails due to... license checks?  Afraid I don't really understand what's going on with all that.

Anyhow— I'll give it a go and see where we get.

Dispader:
And hold off on responding to my personal panic too, too quickly: I just refreshed master and was able to do a `./gradlew distBin` with no errors whatsoever!

PascalSchumacher:
>A cycle was detected in the build path of project 'groovy'. The cycle consists of projects {groovy, groovy-ant, groovy-groovydoc, groovy-templates, groovy-xml, groovy-test, groovy-bsf, groovy-console, groovy-swing, groovy-docgenerator, groovy-groovysh, groovy-jmx, groovy-json, groovy-jsr223, groovy-macro, groovy-nio, groovy-servlet, groovy-sql, groovy-testng}

This is probably fixable, but the problem is that the groovy-eclipse compiler is a fork of the groovy compiler. Most of the time you can not compile the current code with compiler for the previous version (which the groovy-eclipse compiler always is), so having everything working in eclipse is rare.

Dispader:
With apologies, I created [another Pull Request](https://github.com/apache/groovy/pull/286).  (It was the only thing I could figure out.)

@tkruse — I did give you access to my fork, so you should be able to commit/destroy/comment anything we still don't like— and you should be able to apply changes to this Pull (?).  I'm really sorry, guys— I'm still getting used to the workflow here.

Hope you're feeling better, Thibault... and I hope that I'm not annoying you or Pascal too much.

Cheers,
Jake

PR #105, opened 2015-08-30
GROOVY-7519: Fixes JsonOuput.toJson() leads to StackOverflowException

eginez:
JsonOutput overflows when serializing a File due to recursive references to File objects. Namely getAbsoluteFile(), getCanonicalFile(), and getParentFile(). These properties will be properly serialized by their counterpart methods getAbsolutePath(), getCanonicalPath() and getParent().

### DIFF
--- a/subprojects/groovy-json/src/main/java/groovy/json/JsonOutput.java
+++ b/subprojects/groovy-json/src/main/java/groovy/json/JsonOutput.java
@@ -287,7 +287,7 @@ public class JsonOutput {
                 writeArray(objectClass, object, buffer);
             } else if (Enum.class.isAssignableFrom(objectClass)) {
                 buffer.addQuoted(((Enum<?>) object).name());
-            }else if (objectClass == File.class){
+            }else if (File.class.isAssignableFrom(objectClass)){
                 Map<?, ?> properties = DefaultGroovyMethods.getProperties(object);
                 properties.remove("class");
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Hi Esteban,

You talking about getXZYPath() has me wondering if your fix will work with Java 6 (because Groovy still supports that)?

Thanks,
Pascal

eginez:
Good question let me check the jdk 6 java docs I suppose team city will also tell us if the test breaks

eginez:
After checking the documentation, I can confirm this patch will work in the JDK6.
The JDK6 has basically the same methods using File as a return type. These methods will be excluded from the serialization. However their counterpart methods return String(as opposed to Path in JDK7+ )
, which meas they will get properly serialized in json. Eg

File getParentFile() ----> skipped
(String|Path) getParent() ----> serialized to [parent='/some/path/']

Thus no loss of information/ no recursive references to File object.

eginez:
Made a mistake in previous comment, the JDK7 does not use Path as the return type of the methods, it uses String.

graemerocher:
One comment I would make regarding this, if we start adding `isAssignableFrom` checks for every possible type then the performance is going to start to suffer. It may be that there needs to be an extensible codec mechanism added instead.

PascalSchumacher:
Thanks! :+1:

PR #106, opened 2015-08-31
Add Magic Package topic

stavytskyi:
I would like write this topic to manual.


PascalSchumacher:
Thanks for taking this on. :+1:

PR #107, opened 2015-09-01
Fix Groovysh Test failure under Java9, due to new virtual Filesystem …

tkruse:
…structure

This patch only scans the virtual subfolder 'modules' (not 'packages'), thus that the element paths loose the leading slash:
```/modules/foo/xyz/XYZ.class -> modules/foo/xyz/XYZ.class```
So that the stripping of 'modules/<modulename>' works again.

### DIFF
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/PackageHelperImpl.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/PackageHelperImpl.groovy
@@ -108,8 +108,9 @@ class PackageHelperImpl implements PreferenceChangeListener, PackageHelper {
             }
         }
         if (jigsaw) {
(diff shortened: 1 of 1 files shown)


tkruse:
This also contains minor refactorings: not having a special-purpose default closure, and passing around the url "jrt:/" instead of duplicating it.

tkruse:
Related to this:
https://github.com/apache/incubator-groovy/commit/0e384ec3
and this:
http://mail.openjdk.java.net/pipermail/jigsaw-dev/2014-November/004044.html

With this PR, I guess the test would break when running against earlier alphas of Java9 (I was using 1.9.0-ea-b76, ot sure which one was used on buildfarm). Not sure if anyone cares. If so, it might be possible to hack something that tries to act differntly on different early Java9 versions.

PascalSchumacher:
Thanks you very much. :)

No need to worry about earlier alphas imho.

The groovy buildserver builds jdk9 from source once a week.

PR #109, opened 2015-09-01
Groovy-7574: Ranges: Make Range members final. Also improve javadoc.

tkruse:
Else they can easily be changed in Groovy, which leads to undefined states

### DIFF
--- a/src/main/groovy/lang/IntRange.java
+++ b/src/main/groovy/lang/IntRange.java
@@ -102,28 +102,32 @@ public class IntRange extends AbstractList<Integer> implements Range<Integer> {
      * For non-inclusive aware ranges, the first number in the range; <code>from</code> is always less than or equal to <code>to</code>.
      * For inclusive aware ranges, the <code>from</code> argument supplied to the constructor.
      */
-    private int from;
+    private final int from;
 
     /**
      * For non-inclusive aware ranges, the last number in the range; <code>to</code> is always greater than or equal to <code>from</code>.
      * For inclusive aware ranges, the <code>from</code> argument supplied to the constructor.
      */
-    private int to;
+    private final int to;
 
     /**
      * If <code>false</code>, counts up from <code>from</code> to <code>to</code>.  Otherwise, counts down
-     * from <code>to</code> to <code>from</code>. Not used for inclusive aware ranges.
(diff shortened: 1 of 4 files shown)


tkruse:
Rebased, added commits forROOVY-7585, some refactorings, and tests

paplorinc:
Interesting.

On a slightly different topic: how difficult would it be to support unbound ranges also, like in ```https://code.google.com/p/guava-libraries/wiki/RangesExplained```.

I don't know about the syntax, I guess `_..10` or `(..10)` or `<..10`.

Or ranges with non-integer step size, i.e. `0.1, 0.11, 0.11 ... 0.2` as `0.1..0.01..0.2` or at least to support integers in the ObjectRange's step setter :/

Let me know if you consider this off topic and will post it in the issues instead.

tkruse:
non-integer ranges are already kind-of supported in Groovy. Ranges are defined by the "next()" method on the instances, which for 0.1..10 yields [0.1, 0.2, 0.3 ...]
Ranges can either be inspired by for-loops (having an increment step), or by mathematic intervals. Given Guava has ranges as you point out, I am not sure if Groovy should go out of its way to implement the same thing. 

Changing the grammar so that the parse can parse a larger variety of ranges should not be that difficult technically, but you may face resistance from language designers. Changing the code for the ranges an be a bit challenging, in particular since you must not break existing usages.

I am not sure if you are volunteering to work on this, or asking for this to be done. In any case I'd say the mailing lists are a good start to try to get support: http://www.groovy-lang.org/mailing-lists.html

paplorinc:
> `0.1..10` yields `[0.1, 0.2, 0.3 ...]`

actually it yields `[0.1, 1.1, 2.1, 3.1, 4.1, 5.1, 6.1, 7.1, 8.1, 9.1]`, because it has a step size of `1`, so e.g. `(0.1..0.5).collect()` will yield `[0.1]`

> I am not sure if you are volunteering to work on this, or asking for this to be done.

Would gladly contribute, will ask the mailing list for opinions, thanks :)

tkruse:
sorry for wrong info about floats.

tkruse:
superceded by #141 and #142 and #143

PR #111, opened 2015-09-01
GROOVY-5763: Make InvokerHelper output self-consistent

tkruse:
Changes the internal API of the formatting methods in InvokerHelper and also the output in some cases when using verbose or safe arguments.

This PR contains currently controversial changes to the formatting methods in InvokerHelper that could break client code. 

Further analysis and discussion required.

EDIT: This currently contains the commits from #96 (until that one is merged).


blackdrag:
why is GROOVY-5763 mentioned here? isn't that the wrong issue?

tkruse:
thanks, just twisted the numbers, several times by copy&paste. Even the branch has the wrong name, so I reopen with a differently named source branch #112

PR #118, opened 2015-09-17
Provide static reference for indy setting

pickypg:
This enables developers to avoid using a "magic string" in their code and to use the known-to-be-right version of it when setting up `CompilerConfiguration`s to use invokedynamic (aka indy).

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/asm/WriterController.java
+++ b/src/main/org/codehaus/groovy/classgen/asm/WriterController.java
@@ -94,7 +94,7 @@ public class WriterController {
             optimizeForInt=false;
             // set other optimizations options to false here
         } else {
-            if (Boolean.TRUE.equals(optOptions.get("indy"))) invokedynamic=true;
+            if (Boolean.TRUE.equals(optOptions.get(CompilerConfiguration.INVOKEDYNAMIC))) invokedynamic=true;
             if (Boolean.FALSE.equals(optOptions.get("int"))) optimizeForInt=false;
             if (invokedynamic) optimizeForInt=false;
             // set other optimizations options to false here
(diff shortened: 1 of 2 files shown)


keeganwitt:
Really if we're going to do this (and I agree we should), we might as well go all the way and have fields for all the other strings too.

pickypg:
@keeganwitt I agree. I'm not actually sure of the full list, but if one exists then I can update the PR to do that instead.

keeganwitt:
If we did "indy", one option we should add for sure is "int", since most callers will want to disable that if they're enabling indy.  We also should replace all occurrences with the static reference I think.  Which would include
* GroovyMain
* ClassNodeResolver
* FileSystemCompiler
* Groovy (in groovy-ant module)

Others we might consider are
* all (another option for optimizationOptions)
* stubDir
* keepStubs
* namedValues
* flags

For options that are kept in maps, maybe an enum as the key instead of a bunch of static strings would be better anyway.  Not sure how @melix would feel about that idea.

keeganwitt:
I'm thinking ones we probably _wouldn't_ do would be all the system properties (e.g. "groovy.source.encoding", "groovy.target.bytecode", etc).  I don't think those would be programmatically set.

pickypg:
Agreed. It's uncommon to expose system properties.

keeganwitt:
I guess to be as backwards compatible as we can, we should just keep as strings.

pickypg:
As long as the strings don't change, there is no backwards compatibility lost and compatibility is improved for newer clients. It's only for code looking for the

I've been traveling a lot recently, so I'll dig this one back up and start going through to find the various necessary strings.

PascalSchumacher:
Thanks!

PascalSchumacher:
I merged the pull request. Of course you are still welcome to extend it to other constants. :)

pickypg:
@PascalSchumacher Thanks. I've been getting pulled in a lot of directions for the past two months. I'll try to circle back for the other ones noted above!

PR #119, opened 2015-09-17
Allow security managers to prevent system property access for Indy

pickypg:
When using the `"indy"` setting, it tries to read a system property to enable logging.

This change enables a JVM Security Manager to block this setting and therefore simply ignore the setting to use the default: do not log.

From a security manager, you might otherwise catch a `SecurityException` containing:

> access denied ("java.util.PropertyPermission" "groovy.indy.logging" "read")

Not only is this not recoverable, but the `IndyInterface` class will be inaccessible thereafter due to it being in a `static` block that fails (the dreaded `NoClassDefFoundError`).


PascalSchumacher:
@pickypg Thanks for this pull request. I created https://issues.apache.org/jira/browse/GROOVY-7626 to track this.

PascalSchumacher:
Thanks for the pull request. :)

Sorry for taking so long to merge. :(

The fix will be in the next release (2.4.6).

PR #121, opened 2015-09-28
Disabling timestamps in groovydoc

ebourg:
Hi,

There is an ongoing effort in Debian to make every builds [reproducible](https://wiki.debian.org/ReproducibleBuilds) and the timestamps embedded in the files generated by groovydoc is [one](https://reproducible.debian.net/issues/unstable/timestamps_in_documentation_generated_by_groovydoc_issue.html) of the many issues to solve in order to achieve this goal.

I'd like to suggest the addition of a `notimestamp` option for groovydoc similar to the one in javadoc, it would make it possible to remove the timestamps when needed.

Thank you


graemerocher:
+1 this has become a real problem for us too because we use gh-pages to publish our docs so each time we publish it creates huge commits because every javadoc has changed

It also prevents us from publishing snapshots because the change sets are too big.

ebourg:
The patch only removes the timestamps and preserves the version of Groovy used (javadoc's `-notimestamp` option removes both). Removing the version might be desirable too in your case, otherwise you'll still get large change sets when upgrading to a new version of Groovy.

graemerocher:
upgrading Groovy is less frequent, so I'm not as concerned about that

paulk-asert:
It wouldn't be that hard to make both configurable but I'll have a bit of a closer look at what the JavaDoc code does before commenting further here if needed. It seems like this is a fairly straight-forward change even if some minor tweaking might be desirable.

ebourg:
Here is the link to the relevant code in javadoc:

http://hg.openjdk.java.net/jdk8u/jdk8u/langtools/file/f166261986cc/src/share/classes/com/sun/tools/doclets/formats/html/markup/HtmlDocWriter.java#l329

paulk-asert:
JavaDoc doesn't have version as an option, so I think a separate flag would make sense for that. I am happy to merge your request and add the extra flag in the next day or two hopefully.

paulk-asert:
I did a minor tweak to ensure that the new behavior was backwards compatible by default but otherwise merged your proposal. I also then added a `noversionstamp` flag too. Thanks!

ebourg:
Thank you!

ebourg:
Actually the issue is not completely fixed, I didn't notice the templates also include the date with a `<meta>` tag:

    <meta name="date" content="${new Date().format('yyyy-MM-dd')}">

This line should be removed when `-notimestamp` is specified.

paulk-asert:
Amended. Nice pick up.

ebourg:
Thank you for the quick fix

paulk-asert:
I added another commit to enable the same support in the ant task.

PR #122, opened 2015-09-28
MacroGroovy

bsideup:
This PR supercedes https://github.com/groovy/groovy-core/pull/636 : it is the same contents but rebased onto master. See https://github.com/groovy/groovy-core/pull/636 for discussion.


src/main/org/codehaus/groovy/ast/ClassCodeVisitorSupport.java
```diff
@@ -161,11 +161,6 @@ public void visitAssertStatement(AssertStatement statement) {
         super.visitAssertStatement(statement);
     }
 
-    public void visitBlockStatement(BlockStatement block) {
```
bsideup: @melix I need your support on this deletion. Is it supposed to be deleted?

melix: Given the fact that it is still there on master, I assume it shouldn't be deleted. I would restore it and run the tests locally to make sure.

PR #124, opened 2015-09-29
[GROOVY-6854] JsonOutput should be configurable with ISO 8601 UTC time zone instead of GMT

benignbala:
Added a private static final SimpleDateFormat and a private static
TimeZone that gets set when setTimeZone is called.

writeDate uses this static final SimpelDateFormat object and sets the
configured timezone on this object before converting the Date.

Added a test case for the fix as well.


subprojects/groovy-json/src/main/java/groovy/json/JsonOutput.java
```diff
@@ -61,7 +61,8 @@
     private static final String NULL_VALUE = "null";
     private static final String JSON_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
     private static final String DEFAULT_TIMEZONE = "GMT";
-
+    private static final SimpleDateFormat dateFormat = new SimpleDateFormat(JSON_DATE_FORMAT);
```
PascalSchumacher: SimpleDateFormat is not thread-safe (see http://stackoverflow.com/questions/6840803/simpledateformat-thread-safety for an explanation)

benignbala:
Updated the pull request by adding a setDateFormat and making the dateFormat field non-final.

benignbala:
Another update :) overloaded toJson(Date) and toJson(Calendar). There are now methods that take an additional parameter of the time SimpleDateFormat.

subprojects/groovy-json/src/main/java/groovy/json/JsonOutput.java
```diff
@@ -61,7 +61,7 @@
     private static final String NULL_VALUE = "null";
     private static final String JSON_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
     private static final String DEFAULT_TIMEZONE = "GMT";
-
+    private static SimpleDateFormat dateFormat = new SimpleDateFormat(JSON_DATE_FORMAT);
```
jwagenleitner: Due to thread safety issues you really don't want to `new SimpleDateFormat` at the class level.  Creating the format should be done at the method level so multiple threads wont access the same object concurrently.

subprojects/groovy-json/src/main/java/groovy/json/JsonOutput.java
```diff
@@ -119,12 +119,24 @@ public static String toJson(String s) {
      * @return a formatted date in the form of a string
      */
     public static String toJson(Date date) {
+        dateFormat.setTimeZone(TimeZone.getTimeZone(DEFAULT_TIMEZONE));
```
jwagenleitner: See note above, probably should new up a `SimpleDateFormat` here, might be better to provide a way to override the JSON_DATE_FORMAT that is used and construct a new format object using either the provide format string and fall back to JSON_DATE_FORMAT if not provided.

benignbala:
@jwagenleitner Thanks for your inputs, in that case should we provide toJson(Date), toJson(Date, Format), toJson(to, format, timezone), toJson(timezone). ? And, we will have to provide the same set of stuff to Calendar as well.

Otherwise, we could do just what the bug asks - Make timeZone configurable and leave the date format fixed -  JSON_DATE_FORMAT is that ok?

subprojects/groovy-json/src/main/java/groovy/json/JsonOutput.java
```diff
@@ -59,9 +59,8 @@
     private static final char[] EMPTY_STRING_CHARS = Chr.array(QUOTE, QUOTE);
 
     private static final String NULL_VALUE = "null";
-    private static final String JSON_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
+    private static final String JSON_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ssXX";
```
jwagenleitner: This will throw an `java.lang.IllegalArgumentException` in Java 6 and according to http://www.groovy-lang.org/install.html#_prerequisites JDK 6+ works with Groovy 2.4.

jwagenleitner:
As the original poster said, this is tough being that the entire class is static.  Even with this change you would have no way to alter the format of the outputted date unless you call one of the toJson(Date) methods.

```
toJson([a:'foo', b: new Date()]) // will use default, no way to alter format or tz
```
One alternative may be to return to using a `ThreadLocal` (just not a sub-class that would cause leaks in Tomcat per [GROOVY-7143](https://issues.apache.org/jira/browse/GROOVY-7143)) and check a system property to determine if ISO8601 dates should be outputted.

```groovy
.....
.....
    private static final String DATE_FORMAT_KEY = "groovy.json.JsonOutput.dateFormat";
    private static final String DATE_FORMAT_VAL_ISO8601 = "ISO8601";
    private static final String JSON_DATE_FORMAT_ISO8601 = "yyyy-MM-dd'T'HH:mm:ssXX";
    private static final String DEFAULT_TIMEZONE_ISO8601 = "UTC";

    private static final ThreadLocal<DateFormat> dateFormatter = new ThreadLocal<DateFormat>() {
        @Override
        protected DateFormat initialValue() {
            SimpleDateFormat df = null;
            if (DATE_FORMAT_VAL_ISO8601.equals(System.getProperty(DATE_FORMAT_KEY))) {
                df = new SimpleDateFormat(JSON_DATE_FORMAT_ISO8601, Locale.US);
                df.setTimeZone(TimeZone.getTimeZone(DEFAULT_TIMEZONE_ISO8601));
            } else {
                df = new SimpleDateFormat(JSON_DATE_FORMAT, Locale.US);
                df.setTimeZone(TimeZone.getTimeZone(DEFAULT_TIMEZONE));
            }
            return df;
        }
    };
.....
.....
.....
    /**
     * Serializes date and writes it into specified buffer.
     */
    private static void writeDate(Date date, CharBuf buffer) {
        buffer.addQuoted(dateFormatter.get().format(date));
    }
.....
.....
```
Then calling code could turn on ISO8601 dates
```groovy
    // From command line:
    //     groovy -Dgroovy.json.JsonOutput.dateFormat=ISO8601 myscript.groovy
    // or from code
    System.setProperty("groovy.json.JsonOutput.dateFormat", "ISO8601");
    JsonOutput.toJson([a:'foo', b: d]) //== {"a":"foo","b":"2008-03-04T12:50:00Z"}
```
You could provide a setter in addition to or in place of the System Property and update the thread local variable.
```groovy
    public static void setDateFormat(DateFormat dateFormat) {
        dateFormatter.set(dateFormat);
    }
```
This gets around the thread safety issue.  However, it's probably not uncommon to have multiple threads processing JSON payloads concurrently.  With having a static setter any changes affect all threads which may be surprising and with thread local you have to remember to call the setter from each thread which may also be surprising.

benignbala:
Thanks a lot @jwagenleitner , I have implemented your suggestions.

blackdrag:
First of all, sorry, that there has been no communication about this for such a long time. One reason this is open for that long is, because the solution is not really satisfying and a better solution is lacking. I think we could go with the property idea, but only testing once... similar to for example file.encoding. I would not go the route of allowing changes to the format encoding at runtime after first use. I would especially not do anything with ThreadLocal if not absolutely required. If we require a change at runtime I think the category usage might not be the right thing to do for this. Then I think we would need a different solution as a new feature. So my suggestion is to close this request and make a new, simpler one, with only the property for the timezone and one for the format, read on first use.

benignbala:
@blackdrag , Got it - I am pretty to new to contributing to groovy and so was trying to pick easier bugs to fix :) So, thanks for you comments/suggestions. 

Shall I close this one and create another one with just the property for the timezone ? Or will you/other committers be closing this ? Thanks

blackdrag:
It is a bit troublesome to close this issue on our side, because of Apache policies. So I would appriciate you closing this one and make a new pull request

benignbala:
Closed :) thanks for the inputs/suggestions.

PR #125, opened 2015-10-01
Experimental compiler resolving improvement

blackdrag:
Experimental compiler resolving improvement to avoid a myriad of inner class resolves


PascalSchumacher:
I guess this should be closed? (It broke some tests on the team city as far as I remember.)

PR #128, opened 2015-10-06
GROOVY-7546: fix NullPointerException message when trying to add non-…

PascalSchumacher:
…null objects to null

Created out of patch https://issues.apache.org/jira/browse/GROOVY-7546 submitted by Jochen Kemnade


paulk-asert:
+1

PR #136, opened 2015-10-09
Update Groovy Truth doc for arrays

gillius:
Note that non-empty arrays are true, not just collections.


PascalSchumacher:
Thanks!

PR #137, opened 2015-10-09
GROOVY-7623 Use ClassValue by default on IBM Java

candrews:
IBM Java doesn't have the ClassValue garbage collection issue that OpenJDK has (https://bugs.openjdk.java.net/browse/JDK-8136353) so when running on that JVM, enable ClassValue by default.

https://issues.apache.org/jira/browse/GROOVY-7623

### DIFF
--- a/src/main/org/codehaus/groovy/reflection/GroovyClassValueFactory.java
+++ b/src/main/org/codehaus/groovy/reflection/GroovyClassValueFactory.java
@@ -27,8 +27,9 @@ class GroovyClassValueFactory {
 	 * This flag is introduced as a (hopefully) temporary workaround for a JVM bug, that is to say that using
 	 * ClassValue prevents the classes and classloaders from being unloaded.
 	 * See https://bugs.openjdk.java.net/browse/JDK-8136353
+	 * This issue does not exist on IBM Java (J9) so use ClassValue by default on that JVM. 
 	 */
-	private final static boolean USE_CLASSVALUE = Boolean.valueOf(System.getProperty("groovy.use.classvalue", "false"));
+	private final static boolean USE_CLASSVALUE = Boolean.valueOf(System.getProperty("groovy.use.classvalue", "IBM J9 VM".equals(System.getProperty("java.vm.name"))?"true":"false"));
 
 	private static final Constructor groovyClassValueConstructor;
 


melix:
Can you create a separate jira issue for this? We do not reopen bugs to make tracking easier. Also do we have evidence that this works under J9?

candrews:
@melix I've been talking to an IBM employee on the J9 development team, and he tested the issue and says that the ClassValue bug  (https://bugs.openjdk.java.net/browse/JDK-8136353) does not occur on J9.

candrews:
@melix sorry, I didn't mean to reopen an issue to make tracking easier, I figured that another change caused a regression so the issue that regressed should be reopened.
What issues would be helpful to create? One for GROOVY-6704 regressing, and another for IBM J9 working (this pull request)? Or just one issue?

Thanks again!

melix:
> I've been talking to an IBM employee on the J9 development team, and he tested the issue and says that the ClassValue bug (https://bugs.openjdk.java.net/browse/JDK-8136353) does not occur on J9.

Good, I will have to double check before we accept it. As for reopening, I mean that this PR should be associated with its own ticket. If you could create one and reword your commit to include the ticket number it would help, especially because there are bots in actions too, it makes it hard to follow otherwise.

candrews:
@melix I'll ask the IBM fellow to comment here so you can talk to him directly :)

IanPartridge:
Hi - I'm the IBM engineer who has been discussing this with @candrews 

Below is a testcase which I've been investigating.  Running this on Groovy 2.4.5 with IBM's implementation of Java SE 8 shows a memory leak (with -Xmx256m it throws `OutOfMemoryError` in a few minutes).  If I add the system property `-Dgroovy.use.classvalue=true` the memory leak is fixed and the testcase runs indefinitely with no problems.

I think this shows that the `java.lang.ClassValue` code @candrews added for Java 7+ works as designed on IBM's runtime.

Test.java:
```java
package demo;

import groovy.lang.GroovyShell;

public class Test {

		for (int i = 0; i < 10; i++) {
			new Thread(new Runnable() {
				@Override	
				public void run() {
					long i = 0;
					while (true) {
						new GroovyShell().evaluate(i+" * 2", "Script_" + i);
						i++;
					}
				}
			}).start();
		}
				
		while(true);
	}
}
```

PascalSchumacher:
Thanks to everybody who contributed!

PR #138, opened 2015-10-10
syntax documentation: update int + BigDecimal results

gillius:
Per recent discussions on mailing list and experimentation in groovysh
2.4.3, integral types (including BigInteger) with BigDecimal result in
BigDecimal


PascalSchumacher:
Thanks! :+1:

PR #139, opened 2015-10-13
GROOVY-7627 Property calls are not correctly type checked if the sett…

shils:
…er parameter type or getter return type are different from the underlying field type


src/test/groovy/transform/stc/GenericsSTCTest.groovy
```diff
@@ -1613,6 +1613,8 @@ assert result == 'ok'
             class ComponentFixture<T extends Component> {}
             class JButtonFixture extends ComponentFixture<JButton> {}
             class ContainerFixture<T extends Container> extends ComponentFixture<T> {}
+
+            @groovy.transform.TypeChecked(groovy.transform.TypeCheckingMode.SKIP)
```
shils: This class wasn't type checked in the example in the issue (GROOVY-5981), so this should be ok.

blackdrag: you mean the compilation of that class would fail without this?

shils: Yes, I thought it may have been the case that that class should not have compiled with STC in the first place. After you commented, I took another look and realized that it was an actual regression - pseudo property calls within the owning class would break with STC. I've updated my pull request accordingly. 

I also created a simpler unit test for pseudo property calls with STC, but that's probably better off in a separate pr/commit.

shils:
@blackdrag @melix Does this look ok?

PascalSchumacher:
@shils Looks good as far as I can tell. Thanks for all your excellent work on groovy. :+1: Can you rebase this?

shils:
Rebased. Thanks @PascalSchumacher , the feeling is mutual.

shils:
This shouldn't be merged until a decision has been made and carried out for backporting GROOVY-7705.

PR #140, opened 2015-10-14
fix typo

jwilk:


paulk-asert:
applied, thanks! (I ended up changing it further but your attentiveness to detail triggered the change!)

PR #141, opened 2015-10-14
Groovy-7574: Ranges: Make Range members final. Also improve javadoc.

tkruse:
Else they can easily be changed in Groovy, which leads to undefined states

supercedes #109


paulk-asert:
merged, thanks

PR #142, opened 2015-10-14
Groovy-7585 improve behavior of ObjectRange

tkruse:
supercedes #109

PR #143, opened 2015-10-14
GROOVY-7629 ObjectRange duplicate functionality and bad reliance on size()

tkruse:


tkruse:
This also contains an improvement to subList, such that elements are not stepped over twice unnecessarily

tkruse:
One thing to discuss is whether the iterator() method should return a thread-safe operator, or leave that to clients. I believe the existing implementation was not thread-safe, so I am also happy to remove the synchronized wrapper from the PR.

PascalSchumacher:
Thanks for pull request. :) Can you rebase it? Thanks!

tkruse:
Rebased. This PR is bound to strongly conflict with #120 in git. So whichever PR goes second will have to dosome conflict resolving.

paulk-asert:
OK, I merged the StepIterator commit. The tests were already covered off by PR #142. I didn't take the size() change to return Integer.MAX_VALUE on overflow. I am -1 on that change at the moment.

Although a breaking change, I would be in favor of failing fast when overflow is encountered for the Number cases (during construction) and throwing an IllegalStateException if overflow was encountered for the brute force cases where deferred calculation is required. I think we should add a BigRange or similar class that doesn't implement List and handles the size() properly - perhaps even stream-based. But that's for another time/issue/PR.

tkruse:
The javadoc of java.util.List says:
"int size():
Returns the number of elements in this list. If this list contains more than Integer.MAX_VALUE elements, returns Integer.MAX_VALUE."

For java, it is valid for a class to implement List and contain more than MAX_VALUE items. Throwing IllegalStateException in the constructor or in size() might be better in some respects (fail fast), but would violate that contract, and User expectations.

paulk-asert:
Hmm. Yes, you are right. I just did some more analysis of the Java behavior and I think what you proposed is the right way to go - though we should have a BigRange class too down the track I believe.

paulk-asert:
Additional commit merged. Thanks.

PR #145, opened 2015-10-18
GROOVY-5380: Add Ctrl+L/Ctrl+B for line/block selection

jwagenleitner:
Implemented Ctrl+B (Select Block) to handle selecting a word, line and block.  If nothing is currently selected the action will select the next chunk unless at the end of the line then the previous chunk will be selected.  If there is already a partial selection on a line, then the entire line will be selected.  If an entire line or multiple lines are already selected then the block/paragraph will be selected.


PascalSchumacher:
Very nice addition to the Groovy Console. :+1: Thanks!  :)

PR #149, opened 2015-10-19
fix after GROOVY-7574: Always do null checks in ObjectRange constructor

tkruse:
This is a small fix after the recently merged #141 moving null-checks. I hope it is small enough to be quickly merged.


PascalSchumacher:
Thanks! :+1:

PR #150, opened 2015-10-19
fix GROOVY-7637 issues around Number Comparisons

tkruse:


tkruse:
Edit: moved commit for GROOVY-7636 to #151

paulk-asert:
I merged this but kept the behavior from master of throwing CCE in a place you were throwing a GRE. I changed the test accordingly. Let me know if you think this is okay from your side. I'll look at PR#151 next.

tkruse:
I cannot remember if there was a specific reason for me to choose GRE at the time. Probably should have commented.

I guess at the time I noticed the default CCE gives no useful information, so my intent was probably to provide the user with more details of the failure. At the time I must have experimented a lot with corner-case ranges and noticed the default CCE does not contain enough information.
I understand now that this is also an API change and there are performance implications in creating new Exceptions.

So relying on the default CCE should be better from a technical perspective, even if the generated message is sadly not very useful.

tkruse:
Actually, reading the JIRA issue, I remember this was more than just about the message details. I considered the throwing of a CCE a glitch, meaning un undesired undocumented leaking Exception that confronts Users with the internals of Groovy.

This test code shows the problem:

    shouldFail(ClassCastException) {
            compareTo(1, "22")
    }

The question here is whether Groovy as a language **wants** to throw a CCE when the user tries to compare a number and that String, or whether Groovy wants to throw a GRE, as it does for several other error cases. The problem is that the user does not know or care about types he passes to compareTo(), since compareTo is so mighty flexible in comparing apples and oranges in many cases. A CCE should possibly only be thrown if the documentation clearly states that a cast will happen, of that such an Exception will be thrown. Also a CCE may look like a Groovy bug to many users (missing instanceof check before a cast, bad type handling). A GRE looks more like intended behavior.

I think I presumed that the original author of the code had overseen the possibility of a CCE here and did not wrap it out of oversight.

I am somewhat dispassionate about keeping the CCE or using a GRE in the future. I believe semantically the CCE is not desirable for users, both in terms of the message and in terms of the exception type. But I may be wrong, users might actually want to compare all items in a set to check that they are compareTo() compatible, and rely on the CCE to detect pairs that are not compatible.

So with that explanation, I'll let you decide.

tkruse:
for completeness sake: Using an IllegalArgumentException could be a viable third alternative to CCE and GRE, with reasonably semantics.

paulk-asert:
I think a more informative exception would be preferred (probably a GRE). I'll think about making such a change and it's impacts.

paulk-asert:
Proposed additional changes:
https://github.com/apache/groovy/pull/372

PR #152, opened 2015-10-21
GROOVY-7640 fixes

Controlix:
Here is a test that demonstrates the bug and the fixes for Default-, Initializer-, and ExternalStrategy.
I'm still working on the fix for the SimpleStrategy and I might enhance the tests for all strategies.


keeganwitt:
Please rebase rather than merge upstream commits, it keeps the commit log cleaner.

paulk-asert:
The proposed code changes look good. However, for consistency with several other AST transforms, I think we should have an `includeSuperProperties` flag to turn on this behavior. Also, we need to update the `Builder` section of the `core-metaprogramming.adoc` file with doco/examples.

PascalSchumacher:
@paulk-asert Since this pull request seems to be abandoned, I took a look at making the changes you suggested.

It seems like it is not possible to add this feature to the `SimpleStrategy`. When you replace `getInstancePropertyFields(buildee)` with `getSuperPropertyFields(buildee)` the test for the simple strategy still fails with  `The return type of Person setAge(int) in Person is incompatible with void in Mamal`. 

As `SimpleStrategy` does not support most of the other configuration options, I guess this is o.k.? What do you think?

PascalSchumacher:
I merged the pull request and did the updates Paul requested. Thanks! :+1:

PR #155, opened 2015-10-22
Fixed the typo in the documentation on http://www.groovy-lang.org/processing-xml.html page.

rohitbishnoi:
update the line "As you can see there are to types of notations to get attributes"
to "As you can see there are two types of notations to get attributes"

screenshot from 2015-10-22 10-41-33


paulk-asert:
merged, thanks

PR #158, opened 2015-10-23
GROOVY-7363 Reliable type resolution in the static type checker with covariant getters

fpavageau:


PascalSchumacher:
@blackdrag @melix Is this ready to merge?

blackdrag:
I am thinking the patch does not really address the problem. The order must not matter. If it does, then there is a bug in the code that works only with a certain order. And that's the part that needs fixing with either a sorting there or not depending on sorting at all.  Furthermore, the patch addresses code of precompiled classes generated from a class object. We have also ClassNodes objects from code currently under compilation and from classes stored in jars and read by asm. Nothing in this patch asserts the order in those cases, and I think there should not be as well. As the patch is now I must fear, that if the classes inside of Groovy7363Support or moved into the assertScript part, that the test will fail again. So -1 on merging from me in the current state

fpavageau:
OK, I'll move the fix to the method resolution instead of the `ClassNode` configuration.

fpavageau:
I pushed again, moving the fix to `ClassNode.getGetterMethod()` so it doesn't pick the first matching method that comes along, but the "best" one independently of the order, as suggested by @blackdrag.

PascalSchumacher:
Thanks, looks much better to me. @blackdrag what do you think?

blackdrag:
I think we can go with that +1

PascalSchumacher:
Thanks! :+1: 

@blackdrag Thanks for the reviews. :)

PR #160, opened 2015-10-25
GROOVY-4888 - groovy console script execution output appears in all o…

jwagenleitner:
…utput windows

PR #161, opened 2015-10-26
GROOVY-4888 - fix so child threads output/err streams go to correct console

jwagenleitner:
@PascalSchumacher sorry for the extra trouble, I forgot to include this minor change in the original PR #160.  This makes it so that any new Thread's started in the console script itself output stdout/stderr to the correct console output area.


PascalSchumacher:
No problem :) Thanks! :+1:

PR #162, opened 2015-10-26
GROOVY-7616: Fix 'Illegal class name' when collectEntries sees an array

rahulsom:


paulk-asert:
There are two parts to this issue. The first is that ProxyGeneratorAdapter isn't mangling the generated class name correctly in the case of arrays. The second is that getKey doesn't handle arrays (which this PR fixes). I am going to split the original Jira issue to cover the two cases.

paulk-asert:
merged - thanks!

PR #164, opened 2015-10-28
GROOVY-7649: clean up current dependencies state when resolve fails

pluradj:
https://issues.apache.org/jira/browse/GROOVY-7649


PascalSchumacher:
Thanks! :+1:

PR #167, opened 2015-10-29
GROOVY-7639 Comparing primitive to null safe primitive property expre…

shils:
…ssion should not throw NPE

* Correct assertions in tests for GROOVY-5613


aaronzirbes:
:+1:

shils:
Merged, thanks for taking a look.

PR #169, opened 2015-10-31
Non-atomic use of atomicCategoryUsageCounter

ksuderman:
The atomicCategoryUsageCounter was Incremented/decremented on one line and and then read on the next line leaving a small window for race conditions to happen.  This is a possible fix for https://issues.apache.org/jira/browse/GROOVY-7535


PascalSchumacher:
Thanks! :+1:

PR #172, opened 2015-11-01
Groovysh: refactor: Tests robust against race conditions on Preferences

tkruse:
as mentioned on mailing list topic "GroovyshInterpreterModeTest Failures"

PR #173, opened 2015-11-01
fix after #172, Interpretermode tests also need to run all other tests

tkruse:
Sorry, I did not think that through. #172 is bad because it makes many tests from parent test class run without proper configuration.


PascalSchumacher:
Test failed. Did you get a automatic notification for the failures?

tkruse:
No, I get no mail notifications. Working on fixing...

PascalSchumacher:
two fixed, two to go...

PascalSchumacher:
Thanks! :)

PR #175, opened 2015-11-02
GROOVY-4847 - Missing GDK rounding (and trunc) methods for BigDecimal

jwagenleitner:
I tried to address the issues raised by @blackdrag in the original groovy/groovy-core#281 PR.  The `round` methods may be slightly confusing since the `BigDecimal#round(MathContext)` method works on  the entire number, so I tried to call that out in the javadoc.


PascalSchumacher:
Thanks! :+1:

PR #176, opened 2015-11-03
GROOVY-7465: ResourceGroovyMethods/NioGroovyMethods BOM behavior is inconsistent

keeganwitt:


keeganwitt:
I decided to treat the consumption of BOM as a separate issue and have opened [GROOVY-7661](https://issues.apache.org/jira/browse/GROOVY-7661) to track that.

src/main/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
```diff
@@ -831,9 +836,8 @@ public static File leftShift(File file, InputStream data) throws IOException {
     }
 
     /**
-     * Write the text to the File, using the specified encoding.  If the given
-     * charset is "UTF-16BE" or "UTF-16LE" (or an equivalent alias), the
-     * requisite byte order mark is written to the file before the text.
+     * Write the text without a BOM to the File without writing a BOM, using
```
PascalSchumacher: This sentence seems a bit confusing to me.

src/main/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
```diff
@@ -1800,6 +1925,20 @@ public static BufferedWriter newWriter(File file, boolean append) throws IOExcep
...
 
     /**
+     * Helper method to create a buffered writer for a file without writing a BOM.
+     *
+     * @param file    a File
+     * @param charset the name of the encoding used to write in this file
+     * @param append  true if in append mode
+     * @return a BufferedWriter
+     * @throws IOException if an IOException occurs.
+     * @since 1.0
```
PascalSchumacher: should be 2.5.0

src/main/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
```diff
@@ -974,11 +1021,30 @@ public static void append(File self, InputStream stream) throws IOException {
...
+     * @since 2.5.0
+     */
+    public static void append(File file, Object text, String charset, boolean writeBom) throws IOException {
         Writer writer = null;
         try {
+            boolean shouldWriteBom = writeBom && !file.exists();
             FileOutputStream out = new FileOutputStream(file, true);
-            if (!file.exists()) {
-                writeUTF16BomIfRequired(charset, out);
+            if (shouldWriteBom) {
```
PascalSchumacher: Shouldn't we keep the `!file.exists()` check?

PascalSchumacher:
I think the writeUTF16BomIfRequired() methods have the writer/outputstream as the first parameter. What do you think?

subprojects/groovy-nio/src/main/java/org/codehaus/groovy/runtime/NioGroovyMethods.java
```diff
@@ -1540,12 +1709,33 @@ public static BufferedWriter newWriter(Path self, String charset) throws IOExcep
...
+     * <code>writeBom</code> is <code>true</code>, and the file doesn't already exist,
+     * the requisite byte order mark is written to the stream when the writer is created.
+     *
+     * @param self    a Path
+     * @param charset the charset used
+     * @param writeBom whether to write the BOM
+     * @param closure a closure
+     * @return the value returned by the closure
+     * @throws java.io.IOException if an IOException occurs.
+     * @since 2.3.0
```
PascalSchumacher: should be 2.5.0

PascalSchumacher:
great work :+1:

src/main/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
keeganwitt: Oops, copy/paste error.  Good catch.

src/main/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
keeganwitt: The check is still there, I just pulled it out into a separate variable (`shouldWriteBom`).  I could do it without a variable, I only use it once so it's not really necessary.

subprojects/groovy-nio/src/main/java/org/codehaus/groovy/runtime/NioGroovyMethods.java
keeganwitt: Yep, you're right.

keeganwitt:
Yea, I like the idea of switching the parameter order.  Thanks for the suggestion.

src/main/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
PascalSchumacher: I misread that. :( Sorry.

PascalSchumacher:
I guess this ready to merge?

keeganwitt:
The only other question I had was on `FileOutputTool`'s use of BOM.  The change I made was to preserve existing behavior, so I guess should be OK -- just wasn't sure why it was done that way.

PR #178, opened 2015-11-04
GROOVY-7660: gradle/utils.gradle: Remove unused imports and unused variables. 

monsieurp:
This pull request remove unused imports and unused variables in gradle/utils.gradle. Also, instead of importing the whole namespace containing all ASM Opscode, we specify them one by one.


PascalSchumacher:
Thanks!

PR #179, opened 2015-11-04
GROOVY-7656 Statically compiled spread safe method calls on list lite…

shils:
…rals result in the list expression being evaluated twice


blackdrag:
I think that is ok to merge. Frankly this part of the code cries for removal of code duplication in my opinion, but that is unrelated to the issue

shils:
Agreed, and there's also the usages of declared "spreadresult" variables that should be replaced with temporary variables. Thanks for reviewing.

Closing this manually as I forgot to amend my commit message.

PR #181, opened 2015-11-07
GROOVY-2178 - Shell can not handle multi-line list defs

jwagenleitner:
To address @blackdrag's comments on the issue, there was already some code that handles [detecting incomplete lines] (https://github.com/apache/incubator-groovy/blob/1eb37ea96428e14f0418fa003292cce9d3f94f94/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/Parser.groovy#L204-L212) and it looks like it captures all the chars referenced in the comment.  I wasn't able to replicate the problem with opened braces `{` so am just trying to detect (if compilation error is thrown) if there's an unmatched bracket `[` or paren `(`.


PascalSchumacher:
Thanks! :+1: Nice to have this issue fixed after all those years. :)

PR #182, opened 2015-11-08
Fix typo: Missing trailing 's' in design pattern section

DarthPumpkin:
*decorator* is singular, thus a trailing *s* is needed.


PascalSchumacher:
Thanks!

PR #183, opened 2015-11-09
GROOVY-7665 - JsonSlurperCharSource decimal parsing producing unexpected results

jwagenleitner:
Commit 842bcd4e23fefdf5b5e17d8d91af50e1ce565096 introduced the `BigDecimal` return types and reused some of the original parsing methods.  However, in this case I think it makes more sense to just let `BigDecimal` parse the text since converting first to a double can cause some odd results.


PascalSchumacher:
Thanks! :+1:

PR #184, opened 2015-11-09
GROOVY-6922 - JsonSlurper loses trailing 0's in numbers

jwagenleitner:
This change is dependent on the change made in commit 84467fddc64744cc05d2b91b8704c753a9fb3d04 in master to the [CharScanner#parseBigDecimal(char[],int,int)] (https://github.com/apache/incubator-groovy/commit/84467fddc64744cc05d2b91b8704c753a9fb3d04#diff-d0df59d6b1bac924a44389e6b41aac99R687) method.

The code being replaced looked like it was code left over from when the method returned a double for decimal values.  This change uses the [BigDecimal(char[],int,int)] (http://docs.oracle.com/javase/7/docs/api/java/math/BigDecimal.html#BigDecimal(char[],%20int,%20int)) constructor (available since 1.5) for parsing of the decimal string and also eliminates the new String object allocations.

### DIFF
--- a/subprojects/groovy-json/src/main/java/groovy/json/internal/CharScanner.java
+++ b/subprojects/groovy-json/src/main/java/groovy/json/internal/CharScanner.java
@@ -653,23 +653,8 @@ public class CharScanner {
             } else {
                 value = parseLongFromTo(buffer, from, index);
             }
-        } else if (foundDot && simple) {
-            BigDecimal lvalue;
-
-            if (length < powersOf10.length) {
-                if (isInteger(buffer, from, length)) {
-                    lvalue = new BigDecimal(parseIntFromToIgnoreDot(buffer, from, index));
-                } else {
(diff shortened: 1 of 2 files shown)


PascalSchumacher:
Great work! Thanks a lot. :)

Just  a small nitpick: In groovy the assert statement should be `assert actual == expected` (I also get confused from time to time because the order is the opposite of JUnits `assertX(expected, actual)`).

jwagenleitner:
Hi @PascalSchumacher, thanks for the reminder, I'm a little too used to the junit style.  I'd be happy to make the changes to this commit, just not sure if I should rebase the change or add on as an additional commit.

PascalSchumacher:
Hi @jwagenleitner, please rebase (we prefer the changes done of an an issue to be a single commit). Thanks!

jwagenleitner:
Fixed the assert argument order, thanks again for the feedback/review @PascalSchumacher

PascalSchumacher:
Thanks! :)

PR #185, opened 2015-11-10
GROOVY-7668 proposed improvement

sbglasius:


blackdrag:
Looks like MalformedParameterizedTypeException has no String constructor. I suggest wrapping in for example GroovyRuntimeException instead (can't think of a better one right now... I would normally suggest ReflectiveOperationException, but that is Java7 only).

sbglasius:
Thanks, I will update the PR

PascalSchumacher:
Merged: 3b85e8610fa2976dc4430f1930e6a1369a69d016 Thanks!

It would be nice if you could close the pull request (I forgot to add a comment when cherry-picking and we can not do it by hand because this is a read only-mirror.) Thanks!

PR #187, opened 2015-11-12
GROOVY-7672: TupleConstructor should have only SOURCE retention policy

paulk-asert:


src/test/org/codehaus/groovy/transform/TupleConstructorTransformTest.groovy
```diff
@@ -131,10 +131,37 @@ class TupleConstructorTransformTest extends GroovyShellTestCase {
...
+            import groovy.transform.*
+            import java.awt.Color
+
+            class Named {
+                String name
+            }
+
+            @ToString(includeSuperProperties=true, ignoreNulls=true, includeNames=true, includeFields=true)
+            @TupleConstructor(force=true, defaults=false)
+            @TupleConstructor(force=true, defaults=false, includeFields=true)
```
shils: I think it would be helpful to include a multiple usage example like this in the documentation.

paulk-asert:
Good point, I'll add some doco.

paulk-asert:
Added

PR #188, opened 2015-11-14
JSON CharScanner cleanup

jwagenleitner:
While looking into some other issues I noticed quite a few unused methods in the `CharScanner` class.  Even though the class and it's methods are public, since it's in an `internal` named package it seems that it would be ok to treat it as non-public facing API ([I asked on the dev list] (http://mail-archives.apache.org/mod_mbox/incubator-groovy-dev/201511.mbox/%3CCAJ2HCd-ZcuZ%2BD6WUjjVAat8thEmibzW-9y1MchPz3cRR-yVdJg%40mail.gmail.com%3E)).  

Changes:

1. Removed all methods (and those that were left unused after their removal) that were not referenced anywhere in the entire project or were only referenced in the `CharScannerTest` class as the method under test.
1. Removed unused class variables
1. Removed duplicate code in `parseIntFromTo`, both branches of `if` had same code
1. Added additional tests for `parseIntFromTo` to make sure changes above were ok
1. Removed unneeded `final` modifier from static method
1. Removed `== true` in boolean checks
1. Removed one and only call to `#parseBigDecimal(char[], int, int)` in order to remove it since was only referenced in same class and `BigDecimal` constructor offers the same signature.

Besides ensuring a `./gradlew clean test` passed I also did some testing against various large json payloads (https://github.com/jwagenleitner/groovy-json-tests) to make they all parsed successfully.

I know it's a lot of change (deletions) but wanted to offer the PR for consideration.


PascalSchumacher:
Thanks! :+1: 

I'm a big proponent of removing unused/unnecessary code whenever possible.

PR #189, opened 2015-11-15
GROOVY-7630 - JsonSlurper LAX parser with invalid number

jwagenleitner:
Breaking changes proposed here, but current behavior appears incorrect even in terms of relaxed parsing.  In addition to addressing GROOVY-7630, also found some missing cases around decimal parsing which were partially addressed by commit 8063a9dec07329804f3e1770aa31984bc38a694d for GROOVY-7344.

There seems to be quite a bit of duplication between the LAX and INDEX_OVERLAY parsers that I didn't address in this PR (actually adding more duplication :smile:).  I didn't want to introduce too much change, but at some point would like to come back and remove the duplication.

### DIFF
--- a/subprojects/groovy-json/src/main/java/groovy/json/internal/JsonFastParser.java
+++ b/subprojects/groovy-json/src/main/java/groovy/json/internal/JsonFastParser.java
@@ -163,6 +163,8 @@ public class JsonFastParser extends JsonParserCharArray {
         char currentChar;
         boolean doubleFloat = false;
         boolean foundDot = false;
+        boolean foundSign = false;
+        boolean foundExp = false;
 
         if (minus && index + 1 < array.length) {
             index++;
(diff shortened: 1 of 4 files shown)


PascalSchumacher:
Hi John,

What exactly are the breaking changes? That some incorrect numbers are not silently parsed to a number?

jwagenleitner:
Hi Pascal, 

Yes, one change would be that invalid numbers integers will now fail on parse instead of being parsed to some value (`1a` == `59`).

Second change is that invalid decimal values will now fail on parse instead of on access.  This change was already in place for extra decimal places (GROOVY-7344) so this just looks for the other cases which are probably not very likely (extra e/E or -/+) but thought should be added for completeness.  Prior to this change parse would be successful but these would throw exceptions on access since they are lazily evaluated as BigDecimals.

I think the changes for the INDEX_OVERLAY parser definitely make sense, I'm a little less sure about LAX.  I guess the argument could be made that for LAX failing on access would be better than failing on parse.

PascalSchumacher:
Thanks for the info. :)

Any idea how other JSON parsers handle the LAX parsing?

jwagenleitner:
Good question, I'm not sure how the other parsing libraries handle invalid numbers in their relaxed modes.  I'll try to do some quick testing of the relaxed parsing in some libraries like Jackson and Gson in the next couple of days and will report back.

jwagenleitner:
I did some very primitive testing and posted my results on the JIRA ticket.

PascalSchumacher:
+1, as you a committer now, you can merge it yourself :)

jwagenleitner:
Gave it some more thought and I agree that LAX should probably be forgiving and treat invalid numbers as a string.  Any feedback on it would be most welcomed.

Also, I think this qualifies as a bugfix and I misspoke when I said it contains breaking changes.

PascalSchumacher:
>Gave it some more thought and I agree that LAX should probably be forgiving and treat invalid >numbers as a string. Any feedback on it would be most welcomed.

I'm not JSON expert, but I'm fine with this.

>Also, I think this qualifies as a bugfix and I misspoke when I said it contains breaking changes.

I agree. I think it's fine to merge into the 2_4_X branch. :)

PR #193, opened 2015-11-22
Default value of includePackage is True

anshbansal:
The default value of includePackage is True not false


PascalSchumacher:
Thanks! :+1:

PR #194, opened 2015-11-24
GROOVY-7342 Last enum value is hidden if annotated

shils:
* Try syntactic predicate for enum constant before predicate for end of enum constants


paulk-asert:
+1

PR #196, opened 2015-11-25
GROOVY-7538 Resolve placeholder return types correctly

fpavageau:
This solves the first half of [GROOVY-7538](https://issues.apache.org/jira/browse/GROOVY-7538).

I've tried to replace the test in `Groovy7538Bug` by a new one in `GenericsSTCTest`, but ran into another issue while trying to implement the test purely in Groovy (I'll open a new ticket on https://issues.apache.org/).


fpavageau:
I'm closing the PR, as my new fix for the other test case seems to actually fix both at the same time *and* is much, much simpler. I need a bit more time to make sure it really checks out.

PR #197, opened 2015-11-26
GROOVY-7538 Bound unbounded wildcards using the type declaration

fpavageau:
If there is a type declared with bounds, for example

    class NumberList<T extends Number> implements List<T>

but used with an unbounded wildcard

    NumberList<?> getList()

the bounds are actually implicit and need to be taken into account for the
static type checks, to be able to use the type of the parameter directly:

    getList()[0].longValue()

The type of the element is known to be `Number`, not `Object`.


fpavageau:
As suggested by @melix (on a previous version of the commit I overwrote with a force push), the test case continues to implement `StaticCompilationTestSupport` to make sure it not only passes the static type checks but also compiles statically.

PascalSchumacher:
Thanks! :+1:

PR #202, opened 2015-12-02
Remove incubator from the URL of the repositories

fpavageau:
I've also changed the Apache URL to use HTTPS instead of HTTP.

Note that the github mirror still says
> mirrored from git://git.apache.org/**incubator-**groovy.git


paulk-asert:
looks good - I did just try merging it but it looks like some permissions for repo authorisation on the new repo are still pending

PR #204, opened 2015-12-03
GROOVY-7420 Prefer a method signature without boxing or unboxing

fpavageau:
The method resolution mechanism is closer to what is done when following the
Java Language Specification ([JLS 15.2.2](http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.12.2)), and removes the ambiguity when
calling a method having both primitive and object overloads with an object.

### DIFF
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -816,7 +816,9 @@ public abstract class StaticTypeCheckingSupport {
                 && unwrapReceiver!=unwrapCompare) {
             dist = getPrimitiveDistance(unwrapReceiver, unwrapCompare);
         }
-        if (isPrimitiveType(receiver) && !isPrimitiveType(compare)) {
+        // Add a penalty against boxing or unboxing, to get a resolution similar to JLS 15.12.2
+        // (http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.12.2).
+        if (isPrimitiveType(receiver) ^ isPrimitiveType(compare)) {
             dist = (dist+1)<<1;
         }
         if (unwrapCompare.equals(unwrapReceiver)) return dist;
--- a/src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy7420Bug.groovy
+++ b/src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy7420Bug.groovy
@@ -18,12 +18,10 @@
  */
 package org.codehaus.groovy.classgen.asm.sc.bugs
 
-import groovy.transform.NotYetImplemented
 import groovy.transform.stc.StaticTypeCheckingTestCase
 import org.codehaus.groovy.classgen.asm.sc.StaticCompilationTestSupport
 
 class Groovy7420Bug extends StaticTypeCheckingTestCase implements StaticCompilationTestSupport {
-    @NotYetImplemented
     void testOverloadedMethodWithPrimitiveOrObjectParameter() {
         assertScript '''
             class A {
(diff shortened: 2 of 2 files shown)


PascalSchumacher:
@blackdrag @melix What do you think, is this ready to merge?

blackdrag:
I think the method signature change is ok. But I miss a test case for GROOVY-7420. As for the change itself... Not sure I feel too happy about diverging so much from the method resolution algorithm we use at runtime.  The distance calculation was supposed to catch the cases you have fixed here. Also with the advantage not to have to select the method multiple times. On the other hand I think the change will work without causing too much divergence. So all in all I am ok with the change

fpavageau:
I see your point. I can have a second look to see if I can find a way to plug into the parameters/arguments distance calculation to bias against boxing/unboxing by adding a penalty. It did not seem possible the first time, and I didn't really want to impact the other callers of the modified methods (which is why the original method is overloaded with another one adding the `mode` parameter, and calls it with the same mode as before).

Regarding the test cases, I'm not sure what test you feel is missing. There was [already one](https://github.com/fpavageau/groovy/commit/3583d6d95d664dcfc91402ec324905cf731cd1bd#diff-f6d8ec081f0cf2fcc6dfe44644e336ac) with a `@NotYetImplemented` annotation which is now removed, completed to test both primitive and object calls, and I also [added another test](https://github.com/fpavageau/groovy/blob/3583d6d95d664dcfc91402ec324905cf731cd1bd/src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy7420Bug.groovy#L43) to catch a failure I had while running the complete "clean test" build. Do you want more tests to cover the other branches of the method resolution (the varargs cases), maybe also test what happens when the argument is of an narrower primitive type, like `int`? I'll be happy to add some more.

blackdrag:
adding a penalty is what the runtime algorithm does. But I really let you decide. I find it more important having someone how is looking at those sometimes very difficult code parts and fixes some things, than to get the perfect solution.  As for the tests, I was actually not seeing that part before, because I looked only at the big file sorry.So I take the part about missing tests back. 

In total I think the change can be merged now

fpavageau:
You know what? Based on your comment, I've revisited the parameters/arguments calculation (I've spent more time since that PR in the vicinity, so I'm getting to know it better) and I actually have a one liner fix now which passes the complete "clean test" build.

The distance calculation *did* try to catch the case I fixed, except it only penalized the unboxing case instead of both boxing and unboxing.

So please, don't merge yet, I'll push the revised fix later.

fpavageau:
OK, I've pushed this much nicer and simpler fix, thanks @blackdrag for the prod.

Now if the JVM wouldn't crash when trying to build the PR in Jenkins... :)

blackdrag:
the patch looks much better now ;) As for Jenkins... this actually looks like a JVM bug to me. I have seen this kind of error before. I cannot explain why your code triggers it at all. Memory corruption usually may not even tied to the place it happens. And I bet you cannot reproduce the crash locally either. A little change somewhere could be enough to avoid the problem... maybe you want to write an additional test (you could try all primitives for example) and see if that helps

fpavageau:
Well, no luck with a new push adding an extra assertion to the test.

All the builds after [127](https://builds.apache.org/job/Groovy%20Github%20PRs/127/consoleFull) have failed (127 did a fresh install of the JDK and Gradle), and they fail right when the JVM starts since there are no Gradle logs at all, contrary to what can be [seen in 127](https://builds.apache.org/job/Groovy%20Github%20PRs/127/consoleFull#gradle-task-0). The container/VM/slave is probably running short on memory and can't properly start the JVM.

shils:
+1 after a CI build

fpavageau:
I force-pushed a new version of the same commit, which passed this time. Somebody's fixed Jenkins :).

PascalSchumacher:
Thanks! :+1:

PR #205, opened 2015-12-06
GROOVY-7699: collate() on an empty list is an empty list.

rahulsom:
I think GROOVY-7699 should be fixed, but I noticed it's by design that it returns a list of one empty list.

I committed code from IDEA, and it removed trailing spaces from lines. If that's a problem, I'll close this and resend a PR.


paulk-asert:
PR merged, thanks!

PR #203, opened 2015-12-02
Support for Array and Iterable not just Collection in JsonBuilder

graemerocher:
StreamingJsonBuilder currently only supports `Collection` types but not arrays and iterables. This pull request corrects that also improves compatibility with `@CompileStatic`

PR #206, opened 2015-12-06
GROOVY-7703 - Add support for Iterable with Closure to JsonBuilder/StreamingJsonBuilder

jwagenleitner:
This applies similar changes from PR #203 to the JsonBuilder (non-Streaming) class.


PascalSchumacher:
:+1:

PR #207, opened 2015-12-09
GROOVY-6958 null != NullObject when using categories

shils:


shils:
Merged, thanks for taking a look.

PR #208, opened 2015-12-09
StreamingJsonBuilder should handle nested closures

graemerocher:
Currently StreamingJsonBuilder doesn't handle nested closures properly and instead these are delegated to `JsonOutput.toJson` which is not streaming and leads to an unexpected change of delegate in the nested closure. This pull request fixes that erroneous behaviour.


shils:
+1

PascalSchumacher:
Thanks for the fix. :)

Could you please create JIRA issues for any fixes/enhancements (that makes tracking for release notes etc. much easier)? Thanks!

graemerocher:
Issue https://issues.apache.org/jira/browse/GROOVY-7706

jwagenleitner:
Merged, thanks!  Just waiting for the asfgit bot to bring over the changes to the github mirror.

PR #210, opened 2015-12-10
StreamingJsonBuilder - fix IllegalStateException when writing unescaped output

graemerocher:
When writing unescaped JSON the state is not reset, resulting in an exception if another entry is written.


graemerocher:
Issue https://issues.apache.org/jira/browse/GROOVY-7707

jwagenleitner:
Merged, thanks!

PR #213, opened 2015-12-16
GROOVY-7710 GROOVY-7711 Apply the rules to resolve method consistently

fpavageau:
A no-arg call to a varargs method needs to be resolved with the same rules
as when there are exactly the same number of parameters, or even more, by
applying the class distance and extension method biases.


PascalSchumacher:
Build failed because build-server ran out of space `stderr: /home/jenkins/jenkins-slave/workspace/Groovy Github PRs/.git: No space left on device`, that's unrelated for sure. ;)

PR looks good as far as I can tell. What do you think @blackdrag @melix @shils ?

fpavageau:
I've force-pushed to trigger a new build, since the problem of the full disk seems to have been resolved.

PascalSchumacher:
Thanks! :+1:

PR #214, opened 2015-12-16
Clean up in the STC classes

fpavageau:
It's mainly based on IntelliJ IDEA inspections.

### DIFF
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -1574,7 +1574,7 @@ public abstract class StaticTypeCheckingSupport {
         if (orig.isGenericsPlaceHolder()!=copy.isGenericsPlaceHolder()) return false;
         if (!orig.equals(copy)) return false;
         GenericsType[] gt1 = orig.getGenericsTypes();
-        GenericsType[] gt2 = orig.getGenericsTypes();
+        GenericsType[] gt2 = copy.getGenericsTypes();
         if ((gt1==null || gt2==null) && gt1!=gt2) return false;
         if (gt1==gt2) return true;
         if (gt1.length!=gt2.length) return false;
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -1311,7 +1311,6 @@ public abstract class StaticTypeCheckingSupport {
             return typeCheckMethodsWithGenerics(receiver.getGenericsTypes()[0].getType(), arguments, candidateMethod);
         }
         // both candidate method and receiver have generic information so a check is possible
-        Parameter[] parameters = candidateMethod.getParameters();
         GenericsType[] genericsTypes = candidateMethod.getGenericsTypes();
         boolean methodUsesGenerics = (genericsTypes != null && genericsTypes.length > 0);
         boolean isExtensionMethod = candidateMethod instanceof ExtensionMethodNode;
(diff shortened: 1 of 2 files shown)


src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1574,9 +1560,9 @@ private static boolean equalIncludingGenerics(ClassNode orig, ClassNode copy) {
         if (orig.isGenericsPlaceHolder()!=copy.isGenericsPlaceHolder()) return false;
         if (!orig.equals(copy)) return false;
         GenericsType[] gt1 = orig.getGenericsTypes();
-        GenericsType[] gt2 = orig.getGenericsTypes();
-        if ((gt1==null || gt2==null) && gt1!=gt2) return false;
+        GenericsType[] gt2 = copy.getGenericsTypes();
         if (gt1==gt2) return true;
+        if (gt1 == null || gt2 == null) return false;
```
blackdrag: I personally don`t like this reordering of checks here. For me the intend more clear as it was before (of course minus  the orig/copy bug)

src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -171,7 +171,7 @@
 
     protected final ReturnAdder.ReturnStatementListener returnListener = new ReturnAdder.ReturnStatementListener() {
         public void returnStatementAdded(final ReturnStatement returnStatement) {
-            ClassNode returnType = checkReturnType(returnStatement);
+            checkReturnType(returnStatement);
```
blackdrag: while true, that the variable is unused... for debugging purposes it helps a lot to have that information there

blackdrag:
In general what I don`t like so much is having a having method call or complex expression in a return statement. That makes debugging much more difficult than needed in my eyes. Same for the cases in which you go through a list of checks, then I won`t let the last check just return using an expression. There are multiple cases like those in this pull request. What I mean is that we can do all those changes, but frankly I don`t think that most of them really improve the code quality or readability of the code. The loop changes for example and the one generics change are fine. But making return statements more complex.... don`t really like that I must say. The cases of local variables before branches and them having a value set, that is then also set in all cases of the branches... that change is ok too for me

fpavageau:
I have no problem removing the changes on the return statements, it's indeed more a matter of taste and personal workflow for debugging (I don't mind stepping inside methods, but that's me), and I certainly don't want to upset other people's workflows.

blackdrag:
don't worry about upsetting my workflow ;) If you have new code that does this, no problem. For existing code... well, I would prefer to keep the status quo in that case and do such changes really only if it is more than just a matter of taste ;)

src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
fpavageau: Ah well, the intent was actually more obscure for me in the original. If you want a more expressive statement, I'd have found it clearer written like:

    if ((gt1 == null && gt2 != null) || (gt1 != null && gt2 == null)) return false;

or

    if ((gt1 == null) ^ (gt2 == null)) return false;

Still, the `null == null` case is handled by `gt1 == gt2` in any case, whether it comes first or second.

I'm not especially attached to the modification, though.

blackdrag: if you want to change it to the xor version you are welcome

PascalSchumacher:
Merged the commits! Thanks! :+1:

fpavageau:
Thanks Pascal.

For the record, I'm still working on finding a test proving the fix in what is now 197472c, but haven't had enough time until today. I should have more time on Monday.

fpavageau:
Well, I haven't been able to create a test that would fail without the fix, because it always fails because of other generics-related bugs anyway :smile:. So I've tackled those instead.

PR #217, opened 2015-12-21
GROOVY-7717 StaticTypeCheckingTestCase#shouldFailWithMessages erroneo…

shils:
…usly passes

*StaticTypeCheckingTestCase: shouldFailWithMessages should only pass if all and only the expected error messages occur
*PrecompiledExtensionNotExtendingDSL:  fix ASTNode associated with error added in onMethodSelection method


blackdrag:
+1

shils:
Merged, thanks for reviewing.

PR #220, opened 2015-12-22
Fix finalizeReference method override for ManagedReference subclasses

jwagenleitner:
In working on PR #219 I ran across an issue with `ManagedConcurrentMap` not cleaning up `ManagedReference`'s.  It looked like the intent to was to override the `finalizeReference()` method but instead was named `finalizeRef()`.

### DIFF
--- a/src/main/org/codehaus/groovy/util/ManagedConcurrentValueMap.java
+++ b/src/main/org/codehaus/groovy/util/ManagedConcurrentValueMap.java
@@ -22,7 +22,7 @@ import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * This is a basic implementation of a map able to forget its values. This
- * map uses internally a ConcurrentHashMap, thus should be save for concurrency.
+ * map uses internally a ConcurrentHashMap, thus should be safe for concurrency.
  * hashcode and equals are used to find the entries and should thus be implemented
  * properly for the keys. This map does not support null keys.
  * @author <a href="mailto:blackdrag@gmx.org">Jochen "blackdrag" Theodorou</a>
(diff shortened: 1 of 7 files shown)


blackdrag:
could you also test against 7646? Since this fixes a problem in the managed map I would expect this change to really fix the issue

jwagenleitner:
I re-tested 7646 with these changes only and still get an OOME.  A heap dump shows 12k class loaders and they all seem to be rooted via `ClassInfo.klazz`.

If I set `-Dgroovy.use.classvalue=true` then I do not get the OOME.  But that is also true using 2.4.5 with classvalue enabled.

jwagenleitner:
Found a few more issues around `finalizeReference` so added a few more commits.  

I re-tested 7646 with the updated changeset and still get OOME.

PascalSchumacher:
+1

kcsampson:
Wow, thanks so much John!
This is the kind of thing that's going to really make groovy "the Best" language!

jwagenleitner:
@blackdrag just curious if you think these changes look ok?

While they don't directly address some of the issues, I did notice in testing GROOVY-7621 these changes consistently enabled it to go more iterations before an OOME.

src/main/org/codehaus/groovy/util/ManagedConcurrentValueMap.java
```diff
@@ -67,8 +67,8 @@ public void put(final K key, V value) {
         ManagedReference<V> ref = new ManagedReference<V>(bundle, value) {
             @Override
             public void finalizeReference() {
-                super.finalizeReference();
                 internalMap.remove(key, get());
```
jwagenleitner: `get()` here returns V, but internalMap is `ConcurrentHashMap<K,ManagedReference<V>>`, so think it should be:

```
internalMap.remove(key, this);
```

PR #221, opened 2015-12-22
GROOVY-6787 Compare arguments with the generic bounds

fpavageau:
The type inference on arguments shouldn't override the types of the
parameters, especially their bounds. This prevented the bound checks from
occurring.


blackdrag:
+1

fpavageau:
I just added an extra test for the simple unbounded wildcard case: `T extends List<?>`, which uses the `return true` statement I added in `GenericsType.GenericsTypeMatcher.matches()`.

PascalSchumacher:
Thanks! :+1:

PR #222, opened 2015-12-23
Removed unused Local ClassInfo cache objects (ThreadLocalMapHandler and LocalMap)

jwagenleitner:
Related to PR #219, this removes unused ThreadLocal caching from ClassInfo.  Because the `localMapRef` ThreadLocal field was wrapped in a `WeakReference` and nothing else ever references the referent ThreadLocalMapHandler object, I believe the `localMapRef` reference likely always returned `null` causing delegation to the global cache.  In addition, nothing was ever inserted into the LocalMap held by the ThreadLocal so it basically would always just return values from the `globalClassValue/Set`.

### DIFF
--- a/src/main/org/codehaus/groovy/reflection/ClassInfo.java
+++ b/src/main/org/codehaus/groovy/reflection/ClassInfo.java
@@ -25,9 +25,6 @@ import org.codehaus.groovy.reflection.stdclasses.*;
 import org.codehaus.groovy.util.*;
 import org.codehaus.groovy.vmplugin.VMPluginFactory;
 
-import java.lang.ref.PhantomReference;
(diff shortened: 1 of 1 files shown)


jwagenleitner:
On possible use of the local cache that I could think of would be to eliminate/reduce lock contention when accessing the global cache. However, I think it would introduce several potential problems such as another potential to hold refs to class/classinfo and possibly causing problems in thread pool environments.

As it stands I don't think the current implementation is working as a local cache so rather than address the PhantomReference and its use in the ThreadLocal subclass it seemed like removal was the way to go.

blackdrag:
That was more than I asked for, but still +1 ;)

PR #224, opened 2015-12-24
ReferenceManager.createThresholdedIdlingManager should delegate to ca…

jwagenleitner:
…llback manager after thresold is crossed.


blackdrag:
+1

PR #225, opened 2015-12-26
GROOVY-7716 - groovy.json.internal.FastStringUtils fails on jdk9

jwagenleitner:
JDK 9 Compat Strings enhancement changed the internal representation of the value field from a char[] to a combination of a byte[] and an encoding flag field (see http://openjdk.java.net/jeps/254).  As a workaround this change allows fallback to the normal String#toCharArray method and bypasses using direct access.


PascalSchumacher:
+1

PR #226, opened 2015-12-27
GROOVY-7709 - ConvertedClosure/ConvertedMap do not work with interface extending GroovyObject

jwagenleitner:
Just posting for discussion and acknowledge that there is probably a better fix and lots here is not being addressed other the immediate issue.  For this fix I borrowed from what was done in `GroovyResultSetProxy`.

Assuming this is even somewhat close to the right way to address it, it might make sense for the `ConvertedMap` to handle the metaClass in case the map has a `getMetaClass` key.

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/ConversionHandler.java
+++ b/src/main/org/codehaus/groovy/runtime/ConversionHandler.java
@@ -18,7 +18,11 @@
  */
 package org.codehaus.groovy.runtime;
 
+import groovy.lang.GroovyObject;
 import groovy.lang.GroovyRuntimeException;
+import groovy.lang.GroovySystem;
+import groovy.lang.MetaClass;
+import org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl;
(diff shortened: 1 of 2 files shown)


src/main/org/codehaus/groovy/runtime/ConversionHandler.java
```diff
@@ -189,4 +202,15 @@ public static boolean isCoreObjectMethod(Method method) {
...
     }
 
+    private MetaClass setMetaClass(MetaClass mc) {
+        metaClass = mc;
+        return mc;
+    }
+
+    private MetaClass getMetaClass(Object proxy) {
+        if (metaClass == null) {
+            metaClass = ((MetaClassRegistryImpl) GroovySystem.getMetaClassRegistry()).getMetaClass(proxy);
```
jwagenleitner: Not sure if this is the right way to get the `metaClass` for the `Proxy` object.

blackdrag:
I think this will work out, so +1

PR #228, opened 2015-12-29
GROOVY-7036: An interface implementation (override) with a method inc…

shils:
…luding a default parameter value does not compile


blackdrag:
+1

PR #229, opened 2015-12-29
GROOVY-7705: Add bridge methods for setting private fields of outer c…

shils:
…lass from inner class or closure


PascalSchumacher:
No negative feedback, plenty of tests, build passes. I guess this is good to merge?

shils:
@melix @blackdrag could you take a look at this if you haven't already? @PascalSchumacher I also want to add a test for the case of multiple bridge methods before merging.

blackdrag:
I would have liked Cedric to comment here, but if he has no time atm, then I suggest to just merge it. He can still complain later ;) The commit itself looks good to me

shils:
Merged, thanks for reviewing.

PR #230, opened 2016-01-01
GROOVY-7150 - Redirection of output in javax.script.ScriptEngine.invokeFunction() doesn't work

jwagenleitner:
Instead of adding `context` and `out` to the engine context this change provides access to those in the `groovy.lang.Binding` to the script class.  This should allow a user to provide their own variables if they wish to override the behavior.  The problem was caused because the `out` variable was being removed from the context (for cleanup reasons), so the cached `MethodClosure` no longer had access to it since it was executed in another method after the `eval` call.

I also added another commit to ensure the factory returned by the engine is the same factory that created it.  I couldn't see a reason why each engine instance should return a new factory.

### DIFF
--- a/subprojects/groovy-jsr223/src/main/java/org/codehaus/groovy/jsr223/GroovyScriptEngineImpl.java
+++ b/subprojects/groovy-jsr223/src/main/java/org/codehaus/groovy/jsr223/GroovyScriptEngineImpl.java
@@ -221,48 +221,6 @@ public class GroovyScriptEngineImpl extends AbstractScriptEngine implements Comp
 
     // package-privates
     Object eval(Class scriptClass, final ScriptContext ctx) throws ScriptException {
-        // Bindings so script has access to this environment.
-        // Only initialize once.
-        if (null == ctx.getAttribute("context", ScriptContext.ENGINE_SCOPE)) {
-            // add context to bindings
-            ctx.setAttribute("context", ctx, ScriptContext.ENGINE_SCOPE);
-
-            // direct output to ctx.getWriter
-            // If we're wrapping with a PrintWriter here,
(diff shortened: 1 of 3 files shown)


subprojects/groovy-jsr223/src/main/java/org/codehaus/groovy/jsr223/GroovyScriptEngineImpl.java
```diff
@@ -275,6 +238,19 @@ public Object getVariable(String name) {
                     if (scope != -1) {
                         return ctx.getAttribute(name, scope);
                     }
+                    // Redirect script output to context writer, if out var is not already provided
+                    if ("out".equals(name)) {
+                        Writer writer = ctx.getWriter();
+                        if (writer != null) {
+                            return (writer instanceof PrintWriter) ?
+                                    (PrintWriter) writer :
+                                    new PrintWriter(writer, true);
```
jwagenleitner: This could potentially create lots of temporary `PrintWriter` objects if `ctx.getWriter()` returns a different type.  But the alternative of caching the writer may mean keeping a reference to it longer than necessary.

PascalSchumacher:
Hi John,

thanks for taking this on. You write

>I also added another commit to ensure the factory returned by the engine is the same factory that >created it. I couldn't see a reason why each engine instance should return a new factory.

but the pull request is just a single commit?

jwagenleitner:
Hi Pascal, I must have inadvertently squashed the commit I referred to.  I've broken it out again.  Thanks for pointing that out.

PascalSchumacher:
+1

paulk-asert:
+1

PR #232, opened 2016-01-04
Two branches in the same conditional structure should not have exactly the same implementation

kirill-vlasov:
This pull request is focused on resolving occurrences of Sonar rule squid:S1871 - Two branches in the same conditional structure should not have exactly the same implementation
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid:S1871
Please let me know if you have any questions.
Kirill Vlasov

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/asm/OperandStack.java
+++ b/src/main/org/codehaus/groovy/classgen/asm/OperandStack.java
@@ -336,14 +336,12 @@ public class OperandStack {
                 return;
             }
             box();
-        } else if (primTop) {
-            // top is primitive, target is not
-            // so box and do groovy cast
-            controller.getInvocationWriter().castToNonPrimitiveIfNecessary(top, targetType);
         } else if (primTarget) {
             // top is not primitive so unbox
             // leave that BH#doCast later
         } else {
+            // top is primitive, target is not
+            // so box and do groovy cast
             controller.getInvocationWriter().castToNonPrimitiveIfNecessary(top, targetType);
         }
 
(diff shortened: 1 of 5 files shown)


src/main/org/codehaus/groovy/classgen/asm/OperandStack.java
```diff
@@ -336,14 +336,12 @@ private void doConvertAndCast(ClassNode targetType, boolean coerce) {
...
-        } else if (primTop) {
-            // top is primitive, target is not
-            // so box and do groovy cast
-            controller.getInvocationWriter().castToNonPrimitiveIfNecessary(top, targetType);
         } else if (primTarget) {
             // top is not primitive so unbox
             // leave that BH#doCast later
         } else {
+            // top is primitive, target is not
+            // so box and do groovy cast
```
paulk-asert: I am thinking this comment should be improved. It covers both the case the comment mentions and the default case. Let's see if Jochen or anyone else has thoughts on what that default branch is supposed to do with respect to how best to rephrase if needed.

blackdrag: the comment needs a change... "// top might be primitive, target is not" and "// so let invocation writer box if needed and do groovy cast otherwise". I think this case appeared to due to moving some functionality into the invocation writer, I think before there has been really two different actions. Well.. Still can make this change with the changed comment then of course ;)

src/main/org/codehaus/groovy/classgen/asm/sc/StaticInvocationWriter.java
```diff
@@ -602,7 +602,7 @@ private static void pushZero(final MethodVisitor mv, final ClassNode type) {
         boolean isInt = ClassHelper.int_TYPE.equals(type);
         boolean isShort = ClassHelper.short_TYPE.equals(type);
         boolean isByte = ClassHelper.byte_TYPE.equals(type);
-        if (isInt || isShort || isByte) {
+        if (isInt || isShort || isByte || ClassHelper.boolean_TYPE.equals(type)) {
```
blackdrag: actuallly, the  "|| ClassHelper.boolean_TYPE.equals(type)" should be removed. isByte contains this information already.

PascalSchumacher:
Thanks!

@paulk-asert @blackdrag I updated the pull request according to your feedback before merging.

PR #234, opened 2016-01-06
[build] Remove javax.script pre-JDK6 dependency

jwagenleitner:
Minimum jdk is now 6 so no longer need compatibility dependency.


PascalSchumacher:
+1

PR #236, opened 2016-01-09
GROOVY-6327 - normalize line endings

jwagenleitner:
Use `.gitattributes` to normalize line endings (http://git-scm.com/docs/gitattributes).  Besides adding the `.gitattributes` file also had it identify existing files with `CRLF` and converted them to `LF`.


PascalSchumacher:
+1

blackdrag:
This changes 454 files....

jwagenleitner:
Yes, a few files were affected. :smile: 

That is why I left as 2 commits in case such a big change isn't accepted at this time but in case we want to add just the `.gitattributes` file to make sure line endings are ok going forward.  This big of a change should probably have everyone's agreement and be coordinated since it wouldn't be fun to handle merge conflicts.  If you run a `git diff -w master..<this branch>` only the new file should show.  All tests passed locally, waiting for the PR hook to be fixed to see if CI server passes.

jwagenleitner:
Below are the affected file counts by directory, the bulk of EOL changes being in `src/test`.

 52 - src/main/groovy
 70 - src/main/org/codehaus/groovy
291 - src/test
 26 - src/examples
  5 - subprojects/groovy-jmx

Misc (9):

config/bnd/groovy-all.bnd
config/bnd/groovy.bnd
config/checkstyle/codeHeader.txt
gradlew.bat
src/main/META-INF/groovy-release-info.properties
src/main/META-INF/services/org.codehaus.groovy.source.Extensions
src/main/META-INF/services/org.codehaus.groovy.transform.ASTTransformation
src/main/overview.html
src/main/overviewj.html

No need for a big change all at once, probably be best to slowly phase it in.  Maybe starting with `src/examples` and `groovy-jmx` since those probably wouldn't create merge conflicts with open PR's or in-progress work.

Of course it's perfectly valid to not do anything and leave EOL's as is.

PascalSchumacher:
Would .attributes just be for new files, or will it modify the line endings as soon as there is a change to a file?

jwagenleitner:
Both.  It will modify the line endings if you change a file.  For example, if you make a one line change to `src/main/org/codehaus/groovy/runtime/ConversionHandler.java` it will change all line endings when you commit that change.  So normal diff's in git or github will show the whole file as changed unless you add `-w` and `&w=1` respectively.  The line endings of the working copy wont be changed so they'll be warnings on each future commit unless the working copy is reset.  But after the first commit to normalize the line endings future changes should only show what changed (regardless of line endings in the working copy).

blackdrag:
That means we will still get problems with finding out about the history of a specific line. Furthermore I am wondering about files like gradlew.bat. This is a windows file and bat expects windows line endings afaik. So could this change cause problems here?

jwagenleitner:
>That means we will still get problems with finding out about the history of a specific line. 

If we did the conversion there would be one more commit in the history for a line in the files affected.  However, a line is still a line so you should be able to still see the commits of particular line.

>Furthermore I am wondering about files like gradlew.bat. This is a windows file and bat expects windows line endings afaik. So could this change cause problems here?

The `.gitattributes` let you specify an override for certain files so they can retain CRLF endings.  I have `.bat` files set like that in this change.

```
# Windows files
*.bat           text eol=crlf 
```
So that's why `src/bin/*.bat` files were not affected.  The `gradlew.bat` was affected because it actually had Linux line endings so they were converted to Windows.

jwagenleitner:
To really do the conversion would take coordination and this PR was really just to allow review of the potential changes.  Based on the feedback and with releases possibly coming soon and lack of time, it's probably best to defer this.  I am closing this PR and will keep track of the JIRA ticket and work towards conversion in the future.

PR #237, opened 2016-01-10
ConversionHandler improvements

jwagenleitner:
Added final where appropriate and limited plugin variable scope in invoke method.

Related to PR #226 (GROOVY-7709), while full synchronization may not be required the lazily initialized `metaClass` should probably be `volatile`.

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/ConversionHandler.java
+++ b/src/main/org/codehaus/groovy/runtime/ConversionHandler.java
@@ -42,11 +42,15 @@ import java.util.concurrent.ConcurrentHashMap;
  * @author <a href="mailto:blackdrag@gmx.org">Jochen Theodorou</a>
  */
 public abstract class ConversionHandler implements InvocationHandler, Serializable {
-    private Object delegate;
+    private final Object delegate;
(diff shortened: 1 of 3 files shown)


blackdrag:
I do agree with the final changes, they are good. But assuming metaClass is not volatile, where would you expect a problem? Having a partially initialized meta class object being published must not be a problem, or half of Groovy has a big problem. Every meta class is itself responsible for its part being properly visible. So -1 on the volatile

jwagenleitner:
Thanks for the feedback I appreciate it.  In looking at the [MetaClassImpl] (https://github.com/apache/groovy/blob/2046dd76bdbb7398d425e83a7d9b39a0a8707c9a/src/main/groovy/lang/MetaClassImpl.java#L148-L160) I just noticed a few non-final instance variables.  But most aren't really part of the initialization that I can tell other than the `FastArray constructors` and `MetaClassRegistry registry` and both look like candidates to be final.

Guess I was just being overly cautious since lazy initialization can be tricky.  I'll update the PR and remove the `volatile`.

jwagenleitner:
PR updated to remove final.  Included a few minor changes to the subclasses.

blackdrag:
+1

PR #239, opened 2016-01-14
GROOVY-7105 Add try/catch for NoClassDefFoundError

belkaram:
SAM method resolution can cause loading of classes used in method
arguments (or as a return type).
When such a class is not present on a CLASSPATH, it causes a
NoClassDefFoundError even though the class may actually not be needed
for program execution.

This is the naive sollution, but I think it may work. It suppresses the NoClassDefFoundError when resolving SAM method, but in the case when the class is really needed for execution, it will throw the exception then.


blackdrag:
just noticed, that this is for the 2.3.x branch... the fix should of course into master and all maintained versions

PascalSchumacher:
Thanks!

PR #241, opened 2016-01-15
Revert change from c5f17abbe Fixing squid:S2444 - Lazy initialization…

jwagenleitner:
Partial revert of commit c5f17abbe which changed HandleMetaClass#myMetaClass to an instance variable.  This change reverts that change making myMetaClass static and initialized on demand.

See dev list discussion http://mail-archives.apache.org/mod_mbox/groovy-dev/201601.mbox/%3C568CD9D3.7000200@gmx.org%3E

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/HandleMetaClass.java
+++ b/src/main/org/codehaus/groovy/runtime/HandleMetaClass.java
@@ -24,7 +24,6 @@ import java.lang.reflect.Method;
 
 public class HandleMetaClass extends DelegatingMetaClass {
     private Object object;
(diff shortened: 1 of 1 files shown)


PascalSchumacher:
+1

jwagenleitner:
Modified the commit to be a little more inline with what had suggested on the dev list.  It was less code to use a private holder class to do the lazy init (even if not strictly needed for thread safety).  If the other style is preferred I'll change it to use a static check and set method for the static field.  What do you think @blackdrag ?

blackdrag:
+1

PR #242, opened 2016-01-16
GROOVY-7558 Add private field accessors and bridge methods to dynamic…

shils:
… classes with statically compiled inner classes

### DIFF
--- a/src/main/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
+++ b/src/main/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
@@ -28,9 +28,12 @@ import org.codehaus.groovy.ast.stmt.ExpressionStatement;
 import org.codehaus.groovy.ast.stmt.ForStatement;
 import org.codehaus.groovy.ast.stmt.Statement;
 import org.codehaus.groovy.ast.tools.GeneralUtils;
+import org.codehaus.groovy.classgen.GeneratorContext;
 import org.codehaus.groovy.classgen.asm.*;
 import org.codehaus.groovy.classgen.asm.sc.StaticCompilationMopWriter;
 import org.codehaus.groovy.classgen.asm.sc.StaticTypesTypeChooser;
+import org.codehaus.groovy.control.CompilationFailedException;
+import org.codehaus.groovy.control.CompilationUnit;
 import org.codehaus.groovy.control.SourceUnit;
 import org.codehaus.groovy.transform.stc.StaticTypeCheckingSupport;
 import org.codehaus.groovy.transform.stc.StaticTypeCheckingVisitor;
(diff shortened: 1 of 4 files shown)


src/main/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
```diff
@@ -123,9 +123,20 @@ public void visitClass(final ClassNode node) {
...
+        if (isStaticallyCompiled(node)) {
+            ClassNode outer = node.getOuterClass();
+            if (outer != null && !isStaticallyCompiled(outer)) {
+                outer.putNodeMetaData(StaticCompilationMetadataKeys.DYNAMIC_OUTER_NODE, true);
+            }
+        }
         classNode = oldCN;
     }
 
+    public void visitDynamicOuterClass(ClassNode node) {
```
shils: adding a public method here is not ideal, though it seems like the alternative would be to create a separate class responsible for adding bridge methods, which would be used by StaticCompilationVisitor as well as the phase operation added for dynamic outer classes.

blackdrag:
So basically you find, you need to generate those bridge methods, and leave it to a later visitor to actually produce that code... The alternative would be to write a visitor that can adds this information and is run from StaticCompilationVisitor, to let the actual adding of those methods be done by the StaticCompilationVisitor later on. But actually the list of inner classes in the classnode is supposed to contain all inner classes at this stage, so I am not sure you even need a full blown visitor for this. What do you think

shils:
If I understand correctly, your suggestion is to have the StaticCompilationVisitor for an inner class or method add bridge methods to its outer class? The issue there is that there may be multiple statically compiled inner classes/methods, each possibly accessing different private members of the outer class, so we would have to try adding bridge methods (to the outer class) for each statically compiled inner node and keep track of which private members we'd already added bridge methods for. This would add some complexity to the implementation, but I think it would be ok. Does that sound like what you had in mind?

PascalSchumacher:
@blackdrag Any feedback for @shils?

blackdrag:
Sorry, for not answering earlier. It might be more easy to explain by example... Look at StaticTypeCheckingVisitor.VariableExpressionTypeMemoizer. The memoizer is used in the visitor to track something, even though StaticTypeCheckingVisitor is itself a visitor. The memoizer here does a specific task, by collecting information in the subtrees of the AST, and feeds that to the original visitor. IN your case that would be the information of what bridge methods do need to be added

shils:
@blackdrag That's alright, I've been busier than usual as of late anyway. Thanks for the feedback, I'll take a look.

shils:
@blackdrag I'm not sure I understand your comment. Dynamic classes are not visited by any static compilation visitor so the bridge method collecting visitor you suggest would have to be run from some other visitor... Also, the collecting visitor would have to perform method selection across all of the statically compiled inner classes (or be run after method selection.) 

In any case, I changed the implementation to not add a public method at the cost of possibly misusing phase operations.

paulk-asert:
+1, the use of a callback in metadata is a little novel but I can't think of a better way right now and I think correctness trumps the added complexity in this case. We can always refactor and simplify if we can think of an improvement.

shils:
Agreed, thanks for reviewing this.

PR #244, opened 2016-01-19
Sorted the list of contributors alphabetically

anshbansal:


PascalSchumacher:
I guess the list was supposed to be ordered alphabetically by surname.

Is it common to order by first name in the english speaking parts of the world?

anshbansal:
@PascalSchumacher 
Not sure about that. If we are taking cultural differences into account then a simpler thing to do would be to just remove the part about sorting. :smile:

PascalSchumacher:
Thanks for the pull request. I decided to keep the current sorting by surname.

PR #245, opened 2016-01-20
GROOVY-4698 - Possible memory leak in Tomcat (GroovyScriptEngine ThreadLocal)

jwagenleitner:
Recommend reviewing each commit separately.  The first commit I think is fairly safe and addresses the issue reported.  The others I'm a little less sure about.

863247f - main fix is ensuring that `localTh.remove()` (changed from `localTh.set(null)`) is always called even if an exception occurs.  Did some refactoring by pulling out the code used to update dependencies and the script cache.  Even though some changes have been made to the ThreadLocal's in this class since that JIRA, the problem remains because `set(null)` will not be called if `super.parseClass` throws.  I think this is a fairly safe change.

dbfa026- If we always clean up the ThreadLocal by calling remove, I think things can be simplified by removing the `WeakReference` holder and synchronized `getLocalData` method.

c32722a- Since `super.parseClass` is synchronized and the other code in that method just updates thread confined (ThreadLocal) data and ConcurrentHashMap scriptCache, it seemed like that could be done outside of synchronization.  I don't know if there might be a potential problem with ordering of the `scriptCache.put` calls.  However, if it is supposed to be synchronized, then I'm not sure a ThreadLocal is needed since that would mean all access to the ThreadLocal is synchronized.  In that case why not just have a `private LocalData localData;` field and set it with a `new LocalData()` each time `parseClass` is called?

**UNIT TEST NOTE**: I'm not sure how good the added test methods are.  They were helpful in showing the issue before the changes (commit 863247f).  However, they might be brittle since they rely on poking into package-private/private members.

### DIFF
--- a/src/main/groovy/util/GroovyScriptEngine.java
+++ b/src/main/groovy/util/GroovyScriptEngine.java
@@ -81,15 +81,12 @@ public class GroovyScriptEngine implements ResourceConnector {
         Map<String, String> precompiledEntries = new HashMap<String, String>();
     }
 
-    private static WeakReference<ThreadLocal<LocalData>> localData = new WeakReference<ThreadLocal<LocalData>>(null);
-
-    private static synchronized ThreadLocal<LocalData> getLocalData() {
-        ThreadLocal<LocalData> local = localData.get();
-        if (local != null) return local;
-        local = new ThreadLocal<LocalData>();
-        localData = new WeakReference<ThreadLocal<LocalData>>(local);
-        return local;
-    }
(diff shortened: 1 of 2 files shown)


blackdrag:
All commits but dbfa026 are ok with me. For dbfa026 I would like to be sure, that if you load a Groovy instance and use the thread local, that the JVM can still later unload it again and there is no JDK-internal data structure preventing this.

jwagenleitner:
For now I have merged just the main commit that addresses the issue, minus the 2 test cases that relied on internal Thread fields.

The other commits I will break out into separate PR's if I can get more comfortable that they wont cause problems.

PR #246, opened 2016-01-20
correct attribution of `prettyPrint` method and wording of comment ab…

marcpa00:
…out it taking a normal String.


PascalSchumacher:
Thanks! :+1:

PR #247, opened 2016-01-24
Added new class BufferedIterator.

ataylor284:
BufferedIterator is a wrapper for Iterator that adds a new method
`head` to return the next item without consuming it.  A helper method
was also added to DefaultGroovyMethods to create a buffered wrapper by
calling iterator.buffered() on any Iterator.  BufferedIterator is
loosely based on the Scala class of the same name.

### DIFF
--- a/src/main/groovy/util/BufferedIterator.java
+++ b/src/main/groovy/util/BufferedIterator.java
@@ -1,0 +1,69 @@
+/**
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package groovy.util;
+
+import java.util.Iterator;
+
+/**
+ * An iterator that allows examining the next element without consuming it.
+ *
+ * @author Andrew Taylor
+ * @since 2.5.0
+ */
+public class BufferedIterator<T> implements Iterator<T> {
+
+    private final Iterator<T> iter;
+    private boolean hasBuffered;
+    private T buffered;
+
+    public BufferedIterator(Iterator<T> iter) {
+        this.iter = iter;
+        this.hasBuffered = false;
+    }
+
+    public boolean hasNext() {
+        return hasBuffered || iter.hasNext();
+    }
+
+    public T next() {
+        if (hasBuffered) {
+            T buffered = this.buffered;
+            this.buffered = null;
+            hasBuffered = false;
+            return buffered;
+        } else {
+            return iter.next();
+        }
+    }
+
+    public void remove() {
+        throw new UnsupportedOperationException();
+    }
+
+    /**
+     * Return the next element to be returned by next() without consuming it.
+     */
+    public T head() {
+        if (!hasBuffered) {
+            buffered = iter.next();
+            hasBuffered = true;
+        }
+        return buffered;
+    }
+}
--- a/src/main/groovy/util/BufferedIterator.java
+++ b/src/main/groovy/util/BufferedIterator.java
@@ -26,44 +26,6 @@ import java.util.Iterator;
  * @author Andrew Taylor
  * @since 2.5.0
  */
-public class BufferedIterator<T> implements Iterator<T> {
-
-    private final Iterator<T> iter;
-    private boolean hasBuffered;
-    private T buffered;
-
-    public BufferedIterator(Iterator<T> iter) {
-        this.iter = iter;
-        this.hasBuffered = false;
-    }
-
-    public boolean hasNext() {
-        return hasBuffered || iter.hasNext();
-    }
-
-    public T next() {
-        if (hasBuffered) {
-            T buffered = this.buffered;
-            this.buffered = null;
-            hasBuffered = false;
-            return buffered;
-        } else {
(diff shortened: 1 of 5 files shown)


src/main/groovy/util/BufferedIterator.java
```diff
@@ -0,0 +1,69 @@
...
+            this.buffered = null;
+            hasBuffered = false;
+            return buffered;
+        } else {
+            return iter.next();
+        }
+    }
+
+    public void remove() {
+        throw new UnsupportedOperationException();
```
jwagenleitner: Any reason not to delegate this and then clear buffered/hasBuffered?

jwagenleitner:
Looks like useful wrapper.  For a new feature like this it would be good to have a JIRA ticket since that might spark more discussion (and helps tie features to a release).

A couple of things come to mind in working with it.  Not proposing any changes to what's been proposed, just thinking out loud:

1. Would it make sense to also define a `bufferedIterator` method on `Iterable` so you could do something like `[1,2].bufferedIterator`?

2. Since it's an extension of the `Iterator` interface would it make sense to also define this new type `groovy.util.BufferedIterator` as an Interface with the implementation used by `DefaultGroovyMethods` coming from `org.codehaus.groovy.util`?  Don't have a compelling reason for this since I doubt there'd been too many implementation of it, just tend to prefer interfaces for new types rather than concrete classes but this could be considered more a wrapper than new type.

src/main/groovy/util/BufferedIterator.java
ataylor284: If the `head` method has been called and an item is buffered, the underlying iterator is pointing at the wrong element.  Delegating would remove the item after the one it should.  This case could be detected but there's no way to handle it correctly.  Rather than failing in some cases, I think it's better to just not support removal.

ataylor284:
Good point.  I'll create a jira issue to track this.  As for `bufferedIterator`, I don't see much advantage to calling `iterable.bufferedIterator()` over `iterable.iterator().buffered()`.  I guess it's a matter of taste.  I like the idea of exporting interfaces in the official namespace with implementations in `org.codehaus`, but it might be overkill in this instance.  I can't think of a reason you'd want an alternative implementation of `BufferedIterator`.

Thanks for the feedback!

src/main/groovy/util/BufferedIterator.java
jwagenleitner: Good point, my mistake, I was thinking in terms of `remove()` removing the element returned from the last call to either `head` or `next` but this would be confusing if you wanted to delete the current value based on the next value.

```
value = itr.next() // 1
if (itr.head() == 2) {
    itr.remove() // removes 2 not 1
}
```

blackdrag:
it might be worth also looking at http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/PeekingIterator.html They use peek instead of head and they support remove, but say it can be unspecified. Now, while the current implementation here does not really support a combination of remove and head, I can think of a collection class implementation that is able to do so. So it might actually be worth to go through an interface.... especially in combination with `bufferedIterator()`.

ataylor284:
Ok, I've updated the pull request.  As suggested, I made `BufferedIterator` an interface and created implementations in `org.codehaus.groovy.util`.  A list specific version supports the `remove()` operation.  Finally,  new methods `Iterable.bufferedIterator()` and `List.bufferedIterator()` each return an appropriate implementation.

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -16045,6 +16048,51 @@ public void remove() {
...
+     * <pre class="groovyTestCase">
+     * assert [1, 2, 3, 4].iterator().buffered().with { [head(), toList()] } == [1, [1, 2, 3, 4]]
+     * </pre>
+     *
+     * @param self an iterator object
+     * @return a BufferedIterator wrapping self
+     * @since 2.5.0
+     */
+    public static <T> BufferedIterator<T> buffered(Iterator<T> self) {
+        return new IteratorBufferedIterator<T>(self);
```
jwagenleitner: In case `self` is a `BufferedIterator` should this return `self` instead of re-wrapping?

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -16045,6 +16048,51 @@ public void remove() {
...
+    }
+
+    /**
+     * Returns a <code>BufferedIterator</code> that allows examining the next element without
+     * consuming it.
+     * <pre class="groovyTestCase">
+     * assert new LinkedHashSet([1,2,3,4]).bufferedIterator().with { [head(), toList()] } == [1, [1,2,3,4]]
+     * </pre>
+     *
+     * @param self an iterator object
```
jwagenleitner: s/iterator/iterable/

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -16045,6 +16048,51 @@ public void remove() {
...
+    }
+
+    /**
+     * Returns a <code>BufferedIterator</code> that allows examining the next element without
+     * consuming it.
+     * <pre class="groovyTestCase">
+     * assert [1, 2, 3, 4].bufferedIterator().with { [head(), toList()] } == [1, [1, 2, 3, 4]]
+     * </pre>
+     *
+     * @param self an iterator object
```
jwagenleitner: s/iterator/list/

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -16045,6 +16048,51 @@ public void remove() {
...
+     * @return a ListBufferedIterator wrapping self
```
jwagenleitner: Might be good to refer to the general `BufferedIterator` and not the specific implementation type.

jwagenleitner:
Looks good, just had a few very minor comments.  In my opinion I do like how guava allows remove if peek hadn't been previously called.  That seems more flexible by allowing it to be used as a normal iterator if needed.

ataylor284:
Fixed the documentation errors and implemented `remove()` for `IteratorBufferedIterator` in the cases where it would be correct.

PascalSchumacher:
jira to track this: https://issues.apache.org/jira/browse/GROOVY-7746

PascalSchumacher:
Thanks! :+1:

PR #249, opened 2016-01-27
Remove duplicate ReferenceBundle definitions

jwagenleitner:
Removed static initialization of default Soft/Weak bundles in ReferenceManager class.  The methods in that class were not used in the codebase.  Looking at the history of those static bundle references it appeared that 500 was the original threshold setting.  Typical startup creates roughly 1000 managed references, so 500 seems like a more appropriate value than 5000.

Note: the deprecated methods are in the `org.codehaus.groovy.util` package which might be considered non-public even though the methods are public.  But just to be safe left them in place just in case they may be referenced in others code.


blackdrag:
+1

PR #250, opened 2016-01-28
ListHashMap fixes

jwagenleitner:
When removing entries and/or switching to Map the arrays should be cleared.

Methods containsKey/containsValue throw NPE if size == maxListSize.


blackdrag:
+1

PR #252, opened 2016-02-01
Map#removeAll(Closure) and Map#retainAll(Closure)

oreissig:
I added support for a method I missed more than once: `Map.removeAll(Closure)`, which behaves exactly like `Collection.removeAll(Closure)` and takes the same parameters as `Map.findAll(Closure)`. This enables convenient use of key/value parameters in contrast to the workaround of `Map.entrySet().removeAll(Closure)`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -136,7 +136,7 @@ ext {
     commonsHttpClientVersion = '3.1'
     eclipseOsgiVersion = '3.9.1-v20140110-1610'
     gparsVersion = '1.2.1'
-    gradleVersion = '2.8'
+    gradleVersion = '2.10'
     ivyVersion = '2.4.0'
     jansiVersion = '1.11'
     jarjarVersion = '1.3'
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
Hi @oreissig,

I like it and created a jira entry for this enhancement: https://issues.apache.org/jira/browse/GROOVY-7745

It would be very nice if you could also add an example to the javadoc like Map#each(Closure) etc.

Thanks,
Pascal

oreissig:
I also added an implementation incl. test and javadoc for `Map#retainAll(Closure)` for consistency, and also added an example snippet to the according methods on `Collection`.

One question: I'm directly calling `self.entrySet().iterator()`, whereas the existing `Collection` method does `InvokerHelper.asIterator(self)`. Should I also use `InvokerHelper` to get the `entrySet()` as well as its `iterator()`?

paulk-asert:
Nice work. Looks good to me.

paulk-asert:
Merged, thanks!

PR #253, opened 2016-02-05
GROOVY-7742 - Stackoverflow when method parameters have different placeholders

jwagenleitner:
Not sure about this fix, but putting it out there for review.

### DIFF
--- a/src/test/groovy/transform/stc/GenericsSTCTest.groovy
+++ b/src/test/groovy/transform/stc/GenericsSTCTest.groovy
@@ -1800,6 +1800,19 @@ assert result == 'ok'
 '''
     }
 
+    // GROOVY-7742
+    void testMethodParamWithDifferentPlaceholderForTypeArgument() {
+        // We use <T> here while List uses <E>
+        assertScript '''
+            class Foo {
+                public <T> Class<T> firstClass(List<Class<T>> classes) {
+                    return classes.first()
+                }
+            }
+            null
+        '''
+    }
+
     static class MyList extends LinkedList<String> {}
 
     public static class ClassA<T> {
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -1520,6 +1520,9 @@ public abstract class StaticTypeCheckingSupport {
                 GenericsType replacement = connections.get(name);
                 if (replacement==null) {
                     GenericsType value = entry.getValue();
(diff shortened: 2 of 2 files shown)


src/test/groovy/transform/stc/GenericsSTCTest.groovy
```diff
@@ -1800,6 +1800,19 @@ assert result == 'ok'
 '''
     }
 
+    // GROOVY-7742
+    void testMethodParamWithDifferentPlaceholderForTypeArgument() {
+        // We use <T> here while List uses <E>
```
shils: I don't think using a different placeholder than the generic type is the issue here e.g. if 'T' is replaced with 'U' in this test, it passes without your change.

fpavageau: The problem does not appear when using a different placeholder than `List` (i.e. not `E`), it appears when using the same as `Class` (`T`).

shils: That was my thought initially as well, but in that case I would expect the following to fail since AtomicReference uses 'V' as a placeholder:
```
import java.util.concurrent.atomic.AtomicReference

class Foo {
    public <V> AtomicReference<V> firstReference(List<AtomicReference<V>> references) {
        return references.first()
    }
}
```

The code above compiles but, interestingly enough, the variant below does not:

```
import java.util.concurrent.atomic.AtomicReference                                 

class Foo {                                                                        
    public <T> AtomicReference<T> firstClass(List<AtomicReference<T>> references) {
        return references.first()                                                  
    }                                                                              
}
```

jwagenleitner: Looks like when it resolves the placeholders it adds T based on DGM `static <T> T first(List<T> self)` since it uses `T` as placeholder.  The real code that produced this error used `collect` which also uses `T`.  Changing the code to use `return classes.get(0)` fixes it.

PascalSchumacher:
Hopefully we can release 2.4.6 soon, but before we do we should fix this issue.

@ everybody: Should we apply this fix, or rollback the original change?

jwagenleitner:
I don't feel confident enough in this area of the code to say this fix is good.  So unless others agree it's a good fix the best option may be to revert the original change.  

As @shils and @fpavageau have pointed out, I haven't correctly captured the reason for the failure in naming the test.  It almost seems to be related to Extension methods, and in this case the return type being from the DefaultGroovyMethods.

shils:
+1 for reverting https://github.com/apache/groovy/commit/fae29119a1102393ae5d1645c3fc1e06547b0ad8 prior to 2.4.6 and creating an issue to revisit it in the future.

PR #257, opened 2016-02-09
GROOVY-7312 Add ACC_STATIC flag to inner classes of interfaces

shils:


src/main/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -80,7 +80,11 @@ public void writeClosure(ClosureExpression expression) {
 
         // generate closure as public class to make sure it can be properly invoked by classes of the
         // Groovy runtime without circumventing JVM access checks (see CachedMethod for example).
-        ClassNode closureClass = getOrAddClosureClass(expression, ACC_PUBLIC);
+        int mods = ACC_PUBLIC;
+        if (classNode.isInterface()) {
```
shils: Classes for closures defined in static methods could (should?) also be flagged as static, but I left that case as is since it wasn't mentioned in GROOVY-7312.

PascalSchumacher:
@blackdrag Any feedback for @shils?

blackdrag:
+1

PR #259, opened 2016-02-10
Fix the comparison of generic types

fpavageau:
The comparison cannot shortcut when a condition passes, only when one fails,
since it's basically a big AND. E.g. even if both lower bounds are null (and
therefore equal), the upper bounds still need to be compared. The only
obvious shortcut is if the operands are the same object, as the first
condition.

I did not manage to create a test that would fail without the change. However, it's a change similar to fae2911 which ended up causing [GROOVY-7742](https://issues.apache.org/jira/browse/GROOVY-7742), so I think it should only be applied to `master` for the moment.

### DIFF
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -1575,16 +1575,18 @@ public abstract class StaticTypeCheckingSupport {
         if (!equalIncludingGenerics(orig.getType(), copy.getType())) return false;
         ClassNode lower1 = orig.getLowerBound();
         ClassNode lower2 = copy.getLowerBound();
-        if ((lower1==null || lower2 ==null) && lower1!=lower2) return false;
-        if (lower1==lower2) return true;
-        if (!equalIncludingGenerics(lower1,lower2)) return false;
+        if ((lower1 == null) ^ (lower2 ==null)) return false;
+        if (lower1 != lower2) {
+            if (!equalIncludingGenerics(lower1, lower2)) return false;
+        }
         ClassNode[] upper1 = orig.getUpperBounds();
         ClassNode[] upper2 = copy.getUpperBounds();
-        if ((upper1==null || upper2==null) && upper1!=upper2) return false;
-        if (upper1==upper2) return true;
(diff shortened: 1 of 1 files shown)


src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1575,16 +1575,18 @@ private static boolean equalIncludingGenerics(GenericsType orig, GenericsType co
         if (!equalIncludingGenerics(orig.getType(), copy.getType())) return false;
         ClassNode lower1 = orig.getLowerBound();
         ClassNode lower2 = copy.getLowerBound();
-        if ((lower1==null || lower2 ==null) && lower1!=lower2) return false;
-        if (lower1==lower2) return true;
```
fpavageau: This is the shortcut that's incorrect if `lower1` and `lower2` are both null. The other changes are there mainly for consistency.

blackdrag:
the commit is ok, but it could have been a smaller one, if I look at your changes in line 1600 and following. Same for the xor parts, the old logic could have stayed, only the following line with == was really wrong here. You also don't really need to check for == to avoid a call to equalIncludingGenerics, the method will do that check anyway. Where it does makes sense, is of course in the upper bounds check, because they can be null now, thus you need to check or else you cannot get the length (you know that of course, it is just for a third party reader)

fpavageau:
Yes, as I wrote in my earlier comment, I took the opportunity of the fix to make the code a bit more regular and consistent: the lower and upper bounds are compared in a similar way, itself similar to the way the generic types are compared in the second method (replace "upper" by "gt" and the blocks are identical), and the use of `^` is generalized (I introduced it in a previous PR). I felt it made the code easier to understand.

The `lower1 != lower2` check protecting the `equalIncludingGenerics` call is indeed useless and can be removed to simplify.

However, I can also definitely limit the change to removing the `lower1 == lower2` shortcut if you'd rather keep the change as small as possible. I think the code is more brittle (and a bit less readable) as it is -- you couldn't swap the lower and upper bounds checks as they are if you wanted to, for example, because of the `upper1 == upper2` shortcut which would become as incorrect as `lower1 == lower2` currently is -- but it certainly works. I just mean my change is not purely a style change.

PascalSchumacher:
Thanks! :+1:

PR #260, opened 2016-02-11
Ignore all Eclipse project files

oreissig:
I added some missing subprojects to the `.gitignore` file. That's the quick & easy solution. I'd actually prefer ignoring `.classpath`, `.project` and `.settings` globally without hard coding all subprojects into `.gitignore`. Any opinions?

Additionally I reverted the default task, so that gradle shows its generic helpful "splash screen" when being invoked without parameters. Was there a particular reason for the wrapper regeneration being the default task? If so maybe we could comment that with a one liner.


PascalSchumacher:
>I added some missing subprojects to the .gitignore file. That's the quick & easy solution. I'd actually prefer ignoring .classpath, .project and .settings globally without hard coding all subprojects into .gitignore. Any opinions?

I agree.

>Additionally I reverted the default task, so that gradle shows its generic helpful "splash screen" when being invoked without parameters. Was there a particular reason for the wrapper regeneration being the default task? If so maybe we could comment that with a one liner.

no idea, but I agree it should not be the default

oreissig:
I think the CI failure is not related to my changes.
```

* What went wrong:
Unable to start the daemon process.
This problem might be caused by incorrect configuration of the daemon.
For example, an unrecognized jvm option is used.
Please refer to the user guide chapter on the daemon at https://docs.gradle.org/2.11/userguide/gradle_daemon.html
Please read the following process output to find out more:
-----------------------
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF-8
Please use CMSClassUnloadingEnabled in place of CMSPermGenSweepingEnabled in the future
*** Error in `/home/jenkins/jenkins-slave/tools/hudson.model.JDK/latest1.6/bin/java': free(): invalid next size (fast): 0x00007fefe0c06be0 ***
```

PascalSchumacher:
Yes, the failure is not caused by your changes. It's a is a reoccurring temporary issue on builds.apache.org. There is an unsolved infrastructure issue for this: https://issues.apache.org/jira/browse/INFRA-11238

PascalSchumacher:
Thanks!

PR #261, opened 2016-02-12
make private methods static when they are pure functions

oreissig:
Just a minor thing: Most AST transformations are stateless and only pass data via arguments, which allows to treat a lot of their helper methods as static functions.


oreissig:
CI failure seems to be unrelated to my changes

PascalSchumacher:
Yes, the failure is an unrelated reoccurring temporary issue on builds.apache.org. There is an unsolved ticket for this: https://issues.apache.org/jira/browse/INFRA-11238

I also set up travis ci to build pull requests, in the hope that it's more stable than builds.apache.org.

If you like you can force push (push -f), to give it another try.

PR #262, opened 2016-02-13
Long running tests

jwagenleitner:
These are listed as some of the longest running tests on the TeamCity CI server.  One test being removed makes no assertions.  Other changes involve reducing the sleep time or loops performed while sleeping which I don't believe will affect the usefulness of the test.


PascalSchumacher:
+1

PR #263, opened 2016-02-13
GROOVY-7751 - CallableStatement leak in Sql.call() method

jwagenleitner:
Found a similar issue with `BatchingStatementWrapper` not having it's delegate statement closed when statement caching was enabled.  Added fix for that as a separate commit.


PascalSchumacher:
+1

PascalSchumacher:
@jwagenleitner What about merging this?

jwagenleitner:
@PascalSchumacher I was going to test it with a small project I have that uses stored procs, will do that in the next day or two and merge if no issues.

PR #265, opened 2016-02-20
Delegate to getter

oreissig:
Hi there,

what I basically want to do is this:
```diff
-@Target({ElementType.FIELD})
+@Target({ElementType.FIELD, ElementType.METHOD})
 public @interface Delegate {
```
so that I can
```groovy
@Delegate
Foo getSomeFoo() {
    // calcuate foo here
}
```

I rewrote the AST transform and put a lot of the FieldNode-specific context into a simple struct-like object (this rework is what makes the diff look so big, although it's just prepending `delegate.`). I then fill this object either with the field's data or the method's data. The actual code generation is not changed at all, except for the actual "getter", i.e. a field access or a method invocation.

However, I'm having trouble generating the correct AST. Currently the decompiled class looks like this:
```groovy
  @Delegate
  public DelegateFoo getStrategy() {
    // some stuff that works okay
  }

  public Object foo()
  {
    return ((DelegateFoo)getProperty("getStrategy")).foo(); return null;
  }
```
But there is no _property_ named `getStrategy`, the property is named `strategy` with a _method_ `getStrategy`. Yet I struggle to generate the correct expression to actually call the method as a method. Could someone please give me a hint?

Best regards

### DIFF
--- a/src/main/groovy/lang/Delegate.java
+++ b/src/main/groovy/lang/Delegate.java
@@ -26,11 +26,13 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-// TODO update javadoc for use on a getter method
 /**
- * Field annotation to automatically delegate part of the functionality of an owner class to the annotated field.
+ * Annotation to automatically delegate part of the functionality of an owner class to the
+ * annotated delegation target, i. e. a field or getter method's return value.
  * <p>
- * All public instance methods present in the type of the annotated field and not present in the owner class
+ * The delegate type is either the type of the annotated field or the return type of
+ * the annotated getter method.
+ * All public instance methods present in the delegate type and not present in the owner class
  * will be added to owner class at compile time. The implementation of such automatically added
  * methods is code which calls through to the delegate as per the normal delegate pattern.
  * <p>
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
@paulk-asert What do you think?

shils:
To answer your last question: it looks like the problem is that your delegating method calls use a variable with the delegate name as the receiver, even when the delegate is a method. In that case, the receiver should not be a variable, but rather a method call expression itself e.g. `this.getStrategy()`. Here's the block of code I'm referring to: 

```java
MethodCallExpression mce = callX(
        alsoLazy ? propX(varX("this"), delegate.delegateName.substring(1)) :
        varX(delegate.delegateName, correctToGenericsSpecRecurse(genericsSpec, delegate.type)),
        candidate.getName(),
        args);
```

oreissig:
Ah, thanks a lot for pointing that one out. I mindlessly went through the class and added `delegate.` but didn't notice this variable access at all.

oreissig:
I completed the necessary javadoc additions and cleaned up the remaining TODOs.

PascalSchumacher:
I created a jira issue for this: https://issues.apache.org/jira/browse/GROOVY-7769

oreissig:
Any opinions?
Not a good feature? Not a good implementation? I'm open to criticism :)

PascalSchumacher:
@oreissig I honestly do not know. Sorry. :(

I'm hoping that @paulk-asert will offer his opinion of this pull request.

paulk-asert:
will do soon - been a little busy ...

paulk-asert:
Any chance you can rebase against master?

oreissig:
all done

paulk-asert:
Merged with some doco tweaks, thanks!

PR #266, opened 2016-02-20
cache Gradle artifacts for Travis CI

oreissig:
Caching Gradle's user home (except the lock-file which gets touched every time) will increase the CI build's performance.


PascalSchumacher:
Thanks!

PR #267, opened 2016-02-21
Modified `MapWithDefault` to not store the default value

paplorinc:
Please see https://github.com/careercup/CtCI-6th-Edition-Groovy/blob/d116d65469bdf17d1e215e89f3e76ac3a97660a9/src/main/groovy/Ch04_TreesAndGraphs/_04_12_PathsWithSum.groovy#L21 for valid usecases

### DIFF
--- a/src/main/groovy/lang/MapWithDefault.java
+++ b/src/main/groovy/lang/MapWithDefault.java
@@ -29,67 +29,83 @@ import java.util.Set;
  * @since 1.7.1
  */
 public final class MapWithDefault<K, V> implements Map<K, V> {
-
     private final Map<K, V> delegate;
-    private final Closure initClosure;
+    private final Closure<V> initClosure;
 
-    private MapWithDefault(Map<K, V> m, Closure initClosure) {
+    private MapWithDefault(Map<K, V> m, Closure<V> initClosure) {
         delegate = m;
         this.initClosure = initClosure;
     }
 
-    public static <K, V> Map<K, V> newInstance(Map<K, V> m, Closure initClosure) {
+    public static <K, V> Map<K, V> newInstance(Map<K, V> m, Closure<V> initClosure) {
         return new MapWithDefault<K, V>(m, initClosure);
     }
 
+    @Override
     public int size() {
         return delegate.size();
     }
 
+    @Override
     public boolean isEmpty() {
         return delegate.isEmpty();
     }
 
+    @Override
     public boolean containsKey(Object key) {
         return delegate.containsKey(key);
     }
 
+    @Override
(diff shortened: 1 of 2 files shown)


jwagenleitner:
If I put entries in the map, even if I set them to the default value, I would expect to be able to iterate over them using methods such as `keySet()/entrySet()`.  However that would no longer work with this change in place.

paplorinc:
Hello @jwagenleitner :)

I agree, that's what I would expect in case of a normal map also.
But I would also expect `containsKey` to return `false`, if I haven't inserted anything, but that's not the current behavior.
```Groovy
def x = [:].withDefault {0}
println x.containsKey(1) // false
println x[1]                      // 0
println x.containsKey(1) // true
```
Therefore I think this new behavior is consistent with the class.

If you don't agree, I have a few alternative ideas, like:
* removeIfDefault(Object) // removes current key, if value is equal to default
* consolidate() // removes all default values
* constructor parameter `boolean storeDefaultValues = false`

jwagenleitner:
In your example `println x[1]` does insert an entry in the map.  If you used `keySet()/entrySet()` you'd be able to iterate.  I think it's expected that by calling `get` you are ensuring that entry will be added if it does not exist already.

Of the alternatives, I think the constructor parameter might work the best so then you could construct the map like this:

`[:].withDefault(false) { 7 }`

I'd recommend opening a JIRA for this feature, hopefully that would generate some more discussion.

paplorinc:
ok, sure, will do.

I know the `getAt` inserts it, that's why I gave it as an example - I find it confusing that a get will insert.

jwagenleitner:
I would agree that if there is a new option to not store the default value, then `get` should be changed to just return the value and not add to the delegate.

paplorinc:
... but that would be confusing also, because e.g.
```Groovy
def x = [:].withDefault { [] }
x[1] << 0
println x // [1:[0]]
```
wouldn't work anymore

jwagenleitner:
You're right, good point.  So even with the proposed change it not correct to say that the default wont be stored since it will for `get` just not `put`.

paplorinc:
Yeah, that's why I pointed it out.
Where should I open an issue for this?

PascalSchumacher:
@paplorinc You can open issue for groovy at https://issues.apache.org/jira/browse/GROOVY

paplorinc:
Opened an issue for this PR: https://issues.apache.org/jira/browse/GROOVY-7802

paulk-asert:
Closing this as there is now a revised PR over on the Apache Groovy site:
https://github.com/apache/groovy/pull/1692

PR #271, opened 2016-02-24
Link to MrHaki's blog in TupleConstructor javadoc.

pledbrook:
As discussed on Twitter, here's an example of linking to @mrhaki's blog from the javadocs so that people have ready access to more examples for certain classes. If this gets merged, I'll look into adding more.


PascalSchumacher:
Thanks for the pull request.

Cédric commented on it on the developer mailing list:

>I don't think linking to external resources like this is a good idea. We
don't own the end link, it can be dead very easily, especially in the
future. I would rather improve the documentation.

http://mail-archives.apache.org/mod_mbox/groovy-dev/201602.mbox/%3CCADQzvmknP1LUd_VLX8ymQtEJdXq5Mm0HzfSdNGMQPzi5%2BK_GaQ%40mail.gmail.com%3E

PR #274, opened 2016-02-29
Normalize groovysh command descriptions

Dispader:
A `:help` command issued from the Groovy Shell offers command descriptions.  The majority of commands match the same style, beginning with imperative verbs and not ending with punctuation.

I think that the description field is not only an extremely helpful feature, but makes the shell look like a very professional and polished product.

This commit attempts to change the style of all descriptions to match the style in the majority.

Hope this finds you all well, and hope it helps!


PascalSchumacher:
Merged. Thanks! :+1:

PR #276, opened 2016-03-02
GROOVY-7530 and GROOVY-7602: intersect and disjoint broken for non-Co…

paulk-asert:
…mparable objects (closes #276)


blackdrag:
+1

PR #277, opened 2016-03-02
CompilerConfiguration.output doesn't seem to be used for anything

yaccz:


PascalSchumacher:
@blackdrag @melix What do you think, should we merge this?

blackdrag:
The change is not a binary compatible one... I guess the error reporter should have used that output to print the error messages to. So we can remove the property, but not in 2.4. And frankly, for 2.5 I think it should be deprecated only

yaccz:
@PascalSchumacher done.

PascalSchumacher:
Thanks!

PR #278, opened 2016-03-03
More examples for AST transformation classes in Javadoc

mrhaki:
* Give readers of the Javadoc documentation more example code
* Examples based on MrHaki's Groovy Goodness blog posts

### DIFF
--- a/src/main/groovy/transform/BaseScript.java
+++ b/src/main/groovy/transform/BaseScript.java
@@ -85,7 +85,7 @@ import java.lang.annotation.Target;
  * the AST is currently implemented they are not accessible in the script body code.
  * </p>
  * <p>More examples:</p>
- * <pre>
+ * <pre class="groovyTestCase">
  * // Simple Car class to save state and distance.
  * class Car {
  *     String state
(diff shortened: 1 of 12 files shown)


PascalSchumacher:
Thanks!

I guess the examples are executable? If that is the case, could you please use `<pre class="groovyTestCase">` so that they are executed during the build (like Paul did in: https://github.com/apache/groovy/commit/aa38aca0b93036453df7859de3f2a9992ab40358)?

mrhaki:
Hi, I didn't know that appending the CSS class groovyTestCase would trigger a real test. I will add it where it is appropriate.

PascalSchumacher:
Thanks!

It's a well hidden feature ;) :(. 

With `<pre class="groovyTestCase">` you mark the code so that https://github.com/apache/groovy/blob/master/subprojects/groovy-test/src/main/groovy/groovy/util/JavadocAssertionTestBuilder.groovy will find it and build a test-suite. The suite is then run by https://github.com/apache/groovy/blob/master/src/test/MainJavadocAssertionTest.groovy. 

JavadocAssertionTestBuilder is distributed with Groovy, so everybody could use it, but I do not know if it is documented anywhere.

PascalSchumacher:
Thanks! :+1:

PascalSchumacher:
Thanks! :+1:

PR #281, opened 2016-03-05
Fix minor typo

Dispader:
"It's" is used for the contraction "it is."  "Its" is possessive.

I know it's minor, but it was bugging me.  So— for what it's worth... here's a one-character Pull.


PascalSchumacher:
Thanks! Even though it is minor it is still a welcome improvement.

PR #282, opened 2016-03-06
Some bulk refactorings

oreissig:
I went through the codebase of the root project and did the following refactorings:
- make private methods static when they are plain functions
- add @Deprecated when javadoc says the method is @deprecated
- remove unused code
- avoid synthetic accessors by making private fields/methods package-private
- make private fields final where possible
- fix some inconsistent whitespaces

None of this should change the behaviour in any way.

### DIFF
--- a/src/main/groovy/grape/GrapeIvy.groovy
+++ b/src/main/groovy/grape/GrapeIvy.groovy
@@ -76,11 +76,11 @@ class GrapeIvy implements GrapeEngine {
     Set<String> resolvedDependencies
     Set<String> downloadedArtifacts
     // weak hash map so we don't leak loaders directly
-    Map<ClassLoader, Set<IvyGrabRecord>> loadedDeps = new WeakHashMap<ClassLoader, Set<IvyGrabRecord>>()
+    final Map<ClassLoader, Set<IvyGrabRecord>> loadedDeps = new WeakHashMap<ClassLoader, Set<IvyGrabRecord>>()
     // set that stores the IvyGrabRecord(s) for all the dependencies in each grab() call
-    Set<IvyGrabRecord> grabRecordsForCurrDependencies = new LinkedHashSet<IvyGrabRecord>()
+    final Set<IvyGrabRecord> grabRecordsForCurrDependencies = new LinkedHashSet<IvyGrabRecord>()
     // we keep the settings so that addResolver can add to the resolver chain
-    IvySettings settings
+    final IvySettings settings
 
     public GrapeIvy() {
         // if we are already initialized, quit
--- a/src/main/groovy/lang/GroovyCodeSource.java
+++ b/src/main/groovy/lang/GroovyCodeSource.java
@@ -164,7 +164,7 @@ public class GroovyCodeSource {
         this(uri.toURL());
     }
 
-    public GroovyCodeSource(URL url) throws IOException {
+    public GroovyCodeSource(URL url) {
         if (url == null) {
             throw new RuntimeException("Could not construct a GroovyCodeSource from a null URL");
         }
--- a/src/main/groovy/lang/GroovyCodeSource.java
+++ b/src/main/groovy/lang/GroovyCodeSource.java
@@ -58,7 +58,7 @@ public class GroovyCodeSource {
     /**
      * The groovy source to be compiled and turned into a class
      */
-    String scriptText;
+    private String scriptText;
 
     /**
      * The certificates used to sign the items from the codesource
--- a/src/main/groovy/lang/GroovyCodeSource.java
+++ b/src/main/groovy/lang/GroovyCodeSource.java
@@ -58,7 +58,7 @@ public class GroovyCodeSource {
     /**
      * The groovy source to be compiled and turned into a class
      */
-    private String scriptText;
+    String scriptText;
 
     /**
      * The certificates used to sign the items from the codesource
--- a/src/main/groovy/lang/ObjectRange.java
+++ b/src/main/groovy/lang/ObjectRange.java
@@ -42,12 +42,12 @@ public class ObjectRange extends AbstractList implements Range {
     /**
      * The first value in the range.
      */
-    final Comparable from;
+    private final Comparable from;
 
     /**
      * The last value in the range.
      */
-    final Comparable to;
+    private final Comparable to;
 
     /**
      * The cached size, or -1 if not yet computed
@@ -57,7 +57,7 @@ public class ObjectRange extends AbstractList implements Range {
     /**
      * <code>true</code> if the range counts backwards from <code>to</code> to <code>from</code>.
      */
-    final boolean reverse;
+    private final boolean reverse;
 
     /**
      * Creates a new {@link ObjectRange}. Creates a reversed range if
--- a/src/main/groovy/lang/ObjectRange.java
+++ b/src/main/groovy/lang/ObjectRange.java
@@ -42,12 +42,12 @@ public class ObjectRange extends AbstractList implements Range {
     /**
      * The first value in the range.
      */
-    private final Comparable from;
+    final Comparable from;
 
     /**
      * The last value in the range.
      */
-    private final Comparable to;
+    final Comparable to;
 
     /**
      * The cached size, or -1 if not yet computed
@@ -57,7 +57,7 @@ public class ObjectRange extends AbstractList implements Range {
     /**
      * <code>true</code> if the range counts backwards from <code>to</code> to <code>from</code>.
      */
-    private final boolean reverse;
+    final boolean reverse;
 
     /**
      * Creates a new {@link ObjectRange}. Creates a reversed range if
--- a/src/main/groovy/lang/ObjectRange.java
+++ b/src/main/groovy/lang/ObjectRange.java
@@ -270,7 +270,8 @@ public class ObjectRange extends AbstractList implements Range {
     /**
      * protection against calls from Groovy
      */
-    private void setSize(int size) {
+    @Deprecated
+    void setSize(int size) {
         throw new UnsupportedOperationException("size must not be changed");
     }
 
--- a/src/main/groovy/lang/ProxyMetaClass.java
+++ b/src/main/groovy/lang/ProxyMetaClass.java
@@ -18,8 +18,6 @@
  */
 package groovy.lang;
 
-import java.beans.IntrospectionException;
-
 /**
  * As subclass of MetaClass, ProxyMetaClass manages calls from Groovy Objects to POJOs.
  * It enriches MetaClass with the feature of making method invokations interceptable by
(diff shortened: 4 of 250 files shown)


oreissig:
Rebased from master due to conflicts.

PascalSchumacher:
Thanks for the pull request. :)

I already cherry-picked a0aed6c6e523d176ebd480075d54aa9a3e19b472 which fixes some mistakes.

I support merging the rest of the commits, although I'm not sure how much a0aed6c6e523d176ebd480075d54aa9a3e19b472 or a032ed64a21434f12d2d6a560d55c7efb79eb986 add. With 0a78058ebc415e41b9523b77a42a92b907f169f1 there is a small change that somebody calls the methods from dynamic groovy, but of course it is important to remove unused stuff.

I'm thankful for opinions of other commiters on this.

oreissig:
This one particular `getSize()` method in 0a78058 I actually deleted initially, but added it again when I went through the diffs and noticed the comment above. I decided to make the method @Deprecated as a neat trick to indicate in an IDE that this should not be called if written by accident. If you don't like this trick, I'll just leave the method as is.

jwagenleitner:
For commit a032ed6 (avoid synthetic accessors by making private fields/methods package-private), I think removing `private` from members of private nested classes where the member is accessed by the top-level class would be ok.  However, I don't think members of top-level public classes should be changed from `private` to `package-private`.  If they weren't designed to be used by other classes in the package then they shouldn't be made visible just to avoid a synthetic method.  So overall I'd be -1 for this particular commit.

The other commits look good to me.

oreissig:
I also thought about the widening of scope, that may be unintended, but if there's a synthetic accessor method generated, it actually is made available inside the package anyway? Scope-wise there's not much of a difference between a package-private field and a package-private getter, or is there one?

jwagenleitner:
> but if there's a synthetic accessor method generated, it actually is made available inside the package anyway? 

To my knowledge the synthetic accessors are inserted by the Java compiler so that top-level and inner classes can access private members of each other.  From a coding perspective it is not really feasible for another class in the package to use these compiler generated accessors (i.e., they wont show up in IDE code completion).  By widening the scope of the members of the enclosing class encapsulation is being broken in order to avoid having a few synthetic accessors; the only benefit might be slight performance increase and that is even very questionable.  But many people work on the codebase and encapsulation is more important in my opinion in this context.  If someone working in the code see they can access the member they may think it was designed to be accessed by classes in the package when it actually was not.

> Scope-wise there's not much of a difference between a package-private field and a package-private getter, or is there one?

No I don't believe so.  In my original comment I was referring to synthetic accessors in general and not just those generated for methods.  I think again it comes back to encapsulation and breaking it just to avoid the synthetic accessors is not a good thing.

oreissig:
Good point about IDEs respecting the synthetic flag, thanks! So I'll revert all the scoping changes in non-inner classes. For private inner classes the scope should not matter that much, since they as a whole are not accessible to any other classes, therefore I leave my changes there.

oreissig:
Rebased from trunk due to conflicts.

The only cases left where I changed from private to package-private is for private inner classes now.

jwagenleitner:
Because many of the commits touch the same files it is difficult to review just the changes for the synthetic accessors/package-private.  I think for bulk changes like this, even if minor, separate PR's would be easier to review and manage.

PascalSchumacher:
I agree in the future I would prefer separate PR's.

Can you please squash c7fec364237ff496c88d7018b436254bb5b26945 and fcbe33e2e9abcaaed0d0c7b273ff810f5acbb9e6? Thanks!

oreissig:
Yes I already realized this was a bad idea. I would've love to just fixup the initial scope changing commit, but could not do so without major effort because it conflicts with the other commit adding final here and there, that has been done in the meantime :-/

PascalSchumacher:
Another problem with doing so many changes is that there are a lot of merge conflicts when I try to merge the changes into the 2_4_X branch, but without merging it's highly likely that there will be conflicts when backporting bugfixes. :(

oreissig:
If you want I can go through again and restructure the changes. How do you like it? Individual PRs for each of the commits (except the fixup)? Should those individual PRs only contain one commit each, or do you like it to be split up by some metric?

PascalSchumacher:
No further thinking about it I'm not sure if multiple pull request will really help in this case. I guess the pull request would have massive merge conflicts with each other.

I think 69ade9eb71829395e798fd8ba6395287bc44940a it the most valuable commit, because removing code dead code is good for readability and maintainability. I do not know if there is an easy way to untangle it from the rest? c3ad8298bd817e43fd0f56cb898fbce647002298 is also nice. Of course the other commit also have value.

jwagenleitner:
I think that bulk refactorings like these are best done one at a time (serially) with each as it's own PR with a single commit focused on the one change and agree that the unused code commit would be a good one to start with.  For me I'd wait for it to be merged before starting work on the next PR if lots of the same files are to be touched, and work on them in order of most likely to least likely to be merged.  I think that would avoid the headaches of dealing with merge conflicts.

PascalSchumacher:
@oreissig I rebased the remove unused code commit. Thanks!

src/main/groovy/lang/GroovyCodeSource.java
```diff
@@ -164,7 +164,7 @@ public GroovyCodeSource(URI uri) throws IOException {
         this(uri.toURL());
     }
 
-    public GroovyCodeSource(URL url) throws IOException {
```
paulk-asert: I only just spotted this now. While "unused" in the sense that this method didn't throw this exception anymore (I presume it did in the past), removing the exception breaks contract/source code compatibility. :-( I know it is a result of how checked exceptions are handled on the JVM but we should have not done this in a point release - all good to know in hindsight. Anyway, it isn't clear that reverting after a year is going to help things so I'm thinking of just documenting for now with a Jira issue. Apologies if this has already been raised elsewhere.

oreissig: d'oh, I totally forgot about that :(

src/main/groovy/grape/GrapeIvy.groovy
```diff
@@ -76,11 +76,11 @@ class GrapeIvy implements GrapeEngine {
...
     Set<String> downloadedArtifacts
     // weak hash map so we don't leak loaders directly
-    Map<ClassLoader, Set<IvyGrabRecord>> loadedDeps = new WeakHashMap<ClassLoader, Set<IvyGrabRecord>>()
+    final Map<ClassLoader, Set<IvyGrabRecord>> loadedDeps = new WeakHashMap<ClassLoader, Set<IvyGrabRecord>>()
     // set that stores the IvyGrabRecord(s) for all the dependencies in each grab() call
-    Set<IvyGrabRecord> grabRecordsForCurrDependencies = new LinkedHashSet<IvyGrabRecord>()
+    final Set<IvyGrabRecord> grabRecordsForCurrDependencies = new LinkedHashSet<IvyGrabRecord>()
     // we keep the settings so that addResolver can add to the resolver chain
-    IvySettings settings
+    final IvySettings settings
```
blackdrag: not sure those finals are ok, it means there will be no getter for these anymore. So I would like to hear from for example Paul that these are not supposed to have getter

src/main/groovy/lang/ObjectRange.java
```diff
@@ -270,7 +270,8 @@ protected int compareTo(Comparable first, Comparable second) {
     /**
      * protection against calls from Groovy
      */
-    private void setSize(int size) {
+    @Deprecated
+    void setSize(int size) {
         throw new UnsupportedOperationException("size must not be changed");
```
blackdrag: why is it deprecated again?

src/main/groovy/lang/ProxyMetaClass.java
```diff
@@ -44,7 +42,7 @@
     /**
      * convenience factory method for the most usual case.
      */
-    public static ProxyMetaClass getInstance(Class theClass) throws IntrospectionException {
+    public static ProxyMetaClass getInstance(Class theClass) {
```
blackdrag: removal of checked exception in public method

src/main/groovy/lang/ProxyMetaClass.java
```diff
@@ -53,7 +51,7 @@ public static ProxyMetaClass getInstance(Class theClass) throws IntrospectionExc
     /**
      * @param adaptee the MetaClass to decorate with interceptability
      */
-    public ProxyMetaClass(MetaClassRegistry registry, Class theClass, MetaClass adaptee) throws IntrospectionException {
```
blackdrag: removal of checked exception in public method

src/main/org/codehaus/groovy/reflection/GeneratedMetaMethod.java
```diff
@@ -162,7 +162,7 @@ public static void saveDgmInfo(List<DgmMethodRecord> records, String file) throw
             out.close();
         }
 
-        public static List<DgmMethodRecord> loadDgmInfo() throws IOException, ClassNotFoundException {
+        public static List<DgmMethodRecord> loadDgmInfo() throws IOException {
```
blackdrag: removal of checked exception from public api

src/main/org/codehaus/groovy/tools/DgmConverter.java
```diff
@@ -37,7 +37,7 @@
 
 public class DgmConverter implements Opcodes {
 
-    public static void main(String[] args) throws IOException, ClassNotFoundException {
+    public static void main(String[] args) throws IOException {
```
blackdrag: even though this is not likely to cause a problem, still removal of checked exception on public method

src/main/org/codehaus/groovy/tools/shell/IO.java
```diff
@@ -130,7 +130,7 @@ public boolean isDebug() {
     /**
      * Flush both output streams.
      */
-    public void flush() throws IOException {
+    public void flush() {
```
blackdrag: removal of checked exception on public method

src/main/org/codehaus/groovy/vmplugin/v7/IndyArrayAccess.java
```diff
@@ -96,29 +96,6 @@ private static MethodHandle buildSetter(Class arrayClass){
...
-        if (array instanceof char[]) return ((char[])array).length;
-        if (array instanceof short[]) return ((short[])array).length;
-        if (array instanceof int[]) return ((int[])array).length;
-        if (array instanceof long[]) return ((long[])array).length;
-        if (array instanceof float[]) return ((float[])array).length;
-        if (array instanceof double[]) return ((double[])array).length;
-        return 0;
-    }
-
-    private static int normalizeIndex(Object array, int i) {
```
blackdrag: normalizeIndex is used via MethodHandle in line 39. you cannot remove these

PR #283, opened 2016-03-06
avoid rebuilding bootstrapJar

oreissig:
When you just worked on tests in the root project, the bootstrapJar (as well as all groovy classes in the root project) were recompiled. I changed to behaviour to only trigger this process for non-test source changes.


PascalSchumacher:
Thanks! :+1: 

Great change, especially nice for running `gradle testSingleXYZ'.

PR #284, opened 2016-03-06
GROOVY-7640: AST Builder should include superclass properties

PascalSchumacher:
The code is really ugly, but this is caused by `getPropertyInfoFromClassNode` with is in `BuilderASTTransformation` but should have been in `ExternalStrategy`, because it is only used by  `ExternalStrategy`. Now it can not be moved/modified for this feature without breaking binary compatibility.

The pull request is still missing documentation. I will add that if it is accepted.


PascalSchumacher:
Added a commit with the documentation.

PR #285, opened 2016-03-08
GROOVY-6514: groovysh: Add :grab command

Dispader:
This commit adds a first revision `:grab` command to the Groovy Shell.

@tkruse Sorry— I know @PascalSchumacher was asking me to fork from your fork, but I couldn't quite square that circle.  So I created a new branch, and added you as a collaborator (you should be able to force push here, so you can rebase the commits if you like, as well).

In any case, we should be able to see some diffs here and at least critique the approach.  I've tested it, and it does appear to work well.

do not like: The only reliable way I could find to send a message to the existing `ImportCommand` was to invoke property changes (to which it's registered as a listener).  We could make `ImportCommand` open by registering another type of event listener, but this seems like overkill to me.

I did add the bit that verifies that the import completion property has been set before sending the disable/enable events (to force the reload).

Let me know what you think.

Cheers,
Jake

### DIFF
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
@@ -1,0 +1,92 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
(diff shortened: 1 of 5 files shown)


subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
```diff
@@ -0,0 +1,92 @@
...
+ */
+class GrabCommand
+    extends CommandSupport
+{
+    public static final String COMMAND_NAME = ':grab'
+
+    private static final java.util.prefs.Preferences PREFERENCES =
+        java.util.prefs.Preferences.userRoot().node('/org/codehaus/groovy/tools/shell')
+
+    public GrabCommand(org.codehaus.groovy.tools.shell.Groovysh shell) {
```
Dispader: ...and I already found an error in reviewing this...

Dispader:
I already found one explicit package that can be omitted here... thinking of waiting for the CI job to finish checks, closing this Pull, and making another.

subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
Dispader: The explicit `org.codehaus.groovy.tools.shell.Groovysh` shouldn't be required here.  After the CI build completes, I'm going to close this Pull, kill the branch, re-test, and push it back up with this cleaned up.

Dispader:
Closing to clean up import.  (With many apologies.)

PascalSchumacher:
>Closing to clean up import. (With many apologies.)

No need to apologize.

You can also update a pull request by amending a commit and then force pushing. e.g. `commit --amend`, `push -f``.

PR #286, opened 2016-03-08
GROOVY-6514: groovysh: Add :grab command

Dispader:
Take two— notes mostly copied from the first Pull request.

This commit adds a first revision `:grab` command to the Groovy Shell.

@tkruse Sorry— I know @PascalSchumacher was asking me to fork from your fork, but I couldn't quite square that circle.  So I created a new branch, and added you as a collaborator (you should be able to force push here, so you can rebase the commits if you like, as well).

In any case, we should be able to see some diffs here and at least critique the approach.  I've tested it, and it does appear to work well.

do not like: The only reliable way I could find to send a message to the existing `ImportCommand` was to invoke property changes (to which it's registered as a listener).  We could make `ImportCommand` open by registering another type of event listener, but this seems like overkill to me.

I did add the bit that verifies that the import completion property has been set before sending the disable/enable events (to force the reload).

Two changes: I removed an unneeded package specification, and replaced the completer (which gives the user a nice ' ' following the command, but doesn't suggest files as completions for `:grab`.

Let me know what you think.

Cheers,
Jake

### DIFF
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommand.groovy
@@ -1,0 +1,74 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.codehaus.groovy.tools.shell.commands
+
+import groovy.grape.Grape
+import jline.console.completer.Completer;
+
+import org.codehaus.groovy.tools.shell.CommandSupport
+
+import java.util.List;
+
+import org.codehaus.groovy.tools.GrapeUtil
+import org.codehaus.groovy.tools.shell.Groovysh
+
+/**
+ * The 'grab' command.
+ *
+ * @author <a href="mailto:jake.gage@gmail.com">Jake Gage</a>
+ */
+class GrabCommand extends CommandSupport {
+
+    public static final String COMMAND_NAME = ':grab'
+
+    public GrabCommand(Groovysh shell) {
+        super(shell, COMMAND_NAME, ':g')
+    }
+
+    @Override protected List<Completer> createCompleters() { [ null ] }
+
+    @Override Object execute(List<String> args) {
+        validate(args)
+        grab(dependency(args))
+        shell.packageHelper.reset()
+    }
+
+    private void validate(List<String> args) {
+        if ( args?.size() != 1 || 
+             !( args[0] ==~ /^(\w|\.|-)+:(\w|\.|-)+(\w|\.|-)(:+(\w|\.|-|\*)+){0,2}$/ ) ) {
+            fail("usage: @|bold ${COMMAND_NAME}|@ ${usage}")
+        }
+    }
+
+    private String dependency(List<String> args) {
+        validate(args)
+        args[0]
+    }
+
+    private Map<String, Object> dependencyMap(String dependency) {
+        GrapeUtil.getIvyParts(dependency)
+    }
+
+    private void grab(String dependency) {
+        Grape.grab([classLoader: shell.interp.classLoader.parent,
+                    refObject: shell.interp],
+                   dependencyMap(dependency))
+    }
+
+}
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/DefaultCommandsRegistrar.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/DefaultCommandsRegistrar.groovy
@@ -26,6 +26,7 @@ import org.codehaus.groovy.tools.shell.commands.DisplayCommand
 import org.codehaus.groovy.tools.shell.commands.DocCommand
 import org.codehaus.groovy.tools.shell.commands.EditCommand
 import org.codehaus.groovy.tools.shell.commands.ExitCommand
+import org.codehaus.groovy.tools.shell.commands.GrabCommand
 import org.codehaus.groovy.tools.shell.commands.HelpCommand
 import org.codehaus.groovy.tools.shell.commands.HistoryCommand
 import org.codehaus.groovy.tools.shell.commands.ImportCommand
@@ -76,6 +77,7 @@ class DefaultCommandsRegistrar
                 new HistoryCommand(shell),
                 new AliasCommand(shell),
                 new SetCommand(shell),
+                new GrabCommand(shell),
                 // does not do anything
                 //new ShadowCommand(shell),
                 new RegisterCommand(shell),
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/PackageHelper.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/PackageHelper.groovy
@@ -23,4 +23,7 @@ interface PackageHelper {
     public static final String IMPORT_COMPLETION_PREFERENCE_KEY = 'disable-import-completion'
 
     Set<String> getContents(final String packagename)
+
+    void reset()
+
 }
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/PackageHelperImpl.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/util/PackageHelperImpl.groovy
@@ -44,10 +44,18 @@ class PackageHelperImpl implements PreferenceChangeListener, PackageHelper {
 
     PackageHelperImpl(final ClassLoader groovyClassLoader=null) {
         this.groovyClassLoader = groovyClassLoader
-        if (! Boolean.valueOf(Preferences.get(IMPORT_COMPLETION_PREFERENCE_KEY))) {
-            rootPackages = initializePackages(groovyClassLoader)
-        }
+        initializePackages()
         Preferences.addChangeListener(this)
+    }
+
+    void reset() {
+        initializePackages()
+    }
+
+    private void initializePackages() {
+        if (! Boolean.valueOf(Preferences.get(IMPORT_COMPLETION_PREFERENCE_KEY))) {
+            rootPackages = getPackages(this.groovyClassLoader)
+        }
     }
 
     @Override
@@ -56,12 +64,12 @@ class PackageHelperImpl implements PreferenceChangeListener, PackageHelper {
             if (Boolean.valueOf(evt.getNewValue())) {
                 rootPackages = null
             } else if (rootPackages == null) {
-                rootPackages = initializePackages(groovyClassLoader)
+                initializePackages()
             }
         }
     }
 
-    static Map<String, CachedPackage> initializePackages(final ClassLoader groovyClassLoader) throws IOException {
+    private static Map<String, CachedPackage> getPackages(final ClassLoader groovyClassLoader) throws IOException {
         Map<String, CachedPackage> rootPackages = new HashMap()
         Set<URL> urls = new HashSet<URL>()
 
--- a/subprojects/groovy-groovysh/src/main/resources/org/codehaus/groovy/tools/shell/commands/GrabCommand.properties
+++ b/subprojects/groovy-groovysh/src/main/resources/org/codehaus/groovy/tools/shell/commands/GrabCommand.properties
@@ -1,0 +1,21 @@
+#
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+#
+command.description=Add a dependency to the shell environment
+command.usage=<group>:<module>[:<version>|*[:classifier]
+command.help=Fetch a dependency, and add it to the Groovy Shell environment.
--- a/subprojects/groovy-groovysh/src/spec/doc/groovysh.adoc
+++ b/subprojects/groovy-groovysh/src/spec/doc/groovysh.adoc
@@ -202,6 +202,7 @@ Available commands:
   :history   (:H ) Display, manage and recall edit-line history
   :alias     (:a ) Create an alias
   :set       (:= ) Set (or list) preferences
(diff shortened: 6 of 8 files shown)


PascalSchumacher:
Thanks for creating this pull request. :) 

>but doesn't suggest files as completions for :grab.

If I understand this correctly it is an advantage because you do not want files as completions for grab.

I think you should also allow the two argument grab like https://github.com/apache/groovy/pull/104/files and add a four argument version with the fourth argument being the classifier: http://docs.groovy-lang.org/latest/html/documentation/grape.html#Grape-Detail

tkruse:
So I rebased my 8 long-standing PRs by the way.

Regarding this PR, compared to mine at #104, it seems that the allowed values behind :grab seems implemented differently, I think my approach in GrabCommand.groovy:buildSpec might be more versatile but maybe they are equivalent.

My Test class GrabCommandTest.groovy defines more test cases for that purpose. Adding those tests to this PR might help comparing the syntax capabilites.

Regarding the import completer, I believe my rescan command invokes a new reset() function in PackageHelper. I think you can copy those changes from #104, even if you leave out the rescan command.

tkruse:
The classloading seems to work btw, I can reproduce the behavior, so hopefully that means we can have the :grab command soon.

Still no solution for this one https://www.mail-archive.com/dev@groovy.incubator.apache.org/msg00983.html

tkruse:
btw, I notice the imports suggestions are not sorted, not sure since when. Looks messy.

Dispader:
@tkruse You're absolutely right, sir— the `buildSpec` check is more versatile: the only reason I didn't pick it up was that it allowed the `version` component to be empty, and I wasn't sure how the modified `GrabUtil` and `Grab` calls would deal with it.  I only left it out because of my own ignorance, there.  I'll definitely add the input tests from your branch here.

I saw the `reset()` method over in `PackageHandler`!  It'd be a heck of a lot nicer than having to toggle the event— but I couldn't get it to work out for re-loading completions.  What I though* have been happening was that the wrong instance of the class was getting the message— but I think* that's wrong now.  I really think `reset()` is a better way to go.

I don't know how we're feeling, but Pull #104 might wind up being the way we want to go, anyhow; but I'll chug along and put some of the changes above in here.

Dispader:
Oh... @tkruse ... should we take a look at the package import sorting while we're in here?

Dispader:
Yup— @tkruse 's `reset()` solution is far nicer here.

Dispader:
...and the tests fail if we don't implement `reset()` in the test...

The more I look, the more I'm liking Pull #104 .

Dispader:
Wow... that was frustrating.

I can't seem to find a way to get the general Grab() resolver to resolve remotes with only the `module` set in the dependency `Map`.  I don't... think it's possible?

So, it's looking more like we need to construct something that the Grape helper can get to— executing in a sub-shell call loses the classloader.

I'm going to try constructing the canonical short name again and reconstructing it with `GrapeUtil` again before the call.

PascalSchumacher:
I'm not a grape expert. In my opinion supporting the three and four argument version is sufficient. What do you think @paulk-assert ?

PascalSchumacher:
@Dispader Sorry, there is no one argument grab version: https://github.com/apache/groovy/commit/fd567f7e5cd9af35bd53a19af034a5520665b121

paulk-asert:
It would be nice if the same syntax as supported by `GrapeMain#install` was supported:
`<group> <module> [<version> [<classifier>]]`

paulk-asert:
It would also be nice to be able to specify a GrabResolver but that could be done as a separate PR.

Dispader:
Thank you all!

The problem here is that I've been wrong-headed about this— I've been trying to reproduce the behavior of Gradle, using the notation that gets passed in `dependency` sections.  (A mistake that Pull #104 isn't making.)

The way that this code was getting that job done was by putting the single-argument version of `:grab` through the `GrabUtil.getIvyParts` method.  I recall using the notation `@Grab('group:module:version')`.

But a Groovy Shell user can just type `@Grab('...')` right now— *that*'s what you were going for with the `:reload` command, wasn't it, @tkruse ?

Dispader:
One can't register as a listener for `ClassLoader` changes, can one?

tkruse:
Yes, the :rescan command was meant as a trigger to reload the definitions for import completion, after any action the user does to add files to the classloader (such as manually invoking grapes).

The way I knew about using grab manually in groovysh was this:
http://tech.puredanger.com/2010/03/01/groovy-grape-groovysh/
but apparently this may work too:
http://www.groovy-lang.org/mailing-lists.html#nabble-td4632257

However both are too clumsy for my taste to advertize them to novice users (in particular when having to additionally call :rescan next for import completion), hence my wish for a :grab command.

Is the only remaining problem the invocation of GrapeUtil.getIvyParts() and passing only a module name? 
I don't see too much value in reusing this function anyway. It seems that function is not particularly useful for the interaction case anyway, it returns an empty map in error cases. (Also the function is undocumented and I guess untested, as too much of the core groovy code).

I would recommend recreating the expression parsing logic even if that means having duplicate code.
In #104 I was not even aware of GrapeUtil.getIvyParts, else I would probably have copy&pasted it, and made the modifications needed for Groovysh.

Dispader:
Yeah, the place that got me started on this track was Grab command was the [Grape documentation](http://docs.groovy-lang.org/latest/html/documentation/grape.html#Grape-UsingGrapeFromtheGroovyShell).  That led me to the same conclusion sounds like you came to: it's confusing for new users, and still quite a bit of keyboarding even when you know what you want.  It's just a bit ham-handed for my tastes, as well.

The approach I took was that I was looking for something that was playing the role of supplying things that would be understood by the [Grab Annotation](http://docs.groovy-lang.org/latest/html/documentation/grape.html#Grape-Annotation).  That seemed like something that might be familiar, and didn't ask too-too much from the user.

Thus, my original tests for acceptance were 1) passing in things that the annotation could accept and 2) getting added packages to the `import` command's completion.  That's where the `GetIvyParts()` bit came in— the `Grab` Annotation class was looking for either the map... or the short form, but neither is really open to being called from the outside world, but `GrapeUtil.getIvyParts() ` seemed to be.

The `GrapeUtil.getIvyParts()` call seems to chop either the `abbreviated` forms (e.g.: `'com.google.guava:guava:19.0'`, things that come back from [these kinds of searches](http://mvnrepository.com/artifact/com.google.guava/guava/19.0#gradle)), and value-type part calls (e.g.: `:grab com.google.guava guava` assumes the only two-part dependency that Grape would take, `group` and `module`; and `:grab com.google.guava guava 19.0` would take the only three-part.)

I don't know if that's the _right_ way to go, but it's what I was going for.

(I see the `:rescan` command gives a lot of power to the user in it's own right (it'll at least... make the runtime environment look consistent?)  And I'd _love_ to see the `rescan()` method called on all changes to the parent classloaders of the shell, but I'm not finding a way to passively listen for changes to a classloader tree.)

(FWIW, I figured out how to pull and build the branch from Pull #104 — and spent a few hours trying to modify the `buildSpec` to take the Annotation short-form, but I ran into a problem: the `buildSpec` there presumes we can issue the arguments to `Grape.grab()` from a shell runtime environment.  So, we can modify it, but we can't then pass the shell's parent classloader, so `rescan()` wouldn't pick it up (without, of course, modifying Grape).  And I get the feeling that's what's been holding up #104 from being merged.)

So— sorry for the long post— and sorry it's been a few days since I could get back to this, but what do folks think of trying for this as a target contract?

- `:grab` with one string argument
    - attempts the Annotation short-form (Gradle form)
- `:grab` with two string arguments
    - that don't represent Groovy Maps
        - attempts the two-argument version of the Grab install in fd567f7
   - that do represent Groovy Maps (e.g.:
       - `:grab [group: 'com.google.guava'] [module: 'guava']` or
       - `:grab [[group: 'com.google.guava'] [module: 'guava']]`
            - attempts the two-argument map version of the Grab install in fd567f7
- same as above, but for the three-argument constructor with `version`
- same as above, but for the four-argument constructor with `classifier`

If I hear a `yea` or two, I'll start work on the above.

tkruse:
Yes, +#104 could not be merged because it had to modify Grape to work with a rescan.

>:grab with one string argument
> attempts the Annotation short-form (Gradle form)

Consider the unit tests from #104, they test these arguments:
```
// one argument
assert 'group:\'\', module:\'foo\', version:\'\'' == grabCommand.buildSpec(['foo'])
assert 'group:\'\', module:\'foo\', version:\'\'' == grabCommand.buildSpec([':foo'])
assert 'group:\'\', module:\'foo\', version:\'\'' == grabCommand.buildSpec([':foo:'])
assert 'group:\'foo\', module:\'bar\', version:\'\'' == grabCommand.buildSpec(['foo:bar'])
assert 'group:\'foo\', module:\'bar\', version:\'baz\'' == grabCommand.buildSpec(['foo:bar:baz'])
// two arguments
assert 'group:\'foo\', module:\'bar\', version:\'\'' == grabCommand.buildSpec(['foo', 'bar'])
// three arguments
assert 'group:\'foo\', module:\'bar\', version:\'baz\'' == grabCommand.buildSpec(['foo', 'bar', 'baz'])
```
Not sure if that's what you were going for. But a subset would also work, this can later be extended as users play with it and come up with ideas for how to speed up their workflows.

#104 does not consider the 4-argument case merely because I did not know about that possibility.

PascalSchumacher:
In my opinion `:grab 'foo:bar:1.0'` should work as the absolute minimum. Then I would add support  for `:grab 'foo:bar:1.0:jdk7'` and for `:grab 'foo:bar'`. Anything else e.g. maps is optional.

Dispader:
All right, then, I've sketched out test specification on a test truss class.  It tests the private `dependencyMap(String)` method just now, which I'll fix and change to a test on the caller.  For now, it should demonstrate a test of producing a `Grab`-consumable `Map` from a single-argument input.  Hopefully that'll be a bit more clear.

I'll propose:

```Groovy
import spock.lang.Specification

class GrabCommandSpec extends Specification {

    // this test acts on a private method -- if this test ever fails, DELETE IT
    def 'creates valid depenency maps when passed valid inputs'() {
        expect:
            GrabCommand.dependencyMap(input).group == group
            GrabCommand.dependencyMap(input).module == module
            GrabCommand.dependencyMap(input).version == version
            GrabCommand.dependencyMap(input).classifier == classifier
        where:
            input                                  | group             | module     | version | classifier
            'net.sf.json-lib:json-lib:2.2.3:jdk15' | 'net.sf.json-lib' | 'json-lib' | '2.2.3' | 'jdk15'
            'net.sf.json-lib:json-lib:2.2.3'       | 'net.sf.json-lib' | 'json-lib' | '2.2.3' | null
            'net.sf.json-lib:json-lib'             | 'net.sf.json-lib' | 'json-lib' | '*'     | null
    }

    // this test acts on a private method -- if this test ever fails, DELETE IT
    def 'throws IllegalArgumentException when passed invalid inputs'() {
        when:
            GrabCommand.dependencyMap(input)
        then:
            thrown IllegalArgumentException
        where:
            input << 'net.sf.json-lib:json-lib:2.2.3:jdk15'
    }

}
```

One note: the test differs from the `buildSpec` test in #104 because direct calls to `Grape.grab()` require the `Map` inputs.

Iffn we think this describes a decent behavior, I'll shine it up and move it into the test tree.  (I'll also convert it to a `GroovyTestCase` and subclass `CommandTestSupport`, to keep it consistent with the rest of the test tree.)

tkruse:
Spock spec looks nice, but groovy core code should not depend on Spock, btw, so you cannot actually use it in the PR.

In contrast to Pascal I really would love to see support for providing just the module name. The main purpose of groovysh is to enable users to very quickly hack something instead of firing up an IDE. In particular since we cannot offer tab completion. It feewls wrong to have to type out `:grab org.springframework.boot/spring-boot-starter-web` when just `:grab spring-boot-starter-web` is realistically sufficient.

Dispader:
@tkruse As I said, I've no intention of including a Spock test here— and I'd translate it to a `CommandTestSupport`.  The only thing I was going for with the above was to get us to hone in on what we'd like to send to `Grape.grab()`.  Also why I just added as a comment here.

I think we agree on the spirit of the thing: the less a user has to type when they're extemporizing with the shell, the better the experience is going to be for everyone.  (And I couldn't agree more: I'm half into trying to create a completer that can pull in some kind of `group` and `module` for tab completion.)

But­— I don't really understand "to have to type out :grab org.springframework.boot/spring-boot-starter-web when just :grab spring-boot-starter-web is realistically sufficient".

I... really don't know any way to iden a `module` without the `group` at this time.

This may be my fault, and I'll try to be clearer.  The `Grape.grab()` method I'm trying to use won't resolve `module` without `group`

```
groovy:000> import groovy.grape.Grape
===> groovy.grape.Grape
groovy:000> dependencyMap = [module: 'spring-boot-starter-web']
===> [module:spring-boot-starter-web]
groovy:000> classloader = ClassLoader.getSystemClassLoader()
===> sun.misc.Launcher$AppClassLoader@33909752
groovy:000> grapeArgs = [classLoader: classloader]
===> [classLoader:sun.misc.Launcher$AppClassLoader@33909752]
groovy:000> Grape.grab(grapeArgs, dependencyMap)
ERROR java.lang.RuntimeException:
Error grabbing Grapes -- [unresolved dependency: #spring-boot-starter-web;latest.default: not found]
groovy:000> dependencyMap = [group: 'org.springframework.boot', module: 'spring-boot-starter-web']
===> [group:org.springframework.boot, module:spring-boot-starter-web]
groovy:000> Grape.grab(grapeArgs, dependencyMap)
===> null
groovy:000> 
```

I got there by tracing the `Grab` annotation.  But even the `grab` command line utility doesn't seem to be able to resolve just the `module`

```
] grape install spring-boot-starter-web
install requires two to four arguments: <group> <module> [<version>] [<classifier>]
```

tkruse:
Haha, #104 tricks Grape! Try this:
```
groovy:000> import groovy.grape.Grape
groovy:000> Grape.grab(group:'', module:'joda-time', version:'')
```

It seems passing an empty string bypasses some validation maybe.

I don't know if I had done this intentionally in the past, or I never tried to pass a map without the empty string values.

Dispader:
Weeeeird!

I can't get it to work, but I see the confusion now!  And if the empty-map-element works like above, the `buildSpec` in #104 would totally* work.  (And we can easily build the empty elements into all the forms.)  Very* nice catch.

Okay— so­— I can't* get the above to work yet

```
groovy:000> import groovy.grape.Grape
===> groovy.grape.Grape
groovy:000> Grape.grab(group:'', module:'joda-time', version:'')
ERROR java.lang.RuntimeException:
Error grabbing Grapes -- [unresolved dependency: #joda-time;latest.default: not found]
groovy:000> Grape.grab(group:'joda-time', module:'joda-time', version:'')
===> null
groovy:000> Grape.grab(group:'', module:'joda-time', version:'')
ERROR java.lang.RuntimeException:
Error grabbing Grapes -- [unresolved dependency: #joda-time;latest.default: not found]
groovy:000> 
```

  but something*'s going on here.  (The call-with-group in the middle was I thought we might need to have the package in the local Grab cache to pull it by module.)  But `Grab` over there is working, and I think it'd be great* to have a one-arg option (even if it does need to be loaded with a cache, it'd make using groovysh fast.)

  So— can you help try to figure what's special... like, how you're able to trick Grab in the shell over there?  If we can figure it out, I can try to reproduce it here.  I can't help but think that, since it does* work somewhere, we can make it reproducible in the code.

Here's what my env in telling me I'm running, iffn it helps

```shell
~){azazel}07:00P] groovy --version
Groovy Version: 2.4.6 JVM: 1.8.0_60 Vendor: Oracle Corporation OS: Mac OS X
(~){azazel}07:00P] groovysh --version
@|green Groovy Shell|@ 2.4.6
```

tkruse:
Hm, actually it does not work me for anymore, either.  I thought it was the empty strings that made it work back then, but apparently that wasn't it.
Maybe it depends on the remote server returning a result. Maybe that was the case for me back then, and they switched it off at maven central.

So I suggest going with Pascals suggestion to require group and module, and maybe when I find more time, I can figure out what went on back when I implemented #104

Dispader:
All right— sorry about my long absence from this Pull:  I had some problems isolating tests from the collaborators, then managed to let myself fall off the face of the Earth, there.

All that said, I believe we've been able to isolate the tests from collaborators, and get tests that look trustworthy on the public command.

This Pull is rebased from `origin/master` as of today, and implements @PascalSchumacher 's https://github.com/apache/groovy/pull/286#issuecomment-196498228 (great suggestion for the minimal interface).  It lives up to the https://github.com/apache/groovy/pull/286#issuecomment-197187196 (external truss contract here), but uses a https://github.com/Dispader/groovy/blob/add_grab_command/subprojects/groovy-groovysh/src/test/groovy/org/codehaus/groovy/tools/shell/commands/GrabCommandTest.groovy ( GroovyTestCase with collaborator isolation).  It exposes a variant of @tkruse 's `reset()` for `PackageHelper`  It leaves off exposing a package reloading command, and GrabResolver specification for another Pull.

What d'y'all think?

PascalSchumacher:
It looks fine as far as I can tell.

tkruse:
+1

PascalSchumacher:
Thanks! :+1:

PR #287, opened 2016-03-08
Javalike method pointer groovy 7772

uehaj:
This is a change for method pointer semantics in groovy discussed in following issue.
https://issues.apache.org/jira/browse/GROOVY-7772

ref.
http://docs.groovy-lang.org/latest/html/documentation/index.html#method-pointer-operator
https://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html#type

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/MethodClosure.java
+++ b/src/main/org/codehaus/groovy/runtime/MethodClosure.java
@@ -20,6 +20,7 @@ package org.codehaus.groovy.runtime;
 
 import groovy.lang.Closure;
 import groovy.lang.MetaMethod;
+import groovy.lang.MissingMethodException;
 
 import java.util.List;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


src/spec/test/CoercionTest.groovy
```diff
@@ -214,7 +214,7 @@ boolean doFilter(String s) { s.contains('G') }
 Predicate filter = this.&doFilter
 assert filter.accept('Groovy') == true
 
-Greeter greeter = GroovySystem.&getVersion
+Greeter greeter = new GroovySystem().&getVersion
```
blackdrag: this change does not make sense to me. getVersion is a static method, so GroovySystem.&getVersion must be fine

src/main/org/codehaus/groovy/runtime/MethodClosure.java
```diff
@@ -62,9 +74,34 @@ public String getMethod() {
...
+                return InvokerHelper.invokeMethod(insertedReceiver, method, newArgs);
+            }
+            return InvokerHelper.invokeMethod(arguments, method, new Object[]{});
+        }
         return InvokerHelper.invokeMethod(getOwner(), method, arguments);
     }
 
+    @SuppressWarnings("unchecked")
+    public Object call(Object... args) {
+        try {
```
blackdrag: why do we need this method?

paulk-asert:
I merged in the test case additions commit from this PR. The other commits have been superseded. Thanks!

PR #290, opened 2016-03-12
make private methods static when they are plain functions

oreissig:
As discussed in PR #282 I repropose the changes from there in individual pull requests.

### DIFF
--- a/subprojects/groovy-jsr223/src/main/java/org/codehaus/groovy/jsr223/GroovyScriptEngineImpl.java
+++ b/subprojects/groovy-jsr223/src/main/java/org/codehaus/groovy/jsr223/GroovyScriptEngineImpl.java
@@ -416,7 +416,7 @@ public class GroovyScriptEngineImpl extends AbstractScriptEngine implements Comp
     }
 
     // generate a unique name for top-level Script classes
-    private synchronized String generateScriptName() {
+    private static synchronized String generateScriptName() {
         return "Script" + (++counter) + ".groovy";
     }
 
@@ -454,7 +454,7 @@ public class GroovyScriptEngineImpl extends AbstractScriptEngine implements Comp
         return Script.class.getClassLoader();
     }
 
-    private String readFully(Reader reader) throws ScriptException {
+    private static String readFully(Reader reader) throws ScriptException {
         char[] arr = new char[8 * 1024]; // 8K at a time
         StringBuilder buf = new StringBuilder();
         int numChars;
--- a/src/main/groovy/beans/BindableASTTransformation.java
+++ b/src/main/groovy/beans/BindableASTTransformation.java
@@ -182,7 +182,7 @@ public class BindableASTTransformation implements ASTTransformation, Opcodes {
     /*
      * Wrap an existing setter.
      */
-    private void wrapSetterMethod(ClassNode classNode, String propertyName) {
+    private static void wrapSetterMethod(ClassNode classNode, String propertyName) {
         String getterName = "get" + MetaClassHelper.capitalize(propertyName);
         MethodNode setter = classNode.getSetterMethod("set" + MetaClassHelper.capitalize(propertyName));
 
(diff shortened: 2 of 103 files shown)


oreissig:
Travis timed out. I don't think my changes introduced a performance penalty that's responsible for that :)

PascalSchumacher:
Yes, that is highly unlikely ;) :).

I you like you can always make travis retest the pull request be using `commit --amend` e.g. with an empty commit and then force pushing e.g.`push -f`.

oreissig:
Thanks for the tip using `--amend`!

jwagenleitner:
I'm curious, what is the benefit of making this change?

pfmiles:
I'm worrying about that this change might break some reflection code.

oreissig:
The benefit is limiting the possible scope of influence a certain piece of code can have, similar to making stuff private or final. It does not have an immediate effect, but helps distinguishing different concerns (e.g. most private static methods are just little helpers that were not generic enough to be part of a standard library, but nevertheless contain functionality that is not dependent upon the particular object they are used by). This separation of concerns can help programmers comprehend the code better and may even help the JIT make certain assumptions optimizing bytecode.

I admit that any way the impact will probably not be substantial.

jwagenleitner:
Thanks @oreissig I appreciate the explanation.  I agree that it can be a useful indicator for those reading the code.  For me I am not sure if that benefit is worth the change (though I'm kind of leaning towards yes), I would be curious to know what others think.

subprojects/groovy-jsr223/src/main/java/org/codehaus/groovy/jsr223/GroovyScriptEngineImpl.java
```diff
@@ -427,7 +427,7 @@ private Object callGlobal(String name, Object[] args, ScriptContext ctx) {
     }
 
     // generate a unique name for top-level Script classes
-    private synchronized String generateScriptName() {
+    private synchronized static String generateScriptName() {
```
jwagenleitner: This changes the method from being accessed by one thread per instance to one thread per class.  May not be an issue given what this method does, but in general I would probably not make this kind of bulk change on any `synchronized` method.

oreissig: I actually thought about this one, as it sure is a slight semantic change here. The method increments a counter variable, that itself is static. I don't thing having that synchronized on a per-object basis does make sense, or does it?

jwagenleitner: Nevermind, I actually looked at what the method does and the static it increments.  This may actually be a fix. :smile:  Though doesn't static usually come before synchronized?

jwagenleitner: You're right, caught it as soon as I hit the comment button.

oreissig:
Now that you mention it {{private static synchronized}} somehow reads better than {{private synchronized static}}. I'll change it.

@pfmiles When calling private methods from (dynamic) Groovy, it should not matter whether a method is static or not. Statically compiled Groovy doesn't allow calling private methods anyway, so this should only apply to directly using the Java reflection API for invoking private stuff. Do you think this is a concern we should leave stable between minor increments of Groovy?

oreissig:
I just found a way more in-depth explanation for the issue: http://www.beyondjava.net/blog/should-you-avoid-or-embrace-static/

> I recommend to use static as often as possible. There’s a twist: Never use `static` variables, but always use `static` methods.

oreissig:
I rebased the changes from master and squashed the corrections.

jwagenleitner:
Appreciate the rebase and squashing and the good discussion on this.  Thanks!

PR #291, opened 2016-03-12
remove unused or obsolete code

oreissig:
Some cleanups in the root project have already been cherrypicked with PR #282, so I decided to skim over the subprojects as well and remove code that's not used anymore. I also removed code that has been commented out, most of which has actually been commented out 9 years ago.


PascalSchumacher:
Thanks!

blackdrag:
frankly I would not have merged this PR. There are checked exceptions removed from public methods. This is an incompatible change, because Javacode using that method and having a try-catch for the exception will no longer compiler, as we then suddenly try to catch an non existing checked exception. 

And there is this revert... I am not finding that code in this pullrequest, so I am wondering where this actually comes from

PascalSchumacher:
Sorry, there were two similar pull request. This one and 78b6cc54af85988829ecd6120b69f65e0d9a8ade. I mixed them up. 

I know that removing exceptions from signatures is not java source compatible. If you think that keeping this compatibility is important for these classes/methods we can revert both commits.

PR #292, opened 2016-03-12
make private fields final where possible

oreissig:
As discussed in PR #282 I repropose the changes from there in individual pull requests.

### DIFF
--- a/src/main/org/codehaus/groovy/ast/expr/GStringExpression.java
+++ b/src/main/org/codehaus/groovy/ast/expr/GStringExpression.java
@@ -32,13 +32,15 @@ import org.codehaus.groovy.ast.GroovyCodeVisitor;
  */
 public class GStringExpression extends Expression {
 
-    private String verbatimText;
-    private List<ConstantExpression> strings = new ArrayList<ConstantExpression>();
-    private List<Expression> values = new ArrayList<Expression>();
+    private final String verbatimText;
+    private final List<ConstantExpression> strings;
+    private final List<Expression> values;
     
     public GStringExpression(String verbatimText) {
         this.verbatimText = verbatimText;
         super.setType(ClassHelper.GSTRING_TYPE);
+        this.strings = new ArrayList<ConstantExpression>();
+        this.values = new ArrayList<Expression>();
     }
 
     public GStringExpression(String verbatimText, List<ConstantExpression> strings, List<Expression> values) {
--- a/src/main/org/codehaus/groovy/runtime/ProcessGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/ProcessGroovyMethods.java
@@ -433,7 +433,7 @@ public class ProcessGroovyMethods extends DefaultGroovyMethodsSupport {
      * @since 1.0
      */
     protected static class ProcessRunner implements Runnable {
-        Process process;
+        final Process process;
         private boolean finished;
 
         public ProcessRunner(Process process) {
@@ -472,8 +472,8 @@ public class ProcessGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     private static class TextDumper implements Runnable {
-        InputStream in;
-        Appendable app;
+        final InputStream in;
+        final Appendable app;
 
         public TextDumper(InputStream in, Appendable app) {
             this.in = in;
@@ -498,8 +498,8 @@ public class ProcessGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     private static class ByteDumper implements Runnable {
-        InputStream in;
-        OutputStream out;
+        final InputStream in;
+        final OutputStream out;
 
         public ByteDumper(InputStream in, OutputStream out) {
             this.in = new BufferedInputStream(in);
--- a/subprojects/groovy-sql/src/main/java/groovy/sql/BatchingStatementWrapper.java
+++ b/subprojects/groovy-sql/src/main/java/groovy/sql/BatchingStatementWrapper.java
@@ -34,7 +34,7 @@ import java.util.logging.Logger;
  * automatically. If batchSize is zero, then no batching is performed.
  */
 public class BatchingStatementWrapper extends GroovyObjectSupport {
-    private Statement delegate;
+    private final Statement delegate;
     protected int batchSize;
     protected int batchCount;
     protected Logger log;
--- a/src/main/groovy/io/EncodingAwareBufferedWriter.java
+++ b/src/main/groovy/io/EncodingAwareBufferedWriter.java
@@ -29,7 +29,7 @@ import java.nio.charset.Charset;
  */
 
 public class EncodingAwareBufferedWriter extends BufferedWriter {
-    private OutputStreamWriter out;
+    private final OutputStreamWriter out;
     public EncodingAwareBufferedWriter(OutputStreamWriter out) {
         super(out);
         this.out = out;
--- a/src/main/groovy/io/PlatformLineWriter.java
+++ b/src/main/groovy/io/PlatformLineWriter.java
@@ -33,7 +33,7 @@ import java.io.IOException;
  * @author Paul King
  */
 public class PlatformLineWriter extends Writer {
-    private BufferedWriter writer;
+    private final BufferedWriter writer;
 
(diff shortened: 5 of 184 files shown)


src/main/org/codehaus/groovy/control/MultipleCompilationErrorsException.java
```diff
@@ -27,7 +27,7 @@
 public class MultipleCompilationErrorsException extends
         CompilationFailedException {
     
-    protected ErrorCollector collector;
+    protected final ErrorCollector collector;
```
jwagenleitner: There could be subclasses that set this field.

subprojects/groovy-sql/src/main/java/groovy/sql/BatchingStatementWrapper.java
```diff
@@ -34,10 +34,10 @@
  * automatically. If batchSize is zero, then no batching is performed.
  */
 public class BatchingStatementWrapper extends GroovyObjectSupport {
-    private Statement delegate;
+    private final Statement delegate;
     protected int batchSize;
     protected int batchCount;
-    protected Logger log;
+    protected final Logger log;
```
jwagenleitner: While it may not be likely, there could be some subclasses that set this so I think it might be good to avoid changing `protected` fields.

jwagenleitner:
With the exception of a few changes to protected fields, it looks good to me.  With such a large change I would be curious what others think.  Hopefully get some to comment here.

src/main/groovy/grape/GrapeIvy.groovy
```diff
@@ -76,11 +76,11 @@ class GrapeIvy implements GrapeEngine {
     Set<String> resolvedDependencies
     Set<String> downloadedArtifacts
     // weak hash map so we don't leak loaders directly
-    Map<ClassLoader, Set<IvyGrabRecord>> loadedDeps = new WeakHashMap<ClassLoader, Set<IvyGrabRecord>>()
+    final Map<ClassLoader, Set<IvyGrabRecord>> loadedDeps = new WeakHashMap<ClassLoader, Set<IvyGrabRecord>>()
```
shils: making these fields final will break binary compatibility since their synthetic public setters won't be generated

blackdrag:
I agree with @jwagenleitner. We cannot merge this pull request as it is right now.  Frankly I prefer doing a change to final only if I mean to have it final. If for example all but one field is mutated from different places, then making that one field private does not really serve any purpose. Anyway... this should be done only for private fields. Fields of private inner classes are probably save, at least if not written in Groovy.

graemerocher:
Yeah this has huge implications to Groovy libraries and could break existing code

oreissig:
I'll revert the changes to any non-private fields. I agree that we should not change any signiture that's available to others, even if they are unlikely to rely on that.

oreissig:
Done reverting.

I'm not quite sure about the argument with Groovy ignoring the private flag, making this change potentially breaking. By this logic, every change that does not merely add new classes is a breaking change. Maybe we could treak the groovy.* package different than org.codehaus.groovy.*, with the former employing an even stronger compatibility contract?

blackdrag:
Groovy does not ignore private, but private has slightly different semantics in Groovy if compared to Java. I mentioned inner classes only because of some quirks there that may have a different effect from what you expect.

blackdrag:
If you want to change something that has been available in a non-private context, then we have to discuss each and every case.

oreissig:
So you are basically referring to [this issue](https://issues.apache.org/jira/browse/GROOVY-1591)? From what I am reading the "slightly different semantics" means "private is respected with `@CompileStatic` and accessible as if it were public for dynamic Groovy".

Do you mean this change is only safe for private inner classes?

subprojects/groovy-macro/src/main/groovy/org/codehaus/groovy/macro/matcher/internal/MatchingConstraintsBuilder.groovy
```diff
@@ -24,7 +24,7 @@ import org.codehaus.groovy.macro.matcher.TreeContext
 import org.codehaus.groovy.syntax.Token
 
 class MatchingConstraintsBuilder {
-    Set<String> placeholders = new LinkedHashSet<>()
+    final Set<String> placeholders = new LinkedHashSet<>()
```
jwagenleitner: Since this is a Groovy object this is changing a public property to final and I don't think should be done.

src/main/org/codehaus/groovy/ast/expr/GStringExpression.java
```diff
@@ -32,13 +32,15 @@
...
-    private List<Expression> values = new ArrayList<Expression>();
+    private final String verbatimText;
+    private final List<ConstantExpression> strings;
+    private final List<Expression> values;
     
     public GStringExpression(String verbatimText) {
         this.verbatimText = verbatimText;
         super.setType(ClassHelper.GSTRING_TYPE);
+        this.strings = new ArrayList<ConstantExpression>();
+        this.values = new ArrayList<Expression>();
```
jwagenleitner: This would cause a breaking change if any code relied on `#getStrings` or `#getValues` returning `null`.  I would avoid this change.

src/main/org/codehaus/groovy/runtime/ProcessGroovyMethods.java
```diff
@@ -433,7 +433,7 @@ public static Process or(final Process left, final Process right) throws IOExcep
      * @since 1.0
      */
     protected static class ProcessRunner implements Runnable {
-        Process process;
+        final Process process;
```
jwagenleitner: This could be a breaking change.

oreissig:
Thanks for the feedback. I reverted the changes I accidentally made to the Groovy class and changed the expression's behaviour to initialize its fields with null instead of empty list.

oreissig:
Good thing there are test :) The `getString()` or `getValues()` methods of `GStringExpression` are actually assumed to not return null.

jwagenleitner:
I missed that in the original code those were assigned where they were declared and that you had moved it down to the constructor.

oreissig:
I rebased the changes from master and squashed the corrections.

oreissig:
What's the future of this PR?
Maintaining a feature branch with such widespread changes to be conflict free is not rewarding if the outcome is unclear.
Please decide if it should be abandoned at all, if I should just fix the merge conflicts, or if you see the possibility to merge a subset of changes.

blackdrag:
I think this pull request is too big. Try splitting it in bunches of <20 files, for example on a per subproject base would already help. And then maybe package path based for src/main. This will allow more people to do the review, and avoids maintenance if more merge conflicts appear.

oreissig:
The discussion raised concernes that we should avoid this kind of change even for private fields. If so, then I don't need to continue. Otherwise I'm fine with rebasing and splitting the PR.

blackdrag:
I think it is ok to do for private fields

paulk-asert:
I merged this PR but tweaked to exclude changes to these files:
```
src/main/groovy/util/ClosureComparator.java
src/main/org/codehaus/groovy/antlr/LexerFrame.java
src/main/org/codehaus/groovy/control/CompilePhase.java
src/main/org/codehaus/groovy/runtime/ProcessGroovyMethods.java
src/main/org/codehaus/groovy/runtime/metaclass/MetaMethodIndex.java
subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/ClasspathResourceManager.java
subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/MockOutputTool.java
subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/SimpleGroovyExecutableMemberDoc.java
subprojects/groovy-json/src/main/java/groovy/json/internal/ValueList.java
subprojects/groovy-sql/src/main/java/groovy/sql/CallResultSet.java
```
For those files a package-private field was made final in a non-private class. These all break binary compatibility. For some of the internal `org.codehaus.groovy` classes we _might_ decide to do that but it should be on a case-by-case basis, not in the same commit as the non-controversial changes.

PR #295, opened 2016-03-20
ignore eclipse project files for rat

oreissig:
This makes `gradlew check` pass on local systems when using Eclipse instead of IntelliJ.


blackdrag:
+1

paulk-asert:
merged, thanks

PR #296, opened 2016-03-20
remove unnecessary Gradle wrapper configuration

oreissig:
`wrapper` is an implicit Gradle task, you don't need to create one manually, just call e.g.
`./gradlew wrapper --gradle-version=2.12`

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -74,7 +74,7 @@ If obtaining the source from this distribution and you intend to build from sour
 you also need to https://gradle.org/downloads/[download] and install http://gradle.org/[Gradle] and execute one bootstrap step.
 At the top directory of your unpacked source, you need to run the command:
 
-    gradle
+    gradle wrapper
 
 This sets up the Gradle wrapper and from then on you just need the `gradlew` command instead of `gradle`.
 
--- a/gradle/wrapper.gradle
+++ b/gradle/wrapper.gradle
@@ -16,9 +16,6 @@
  *  specific language governing permissions and limitations
  *  under the License.
  */
-task downloadWrapper(type: Wrapper) {
-    description = "Download the gradle wrapper and requisite files. Overwrites any existing wrapper files."
+tasks.withType(Wrapper) {
     gradleVersion = project.gradleVersion
 }
-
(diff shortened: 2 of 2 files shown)


blackdrag:
not sure about this commit. You are right in that there is a builtin wrapper task and that our wrapper task should not be needed. But the soruce distribution comes without the gradlew files. So how does a user know what gradle version to use to ensure there are no problems with the built related to the built tool version? Right now we can tell people to use any gradle to execute the wrapper task and then work with the wrappers from there on

oreissig:
Thanks for the feedback. Actually I forgot about the source distribution. Why does it miss the gradle wrapper at all?

Now that you mentioned it, one of my previous pull requests removed the declaration of wrapper generation as default task in Gradle, thereby breaking the documentation describing the bootstrap of the source distribution. We should either

1. include the Gradle wrapper in the source distribution, or
2. change the README.adoc to call `gradle wrapper` instead of `gradle`.

paulk-asert:
-1 on this change. Apache source distro's can't have the wrapper jar (a binary). They make some minor exceptions but to date the gradle wrapper isn't one of the allowable exceptions. The `downloadWrapper` task was meant to be a default task but that must have been overlooked in earlier commits. I just fixed that, so now the doco should be correct. This still leaves the implicit wrapper task with an explicit version free should someone wish to not abide by our recommendations.

oreissig:
Thanks for the explanation.

I still think it is a good idea to not use a default task, so that Gradle prints a helpful splash screen when invoked without arguments. I propose to document the use of the implicit Gradle wrapper task. Its version can still be overridden on the command line as described in the Gradle manual. I may also link to the Gradle documentation describing this use case in the README.adoc if desired.

oreissig:
rebased from master

paulk-asert:
This looks better. Nice work. I'll just double check that everything works as expected before merging.

paulk-asert:
Looks like you've deleted the repo. Okay if I apply manually?

oreissig:
Please feel free 👍

paulk-asert:
merged, thanks!

PR #298, opened 2016-03-23
Guard against recursive calls to ReferenceManager#removeStallEntries …

jwagenleitner:
…that could cause a StackOverflowError.  A ManagedReference#finalizeReference method may call ManagedReference#clear on other references as part of the clean up process and #clear in turns calls #removeStallEntries.  This change makes it so that a queue is not re-entered for processing once it has started to be processed.

This fix is related to PR #219 because that PR fixes the finalization of the ClassInfo instance.  As part of the cleanup a call is made to [ManagedReference#clear] (https://github.com/apache/groovy/blob/91406a76c6bf00ac27a102fe5b55362238fa8af5/src/main/org/codehaus/groovy/reflection/ClassInfo.java#L196)  which in turn calls [removeStallEntries] (https://github.com/apache/groovy/blob/91406a76c6bf00ac27a102fe5b55362238fa8af5/src/main/org/codehaus/groovy/util/ManagedReference.java#L49).

An option might be to remove the call to removeStallEntries from ManagedReference#clear and that is what was done in PR #219.  But it may be difficult to always guarantee no recursive calls in the Managed classes in which case having a guard in place removes the need of the managed classes from having to worry about it.

### DIFF
--- a/src/main/org/codehaus/groovy/util/ReferenceManager.java
+++ b/src/main/org/codehaus/groovy/util/ReferenceManager.java
@@ -19,6 +19,7 @@
 package org.codehaus.groovy.util;
 
 import java.lang.ref.ReferenceQueue;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 
 public class ReferenceManager {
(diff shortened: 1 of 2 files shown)


blackdrag:
+1

PR #302, opened 2016-04-03
GROOVY-7781 - Cannot retrieve XML attribute with namespace

jwagenleitner:
This change broke another test that asserted that an unprefixed attribute's namespace is the default namespace, but default namespace declarations do not apply to attribute names, see https://www.w3.org/TR/xml-names/#defaulting.

> The namespace name for an unprefixed attribute name always has no value.


gillius:
Awesome thanks for doing this! I'm not a Groovy developer, but I did look at the code and it looks like the changes I'd expect for this, but more so I can confirm the unit test added should both replicate the original bug as well as confirm the fix.

PascalSchumacher:
+1

jwagenleitner:
I created https://issues.apache.org/jira/browse/GROOVY-7808 since part of this fix includes having to change an existing test.  Will give it a couple days to see if there is any feedback.

PR #304, opened 2016-04-03
Added 0o prefix support for octal literals

hlaaftana:


blackdrag:
If you add a test, then you make the pull request perfect

blackdrag:
Can you explain a bit what the motivation for this pull request is? I mean we do have a octals exactly the same way as Java. This would add another way, but with the problem that 0 and O are difficult to differentiate. On the other hand if you did write 0010 instead of 0O10 you would still get 8, so maybe it is not so bad. Still a bit explanation would be good. And not to forget... there should be a test for this

hlaaftana:
languages like python exclusively have 0o for octals, so it'd be easy for people from those languages to get adapted

graemerocher:
@hlaaftana breaking language changes require more information, explanation and tests than simply saying Python does it, so so should Groovy

blackdrag:
The python PEP: https://www.python.org/dev/peps/pep-3127 This also has the motivation:

> The default octal representation of integers is silently confusing to people unfamiliar with C-like languages. It is extremely easy to inadvertently create an integer object with the wrong value, because '013' means 'decimal 11', not 'decimal 13', to the Python language itself, which is not the meaning that most humans would assign to this literal. 

I am actually surprised they want to support 0O. It should also be noted, that they go with not supporting 010 anymore. I would have no problem supporting the 0o variant, but I don't like 0O at all. just using the leading 0 is also not that nice, so I am unhappy with the Java variant actually. Frankly I would have chosen a different letter like c instead. I mean it is x for hexadecimal, and only for binary it is b. I see no problem in using some letter different than o here... unless of course we want to align with python.

A bit google usage found me that not many languages support octal literals, C doesn't have them. Scala removed them and Swift for example supports only the version with small o. My opinion is, the Python people actually made a mistake for allowing 0O, 0o is ok, but is it ok to have 0O10 and 0010 at the same time? Python's answer seems to be no. The VHDL variant would be O"010", but there you really have a clear understanding what is prefix and what is number. 

So all in all I have no problem with allowing 0o for octals, I am against 0O as long as we have the version with the simple 0 prefix though

paulk-asert:
Given that both 0x and 0X are supported for hexadecimal (similarly 0b and 0B for binary), I have no problem with supporting 0O, but I would expect that we would regard it as bad style, the same way we regard lowercase L bad style for long since it can be confused more easily with number 1. That kind of thing could be allowed in the grammar but flagged as poor style by a codenarc rule.

jwagenleitner:
@hlaaftana if you were interested in adding a test to this PR I think the appropriate place for it would be in [SyntaxTest#testOctalLiteral] (https://github.com/apache/groovy/blob/05e9f5639c9fed6fc3406d594b85aca9892faf76/src/spec/test/SyntaxTest.groovy#L23-L43).  Would also be good to update the [core-syntax.adoc guide] (https://github.com/apache/groovy/blob/05e9f5639c9fed6fc3406d594b85aca9892faf76/src/spec/doc/core-syntax.adoc) in the Octal literal section.

jwagenleitner:
Forgot to mention that this would also require a [grammar change in groovy.g] (https://github.com/apache/groovy/blob/05e9f5639c9fed6fc3406d594b85aca9892faf76/src/main/org/codehaus/groovy/antlr/groovy.g#L4201).  I think something like following should do but I'm not a ANTLR person:

```groovy
('0'..'7'|'o'|'O') (('0'..'7'|'_')* ('0'..'7'))?
```
IDE's may report some problems with the syntax, for example in IntelliJ code inspection reports `'0' cannot be applied to (?)` but it still compiles and runs ok.

PR #305, opened 2016-04-05
GROOVY-7810 fix XmlSlurper default ctor doc

gillius:
Javadoc for default constructor says it creates a non-namespace-aware slurper, when it creates a namespace-aware one.


jwagenleitner:
Thanks!

PR #307, opened 2016-04-09
GROOVY-7723 - propertyMissing(String,Object) called for missing getter

jwagenleitner:
Looks like commit 97a45dfb3e3c5212c36 introduced [this change] (https://github.com/apache/groovy/commit/97a45dfb3e3c5212c3610cd90fe1e7434614b260#diff-c1354c0d7df171ad8cafcb3d1bf63857L535). The original code would not have delegated to the setter method. I don't think the intent of the commit was to change the behavior but resulted inadvertently due to the refactoring.


blackdrag:
+1

PR #308, opened 2016-04-09
CliBuilder enhancements

paulk-asert:


src/main/groovy/cli/Unparsed.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package groovy.cli;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Indicates that a method will contain the remaining arguments.
```
shils: I think 'method or field' would be more appropriate here, given the usage within an annotation class spec e.g. GreeterC in CliBuilderTest.

paulk-asert: Nice suggestion

PR #309, opened 2016-04-09
GROOVY-7813: Calling a non-static outer class method from a static cl…

shils:
…ass should fail STC


shils:
This fix is incorrect, closing for now.

PR #311, opened 2016-04-10
Generate tuple constructor with all args and annotations

alien11689:


paulk-asert:
I like the idea but I think we can make it more generic. I have some thoughts on what can be done but I won't get time until next week to expand on the details.

bubach:
God damn it, I was just about to get fond of the Groovy syntax .... whoever came up with the "brilliant" idea of annotations should be hanged, drawn and quartered.

Looking at the homepage documentation, it actually seems like you can't even set a property to protected/private? But public works?  I'm hoping I'm wrong, b/c that's just ... cancer.

I can't for the love of me understand how _anybody_ could think it's a good idea. And yet this bullshit is spreading like locus between the programming dialects.

I managed to stay away with PHP for a while, but that language syntax is like a drove of old Java and C# dev's on a blind stampede, no idea what their doing except it should look like J/C#...
</ end half-drunk rant>

paulk-asert:
Hi Christoffer, perhaps you can explain what you are trying to do and we can show you how to do it in Groovy.

The Java approach to properties is to write the getters, setters and backing field yourself. Groovy properties are a short-hand for the 99% common case of public getters/setters and private backing field but you can always write yourself as per Java for special cases or write your own meta-programming to give you other combinations. When writing yourself fields and methods can be private, protected, package-private or public.

Annotations aren't required but many developers like the magnitude productivity increase that compile-time meta-programming offers.

paulk-asert:
I know it doesn't cover all use cases, but for records in Groovy 4, annotations on the record are carried over to the constructor if placed on the record. It would cover the original use case:
```
import groovy.transform.Generated
import java.lang.annotation.*

@Target(ElementType.CONSTRUCTOR)
public @interface Example1 {}

@Target(ElementType.CONSTRUCTOR)
public @interface Example2 {}

@Example1 @Example2
record Person(String firstName, String lastName) { }

Person.declaredConstructors[0].each {
    assert it.declaredAnnotations*.annotationType() == [Generated, Example1, Example2]
}
```

PR #312, opened 2016-04-13
GROOVY-7814: Support Arrays of arbitrary annotations as the return ty…

paulk-asert:
…pe of an annotation definition


blackdrag:
I think I am now -1 on this change

paulk-asert:
Any particular concern @blackdrag? I like very much the many options this opens up but also wish there was a nicer way to implement. I just can't think of any other way to implement to allow this feature.

paulk-asert:
I'll close this for now. I think it is a nice feature but we have other changes for annotations with higher priority, so we can come back to this at a later date if we progress other items first.

PR #314, opened 2016-04-17
GROOVY-7783 - Redundant computation in ObservableList

jwagenleitner:
Patch provided thanks to Monika Dhok.


jwagenleitner:
Used commit c215e5d31922f24b36a26daf123ef293752669bb as a reference and decided against including tests that assert based on timings.  I did test locally and verified that performance for large collections improved substantially and the overhead of converting smaller collections to Sets were negligible.

paulk-asert:
+1

PR #315, opened 2016-04-21
GROOVY-7820: CachedMethod.compareToCachedMethod throws "Should never …

paulk-asert:
…happen"


blackdrag:
may I suggest to move the declaring class check to just before the throw? That way it will not impose a difference to whatever code is using this already right now

paulk-asert:
Sure, will do.

fmamud:
@paulk-asert thanks for PR by my bug report. 😄

PR #316, opened 2016-04-21
GROOVY-7822 - Redundant computation in ObservableSet

jwagenleitner:
Related to PR #314, performance significantly improved for `retainAll`, similar change to `removeAll` did not yield any significant performance difference.


paulk-asert:
+1

PR #317, opened 2016-04-23
Allow TargetBytecode to be set in Groovyc Ant task to all supported versions

pavelsigarteu:
The parameter now supports 1.8, 1.7, 1.6, 1.5 or 1.4 as opposed to previously 1.4 and 1.5 only. This allows to compile to bytecodeversion 1.8 from ant and not just max 1.5 as before.


paulk-asert:
merged, thanks

paulk-asert:
Tracked in: https://issues.apache.org/jira/browse/GROOVY-7827

PR #324, opened 2016-04-30
Remove unused utility classes

jwagenleitner:
These are public classes but under the `org.codehaus.groovy.util` package so am not sure if that makes them eligible to be removed or if they must be deprecated.  But since there are no references to them in the codebase it would be nice to do the clean-up.

If they are kept, then it would be good to fix the bug in `ManagedDoubleKeyMap#put` (commit 240b915).


keeganwitt:
My understanding is `org.codehaus` classes are regarded as internal (mostly anyway) so I think we wouldn't have to go through the usual deprecation.

blackdrag:
I think it is ok to remove them

jwagenleitner:
Thanks for the feedback, appreciate it.

PR #325, opened 2016-05-01
GROOVY-7646 - Classes generated by Eval() never collected from Permgen/Metaspace

jwagenleitner:
This is also fixed by PR #219 but that requires SoftReferences to be collected (because of beaninfocache).   It seems for cases where a Script class is parsed, executed for its result and the Script class is not retained we can be more aggressive about clearing it from caches so it doesn't require the heap to fill before being collected.

A test is included just to demonstrate the changes do allow the classes to be GC'd but is not a test that is meant to be included if the PR is accepted.

### DIFF
--- a/src/main/groovy/lang/GroovyShell.java
+++ b/src/main/groovy/lang/GroovyShell.java
@@ -515,7 +515,11 @@ public class GroovyShell extends GroovyObjectSupport {
      */
     public Object run(GroovyCodeSource source, String[] args) throws CompilationFailedException {
         Class scriptClass = parseClass(source);
-        return runScriptOrMainOrTestOrRunnable(scriptClass, args);
+        try {
+            return runScriptOrMainOrTestOrRunnable(scriptClass, args);
+        } finally {
+            InvokerHelper.removeClass(scriptClass);
+        }
     }
 
     /**
@@ -563,7 +567,11 @@ public class GroovyShell extends GroovyObjectSupport {
                     }
         });
         Class scriptClass = parseClass(gcs);
-        return runScriptOrMainOrTestOrRunnable(scriptClass, args);
+        try {
+            return runScriptOrMainOrTestOrRunnable(scriptClass, args);
+        } finally {
+            InvokerHelper.removeClass(scriptClass);
+        }
     }
 
     public Object getVariable(String name) {
@@ -582,7 +590,11 @@ public class GroovyShell extends GroovyObjectSupport {
      */
     public Object evaluate(GroovyCodeSource codeSource) throws CompilationFailedException {
         Script script = parse(codeSource);
-        return script.run();
+        try {
+            return script.run();
+        } finally {
+            InvokerHelper.removeClass(script.getClass());
+        }
     }
 
     /**
--- a/src/main/org/codehaus/groovy/reflection/ClassInfo.java
+++ b/src/main/org/codehaus/groovy/reflection/ClassInfo.java
@@ -115,6 +115,22 @@ public class ClassInfo {
         return globalClassValue.get(cls);
     }
 
+    /**
+     * Removes a {@code ClassInfo} from the cache.
+     *
+     * This is useful in cases where the Class is parsed from a script, such as when
+     * using GroovyClassLoader#parseClass, and is executed for its result but the Class
+     * is not retained or cached.  Removing the {@code ClassInfo} associated with the Class
+     * will make the Class and its ClassLoader eligible for garbage collection sooner that
+     * it would otherwise.
+     *
+     * @param cls the Class associated with the ClassInfo to remove
+     *            from cache
+     */
+    public static void remove(Class<?> cls) {
+        globalClassValue.remove(cls);
+    }
+
     public static Collection<ClassInfo> getAllClassInfo () {
         return getAllGlobalClassInfo();
     }
--- a/src/main/org/codehaus/groovy/runtime/InvokerHelper.java
+++ b/src/main/org/codehaus/groovy/runtime/InvokerHelper.java
@@ -35,6 +35,7 @@ import groovy.lang.SpreadMap;
 import groovy.lang.SpreadMapEvaluatingException;
 import groovy.lang.Tuple;
 import groovy.lang.Writable;
+import org.codehaus.groovy.reflection.ClassInfo;
 import org.codehaus.groovy.runtime.metaclass.MetaClassRegistryImpl;
 import org.codehaus.groovy.runtime.metaclass.MissingMethodExecutionFailed;
 import org.codehaus.groovy.runtime.powerassert.PowerAssertionError;
@@ -83,6 +84,7 @@ public class InvokerHelper {
 
     public static void removeClass(Class clazz) {
         metaRegistry.removeMetaClass(clazz);
+        ClassInfo.remove(clazz);
         Introspector.flushFromCaches(clazz);
     }
 


blackdrag:
+1 to the change itself, but I feel it does not solve GROOVY-7646, because that should have not been happening even without you removing classes from the cache actively. I can for example change the eval script to contain a class besides the script class and then it should blow up again. Furthermore, Eval just uses GroovyShell. So you should get the same problem with GroovyShell too. So in my opinion your fix is not addressing the underlying issue

jwagenleitner:
Hi @blackdrag thanks for the feedback.  I agree with you that this does not address the underlying issue, for that I believe that PR #219 does.  The root problem I believe is that `ClassInfo` keeps a strong reference to it's `Class` so it can never be removed from the static `globalClassValue` weak keyed map making the Class and it's inner GroovyClassLoader reachable.  On a side note, it may also allow us revert to using the Java 7 ClassValue as the default where available.

However, based on what I've seen in the heap dumps, I believe even with fixing the ClassInfo/Class hard ref the Introspector's beanInfoCache (Java problem not Groovy) is keeping soft references to the class which I think means that heap will continually fill up to the point of OOME until those are freed.  My thought with this fix (which may be naive) is that in cases where we know the parsed Class is no longer needed we can do some cleanup to assist the GC and prevent filling heap.

blackdrag:
you are very right about the bean cache of course. It is a problem known for years... that and the stupid synchronization mechanism inside the Introspector.  If you now can tell me the performance impact of this change, then I will soon be able to give my ok.

jwagenleitner:
Forgot to mention that my assumption is that `#GroovyShell(GroovyCodeSource)` always creates a unique class/classloader on each call and that no caching is involved based on `GroovyShell#parseClass(GroovyCodeSource)` call to `GroovyClassLoader#parseClass` passing `false` for `shouldCacheSource`.

The only other spot I see where doing this kind of call to the clean-up method is in the `GroovySocketServer` since it creates a unique class/classloader per invocation and never unloads them as reported in GROOVY-3476.

jwagenleitner:
I will try to come up with some reasonable benchmarks in order to assess the performance impact.

jwagenleitner:
With a very [simplistic JMH benchmark] (https://github.com/jwagenleitner/testing-groovy/blob/1685d0d72e3f5716410ac605c7f5e888e72156c6/issues/GROOVY7646/src/jmh/java/bugreport/Groovy7646Perf.java) I get the following numbers:

Pre-change:
```
Benchmark                          Mode  Samples    Score    Error  Units
b.Groovy7646Perf.shellEvaluate    thrpt       50  568.820 ? 94.488  ops/s
b.Groovy7646Perf.shellEvaluate     avgt       50    0.004 ?  0.001   s/op
b.Groovy7646Perf.shellEvaluate       ss       50    0.008 ?  0.001      s
```

Post-change:
```
Benchmark                          Mode  Samples    Score    Error  Units
b.Groovy7646Perf.shellEvaluate    thrpt       50  728.420 ? 58.170  ops/s
b.Groovy7646Perf.shellEvaluate     avgt       50    0.003 ?  0.000   s/op
b.Groovy7646Perf.shellEvaluate       ss       50    0.010 ?  0.002      s
```
Not sure the impact under large number of threads, but suspect if there were a large number of threads the impact of throwing OOME's and reclaiming SoftRefs would outweigh the cost of the added changes.  In fact I wonder if that somehow plays into the results above since with the additional changes the speed improved.

jwagenleitner:
I found that [`GroovyShell#evaluate(Reader,String)`] (https://github.com/apache/groovy/blob/05e9f5639c9fed6fc3406d594b85aca9892faf76/src/main/groovy/lang/GroovyShell.java#L666) was already trying to do this kind of cleanup using [`InvokerHelper#removeClass`] (https://github.com/apache/groovy/blob/05e9f5639c9fed6fc3406d594b85aca9892faf76/src/main/org/codehaus/groovy/runtime/InvokerHelper.java#L67-L70) for GROOVY-649 as part of commit 5724870025c25622015ba13c0310def5742d0b2f.  So I fixed up this PR to use that and added the call in order to remove the ClassInfo from the globalClassValue cache.  I benchmarked again and am still seeing better throughput after the change.

jochenberger:
Is there anything in the way of merging this? We're doing something quite similar using reflection and it improves the memory leak situation dramatically.

blackdrag:
what needs to be done to merge this is, remove the demo test case and resolve the small merge conflict

jwagenleitner:
I removed the demo test case and rebased.  I made a slight change using try/finally to ensure the clean method is called if the script causing an exception.

jochenberger:
Could this please be merged then?

jwagenleitner:
@jochenberger is your issue with `Eval` or more generally `GroovyShell`?

My hesitation with this as it is now is as Jochen pointed out, this doesn't handle scripts that contain class definitions so is not really a good fix, just a bandaid.  Also, I need to do more testing but this may cause problems if metaclasses are touched in the scripts.  I think it would be safe to do this in `Eval` since it doesn't save the shell between invocations, so eval can grab the shell's `GroovyClassLoader` and flush the caches for all the loaded classes.

jochenberger:
I compile and `run()` script classes manually. When compiling, I keep track of the classes that are generated for a script and manually unload them using `InvokerHelper.removeClass()` afterward. Then I have to use reflection to remove them from `globalClassSet`/`globalClassValue`. So what I really need from this PR is the added `ClassInfo.remove()` method and its being called from `InvokerHelper.removeClass()`.

jochenberger:
What are the plans for this now that #219 is merged? There should be much less of an issue, but it would still be nice to be able to eagerly evict classes from the cache when they are not used anymore instead of waiting for the GC to kick in.

jwagenleitner:
I am thinking of going ahead and adding the new `ClassInfo#remove` method and adding the call to it in the `InvokerHelper#removeClass` method.   Those seem like reasonable changes that would not cause any issues and provides the programmer a way to clean up if they choose assuming they have a reference to the Script class.

I am less sure about the clean up done in GroovyShell itself.  In rethinking this, I definitely think the `Eval` class can do some automated cleanup after execution but am not comfortable with doing it in GroovyShell.  So part two after the changes mentioned above would be to see about redoing this PR for `Eval` (as reported in the issue) and not `GroovyShell`.

It might be nice to have a method on `GroovyClassLoader`, `GroovyShell` and `GroovyScriptEngine` (such as `flushCaches` or maybe even make it `Closeable`) so that (more aggressive) cleanup could be initiated if desired but not automatically since they may cause unforeseen issues.

Would be curious to know everyone's thoughts.

blackdrag:
ClassInfo#remove is fine, adding it to InvokerHelper#removeClass is probably fine too. Doing this to Eval is probably working out, since you rarely will create a class within the eval and then using meta programming on it. For the other three this is different, but adding a flushCaches we could have. Of course if we use Closeable, we could use close instead. Adding this to GroovyShell and GroovyScriptEngine would be actually a good idea I think. GroovyClassLoader is, because it is extending URLClassloader, already closeable... so it could go there even without adding a new interface.

jochenberger:
Maybe we should split this into smaller chunks. I think we all agree about `ClassInfo#remove` and adding it to `InvokerHelper#removeClass`, so that could be one PR.
Using `flushCaches/close` in `GroovyShell`, `GroovyScriptEngine`, and `GroovyClassLoader` seems reasonable to me too. The latter would just do a `classCache.values().forEach(InvokerHelper::removeClass)`, right?
`GroovyShell` and `GroovyScriptEngine` would probably just have to `close()` their `GroovyClassLoader` from their `close()` methods then, but since they are both extensible public API, adding the `Closeable` interface would probably have to wait for 3.x.

jochenberger:
I created PRs #444 and #445, I hope we can get these in for 2.4.7.

jwagenleitner:
@jochenberger thanks for the PRs, both merged.  The `ClassInfo#remove` should get in 2.4.8 but the `GroovyClassLoader#close` will be in 2.5.0 since 2.4.8 still has Java 6 as supported.

PR #327, opened 2016-05-04
GROOVY-7604: traits docs diamond problem explanation

andresteingress:
Fixes a (very) tiny text issue raised in GROOVY-7604.

### DIFF
--- a/src/spec/doc/core-traits.adoc
+++ b/src/spec/doc/core-traits.adoc
@@ -199,8 +199,7 @@ WARNING: While traits support public fields, it is not recommended to use them a
 
 == Composition of behaviors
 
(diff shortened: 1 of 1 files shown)


blackdrag:
can you explain why you don´t want to mention the diamond problem?

andresteingress:
The diamond problem is actually mentioned in one of the following sections. 7604 was just about that the ticket author said that the sub-sentence this PR removes would hint the reader that the following example would be show how to solve the diamon problem, which in fact is not the case.

It's just a tiny correction, I also wont be mad if this ticket is simply ignored and closed.

blackdrag:
+1 for merging then

PascalSchumacher:
Thanks!

PR #328, opened 2016-05-05
GROOVY-7793: Compiler compiles class with private abstract method

andresteingress:
This PR adds some logic to the class completion verifier that adds an error to the error collector when private abstract methods are detected in abstract classes or enums (for interfaces such a check is already available).

### DIFF
--- a/src/test/org/codehaus/groovy/classgen/ClassCompletionVerifierTest.java
+++ b/src/test/org/codehaus/groovy/classgen/ClassCompletionVerifierTest.java
@@ -73,11 +73,20 @@ public class ClassCompletionVerifierTest extends TestSupport {
             "Method 'prom' is protected but should be public in interface 'zzz'.";
     private static final String EXPECTED_PRIVATE_METHOD_ERROR_MESSAGE =
             "Method 'prim' is private but should be public in interface 'zzz'.";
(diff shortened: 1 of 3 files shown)


shils:
+1

blackdrag:
+1

src/test/org/codehaus/groovy/classgen/ClassCompletionVerifierTest.java
```diff
@@ -73,13 +73,22 @@
             "Method 'prom' is protected but should be public in interface 'zzz'.";
     private static final String EXPECTED_PRIVATE_METHOD_ERROR_MESSAGE =
             "Method 'prim' is private but should be public in interface 'zzz'.";
+    private static final String EXPECTED_ABSTRACT_PRIVATE_METHOD_ERROR_MESSAGE =
+            "Method 'y' from class 'X' must not be private as it is declared as abstract method.";
```
PascalSchumacher: Should this be "... as an abstract method"?

andresteingress: Yes, "as an" sounds more correct, I will fix it.

PascalSchumacher:
Thanks! :+1:

PR #329, opened 2016-05-07
GROOVY-7833: let grape ignore runner lines with hash to enable comments.

blackdrag:
When getting a groovy jar via grab, there will be a TestNG runner configuration in the jar. This configuration contains the Apache License 2 text, as required by the foundation. The code expected the class only. With this change the comments will be ignored.

### DIFF
--- a/src/main/groovy/grape/GrapeIvy.groovy
+++ b/src/main/groovy/grape/GrapeIvy.groovy
@@ -339,7 +339,10 @@ class GrapeIvy implements GrapeEngine {
 
     void processRunners(InputStream is, String name, ClassLoader loader) {
         is.text.readLines().each {
-            GroovySystem.RUNNER_REGISTRY[name] = loader.loadClass(it.trim()).newInstance()
+            def line = it.trim()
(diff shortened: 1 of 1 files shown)


paulk-asert:
+1, but also, perhaps checking for blank lines, e.g. by adding an explicit check or a grep() would also be useful (since blank line(s) now seem to be in the file), i.e. something like:
`is.text.readLines()*.trim().grep().each { line ->`
(with the subsequent `def line` no longer required)

I'd also make the comment that runners are the only "module feature" that we check for only within grab rather than when loading the jar. I think we should make it like the other features and always check. But still +1 for the change - just where it is called from would change slightly.

blackdrag:
How about `is.text.readLines()*.trim().grep { !(it=="" || it[0]=="#") }.each { line ->` ?

PascalSchumacher:
@Jochen: I guess you should update and merge this, so it will be included in the 2.4.7 release.

paulk-asert:
Merged (including the empty line tweak)

PR #330, opened 2016-05-07
GROOVY-7697: GroovyScriptEngine.loadScriptByName doesn't support envi…

andresteingress:
…ronment variable "groovy.ast"

This PR fixes an issue where the "groovy.ast" dump (only god, its author and the ticket author knew this existed) would not be saved when a script was loaded by name via the GSE. In that case, the `XStreamUtils` instance was given an URI (String) instead of an absolute path and the code there broke.

Regarding error handling in `XStreamUtils` (as this was also mentioned in the ticket): I did not change anything into that direction as I am not convinced that such a secondary "add-on" feature should in any way influence behaviour in `SourceUnit#convert`.

### DIFF
--- a/src/main/org/codehaus/groovy/control/XStreamUtils.java
+++ b/src/main/org/codehaus/groovy/control/XStreamUtils.java
@@ -21,17 +21,44 @@ package org.codehaus.groovy.control;
 import com.thoughtworks.xstream.XStream;
 import com.thoughtworks.xstream.io.xml.StaxDriver;
 
+import java.io.File;
 import java.io.FileWriter;
+import java.net.URI;
 
 public abstract class XStreamUtils {
+
     public static void serialize(final String name, final Object ast) {
+        if (name == null || name.length() == 0) return;
+
         XStream xstream = new XStream(new StaxDriver());
         try {
-            xstream.toXML(ast, new FileWriter(name + ".xml"));
+            File astFile = astFile(name);
+            if (astFile == null) {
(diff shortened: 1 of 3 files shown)


PascalSchumacher:
When I run the build with the changes from the pull request locally (where tests run in parallel), this test always fails:

>groovy.util.ProxyGeneratorTest > testUnknownMethodThrowsUnsupportedOperationException FAILED
    groovy.lang.MissingMethodException: No signature of method: static groovy.util.ProxyGenerator.instantiateAggregateFromInterface() is applicable fo
r argument types: (java.util.LinkedHashMap, java.lang.Class) values: [[myMethodA:groovy.util.ProxyGeneratorTest$_testUnknownMethodThrowsUnsupportedOpe
rationException_closure16@6cb59bd9], ...]
    Possible solutions: instantiateAggregateFromInterface(java.lang.Class), instantiateAggregateFromInterface(java.util.Map, java.lang.Class)

not sure how it could be related.

andresteingress:
Thanks for hinting on that, @PascalSchumacher. I will have a look at the code again once I find some time this week(end).

PascalSchumacher:
Thanks. I'm not sure this is caused by the changes of this pull request, maybe the additional tests just reveal a thread-safety issue of other tests (which now run in parallel because of the new tests).

jwagenleitner:
Thanks!

I tested a build with this PR on several different machines and didn't hit the test failure that @PascalSchumacher had.

PR #331, opened 2016-05-09
GROOVY-7291 - Declaration of double variable without value assignment referrenced in closure

jwagenleitner:
Commit e08c389d67809a421f4f33c27012c54294f98259 made the change that broke this by resolving the type (Double) whereas the original method call resolved to `getOriginType` (double).  This change tries to preserve the call to the type resolver where the origin is not a primitive.  It should also be noted that fully reverting the call back to the original (`compileStack.defineVariable(ve, false)`) version prior to commit e08c389d67809a421f4f33c27012c54294f98259  also passes all tests.

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/asm/BinaryExpressionHelper.java
+++ b/src/main/org/codehaus/groovy/classgen/asm/BinaryExpressionHelper.java
@@ -297,7 +297,8 @@ public class BinaryExpressionHelper {
                 !(leftExpression instanceof TupleExpression) )
         {
             VariableExpression ve = (VariableExpression) leftExpression;
-            BytecodeVariable var = compileStack.defineVariable(ve, controller.getTypeChooser().resolveType(ve, controller.getClassNode()), false);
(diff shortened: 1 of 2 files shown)


blackdrag:
hmm.... I think this fix goes the wrong way. We have those resolve calls in there for a reason after all. Did you consider fixing the resolve call? Or does that break things everywhere else?

jwagenleitner:
I had tried to fix it in [StatementMetaTypeChooser#resolveType(Expression,ClassNode)] (https://github.com/apache/groovy/blob/b582ef9bafa92725342797a93a15003a7f9fba1a/src/main/org/codehaus/groovy/classgen/asm/StatementMetaTypeChooser.java#L44) (basically adding condition where originType != primitive) but that broke the build pretty badly.  I sort of figured that from there there would be no way to determine it was only a declaration with an empty RHS expression.  But I'm not knowledgeable in this part of the code, so possible I just didn't apply the right fix.

jwagenleitner:
I tried to fix the resolve call but had no luck.  I will close this since it's not an area I'm familiar with and not sure what the desired bytecode should look like for a closure shared variable vs one that isn't.

PR #332, opened 2016-05-11
GROOVY-7837: TupleConstructor should provide pre and post annotation …

paulk-asert:
…attributes like MapConstructor


paulk-asert:
PR merged

PR #333, opened 2016-05-12
GROOVY-7826: Break infinite recursion in ClassNode.genericTypeAsString

daspilker:


daspilker:
@paulk-asert I can't comment on GROOVY-7826 in JIRA, so I'm doing it here.

See https://github.com/apache/groovy/pull/333/commits/62382ba0604880a5fb6043483ee1aa22d0ce949f for a reproducer on master. On master the `ClassNode`s of the generic type bounds are `org.codehaus.groovy.ast.decompiled.DecompiledClassNode` which do not carry the generic type information. And that breaks the recursion. When disabling ASM resolving, the problem can be reproduced on master.

It would be great if this could be fixed in 2.4. I backported my fix to 2.4 and testing this Jenkins ([JENKINS-34751](https://issues.jenkins-ci.org/browse/JENKINS-34751)) has been successful.

blackdrag:
+1

PascalSchumacher:
Thanks! :+1:

jglick:
Thanks!

PR #335, opened 2016-05-19
GROOVY-6203: GroovyScriptEngine does not take script base class into …

roded:
…account

https://issues.apache.org/jira/browse/GROOVY-6203

The alternative as I see it was to remove the config's setter and add it to the constructors. 
Comments welcome.


paulk-asert:
+1, this looks correct to me

PascalSchumacher:
Thanks! :+1:

PR #338, opened 2016-05-24
GROOVY-7841: Assert fails when accessing particular primitive values …

paulk-asert:
…with @CompileStatic

### DIFF
--- a/src/main/org/codehaus/groovy/transform/sc/transformers/BooleanExpressionTransformer.java
+++ b/src/main/org/codehaus/groovy/transform/sc/transformers/BooleanExpressionTransformer.java
@@ -134,15 +134,19 @@ public class BooleanExpressionTransformer {
                             // int on stack
                         } else if (top.equals(ClassHelper.long_TYPE)) {
                             MethodVisitor mv = controller.getMethodVisitor();
-                            mv.visitInsn(L2I);
+                            mv.visitInsn(LCONST_0);
+                            mv.visitInsn(LCMP);
                             controller.getOperandStack().replace(ClassHelper.boolean_TYPE);
                         } else if (top.equals(ClassHelper.float_TYPE)) {
                             MethodVisitor mv = controller.getMethodVisitor();
-                            mv.visitInsn(F2I);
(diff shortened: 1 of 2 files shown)


src/main/org/codehaus/groovy/transform/sc/transformers/BooleanExpressionTransformer.java
```diff
@@ -134,15 +134,19 @@ public void visit(final GroovyCodeVisitor visitor) {
...
                         } else if (top.equals(ClassHelper.long_TYPE)) {
                             MethodVisitor mv = controller.getMethodVisitor();
-                            mv.visitInsn(L2I);
+                            mv.visitInsn(LCONST_0);
+                            mv.visitInsn(LCMP);
                             controller.getOperandStack().replace(ClassHelper.boolean_TYPE);
                         } else if (top.equals(ClassHelper.float_TYPE)) {
                             MethodVisitor mv = controller.getMethodVisitor();
-                            mv.visitInsn(F2I);
+                            mv.visitInsn(F2D);
```
shils: Why not use a compare float operation here?

paulk-asert: We probably could. In general, comparisons with floats are notoriously ugly, e.g. see:
https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/
But given we are comparing against 0.0, it is probably an edge case that isn't as bad as the general case (but see 'Infernal zero' in the aforementioned link). A general recommendation is to use doubles instead of floats but I don't really know for sure whether it does in this case. My rationale was to look at what bytecode was produced from `assert floatValue != 0.0f` and then replicating that - which is where the `F2D` and `DCMPG` came from.

paulk-asert:
merged

src/main/org/codehaus/groovy/transform/sc/transformers/BooleanExpressionTransformer.java
shils: Thanks for the explanation.

PR #341, opened 2016-05-30
Groovy6950

paulk-asert:


paulk-asert:
I am in two minds about 'borrowing' the Commons IO class and I also haven't done any performance testing as yet. It _should_ be much faster.

PascalSchumacher:
I think it's o.k. (and not that uncommon) for apache projects to copy code from other apache projects.

paulk-asert:
I am not too worried about the legality of copying - no problems there. Just the duplication warning bell going off in the back of my mind... Another class to maintain and possibly deprecate should would revamp all such functionality to use a stream-based solution down the track. I guess I should just do some performance tests and if the gain is there, put in some comments to discourage direct usage of the class outside our internal usage.

paulk-asert:
OK, some local testing gave me about a 20% performance improvement in performance (albeit processing one sample text many times over) with the changes for this PR, so it seems worth putting in.

PR #342, opened 2016-05-31
GROOVY-7620: No error if abstract getter is not implemented but stati…

paulk-asert:
…c field exists


paulk-asert:
This is the fix proposed for 2_4_X. I am actually wondering whether for master we actually implement a stricter version which disallows inherited non-abstract methods which clash with static methods. The PR just fixes the inherited abstract method case to avoid the reported `AbstractMethodError`. This would involve removing ClassCompletionVerifier#L100 and tweaking the test for Groovy-4265.

paulk-asert:
I haven't at this point done anything special for 2.5+. I think we should have the same behavior as Java and reject all static/instance clashes but we can do that as another issue.

PR #344, opened 2016-06-02
Adding Android Classifier

pieces029:
Adding an android classifier in addition to grooid classifier.
They are both the same jar file, just a more specific/descriptive name

@melix hope this change is okay with you. I'm on the home stretch of groovy-android-gradle-plugin, and will be updating docs soon both here and for android soon, and figure I might as well try and get this in while I'm updating everything. If we want to leave grooid I'm cool with it too.


melix:
I'm -1 on this one. We already have too many combinations, I don't think adding one more will help.

PR #345, opened 2016-06-03
GROOVY-7853: o.c.g.r.t.DefaultTypeTransformation does not apply the r…

paulk-asert:
…ight toString on primitve arrays when transforming to String

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/typehandling/ShortTypeHandling.java
+++ b/src/main/org/codehaus/groovy/runtime/typehandling/ShortTypeHandling.java
@@ -20,12 +20,12 @@ package org.codehaus.groovy.runtime.typehandling;
 
 import groovy.lang.GString;
 
+import java.util.Arrays;
+
 /**
  * Class providing various short paths for type conversions. Read the comments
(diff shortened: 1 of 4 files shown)


src/main/org/codehaus/groovy/runtime/typehandling/ShortTypeHandling.java
```diff
@@ -41,7 +43,15 @@ public static Class castToClass(Object object) {
 
     public static String castToString(Object object) {
         if (object==null) return null;
-        if (object instanceof Class) return (String) object;
+        if (object instanceof boolean[]) return Arrays.toString((boolean[])object);
```
gillius: Would it be more efficient to put this within an `if(object.getClass().isArray)` block?

paulk-asert: Good catch. Yes, that's probably a good optimization. Thanks for the
suggestion.

paulk-asert:
Updated per (I believe) Jochen's comment in the mailing list.

PR #346, opened 2016-06-03
GROOVY-7849: Verifier should be aware of array type covariance when c…

shils:
…hecking overriding method return types


blackdrag:
+1

paulk-asert:
+1

shils:
Just noticed your comment; incidentally I did add a test for GROOVY-7185 (https://github.com/apache/groovy/commit/0b3311357a8bcd4b17a97ddf212724ffd96300f2), though it didn't include an assertion as in your example. Thanks for the suggestion.

PR #347, opened 2016-06-12
Fix jigsaw failure IllegalAccessException cannot access JavacTool

tkruse:
I noticed the TeamCity build still fails jigsaw builds:
http://ci.groovy-lang.org/viewLog.html?buildId=33662&buildTypeId=Groovy_Jdk9Build&tab=buildLog

    Caused by: java.lang.IllegalAccessException: class org.gradle.internal.reflect.DirectInstantiator cannot access class com.sun.tools.javac.api.JavacTool (in module jdk.compiler) because module jdk.compiler does not export com.sun.tools.javac.api to unnamed module ...

Since I found a solution, I make this PR to save someone else's time.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,6 +76,14 @@ allprojects {
     if (JavaVersion.current().java7Compatible) {
         apply from: "${rootProject.projectDir}/gradle/asciidoctor.gradle"
     }
+    if (JavaVersion.current().java9Compatible) {
+        tasks.withType(JavaCompile) {
+            options.forkOptions.jvmArgs << '-XaddExports:jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED'
+        }
+        tasks.withType(GroovyCompile) {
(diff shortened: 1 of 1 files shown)


tkruse:
Note you'd still need `-XaddExports:jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED` in the GRADLE_OPTS for jigsaw builds for compilations subtasks without a fork, or you need to start all compilation subtasks as forks using `options.fork = true` in the above. Not sure which solution is preferable.

blackdrag:
There are two kinds f failures wit this one. One is gradle, which needs those switches till Cedrics patch is included. The other is the use of javac in our tests, which should use the compiler API instead. I am basically ok with the patch, since we don´t know wen Cedirc´s patch is included

jwagenleitner:
Now that master is using Gradle 3.0-milestone-2 is this patch still required?  The jigsaw is running but currently failing on `javax.xml.bind` since it's no longer in the base java.se module group.

tkruse:
Maybe not, in any case the Pr is not required to fix it if still broken, so I'll close.

PR #348, opened 2016-06-12
Groovy7860 AutoImplement transform

paulk-asert:


paulk-asert:
Still needs asciidoc documentation but PR created to solicit any potential early feedback.

paulk-asert:
doco added

src/main/groovy/transform/Undefined.java
```diff
@@ -28,6 +28,10 @@
     private Undefined() {}
     public static final String STRING = "<DummyUndefinedMarkerString-DoNotUse>";
     public static final class CLASS {}
+    public static final class UNDEFINED_EXCEPTION extends RuntimeException {
```
shils: I think it would be more in line with the other Undefined fields if the 'UNDEFINED_' prefix were dropped from the field name e.g. 'EXCEPTION_CLASS' or something to that effect.

paulk-asert: Changed. Thanks for the suggestion.

PR #349, opened 2016-06-12
GROOVY-7235-workaround: Ignore tests failing with Java 9 

tkruse:
I won't have time to fix this anytime soon, and in the meantime groovysh import completion might not be crucial for Groovy.

So this is the best I can offer for now, to help with getting Groovy ready for java 9.


jwagenleitner:
Seems like having these tests fail is a good thing, as they serve as a reminder that they probably will need to be addressed at some point.  They will be in good company with many other failing tests under jigsaw. :smile: 

If they are bypassed like this they could be easily forgotten.  I see you've made many contributions in this area so will defer to what you think is best, just playing devil's advocate.

tkruse:
My thinking is that the feature that breaks with java9 is not crucial for groovy. Anyway, open PRs are ugly, might as well close this either way.

tkruse:
My thinking is that the feature that breaks with java9 is not crucial for groovy. Anyway, open PRs are ugly, might as well close this either way.

PR #352, opened 2016-06-17
GROOVY-7865: Better error message in the presence of Generics arity e…

paulk-asert:
…rrors (spike)


paulk-asert:
Just a preliminary spike. It still needs tests but I'd prefer to tidy up the code quite a bit before committing. Currently the GenericsVisitor is invoked twice. There must be a simpler more efficient way.

paulk-asert:
a little bit of rework and some tests added

paulk-asert:
I will probably close this in preference to PR #353 which is a slightly more strict but cleaner alternative.

paulk-asert:
closed as per last comment

PR #354, opened 2016-06-23
GROOVY-7872 Fix nested calls between @Lazy static properties

endSly:
Variable expressions lost owner class context cause they are wrapped into the holder class. Now variable initialization is handled by a private static method in the owning class.


paulk-asert:
Looks good on initial glance.

Just confirming, non-static nesting was working but static not?

Also, has a Jira been created to track this?

endSly:
It only affects to static properties since inner class is used as wrapper of variable for lazy initialization.

I have just reported the issue (GROOVY-7872).

paulk-asert:
merged, thanks!

PR #356, opened 2016-06-25
GROOVY-7873 - Regression performance issue in string methods

jwagenleitner:
In order to implement CharSequence the GString implementation must call toString each time a CharSequence method is invoked.  This is expensive, so GStrings should be converted toString if methods such as charAt are going to be called from a loop.


paulk-asert:
+1

PR #358, opened 2016-06-29
GROOVY-7842 - MarkupTemplateEngine Totally Broken

jwagenleitner:
When running under a Security Manager an exception would be thrown because
the empty codebase value was added as a Permission name which can not be
empty.  So use the same codebase value that is used by GStringTemplateEngine
and StreamingTemplateEngine for unnamed code sources.


jwagenleitner:
For reference, here is where the same codebase value is used in the [GStringTemplateEngine](https://github.com/apache/groovy/blob/941b3c57f7fb9083a2dab9448afb27010b010271/subprojects/groovy-templates/src/main/groovy/groovy/text/GStringTemplateEngine.java#L196) and the [StreamingTemplateEngine](https://github.com/apache/groovy/blob/941b3c57f7fb9083a2dab9448afb27010b010271/subprojects/groovy-templates/src/main/groovy/groovy/text/StreamingTemplateEngine.java#L607).

paulk-asert:
+1

PR #359, opened 2016-06-29
GROOVY-7877: Range step method could work with Number not just int (a…

paulk-asert:
…lternative utility method implementation)

### DIFF
--- a/src/main/groovy/lang/StepUtil.java
+++ b/src/main/groovy/lang/StepUtil.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+
+package groovy.lang;
+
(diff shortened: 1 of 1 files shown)


paulk-asert:
Still needs tests - useful as is for discussion purposes. We could have additional variants which took a Range instead of the initial/max params.

tkruse:
Some Javadoc about the purpose of the class and methods would make a review easier.

paulk-asert:
I will add doco at some point. Basically, a moire functional style for the proposed Number variants for step on Range.

tkruse:
Comments: 
* The name "maxAllowedValue" is confusing, since this may be a minimum, something like "limit" might work better
* Not sure how to reuse these methods for a lazy / streaming approach of reading n steps and then aborting prematurely (like StepIterator inner class in ObjectRange)
* 2nd Exception should include feedback information about the values used
* the code path at the bottom half of step when (initialValue == maxAllowedValue and step!=0) looks dodgy and overly complicated, seems to me the final loop will never execute in that case, so code could return early in that case
* inclusive / exclusive limit distinction not yet there
* Not sure if initialValue, limit and step have all to be of same generic type T
* Not sure if while loop with comparison each time might be safely optimized by calculating how many numbers should be passed to closure, saving the time of comparisons (safely means rounding errors safe). 
* Note that adding step n times is also prone to some rounding "influence" compared to adding (i * step) to initialValue in the loop (many rounding errors for each addition vs. 2 rounding errors (1 multiplication and 1 addition).

paplorinc:
Superseded by https://github.com/apache/groovy/pull/360, I think.
If not, let me know how I should include it.

paulk-asert:
Only possibly superseded by #360 but should not be included in #360. The idea was to consider an alternative to #360 that stopped the Range class from getting too complex and didn't suffer from backwards compatibility issues. So, the idea was one or the other. If we go ahead with #360, we would close this one.

paulk-asert:
@tkruse which line are you worried about being "dodgy"?
The "ascending = ascendingStep" line?
That just covers the case when a single number will be selected and makes sure we don't head off in an infinite loop in the wrong direction.

tkruse:
yes, that line. Unless I am mistaken, the while loop cannot be entered anyway in that case, because neither of the two conditions greaterThan and lessThan can be true if both values are same.

paulk-asert:
I am going to close this PR. I have an alternative spike which I'll turn into a PR once finished which takes a slightly different approach.

PR #360, opened 2016-06-29
`ObjectRange` and `IntRange` iteration and stepping unification

paplorinc:
This commit started out providing non-integer step sizes, but it generated more confusion than expected, therefore only the non-conflicting commits were kept: unifications for Range's iteration and stepping.

### DIFF
--- a/src/main/groovy/lang/IntRange.java
+++ b/src/main/groovy/lang/IntRange.java
@@ -385,26 +385,22 @@ public class IntRange extends AbstractList<Integer> implements Range<Integer> {
             return; // from == to and step == 0, nothing to do, so return
         }
 
-        if (isReverse()) {
-            step = -step;
+        final boolean isAscending;
+        if (step < 0) {
+            step *= -1;
+            isAscending = isReverse();
+        } else {
+            isAscending = !isReverse();
         }
-        if (step > 0) {
-            int value = getFrom();
-            while (value <= getTo()) {
+
+        final long from = getFrom().longValue(), to = getTo().longValue();
+        if (isAscending) {
+            for (long value = from; value <= to; value += step) {
                 closure.call(value);
-                if (((long) value + step) >= Integer.MAX_VALUE) {
-                    break;
-                }
-                value = value + step;
             }
         } else {
-            int value = getTo();
-            while (value >= getFrom()) {
+            for (long value = to; value >= from; value -= step) {
                 closure.call(value);
-                if (((long) value + step) <= Integer.MIN_VALUE) {
-                    break;
-                }
-                value = value + step;
             }
         }
     }
--- a/src/main/groovy/lang/ObjectRange.java
+++ b/src/main/groovy/lang/ObjectRange.java
@@ -237,16 +237,16 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
         if (index < 0) {
             throw new IndexOutOfBoundsException("Index: " + index + " should not be negative");
         }
-        final StepIterator iter = new StepIterator(this, 1);
 
-        Comparable value = iter.next();
-        for (int i = 0; i < index; i++) {
-            if (!iter.hasNext()) {
+        int i = 0;
+        for (Iterator<Comparable> iter = new LazySteppingIterator(this, 1); ; i++) {
+            final Comparable next = iter.next();
+            if (i == index) {
+                return next;
+            } else if (!iter.hasNext()) {
                 throw new IndexOutOfBoundsException("Index: " + index + " is too big for range: " + this);
             }
-            value = iter.next();
         }
-        return value;
     }
 
     /**
@@ -303,8 +303,7 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
                 }
             } else {
                 // let's brute-force calculate the size by iterating start to end
-                final Iterator<Comparable> iter = new StepIterator(this, 1);
-                while (iter.hasNext()) {
+                for (Iterator<Comparable> iter = new LazySteppingIterator(this, 1); iter.hasNext(); ) {
                     tempsize++;
                     // integer overflow
                     if (tempsize < 0) {
@@ -336,7 +335,7 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
 
         // Performance detail:
         // not using get(fromIndex), get(toIndex) in the following to avoid stepping over elements twice
-        final Iterator<Comparable> iter = new StepIterator(this, 1);
+        final Iterator<Comparable> iter = new LazySteppingIterator(this, 1);
 
         Comparable toValue = iter.next();
         int i = 0;
@@ -374,15 +373,13 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
      */
     @Override
     public boolean contains(Object value) {
-        final Iterator<Comparable> iter = new StepIterator(this, 1);
         if (value == null) {
             return false;
         }
-        while (iter.hasNext()) {
-            try {
-                if (compareEqual(value, iter.next())) return true;
-            } catch (ClassCastException e) {
-                return false;
+
+        for (Iterator<Comparable> iter = new LazySteppingIterator(this, 1); iter.hasNext(); ) {
+            if (compareEqual(value, iter.next())) {
+                return true;
             }
         }
         return false;
@@ -393,8 +390,8 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
         if (step == 0 && compareEqual(from, to)) {
             return; // from == to and step == 0, nothing to do, so return
         }
-        final Iterator<Comparable> iter = new StepIterator(this, step);
-        while (iter.hasNext()) {
+
+        for (Iterator<Comparable> iter = new LazySteppingIterator(this, step); iter.hasNext(); ) {
             closure.call(iter.next());
         }
     }
@@ -404,106 +401,89 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
      */
     @Override
     public Iterator<Comparable> iterator() {
-        // non thread-safe iterator
-        final Iterator<Comparable> innerIterator = new StepIterator(this, 1);
         return new Iterator<Comparable>() {
+            private final Iterator<Comparable> delegate = new LazySteppingIterator(ObjectRange.this, 1);
+
             @Override
             public synchronized boolean hasNext() {
-                return innerIterator.hasNext();
+                return delegate.hasNext();
             }
 
             @Override
             public synchronized Comparable next() {
-                return innerIterator.next();
+                return delegate.next();
             }
 
             @Override
             public synchronized void remove() {
-                innerIterator.remove();
+                delegate.remove();
             }
         };
     }
 
     /**
-     * convenience class to serve in other methods.
-     * It's not thread-safe, and lazily produces the next element only on calls of hasNext() or next()
+     * Iterator with configurable step size
+     * It's not thread-safe, and lazily produces the current element (including the first one)
+     * only on hasNext() or next() calls.
      */
-    private static class StepIterator implements Iterator<Comparable> {
-        // actual step, can be +1 when desired step is -1 and direction is from high to low
+    class LazySteppingIterator implements Iterator<Comparable> {
+        private final Range<?> range;
         private final int step;
-        private final ObjectRange range;
-        private int index = -1;
-        private Comparable value;
-        private boolean nextFetched = true;
+        private final boolean isAscending;
 
-        private StepIterator(ObjectRange range, int desiredStep) {
+        private boolean isNextFetched = false;
+        private Comparable next = null;
+
+        LazySteppingIterator(Range<?> range, int desiredStep) {
             if (desiredStep == 0 && compareNotEqual(range.getFrom(), range.getTo())) {
                 throw new GroovyRuntimeException("Infinite loop detected due to step size of 0");
             }
+
             this.range = range;
-            if (range.isReverse()) {
-                step = -desiredStep;
+            if (desiredStep < 0) {
+                step = -1 * desiredStep;
+                isAscending = range.isReverse();
             } else {
                 step = desiredStep;
+                isAscending = !range.isReverse();
             }
-            if (step > 0) {
-                value = range.getFrom();
-            } else {
-                value = range.getTo();
+        }
+
+        @Override
+        public boolean hasNext() {
+            fetchNextIfNeeded();
+            return (next != null) && (isAscending ? compareLessThanEqual(next, range.getTo())
+                                                  : compareGreaterThanEqual(next, range.getFrom()));
+        }
+
+        @Override
+        public Comparable next() {
+            if (!hasNext()) {
+                return null;
+            }
+
+            fetchNextIfNeeded();
+            isNextFetched = false;
+            return next;
+        }
+
+        private void fetchNextIfNeeded() {
+            if (!isNextFetched) {
+                isNextFetched = true;
+
+                if (next == null) {
+                    next = isAscending ? range.getFrom() // make the first fetch lazy too
+                                       : range.getTo();
+                } else {
+                    next = isAscending ? increment(next, step)
+                                       : decrement(next, step);
+                }
             }
         }
 
         @Override
         public void remove() {
-            range.remove(index);
-        }
-
-        @Override
-        public Comparable next() {
-            // not thread safe
-            if (!nextFetched) {
-                value = peek();
-                nextFetched = true;
-            }
-            nextFetched = false;
-            index++;
-            return value;
-        }
-
-        @Override
-        public boolean hasNext() {
-            // not thread safe
-            if (!nextFetched) {
-                value = peek();
-                nextFetched = true;
-            }
-            return value != null;
-        }
-
-        private Comparable peek() {
-            if (step > 0) {
-                Comparable peekValue = value;
-                for (int i = 0; i < step; i++) {
-                    peekValue = (Comparable) range.increment(peekValue);
-                    // handle back to beginning due to modulo incrementing
-                    if (compareLessThanEqual(peekValue, range.from)) return null;
-                }
-                if (compareLessThanEqual(peekValue, range.to)) {
-                    return peekValue;
-                }
-            } else {
-                final int positiveStep = -step;
-                Comparable peekValue = value;
-                for (int i = 0; i < positiveStep; i++) {
-                    peekValue = (Comparable) range.decrement(peekValue);
-                    // handle back to beginning due to modulo decrementing
-                    if (compareGreaterThanEqual(peekValue, range.to)) return null;
-                }
-                if (compareGreaterThanEqual(peekValue, range.from)) {
-                    return peekValue;
-                }
-            }
-            return null;
+            throw new UnsupportedOperationException();
         }
     }
 
@@ -515,23 +495,41 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
     }
 
     /**
-     * Increments by one
+     * Increments by step size
      *
      * @param value the value to increment
-     * @return the incremented value
+     * @param step  the value to increment by
+     * @return the incremented value or null, if there isn't any
      */
-    protected Object increment(Object value) {
-        return InvokerHelper.invokeMethod(value, "next", null);
+    @SuppressWarnings("unchecked")
+    private Comparable increment(Comparable value, int step) {
+        for (int i = 0; i < step; i++) {
+            final Comparable next = (Comparable) InvokerHelper.invokeMethod(value, "next", null);
+            if (!compareGreaterThan(next, value)) /* e.g. `next` of the last element */ {
+                return null;
+            }
+            value = next;
+        }
+        return value;
     }
 
     /**
-     * Decrements by one
+     * Decrements by step size
      *
      * @param value the value to decrement
-     * @return the decremented value
+     * @param step  the value to decrement by
+     * @return the decremented value or null, if there isn't any
      */
-    protected Object decrement(Object value) {
-        return InvokerHelper.invokeMethod(value, "previous", null);
+    @SuppressWarnings("unchecked")
+    private Comparable decrement(Comparable value, int step) {
+        for (int i = 0; i < step; i++) {
+            final Comparable previous = (Comparable) InvokerHelper.invokeMethod(value, "previous", null);
+            if (!compareLessThan(previous, value)) /* e.g. `previous` of the first element */ {
+                return null;
+            }
+            value = previous;
+        }
+        return value;
     }
 
     /**
--- a/src/main/groovy/lang/ObjectRange.java
+++ b/src/main/groovy/lang/ObjectRange.java
@@ -18,17 +18,16 @@
  */
 package groovy.lang;
 
-import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 import org.codehaus.groovy.runtime.InvokerHelper;
 import org.codehaus.groovy.runtime.IteratorClosureAdapter;
-import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
-import org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation;
 
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.util.AbstractList;
 import java.util.Iterator;
 import java.util.List;
+
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.*;
 
 /**
  * Represents an inclusive list of objects from a value to a value using
@@ -195,7 +194,7 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
 
     private static boolean areReversed(Comparable from, Comparable to) {
         try {
-            return ScriptBytecodeAdapter.compareGreaterThan(from, to);
+            return compareGreaterThan(from, to);
         } catch (ClassCastException cce) {
             throw new IllegalArgumentException("Unable to create range due to incompatible types: " + from.getClass().getSimpleName() + ".." + to.getClass().getSimpleName() + " (possible missing brackets around range?)", cce);
         }
@@ -214,8 +213,8 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
     public boolean equals(ObjectRange that) {
         return that != null
                 && reverse == that.reverse
-                && DefaultTypeTransformation.compareEqual(from, that.from)
-                && DefaultTypeTransformation.compareEqual(to, that.to);
+                && compareEqual(from, that.from)
+                && compareEqual(to, that.to);
     }
 
     @Override
@@ -259,14 +258,10 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
     @Override
     public boolean containsWithinBounds(Object value) {
         if (value instanceof Comparable) {
-            final int result = compareTo(from, (Comparable) value);
-            return result == 0 || result < 0 && compareTo(to, (Comparable) value) >= 0;
+            final int result = compareTo(from, value);
+            return result == 0 || result < 0 && compareGreaterThanEqual(to, value);
         }
         return contains(value);
-    }
-
-    protected int compareTo(Comparable first, Comparable second) {
-        return DefaultGroovyMethods.numberAwareCompareTo(first, second);
     }
 
     /**
@@ -385,7 +380,7 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
         }
         while (iter.hasNext()) {
             try {
-                if (DefaultTypeTransformation.compareEqual(value, iter.next())) return true;
+                if (compareEqual(value, iter.next())) return true;
             } catch (ClassCastException e) {
                 return false;
             }
@@ -395,7 +390,7 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
 
     @Override
     public void step(int step, Closure closure) {
-        if (step == 0 && compareTo(from, to) == 0) {
+        if (step == 0 && compareEqual(from, to)) {
             return; // from == to and step == 0, nothing to do, so return
         }
         final Iterator<Comparable> iter = new StepIterator(this, step);
@@ -441,8 +436,8 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
         private Comparable value;
         private boolean nextFetched = true;
 
-        private StepIterator(ObjectRange range, final int desiredStep) {
-            if (desiredStep == 0 && range.compareTo(range.getFrom(), range.getTo()) != 0) {
+        private StepIterator(ObjectRange range, int desiredStep) {
+            if (desiredStep == 0 && compareNotEqual(range.getFrom(), range.getTo())) {
                 throw new GroovyRuntimeException("Infinite loop detected due to step size of 0");
             }
             this.range = range;
@@ -491,9 +486,9 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
                 for (int i = 0; i < step; i++) {
                     peekValue = (Comparable) range.increment(peekValue);
                     // handle back to beginning due to modulo incrementing
-                    if (range.compareTo(peekValue, range.from) <= 0) return null;
+                    if (compareLessThanEqual(peekValue, range.from)) return null;
                 }
-                if (range.compareTo(peekValue, range.to) <= 0) {
+                if (compareLessThanEqual(peekValue, range.to)) {
                     return peekValue;
                 }
             } else {
@@ -502,9 +497,9 @@ public class ObjectRange extends AbstractList<Comparable> implements Range<Compa
                 for (int i = 0; i < positiveStep; i++) {
                     peekValue = (Comparable) range.decrement(peekValue);
                     // handle back to beginning due to modulo decrementing
-                    if (range.compareTo(peekValue, range.to) >= 0) return null;
+                    if (compareGreaterThanEqual(peekValue, range.to)) return null;
                 }
-                if (range.compareTo(peekValue, range.from) >= 0) {
+                if (compareGreaterThanEqual(peekValue, range.from)) {
                     return peekValue;
                 }
             }


src/main/groovy/lang/Range.java
```diff
@@ -73,23 +73,32 @@
     boolean containsWithinBounds(Object o);
 
     /**
+     * Replaced by <code>step(Number, Closure)</code>
```
paplorinc: This is the essence of the change: allow non-integer step sizes also

src/main/groovy/lang/IntRange.java
```diff
@@ -355,40 +360,54 @@ public boolean containsAll(Collection other) {
...
-            int value = getFrom();
-            while (value <= getTo()) {
-                closure.call(Integer.valueOf(value));
-                if ((0L + value + step) >= Integer.MAX_VALUE) {
-                    break;
-                }
-                value = value + step;
+
+        if (isAscending) {
+            for (Long value = from; value <= to; value = (Long) plus(value, step)) {
```
paplorinc: iterating over a long, instead of constantly checking for overflow

src/main/groovy/lang/ObjectRange.java
```diff
@@ -331,24 +331,23 @@ public List subList(int fromIndex, int toIndex) {
 
         // Performance detail:
         // not using get(fromIndex), get(toIndex) in the following to avoid stepping over elements twice
-        StepIterator iter = new StepIterator(this, 1);
+        Iterator it = iterator();
```
paplorinc: `iterator()` already returns a `new StepIterator(this, 1)`

src/main/groovy/lang/ObjectRange.java
```diff
@@ -368,149 +367,152 @@ public String inspect() {
...
-        while (it.hasNext()) {
-            try {
-                if (DefaultTypeTransformation.compareEqual(value, it.next())) return true;
-            } catch (ClassCastException e) {
-                return false;
+        if (value == null)
+            return false;
+
+        for (Iterator it = iterator(); it.hasNext(); ) {
+            if (compareEqual(value, it.next())) {
```
paplorinc: the replaced `compareEqual` returns `false` instead of throwing

src/main/groovy/lang/ObjectRange.java
```diff
@@ -368,149 +367,152 @@ public String inspect() {
...
     public Iterator iterator() {
-        // non thread-safe iterator
-        final Iterator innerIterator = new StepIterator(this, 1);
-        Iterator safeIterator = new Iterator() {
+        return new Iterator() {
+            final Iterator innerIterator = new StepIterator(ObjectRange.this, 1);
+
             @Override
-            public synchronized boolean hasNext() {
+            public boolean hasNext() {
```
paplorinc: why synchronize, if it's not thread anyway???

src/main/groovy/lang/ObjectRange.java
```diff
@@ -368,149 +367,152 @@ public String inspect() {
...
         public Object next() {
-            // not thread safe
-            if (!nextFetched) {
-                value = peek();
-                nextFetched = true;
-            }
-            nextFetched = false;
-            index++;
-            return value;
+            if (!hasNext()) return null;
```
paplorinc: apparently this was a separate requirement - the reason for the original test failures

paplorinc:
@blackdrag, @PascalSchumacher, @tkruse, @paulk-asert, @vladmm, @terebesirobert, please review :).
The original PR was pending for 1 year, I rebased it [upon request](https://github.com/apache/groovy/pull/120#issuecomment-229251445).
Could we please review and merge this ASAP? (the [Java 6 Travis node](https://travis-ci.org/apache/groovy/jobs/141155140) should be fixed...)

src/test/groovy/RangeTest.groovy
```diff
@@ -262,7 +262,24 @@ class RangeTest extends GroovyTestCase {
         assert (0..<upper2).size() == 4
     }
 
-    protected void assertIterate(range, expected) {
+    void testDecimalRanges() {
```
paplorinc: The rest of my tests were already merged in by @paulk-asert in https://github.com/apache/groovy/commit/d6252c31ebc2caf98d546af2bdfd45aa53166085

.travis.yml
```diff
@@ -16,8 +16,8 @@
 language: groovy
 
 jdk:
-  - openjdk6
```
paplorinc: I hope this will fix the [build](https://travis-ci.org/apache/groovy/jobs/141155140)

tkruse:
I think this would be easier to review and discuss if you made individual commits for individual refactorings and changes. As an example, look at the commits I created of your PR so far: https://github.com/tkruse/incubator-groovy/commits/RangeNumberStepSize

I think the rest could be split further into the changes you made to the synchronized StepIterator, overflow handling, making methods static, and finally the additional step methods with Number. Possibly even more smaller commits.

paplorinc:
@tkruse, I have a strong preference for small, but compilable commits, each having a well specified theme - no too big, but not too small either.

By separating each [import](https://github.com/tkruse/incubator-groovy/commit/2aecc87b5f812b76b8d25dd94a7f6d82f505ebe7), I don't see the story that is being told - i.e. why it's important to do that change.

If however you also have a strong preference for the way you's shown it, I will accommodate.

src/main/groovy/lang/IntRange.java
```diff
@@ -187,7 +187,7 @@ public IntRange(boolean inclusive, int from, int to) {
 
     private void checkSize() {
         // size() in the Collection interface returns an integer, so ranges can have no more than Integer.MAX_VALUE elements
-        Long size = (long) this.to - this.from + 1;
+        final Long size = (long) to - from + 1;
```
paplorinc: very boring commit to make the review of the high risk commit simpler

src/main/groovy/lang/ObjectRange.java
```diff
@@ -493,9 +491,9 @@ private Comparable peek() {
                 for (int i = 0; i < step; i++) {
                     peekValue = (Comparable) range.increment(peekValue);
                     // handle back to beginning due to modulo incrementing
-                    if (range.compareTo(peekValue, range.from) <= 0) return null;
+                    if (compareLessThanEqual(peekValue, range.from)) return null;
```
paplorinc: these will disappear in the next commit anyway :)

src/main/groovy/lang/ObjectRange.java
```diff
@@ -240,16 +237,16 @@ public Object get(int index) {
         if (index < 0) {
             throw new IndexOutOfBoundsException("Index: " + index + " should not be negative");
         }
-        final StepIterator iter = new StepIterator(this, 1);
 
-        Object value = iter.next();
-        for (int i = 0; i < index; i++) {
-            if (!iter.hasNext()) {
+        int i = 0;
+        for (Iterator it = iterator(); ; i++) {
```
paplorinc: `new StepIterator(this, 1)` == `iterator()`

src/main/groovy/lang/ObjectRange.java
```diff
@@ -378,15 +373,13 @@ public String inspect() {
...
             return false;
         }
-        while (it.hasNext()) {
-            try {
-                if (compareEqual(value, it.next())) return true;
-            } catch (ClassCastException e) {
-                return false;
+
+        for (Iterator it = iterator(); it.hasNext(); ) {
+            if (compareEqual(value, it.next())) {
```
paplorinc: this comparison doesn't throw anymore

src/main/groovy/lang/ObjectRange.java
```diff
@@ -426,89 +419,52 @@ public void remove() {
...
-            index++;
-            return value;
+
+            final Object result = next;
+            next = isAscending ? increment(next) : decrement(next);
+            return result;
         }
 
         @Override
         public boolean hasNext() {
```
paplorinc: this seems way simpler...

src/main/groovy/lang/ObjectRange.java
```diff
@@ -482,7 +484,11 @@ public List step(int step) {
      * @return the incremented value
      */
     private static Object increment(Object value) {
-        return InvokerHelper.invokeMethod(value, "next", null);
+        if (value instanceof Number) {
+            return plus((Number) value, 1);
```
paplorinc: this makes it faster for e.g. double ranges (even with integer increments)

src/main/groovy/lang/ObjectRange.java
```diff
@@ -472,37 +484,61 @@ public boolean hasNext() {
...
         if (value instanceof Number) {
-            return plus((Number) value, 1);
-        } else {
-            return InvokerHelper.invokeMethod(value, "next", null);
+            return (Comparable) plus((Number) value, step);
+        }
+
+        for (int i = 0; compareLessThan(i, step); i++) {
+            final Comparable next = (Comparable) InvokerHelper.invokeMethod(value, "next", null);
+            if (!compareGreaterThan(next, value)) /* e.g. `next` of the last element */ {
```
paplorinc: should work for enums also, where after the last element the first should be returned

paplorinc:
Thank you for your suggestion, @tkruse, it made the diff of the critical part a lot simpler indeed.
I pushed the new version, I recommend a [commit-by-commit](https://github.com/apache/groovy/pull/360/commits) review

src/main/groovy/lang/ObjectRange.java
tkruse: The innerIterator is not thread-safe, meaning if two threads access next() in parallel, they may get junk. By synchronizing the wrapper, only one thread at a time gets through to the inner interator, other threads are blocked in the wrapper until the thread inside is done. That way, this method returns a thread-safe iterator.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -357,160 +363,182 @@ public String toString() {
...
+            closure.call(it.next());
         }
     }
 
     /**
      * {@inheritDoc}
      */
+    @Override
     public Iterator iterator() {
-        // non thread-safe iterator
```
paplorinc: @tkruse, I was referring this comment. If it's no thread-safe anyway, why synchronize?

src/main/groovy/lang/ObjectRange.java
tkruse: You change the iterator from being lazy to being eager (see it's javadoc, which you would need to change as well). Eager is simpler, but behaves differently. Performance is differenct, also when fetching the next element throws an exception, then eager approach will throw the exception earlier, meaning if there are three valid entries in the range and one invalid, then lazy will return 3 and then throw an exception, while eager will only return 2 and then throw an exception.

src/main/groovy/lang/ObjectRange.java
tkruse: the comment says the INNER iterator "innerIterator " is not thread safe. The OUTER iterator "safeIterator " should be thread safe. That's why the outer one is synchronized, to make it thread-safe.

tkruse: the comment is not a javadoc that says what the iterator() method should return. it comments the next line.

paplorinc: Understood, will put the synchronization back :)

src/main/groovy/lang/ObjectRange.java
paplorinc: If `hasNext` is used before every `next` (as it's meant to be used), I think they should behave in the same way.
If not, please let me know if I should change something.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -408,107 +401,66 @@ public void step(int step, Closure closure) {
...
-        final Iterator innerIterator = new StepIterator(this, 1);
-        final Iterator safeIterator = new Iterator() {
+        return new Iterator() {
+            private final Iterator delegate = new StepIterator(ObjectRange.this, 1);
+
             @Override
             public synchronized boolean hasNext() {
-                return innerIterator.hasNext();
+                return delegate.hasNext();
```
paplorinc: Thanks @tkruse, force pushed a new version :)

src/main/groovy/lang/ObjectRange.java
tkruse: In general, calling hasNext() is not mandatory, clients are free to just call next() many times. But this should not matter here anyway. To explain it differently: Imagine calling increment takes 5 minutes, because it goes to a database in China or something. Then with your implementation, the first call to next() will take 5 minutes to fetch the future value, whereas the current implementation returns at once, because it only returns the initial value. That's the difference between lazy and eager computation of the next value.

paplorinc: If I understand it correctly, both would return after 5 minutes, because the initial `value` is fetched in both cases at start - hence the initial value of `nextFetched` is `true`.

tkruse: There is a difference in any case, and your commit message does not point it out to reviewers, and the javadoc of StepIterator was not changed.
Such changes make your PR more difficult to merge.

paplorinc: You're perfectly right would you like me to document it or make it lazy?

src/main/groovy/lang/ObjectRange.java
```diff
@@ -408,107 +401,86 @@ public void step(int step, Closure closure) {
...
 
-        @Override
-        public boolean hasNext() {
-            // not thread safe
-            if (!nextFetched) {
-                value = peek();
-                nextFetched = true;
-            }
-            return value != null;
+            fetchNextIfNeeded();
```
paplorinc: @tkruse, thanks for your help, made it lazy again - including the fetching of the first element :)

src/main/groovy/lang/ObjectRange.java
```diff
@@ -357,160 +363,198 @@ public String toString() {
...
             @Override
             public synchronized Object next() {
-                return innerIterator.next();
+                return delegate.next();
             }
+
             @Override
             public void remove() {
-                innerIterator.remove();
+                throw new UnsupportedOperationException();
```
tkruse: while this does not change functionality, it is bad style. If anyone later changes stepiterator to support removal, we'd have to remember chaning here as well. So I'd say keep the previous approach for clarity.

paplorinc: ok, will also synchronize it then

src/main/groovy/lang/ObjectRange.java
```diff
@@ -408,107 +401,86 @@ public void step(int step, Closure closure) {
...
             @Override
             public synchronized Object next() {
-                return innerIterator.next();
+                return delegate.next();
             }
 
             @Override
-            public void remove() {
-                innerIterator.remove();
+            public synchronized void remove() {
```
paplorinc: It wasn't synchronized before, since it throws an exception anyway.
It's more consistent this way, though.

tkruse: no, don't synchronize it. Call innerIterator.remove(); like before. I know innerIterator.remove() also always throws, but still it's better to have just one location that throws.

paplorinc: If we implement `remove` later, it will need synchronization, right?

tkruse: true, I did not think about that.

paplorinc:
Rebased.
If there are no more suggestions, could we please merge it?

src/main/groovy/lang/IntRange.java
```diff
@@ -377,40 +382,52 @@ public boolean containsAll(Collection other) {
...
                 return; // from == to and step == 0, nothing to do, so return
             }
         }
 
-        if (isReverse()) {
-            step = -step;
+        boolean isAscending = !isReverse();
+        if (compareLessThan(step, 0)) {
+            step = multiply(step, -1);
+            isAscending = !isAscending;
```
tkruse: why invert this? Looks dodgy, should be commented in the code if correct

paplorinc: if the step is negative, we're going backwards, i.e. it's not ascending anymore

src/main/groovy/lang/IntRange.java
tkruse: Maybe nitpicking, but step can be a very high number now, right? Like twice Long.MAX_VALUE. So for consistency, overflow via large step values could be a concern.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -488,37 +500,61 @@ public void remove() {
...
-     * Increments by one
+     * Increments by step size
      *
      * @param value the value to increment
-     * @return the incremented value
+     * @param step  the value to increment by
+     * @return the incremented value or null, if there isn't any
      */
-    private static Object increment(Object value) {
+    protected static Comparable increment(Comparable value, Number step) {
```
tkruse: why is this protected now?

src/main/groovy/lang/IntRange.java
paplorinc: It could even be `Double.POSITIVE_INFINITY` or `Double.NaN` or `Double.MIN_VALUE` and probably none would work "correctly".

src/main/groovy/lang/ObjectRange.java
paplorinc: because it got messed up with rebases (changed to private and when I changed the order of commits, it got changed back):
https://github.com/apache/groovy/pull/360/files#diff-edcd680998449caffc50809917864ceeR523

Thanks, fixing it now.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -488,37 +500,61 @@ public void remove() {
...
         if (value instanceof Number) {
-            return plus((Number) value, 1);
-        } else {
-            return InvokerHelper.invokeMethod(value, "next", null);
+            return (Comparable) plus((Number) value, step);
+        }
+
+        for (int i = 0; compareLessThan(i, step); i++) {
```
tkruse: What is the general semantics of stepping by 2.5 in an ObjectRange anyway? It seems to me this cannot do anything reasonable unless we start stepping from a number, and even then, invoking next several times until the loop counter is greater than step seems arbitrary. It might be better to be strict about datatypes and just fail if the user tries to do funny stepping, that way it's easier to change behavior later without breaking code.

paplorinc: Not sure what you mean by `funny stepping`, please take a look at the tests for examples:

https://github.com/apache/groovy/pull/360/commits/dbe78b86a119e85722868a7a8c115e766c363188#diff-4db89e6d38a08032807c8360d4d99199R266

src/main/groovy/lang/IntRange.java
paplorinc: I think github compares the commits in the wrong order, this is the final comparison:
https://github.com/apache/groovy/pull/360/files#diff-edcd680998449caffc50809917864ceeR454
```java
if (compareLessThan(desiredStep, 0)) {
    step = multiply(desiredStep, -1);
    isAscending = range.isReverse();
} else {
    step = desiredStep;
    isAscending = !range.isReverse();
}
```

tkruse:
At least the smaller changes should be mergable early(but I am no committer). The large commit I don't know. 

In general, I am not sure if stepping by arbitrary numbers is a good idea. A range currently extends List, meaning we consider a range to be equivalent to a series of items. [1..5] means [1,2,3,4,5].
Writing [1..5].step(0.5) thus may be regarded as equivalent to [1,2,3,4,5].step(0.5), but that is really weird.

I understand the problem of wanting to represent [0.1, 0.2, 0.3, 0.4] as a range, but I think that extending ranges with custom steps is maybe not the ideal approach.

paplorinc:
Thanks @tkruse for your comments.
@blackdrag, @PascalSchumacher, @paulk-asert, please review.

src/main/groovy/lang/IntRange.java
tkruse: github lists commit in an arbitrary order, but each commit is shown correctly by github, I think.

src/main/groovy/lang/ObjectRange.java
tkruse: because of synchronization, iterator() returns a slower implementation. So should be better to keep raw StepIterator

tkruse: and generally it is unclear to me why you changed the loop style here. I am not sure the new structure is objectively better than before.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -306,14 +303,13 @@ public int size() {
                 }
             } else {
                 // let's brute-force calculate the size by iterating start to end
-                final Iterator iter = new StepIterator(this, 1);
```
tkruse: same comment as before, iterator() is synchronized, thus slower

src/main/groovy/lang/ObjectRange.java
```diff
@@ -339,24 +335,23 @@ public List subList(int fromIndex, int toIndex) {
 
         // Performance detail:
         // not using get(fromIndex), get(toIndex) in the following to avoid stepping over elements twice
-        final StepIterator iter = new StepIterator(this, 1);
```
tkruse: same as before, iterator() is synchronized

src/main/groovy/lang/ObjectRange.java
```diff
@@ -378,15 +373,13 @@ public String inspect() {
      */
     @Override
     public boolean contains(Object value) {
-        final Iterator it = new StepIterator(this, 1);
```
tkruse: same here

src/main/groovy/lang/ObjectRange.java
paplorinc: We're not in a critical section here, and uncontended synchronization is pretty cheap anyway.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -499,7 +512,7 @@ public List step(int step) {
      * @param value the value to increment
      * @return the incremented value
      */
-    protected Object increment(Object value) {
+    private Object increment(Object value) {
```
tkruse: not sure if you forgot a force push, but this commit should not change visibility. The groovy-team cherry-picks commits, so they must be valid individually

src/main/groovy/lang/ObjectRange.java
paplorinc: I answered your first related comment instantly, no need to repeat it :).

src/main/groovy/lang/ObjectRange.java
paplorinc: > you changed the loop style here

It's more compact this way, enhances readability by increasing the signal to noise ratio: related looping constructs occupy fewer lines.

src/main/groovy/lang/ObjectRange.java
paplorinc: force pushed now.

src/main/groovy/lang/ObjectRange.java
paplorinc: I measure the difference using JMH on a single thread:

```java
final static List<Integer> iterable = List.range(1, 1000000);

public static class Iter {
    @Benchmark
    public int synchronized_iteration() {
        final Iterator<Integer> it = new Iterator<Integer>() {
            private final Iterator<Integer> delegate = iterable.iterator();

            @Override
            public synchronized boolean hasNext() {
                return delegate.hasNext();
            }

            @Override
            public synchronized Integer next() {
                return delegate.next();
            }
        };

        int result = 0;
        while (it.hasNext()) {
            result ^= it.next();
        }
        return result;
    }

    @Benchmark
    public int normal_iteration() {
        final Iterator<Integer> it = new Iterator<Integer>() {
            private final Iterator<Integer> delegate = iterable.iterator();

            @Override
            public boolean hasNext() {
                return delegate.hasNext();
            }

            @Override
            public Integer next() {
                return delegate.next();
            }
        };

        int result = 0;
        while (it.hasNext()) {
            result ^= it.next();
        }
        return result;
    }
}
```
resulting in:
```
Impl                    Score    ±  Error  Unit   normal_iteration  synchronized_iteration
normal_iteration        193.828  ±  0.82%  ops/s                               1.01x
synchronized_iteration  192.702  ±  0.44%  ops/s       0.99x                              
```
i.e. same speed in a non-contended environment (the synchronized is probably JITed away)

src/main/groovy/lang/ObjectRange.java
tkruse: By funny stapping I mean stepping through an enumeration or a range of characters by step 2.7. I believe the current added tests do not cover the changes to ObjectRange well.

paplorinc:
@paulk-asert, thanks for partial merges, I rebased the commit.

src/main/groovy/lang/ObjectRange.java
paplorinc: Feel free to suggest better tests :)

tkruse: Testing with an enum and custom steps sizes (-2, -1, -0.5, 0, 0.5, 1, 2) would be useful, in particular for circular behavior (range starts at beginning) and reverse steps. Similar tests for char ranges with custom step sizes (-2, -1, -0.5, 0, 0.5, 1, 2).

paplorinc: My enum tests were already merged in by @paulk-asert in https://github.com/apache/groovy/commit/d6252c31ebc2caf98d546af2bdfd45aa53166085#diff-4db89e6d38a08032807c8360d4d99199R109

paulk-asert:
@paplorinc, I merged in the less controversial commits to reduce some noise.

As a general comment, we don't have a strict preference for adding "final" everywhere. It used to be the case that we'd have strongly encouraged it. It is after all a statement of programmer intent and used to be a great hint to the compiler. But it does make the code slightly more noisy and the compiler is now usually smart enough to work out when it can optimize things regardless of the keyword. One take is that if your code is so complex that you need to have final keywords to help understand what things might change and what doesn't, then you need to refactor to more simple code. We certainly don't take it that far and so don't actively discourage its use when it makes sense. In any case, this is just for background information, not trying to stop you adding them if you feel strongly about them. Some reviewers might love them, others will see it as noise - so best to keep them in separate commits - which is what you did, so thanks.

src/main/groovy/lang/ObjectRange.java
tkruse: Yes, it would be nice to extend them with steps 0.1, 0.5, 2.7 or so. That's what bI mean with funny stepping.

paplorinc:
@paulk-asert, I used to have a strong opinion about finals also, but I fully agree with your statement, it's how I view comments also: if you need them to make things more obvious (i.e. whether it changed meaning or not), you should probably redesign.

Could you please let me know what you would like me to change for this PR to be merged (again, it's pending for almost 1 year).

src/main/groovy/lang/ObjectRange.java
paplorinc: Please advise on a usecase when you would want something like that (the API has other flexibility also, that I can't find a good usecase for).

Stepping over enums 1.5 at a time should work, but why would I test it? Stepping with `Double.MIN_VALUE` should also work but it doesn't mean anything and would take ages to run: I won't create a test for it if I don't understand what it means ...

paulk-asert:
So, my observations at this point:

- there seems to be a problem with `IntRange` and decimal step values:
```
(0..9).step(1.1)
// => java.lang.ClassCastException: java.math.BigDecimal cannot be cast to java.lang.Long
```

- I still think using `StepIterator` (or `LazySteppingIterator`) directly rather than `iterator()` is the way to go internally. Regardless of the performance results of adding in the extra synchronization, it seems better to avoid it when we know it isn't needed.

- I still wonder about decimal step sizes for non-Numbers. I wonder whether people will be surprised by the following results:
```
def r1 = 'a'..'j'
assert r1.step(1.01) == ['a', 'c', 'e', 'g', 'i']
assert r1.step(1.99) == ['a', 'c', 'e', 'g', 'i']
```
and also:
```
def r1 = 'a'..'j'
assert r1.size() == 10
assert r1.step(1.1).size() == 5
def r2 = 0L..9L
assert r2.size() == 10
assert r2.step(1.1).size() == 9
```

- We also have the breaking change which (if acceptable) would need to be documented. The following script currently executes:
```
assert (1G..9G).step(2) == [1, 3, 5, 7, 9]
BigInteger.metaClass.next << { delegate + 2 }
assert (1G..9G).step(2) == [1, 5, 9]
```
But fails after the changes. This is to be expected since we removed the "reflective" path and replaced it with an optimized path. In other places in Groovy we do some optimizations like this when we know metaprogramming isn't in play but such checking currently isn't easy to do within the Range classes.

src/main/groovy/lang/ObjectRange.java
tkruse: I would never want to do that. That's my point. ObjectRange should not offer any behavior for undesired cases, but rather thrown an InvalidArgumentExceptions in all cases that have no reasonable outcome, in my opinion. Leaving things undefined is technical debt that Groovy does not need any more of. Writing tests would demonstrate the problem to reviewers, that your PR currently introduces behavior that cannot be understood, but may have to be supported in future versions.

paplorinc: Could you please provide alternatives also, when writing criticism, otherwise it seems just nitpicking.

I consider that the advantages outweigh the above mentioned problems for the above mentioned reasons.

paplorinc:
> it seems better to avoid it when we know it isn't needed

I find it premature optimization, but will accommodate.

> ```groovy
def r1 = 'a'..'j'
assert r1.step(1.01) == ['a', 'c', 'e', 'g', 'i']
assert r1.step(1.99) == ['a', 'c', 'e', 'g', 'i']
```

Not sure what people would expect to happen in this case, as characters are not numbers? :/

> (0..9).step(1.1)

Thanks, my mistake. Fixed it:
```groovy
def intRange = (0..10)
assertIterate(intRange.step(1.5), [0, 1, 3, 4, 6, 7, 9])
```

> We also have the breaking change which (if acceptable) would need to be documented

I added documentation to `increment` and `decrement`, though it wasn't documented previously in `ObjectRange` that they were using `next`/`previous` internally.

Force pushed a new version (with a separate `Comparable`-typing commit).

src/main/groovy/lang/ObjectRange.java
tkruse: I consider throwing an IllegalArgumentException in ObjectRange for non-Integer steps when the starting value is not a number an alternative. Same for IntRange, it may be argued that an IntRange should not be steppable by anything else than an int, i.e. clients may expect the output of stepping an IntRange to be ints only.

Overall, I think I might generally prefer to rename the method step() to something else for non-int "step" values, because it does not step through the values the range represents, but instead it creates a new range with different values. something like scale() or toNewRange().

paplorinc: > it does not step through the values the range represents

what values does (0.1..0.9) represent?

> I consider throwing an IllegalArgumentException in ObjectRange

I don't mind doing this, @paulk-asert?

tkruse: (0.1..0.9) == [0.1]. That's not intuitive, but that's the definition according to javadoc. Call that historic reasons. It might be nice to have Range objects with variable item distance, with a syntax such as 1..2..8, 0..5..100, 0.0..0.1..0.9. 

But creating a range with a variable item distance and traversing through an interval (not range) using a flexible iteration distance are very different things, and it seems to me this PR mixes them up.

Also see this for similar issues in Swift:
* http://oleb.net/blog/2015/09/swift-ranges-and-intervals/
* https://www.hackingwithswift.com/example-code/language/using-stride-to-loop-over-a-range-of-numbers

paulk-asert:
Hmm... what do others think about instead of messing with the step method we introduced another constructor just in ObjectRange.
ObjectRange (Number from, Number to, Number step, boolean inclusive)

tkruse:
@paulk-asert : A new constructor does not help with the "0.1..0.9" syntax use-case targeted with this PR. 
And such a series of numbers NOT using "next()" and "previous()" goes against the javadoc of groovy.lang.Range: "A Range represents the list of all items obtained by starting from a from value and calling next() successively until you reach the to value. [...previous() ...]"

One conceptually clean solution is to create an Object Range starting with a wrapped Number that also has a numeric value to be used in "next", like ```new StridingNumer(0.1, 0.1)..0.9```, but that seems clumsy.

I can also imagine a swift-like alternative enhancing Numbers like this: 0.1.stride(0.1, 0.9), as in the link I posted earlier https://www.hackingwithswift.com/example-code/language/using-stride-to-loop-over-a-range-of-numbers

paulk-asert:
@tkruse, agreed I am not trying to solve syntax use case but trying to get the abstraction right first.

paplorinc:
> 0.1.stride(0.1, 0.9)

so `1.stride(0, 9)` would be `(0..9).step(1)`? ... the first seems very counter intuitive to me

paplorinc:
> 0.1.stride(0.1, 0.9)

so `1.stride(0, 9)` would be `(0..9).step(1)`? ... the first seems very counter intuitive to me.

I think the problem originates from the design decision of making `IntRange` and `ObjectRange`, and not `NumberRange` vs `ComparableRange`. I guess it's too late to redesign that, right?

tkruse:
0.stride(1, 9) would be similar to 0..9, not 1.stride...

paplorinc:
So it's even more confusing...

tkruse:
@paulk-asert : 
Generally we should avoid the word "step" for the extension you propose. I recommend the article on ranges in Swift. Ranges represent indexed lists, that can be represented as a list of pairs:
[(0, a), (1, b), (2, c), ...]. A "step" is the distance between 0 and 1 as an example, but you want to introduce a custom distance between "a" and "b". Maybe call that a "stride" instead of a "step".

The abstraction of groovy.lang.Range is based on next() and previous() as "strides", meaning for domains where elements have a useful inherent indexing order, a segment within that domain can be represented by a lower and upper boundary. That happens to be useful in the case of Integers, single characters, and little else (some obscure custom domain classes maybe), where the implicit "stride" is obvious. Even enums are dubious candidates IMO, because their order is somewhat inexplicit.

Your suggestion is to open this up to a different abstraction, to describe series of elements using a lower boundary, and upper boundary, and a custom stride.

As mentioned trying to represent such a series may be done outside the scope of groovy Ranges, extending Numbers with a stride function.

A similar way to create number series with a fixed stride is:

    (1..9).collect {it * 0.1}   ==    [0.1, 0.2, 0.3, ..., 0.9]

It is more typing and a bit less obvious than

  (0.1..0.9).stride(0.1)   ==    [0.1, 0.2, 0.3, ..., 0.9]

but on the other hand the closure notation is way more extensible than a fixed stride Range. In general I would favour teaching Groovy Programmers to avoid any Range unless based on Integers or special custom classes with custom next() and previous() methods, and use the integer-based range + closure instead. The concept of a range simply does not fit much else than integers. In other words, for me, Groovy 3 could ship without the ObjectRange class, as it brings more problems than benefits. It would even be debatable for me if the ".." operator is useful enough compared to a mere range function like pythons.

blackdrag:
Let us go on the concept side and do step-by-step..

Definition1:
Normally I would define a range along this: If I have a Range X..Y, then for all i<=Y && i>=X, the Range contains i. And that's it. For this I need compareTo on X,Y as well as a "contains" on Range. 

The problem now of course is that the Groovy range is almost never used in this manner, so this clean definition will not do the job. Let us then define the Rang as a List with a lower and upper bound. That makes X, the first element of the list. But how to define the other elements?

Definition2: use next(). This easily realizes a stepping of 1, not as a numeric value as such, but as one-by-one. Now we can define l_0 as X, and l_1 as l_0.next(). For l_(n+1) we define l_n.next(). To met the old requirement of the range we would also have to have X<=l_i<=Y via compareTo to realize a contains. This still leaves the question of how to handle an i, for which there is no list element, but allows X<=i<=Y. Defining "contains" to behave like in the first range definition makes it clear, that there will be nothing clear to the user, if i has to be a list element from a ranges perspective, but that is exactly the definition imposed by List. Leaving that conflict aside a stepping s (of integer type) can be done by multiple executions of next(). A non-int-stepping will not work for this. For ranges based on numbers an optimization could be done by using plus(n), instead of multiple uses of next().

Definition3: use plus(stepping). So instead of using next, we would here use the plus method as defined through X and Y. l_0 is then still X, l_1 is l_0+stepping. and l_(n+1) = l_n+stepping. This can in theory allow arbitrary steppings, for as long as the plus method allows it. Like before we have a conflict for "contains". Again an optimization could be done for number based ranges with l_n = l_0+stepping*n, which requires the definition of multiply of course.

The aim of this pullrequest (well it got modified so much by now, that talking of "this" pullrequest is not really right anymore, but let us leave it at that) seems for me to go in the direction of Definition3, while Definition2, with i having to be a list element for contains, is more of what we use in Groovy now. But it is a big problem for me to realize Definition3 using logic based on "next" instead of plus. Something like 'a'..'j'.step(1.1) makes no sense to me. Also plus doesn´t have to take a number as well.... not to forget, that using plus for String based ranges would lead to actually entirely different results than today.

blackdrag:
So my problem with the intention of this pull request to allow Number as stepping is, that it does not make sense according to Definition 2 and according to Definition 3 Number looks too limiting.... totally ignoring the Definition 3 is a breaking change in the semantics of the range.

paplorinc:
I understand. Should I remove the `Number` iteration and keep just the cleanup part?
Or provide a `Number` stepping only to `ObjectRange`, and only in case of `Numeric` ranges and throw in other cases?

src/main/groovy/lang/ObjectRange.java
```diff
@@ -37,7 +37,7 @@
  * Note: This class is similar to {@link IntRange}. If you make any changes to this
  * class, you might consider making parallel changes to {@link IntRange}.
  */
-public class ObjectRange extends AbstractList implements Range {
+public class ObjectRange extends AbstractList<Comparable> implements Range<Comparable> {
```
blackdrag: Range uses <T extends Comparable>, I guess ObjectRange should do the same and then let get return T for example

src/main/groovy/lang/ObjectRange.java
blackdrag: synchronization serves as an inlining barrier, thus I do not agree to it being cheap in all cases. If not needed it should be avoided.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -397,7 +392,10 @@ public boolean contains(Object value) {
 
     @Override
     public void step(int step, Closure closure) {
-        if (step == 0 && compareTo(from, to) == 0) {
+        if (compareEqual(step, 0)) {
+            if (compareNotEqual(from, to)) {
+                throw new GroovyRuntimeException("Infinite loop detected due to step size of 0");
```
blackdrag: why make this an error instead of an no-op

tkruse: That's consistent with similar lines throughout this class

src/main/groovy/lang/ObjectRange.java
```diff
@@ -409,106 +406,86 @@ public void step(int step, Closure closure) {
...
-                    // handle back to beginning due to modulo decrementing
-                    if (compareGreaterThanEqual(peekValue, range.to)) return null;
-                }
-                if (compareGreaterThanEqual(peekValue, range.from)) {
-                    return peekValue;
+        private void fetchNextIfNeeded() {
+            if (!isNextFetched) {
+                isNextFetched = true;
+
+                if (next == null) {
```
blackdrag: using a check with next = null might not be good enough. What happens if null is used as lower bound? I would like to avoid this leading to endless lopps

src/main/groovy/lang/ObjectRange.java
```diff
@@ -338,22 +337,22 @@ public int size() {
 
         // Performance detail:
         // not using get(fromIndex), get(toIndex) in the following to avoid stepping over elements twice
-        final Iterator<Comparable> iter = new StepIterator(this, 1);
+        final Iterator<Comparable> it = new LazySteppingIterator(this, 1);
```
blackdrag: why bloating the pull request by renaming iter to it? We tend to use iter for iterator, since "it" is often used as iteration element in Closures

src/main/groovy/lang/ObjectRange.java
```diff
@@ -409,106 +406,86 @@ public void step(int step, Closure closure) {
...
-                    return peekValue;
+        private void fetchNextIfNeeded() {
+            if (!isNextFetched) {
+                isNextFetched = true;
+
+                if (next == null) {
+                    next = isAscending ? range.getFrom() // make the first fetch lazy too
+                                       : range.getTo();
+                } else {
+                    next = isAscending ? increment(next)
```
blackdrag: I am missing the usage of step here. Shouldn't SteppingIterator skip elements if Math.abs(step) is >1?

paplorinc: You're right, though I meant it as an intermediate commit - it's corrected in the next commit: https://github.com/apache/groovy/pull/360/files#diff-edcd680998449caffc50809917864ceeR494
.
I will gladly fix this commit also, I just would like someone to tell me what I have to do to make it merged, as it's in review for almost 1 year.

paulk-asert:
I'll have time to look at this again in a couple of days. I think the overall goal you are trying to achieve is very worthy. For me we just need to think about how we slice and dice that functionality in the right way to have nice abstractions and a good story around backwards (or breaking) compatibility. I have a little spike that I think would add to the conversation but I'm only half way through and have other things to do over the next couple of days.

src/main/groovy/lang/ObjectRange.java
paplorinc: I notices that they were inconsistent so I went one way. Thanks for noticing, renamed all to `iter`.

src/main/groovy/lang/ObjectRange.java
paplorinc:
```java
        if (smaller == null) {
            throw new IllegalArgumentException("Must specify a non-null value for the 'from' index in a Range");
        }
        if (larger == null) {
            throw new IllegalArgumentException("Must specify a non-null value for the 'to' index in a Range");
        }
```

src/main/groovy/lang/ObjectRange.java
paplorinc: Not sure what you mean...

tkruse: github interprets html, so I guess Jochen wrote ```<Object>``` or ```<T>```

paplorinc:
@blackdrag, @tkruse, @paulk-asert, thanks for your comments, I removed the Number stepping commit and left all the refactorings.
If later someone decides to add the Number stepping, the code is easier to modify.

Let me know if the [last commit (the Number stepping optimization)](https://github.com/apache/groovy/pull/360/commits/fbfb3621c85f988c25691036696541a2d9cfd16b) should be kept or removed.

src/main/groovy/lang/IntRange.java
```diff
@@ -385,26 +385,23 @@ public void step(int step, Closure closure) {
...
+            step *= -1;
+            isAscending = isReverse();
+        } else {
+            isAscending = !isReverse();
         }
-        if (step > 0) {
-            int value = getFrom();
-            while (value <= getTo()) {
+
+        if (isAscending) {
```
paplorinc: its easier to think about ascending and descending iterations than with bigger from values with negative step sizes...

src/main/groovy/lang/IntRange.java
```diff
@@ -385,26 +385,23 @@ public void step(int step, Closure closure) {
...
                 closure.call(value);
-                if (((long) value + step) >= Integer.MAX_VALUE) {
-                    break;
-                }
-                value = value + step;
             }
         } else {
-            int value = getTo();
-            while (value >= getFrom()) {
+            for (long value = to; value >= from; value -= step) {
```
paplorinc: the `long` values prevent overflow

src/main/groovy/lang/ObjectRange.java
blackdrag: That might be the case, but that does  not mean the other cases are right, not does it mean it should be done like that here as well. I think this should be looked at on a case-by-case base, as well as conceptually.

src/main/groovy/lang/ObjectRange.java
blackdrag: ups, did not notice, yes... `public class ObjectRange<T extends Comparable> extends AbstractList<t> implements Range<T> {` was what I was thinking about

src/main/groovy/lang/ObjectRange.java
paplorinc: it was like this [in `IntRange`](https://github.com/paplorinc/incubator-groovy/blob/master/src/main/groovy/lang/IntRange.java#L360) also.
Would you like me to simplify it in both cases to no-ops?

blackdrag: yes

src/main/groovy/lang/ObjectRange.java
paplorinc: How could we take advantage of the `T` in case of an `0.1..0.9`?

src/main/groovy/lang/IntRange.java
```diff
@@ -379,32 +379,25 @@ public boolean containsAll(Collection other) {
     @Override
     public void step(int step, Closure closure) {
         if (step == 0) {
-            if (!getFrom().equals(getTo())) {
-                throw new GroovyRuntimeException("Infinite loop detected due to step size of 0");
-            }
-            return; // from == to and step == 0, nothing to do, so return
+            return;
```
paplorinc: @blackdrag

src/main/groovy/lang/ObjectRange.java
```diff
@@ -376,27 +375,21 @@ public String inspect() {
...
     }
 
     @Override
     public void step(int step, Closure closure) {
-        if (step == 0 && compareTo(from, to) == 0) {
-            return; // from == to and step == 0, nothing to do, so return
-        }
-        final Iterator<Comparable> iter = new StepIterator(this, step);
-        while (iter.hasNext()) {
+        for (Iterator<Comparable> iter = new LazySteppingIterator(this, step); iter.hasNext(); ) {
```
paplorinc: no need for `step == 0` check here as the iterator already does it

src/main/groovy/lang/ObjectRange.java
```diff
@@ -406,106 +399,87 @@ public void step(int step, Closure closure) {
...
-            if (desiredStep == 0 && compareNotEqual(range.getFrom(), range.getTo())) {
-                throw new GroovyRuntimeException("Infinite loop detected due to step size of 0");
-            }
+        private final boolean isAscending;
+
+        private boolean isNextFetched = false;
+        private Comparable next = null;
+
+        LazySteppingIterator(Range<? extends Comparable> range, int desiredStep) {
+            isNextFetched = (desiredStep == 0);
```
paplorinc: @blackdrag, `step == 0` is a no-op now (i.e. `hasNext` starts with `false`)

src/main/groovy/lang/ObjectRange.java
blackdrag: The main change is that you use T instead of Comparable. In case of `0.1..0.9` T is supposed to be BigDecimal and something like `'a'...100` is not legal from the Java side. it will still work, because ScriptBytecodeAdapter does not care about the static type at all. But if somebody uses the ObjectRange from Java, then why not generify it in that manner for those people?

paplorinc: Done, though I couldn't change every `Comparable` to `T`, because of the constructor (changing char to int and float to double).

src/main/groovy/lang/ObjectRange.java
```diff
@@ -390,10 +390,6 @@ public boolean contains(Object value) {
 
     @Override
     public void step(int step, Closure closure) {
-        if (step == 0 && compareTo(from, to) == 0) {
```
paplorinc: the stepping iterator takes care of this

src/test/groovy/RangeTest.groovy
```diff
@@ -262,6 +264,21 @@ class RangeTest extends GroovyTestCase {
...
+
+        ('b'..'b').step 0, { assert it != null } //ObjectRange
+
+        5.step(5, 1) { assert it != null } // DGM.step(), int
+
+        5.0.step(5.0, 1) { assert it != null } // DGM.step(), BigDecimal
+
+        new BigInteger(5).step(new BigInteger(5), 1) { assert it != null }  // DGM.step(), BigInteger
+
+        (1..2).step 0, { assert it != null } //IntRange
```
paplorinc: these two are changed, as requested by @blackdrag

paplorinc:
@blackdrag, @paulk-asert, could we please merge this?

src/main/groovy/lang/IntRange.java
paulk-asert: This seems controversial to me. If I had a for loop:
```
for (int x = 0; x < 5; x += 0) { }
```
 I would expect an infinite loop (except for the case of the upper bound being the same as the initial value which we handle separately anyway) rather than the compiler optimizing it away to a no-op. So the previous early detection and exception makes more sense to me. Groovy does often give shortcuts to handle cases that might be errors in a more elegant way but this seems to be a bigger magical step than normal.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -493,40 +495,52 @@ public void remove() {
...
+        if (value instanceof Number) {
+            return (T) minus((Number) value, step);
+        } else {
+            for (int i = 0; i < step; i++) {
+                final T previous = (T) InvokerHelper.invokeMethod(value, "previous", null);
+                if (!compareLessThan(previous, value)) /* e.g. `previous` of the first element */ {
+                    return null;
+                }
+                value = previous;
             }
```
paulk-asert: So the main point of consideration now is do we want this breaking change? If so, how we might document it? Is it possible (and would it be worthwhile) to provide a compatibility story or some kind of support for people wanting the old behavior?

src/main/groovy/lang/ObjectRange.java
```diff
@@ -261,16 +260,12 @@ public T get(int index) {
...
         if (value instanceof Comparable) {
-            final int result = compareTo(from, (Comparable) value);
-            return result == 0 || result < 0 && compareTo(to, (Comparable) value) >= 0;
+            final int result = compareTo(from, value);
+            return result == 0 || result < 0 && compareGreaterThanEqual(to, value);
         }
         return contains(value);
     }
 
-    protected int compareTo(Comparable first, Comparable second) {
```
paulk-asert: An earlier request was to make compareTo protected to allow subclasses of ObjectRange to change the behavior. By optimizing this we have taken away that capability. But the question is, was the real motivation behind the previous request so that they could optimize the method themselves? If that is the case, we are in a better position.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -220,13 +219,15 @@ public boolean equals(ObjectRange that) {
     }
 
     @Override
-    public Comparable getFrom() {
-        return from;
+    @SuppressWarnings("unchecked")
```
paulk-asert: This looks a little more complex than I was expecting. Does changing constructors and to/from to be of type T simplify? I was planning to look into this myself but haven't had time yet. The need for SuppressWarnings is a smell but I understand it isn't always easy to get rid of.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -518,25 +496,41 @@ private T peek() {
...
-     * Decrements by one
+     * Decrements by step size
      *
      * @param value the value to decrement
-     * @return the decremented value
+     * @param step  the value to decrement by
+     * @return the decremented value or null, if there isn't any
      */
     @SuppressWarnings("unchecked")
-    private T decrement(T value) {
```
paulk-asert: I'd like to see this rebased. That would help me understand whether we have broken binary compatibility.

src/main/groovy/lang/ObjectRange.java
paplorinc: This wasn't used consistently, only in a few places.
Now it's unified.

src/main/groovy/lang/ObjectRange.java
paplorinc: This is an optimization for e.g. double. It's in a separate commit and is documented on the function's javadoc.

src/main/groovy/lang/ObjectRange.java
paplorinc: I just rebased it, it's up to date and mergeable.

src/main/groovy/lang/ObjectRange.java
paulk-asert: So we still need to decide if we want this breaking change and whether javadoc on a private method (which won't be included in the generated javadoc) is sufficient to advertise the change and as stated previously consider whether to have a compatibility story.

paplorinc: Sure, please let me know if I should delete the commit or not.

src/main/groovy/lang/IntRange.java
paplorinc: I don't mind either, but this is what @blackdrag requested.
Please decide and let me know.

src/main/groovy/lang/ObjectRange.java
```diff
@@ -537,8 +539,9 @@ protected Object increment(Object value) {
      * @param value the value to decrement
      * @return the decremented value
      */
-    protected Object decrement(Object value) {
-        return InvokerHelper.invokeMethod(value, "previous", null);
+    @SuppressWarnings("unchecked")
+    private T decrement(T value) {
```
paulk-asert: This represents a binary incompatible change. We try really hard to avoid those - it leads to lots of problems, e.g. classes compiled with one version of Groovy won't work with other versions.

paplorinc: Again, I did this upon request.

blackdrag: @paplorinc you did this on request, yes, but the pullrequest got a bit messy and it is really difficult to see all the implications before the actual change right now. Lot's of these problems stem also from the API here being such a mess.

There is two possibly problematic parts in this. One is the change from protected to private, the other is the change from Object to T. Since T has Comparable as base, we have a type change as well.  

Which actually leads me back to this API being a mess. Range is based on Comparable and this change was done in 2011... IntRange got changed to conform to this, but ObjectRange did not. 

And then let us look for example at get(int). The method is defined through AbstractList for once, Range sets the List type to "T extends Comparable", which means get(int) is supposed to return at least a Comparable. Now ObjectRange implements Range and defines get(int) as returning Object. This works only because Range is an interface. Had it been an abstract class, it would most probably define get(int) as returning Comparable and then ObjectRange would no longer be able to define an Object returning version. Fun fact:
`interface R<T extends Comparable> extends List<T>{}
class O extends AbstractList implements R{
  def get(int  i){null}
}` will in Groovy fail compilation with `The return type of java.lang.Object get(int) in O is incompatible with java.lang.Comparable in java.util.List`. And does rightfully do so. I really wonder why javac is not complaining here

paplorinc: I understand, please let me know what to change (if nothing else, revert problematic commits) for this to be merged.

tkruse:
I believe my own analysis was that this PR is problematic conceptually, so as a vote that would be -1 (though my opinion may not matter).

I believe Jochens detailed analysis to also be a -1. Also Jochen stresses the point that the API here is a mess, and I don't think this PR would improve that situation.

Paul stated that "the overall goal you are trying to achieve is very worthy". So it's controversial.

I did not see a discussion or vote on the user or dev mailing list.

blackdrag:
I have not yet decided actually if I want to give this a definitive -1, if this is 2.5 or even 3.0 only or if we should just merge.

paplorinc:
Let me know which commit is problematic so that I can revert it.

paulk-asert:
@paplorinc I created an alternative spike here: PR #366
Some snippets of the code is borrowed from some of your work but it wasn't possible to mix in any of your commits as a whole. If we end up preferring #366 then I am more than happy if you want to submit any of that under your name. In any case, interested in your thoughts given the amount of time you have put into this issue. Cheers, Paul.

paplorinc:
Last attempt: kept only the comparator and iterator unification.
If that's still controversial, please merge-and-fix or close this PR.
Thanks!

src/main/groovy/lang/IntRange.java
paulk-asert: The issue with the long value being passed to the closure is that it breaks existing code, e.g.:
`(0..3).step(2){ int i -> println i * 2 }
`

blackdrag: if the doCall method is called dynamically the call is still possible, even if the doCall method takes an int and you supply it with a long

blackdrag: And of course the most simple fix is to cast the long to int here, just before calling the closure. since the long is only for overflow, it should be perfectly valid to do so

paulk-asert: Yes, a cast to int should get around the MissingMethodException in my example and should be safe to do.

paplorinc:
Closing this PR, too bad.

PR #364, opened 2016-07-03
Travis CI - run with latest JDK8

jwagenleitner:
See https://github.com/travis-ci/travis-ci/issues/4042

This updates the jdk8 version from the default 1.8.0_31 to 1.8.0_91 and seems to speed up the test with that jdk by several minutes in my testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,16 @@
 
 language: groovy
 
-jdk:
-  - oraclejdk7
-  - oraclejdk8
+sudo: false
+
+matrix:
+  include:
+    - jdk: oraclejdk7
+    - jdk: oraclejdk8
+      addons:
+        apt:
+          packages:
+            - oracle-java8-installer
 
 install: true
 


paulk-asert:
Hi John, nice work. What are the implications when newer versions of Java 8 come out ? Does having that installer specified mean we always get the latest?

jwagenleitner:
Yes I believe that is true that it will always grab the latest.  It will run the apt-get command on each build and since the WebUpd8 repo package is just a script that downloads the tar.gz from Oracle's site new releases will probably show up pretty quick.

Main motivation for looking at this was to try to speed up the build time.  I also looked at splitting the build in two parts so that core tests `./gradlew :test` and subproject tests `./gradlew test -x :test` could run in parallel as two separate jobs.  This worked and did drop the total test time but it may be confusing to those reviewing the build details if there's a failure.

Just some additional info, the reason for using the matrix.include was to get it to only update jdk8 for the jdk8 build.

paulk-asert:
While I generally like to be in control of (or have some stability around) when the jdk might change underneath the build, I guess we already have that on the non-travis CI server for now, so +1 for this PR from me.

PR #365, opened 2016-07-10
GROOVY-5985 - Invalid hash key serialVersionUID

jwagenleitner:
Using a map key with the same name as a read-only property on the Map
class would throw a groovy.lang.ReadOnlyPropertyException.  The fix for
GROOVY-3471 (commit 18f6f8271ed3bfb) moved the map handling
below the handling for read-only properties.  In order to avoid the
issue in GROOVY-3471 we also need to check and make sure the
method is null so we are not in a setX call.


paulk-asert:
+1

PR #366, opened 2016-07-13
GROOVY-7877: The Range abstraction could support numeric ranges where…

paulk-asert:
… the items in the range differ by some step size different to 1

### DIFF
--- a/src/main/groovy/lang/NumberRange.java
+++ b/src/main/groovy/lang/NumberRange.java
@@ -1,0 +1,629 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ */
+package groovy.lang;
+
+import org.codehaus.groovy.runtime.InvokerHelper;
+import org.codehaus.groovy.runtime.IteratorClosureAdapter;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.AbstractList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.NoSuchElementException;
+
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareEqual;
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareGreaterThan;
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareGreaterThanEqual;
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareLessThan;
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareLessThanEqual;
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareNotEqual;
+import static org.codehaus.groovy.runtime.ScriptBytecodeAdapter.compareTo;
+import static org.codehaus.groovy.runtime.dgmimpl.NumberNumberMinus.minus;
+import static org.codehaus.groovy.runtime.dgmimpl.NumberNumberMultiply.multiply;
+import static org.codehaus.groovy.runtime.dgmimpl.NumberNumberPlus.plus;
+
+/**
+ * Represents an immutable list of Numbers from a value to a value with a particular step size.
+ *
+ * In general, it isn't recommended using a NumberRange as a key to a map. The range
+ * 0..3 is deemed to be equal to 0.0..3.0 but they have different hashCode values,
+ * so storing a value using one of these ranges couldn't be retrieved using the other.
+ *
+ * @since 2.5
+ */
+public class NumberRange extends AbstractList<Comparable> implements Range<Comparable> {
+
+    /**
+     * The first value in the range.
+     */
+    private final Comparable from;
+
+    /**
+     * The last value in the range.
+     */
+    private final Comparable to;
+
+    /**
+     * The step size in the range.
+     */
+    private final Number stepSize;
+
+    /**
+     * The cached size, or -1 if not yet computed
+     */
+    private int size = -1;
+
+    /**
+     * The cached hashCode (once calculated)
+     */
+    private Integer hashCodeCache = null;
+
+    /**
+     * <code>true</code> if the range counts backwards from <code>to</code> to <code>from</code>.
+     */
+    private final boolean reverse;
+
+    /**
+     * <code>true</code> if the range includes the upper bound.
+     */
+    private final boolean inclusive;
+
+    /**
+     * Creates an inclusive {@link NumberRange} with step size 1.
+     * Creates a reversed range if <code>from</code> &lt; <code>to</code>.
+     *
+     * @param from the first value in the range
+     * @param to   the last value in the range
+     */
+    public <T extends Number & Comparable, U extends Number & Comparable>
+    NumberRange(T from, U to) {
+        this(from, to, null, true);
+    }
+
+    /**
+     * Creates a new {@link NumberRange} with step size 1.
+     * Creates a reversed range if <code>from</code> &lt; <code>to</code>.
+     *
+     * @param from start of the range
+     * @param to   end of the range
+     * @param inclusive whether the range is inclusive
+     */
+    public <T extends Number & Comparable, U extends Number & Comparable>
+    NumberRange(T from, U to, boolean inclusive) {
+        this(from, to, null, inclusive);
+    }
+
+    /**
+     * Creates an inclusive {@link NumberRange}.
+     * Creates a reversed range if <code>from</code> &lt; <code>to</code>.
+     *
+     * @param from start of the range
+     * @param to   end of the range
+     * @param stepSize the gap between discrete elements in the range
+     */
+    public <T extends Number & Comparable, U extends Number & Comparable, V extends
+            Number & Comparable<? super Number>>
+    NumberRange(T from, U to, V stepSize) {
+        this(from, to, stepSize, true);
+    }
+
+    /**
+     * Creates a {@link NumberRange}.
+     * Creates a reversed range if <code>from</code> &lt; <code>to</code>.
+     *
+     * @param from start of the range
+     * @param to   end of the range
+     * @param stepSize the gap between discrete elements in the range
+     * @param inclusive whether the range is inclusive
+     */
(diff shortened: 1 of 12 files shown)


paulk-asert:
A spike showing an alternative to PR#360. Basically it takes some of the optimizations desired for ObjectRange from that PR but only implements them in a new class NumberRange. It does also change the default class for creating ranges involving two (non-Integer) Numbers to use NumberRange instead of ObjectRange.

The main thing which is supported is fast calculation of range items rather than relying on next() calls and ranges with a non-integral step size like this:
```
def r = (0.1..0.9).by(0.1)
assert r == [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
```
It is a breaking change in that some weird edge cases involving metaprogramming on Numbers will behave differently. There is a slightly better backwards compatibility story than PR#360 however since users are still free to create ObjectRange's manually if they need the old behavior.

I call it a spike because I still have some questions on the current design. For instance, with ObjectRange we have the following behavior:
```
def or = 'a'..<'f'
assert or.step(2) == ['a', 'c', 'e']
assert or.step(-2) == ['e', 'c', 'a']
```
But currently inclusivity is remembered and applied to the 'to' end of the range when going in different directions, e.g.
```
def nr0 = (1..<6).by(2)
assert nr0.step(1) == [1, 3, 5]
assert nr0.step(-1) == [6, 4, 2]

def nr1 = (1..<6).by(-1)
assert nr1.step(2) == [6, 4, 2]
assert nr1.step(-2) == [1, 3, 5]

def nr2 = (1..<6).by(1)
assert nr2.step(1) == [1, 2, 3, 4, 5]
assert nr2.step(-1) == [6, 5, 4, 3, 2]
```
This may seem unintuitive at first but is kind of useful once understood. It is also somewhat essential to creating efficient ranges that can be traversed in either direction. But it's not the only possible design. We could brute force calculate the final to value (going in the forward direction) and use that as the from value when iterating in reverse.

Another question as brought up by @tkruse in an earlier PR was whether for inaccurate number types like float whether it is better to continually add the step values when iterating or multiply by an index - giving at most two rounding errors rather than a compounding scenario. I also haven't tried to fast calculate all possible cases - it might be possible to cover additional cases. But that can be done I believe as a post optimization. It might also be possible to precalculate the number of items within the range during iteration which might save the cost of comparisons for each iteration.

Also, there are some other optimizations possible. The expression `(0.1..0.9).by(0.1)` creates two NumberRanges. The first with a step size of one, the second with the desired 0.1 step size. This is just a convenience to piggy back on existing syntax features. It would be possible to enhance the compiler to detect cases like that and instead just create a single NumberRange. Again, I consider this an optimization for later.

src/main/groovy/lang/NumberRange.java
```diff
@@ -0,0 +1,536 @@
...
+        return (that instanceof NumberRange) ? equals((NumberRange) that) : super.equals(that);
+    }
+
+    /**
+     * Compares an {@link NumberRange} to another {@link NumberRange}.
+     *
+     * @param that the object to check equality with
+     * @return <code>true</code> if the ranges are equal
+     */
+    public boolean equals(NumberRange that) {
```
jwagenleitner: I wonder if `hashCode()` should be implemented, though not sure how it would be implemented.  Just mentioning it because of GROOVY-7877.  It may be in this case iterating through the entire range (AbstractList#hashCode) is the best way.

jwagenleitner: Maybe too simplistic but wonder if following would work:

```
    @Override
    public int hashCode() {
        int result = 17;
        result = result * 31 + (reverse ? 1 : 0);
        result = result * 31 + from.hashCode();
        result = result * 31 + to.hashCode();
        result = result * 31 + stepSize.hashCode();
        return result;
    }
```
That would mean the following would have different hash codes, but since inclusive isn't used in the equals they aren't equal.
```
Range n1 = new NumberRange(1L, 5L)
Range n2 = new NumberRange(1L, 6L, false)
```

src/main/groovy/lang/NumberRange.java
```diff
@@ -0,0 +1,536 @@
...
+
+    private static boolean areReversed(Number from, Number to) {
+        try {
+            return compareGreaterThan(from, to);
+        } catch (ClassCastException cce) {
+            throw new IllegalArgumentException("Unable to create range due to incompatible types: " + from.getClass().getSimpleName() + ".." + to.getClass().getSimpleName() + " (possible missing brackets around range?)", cce);
+        }
+    }
+
+    public boolean equals(Object that) {
```
paulk-asert: Added, thanks.

src/main/groovy/lang/NumberRange.java
```diff
@@ -0,0 +1,536 @@
...
+                throw new IndexOutOfBoundsException("Index: " + i + " is too big for range: " + this);
+            }
+            value = iter.next();
+        }
+        final Comparable toValue = value;
+
+        return new NumberRange(comparableNumber(fromValue), comparableNumber(toValue), comparableNumber(stepSize), true);
+    }
+
+    public String toString() {
```
paulk-asert: Added, thanks.

src/main/groovy/lang/NumberRange.java
paulk-asert: I think it should be possible to have a streamlined calculation for the fast-path cases in size(). I'll look into that.

paulk-asert: OK, I added a hashCode along the lines you suggested and tweaked equals. Comments welcome.

paulk-asert: I went a little further (and possibly into controversial territory) by providing fastEquals, fastHashCode and canonicalHashCode methods.

jwagenleitner: I see what you mean about the `==` operator making it tricky to respect the Java equals/hashCode contract.  I can't think of another use for the hashCode outside of hashed-based Maps and Collections, so having the alternate implementations may not prove useful.  So it looks like the default behavior of `AbstractList` may be best and this type of range just isn't a good candidate for a key in a HashMap or an item in a HashSet.

I know it bad form, and just thinking out loud, but wonder if this is a situation where returning a constant value from hashCode would be appropriate or when the size of the range exceeds a certain max value.

src/main/groovy/lang/NumberRange.java
```diff
@@ -0,0 +1,603 @@
...
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public Iterator<Comparable> iterator() {
+        // non thread-safe iterator
+        final Iterator<Comparable> innerIterator = new StepIterator(this, stepSize);
+        return new Iterator<Comparable>() {
+            @Override
+            public synchronized boolean hasNext() {
```
jwagenleitner: Just curious, why `synchronized` for the iterator?  I see the same on the `ObjectRange` iterator but can't understand why it would be needed.

paulk-asert: It isn't strictly necessary. It is providing a thread-safe iteration mechanism. This is something that was introduced recently in ObjectRange and I just carried it over.

src/main/groovy/lang/NumberRange.java
paulk-asert: I would favor the current implementation for hashCode compared with returning a constant value. I think the common case would be to use 1..3 and [1, 2, 3] interchangeably which is what the current implementation allows (and honors the equals/hashCode contract). It's only when you start wanting to treat integer 1 and some other Number 1 as the same where things are a little strange wrt that contract.

src/main/groovy/lang/NumberRange.java
jwagenleitner: I figured it was carried over from ObjectRange.  I just couldn't figure out why you'd want 2 or more threads to share an iterator.  Figured the typical use would be in a foreach loop and each thread would get their own instance of an iterator.   Otherwise, if sharing an instance a call to `hasNext()` followed by `next()` seems like it could return `null` for one of the threads after another thread got the last Number.

src/main/groovy/lang/NumberRange.java
paulk-asert: I also agree that the usefulness of the alternate implementations is worthy of discussion. Certainly those implementations will be ignored by the existing map classes and only become relevant for people wanting to write their own map-friendly map classes or for sorting/comparators. Also, whether they belong within NumberRange or in some Range utility or ListEquivalence class is worthy of consideration. Having said that, the implementations might at least better clarify the current behavior to some degree.

src/main/groovy/lang/NumberRange.java
paulk-asert: We could possibly just have a `safeIterator` method which returned the thread-safe version and not bother normally.

src/main/groovy/lang/NumberRange.java
```diff
@@ -0,0 +1,637 @@
...
+    /**
+     * iterates over all values and returns true if one value matches.
+     * Also see containsWithinBounds.
+     */
+    @Override
+    public boolean contains(Object value) {
+        final Iterator it = new StepIterator(this, stepSize);
+        if (value == null) {
+            return false;
+        }
```
jwagenleitner: Minor thing, but should the null check come first.

src/main/groovy/lang/NumberRange.java
```diff
@@ -0,0 +1,637 @@
...
+            return (next != null) && (isAscending
+                    ? (range.inclusive ? compareLessThanEqual(next, range.getTo()) : compareLessThan(next, range.getTo()))
+                    : (range.inclusive ? compareGreaterThanEqual(next, range.getFrom()) : compareGreaterThan(next, range.getFrom())));
+        }
+
+        @Override
+        public Comparable next() {
+            if (!hasNext()) {
+                return null;
+            }
```
jwagenleitner: Should it be a `NoSuchElementException` instead of returning `null`?

src/main/groovy/lang/NumberRange.java
jwagenleitner: To me it seems like a true thread-safe iterator would need to lock in `hasNext()` and unlock when it returns false (or throws) or after a subsequent call by the same lock-holding thread to `next()`.  Though I haven't test it, it appears with these synchronized methods that a thread could have `hasNext()` return `true` and a subsequent call to `next()` return `null` (though maybe should be `NoSuchElementException`).

But I still do not understand the benefit of having a thread safe version of the iterator that multiple threads can share and wonder if maybe those synchronized methods were put in by mistake (especially given the nearby comments about it not being thread safe).  Also, I don't see similar synchronization in the `IntRange` class.

jwagenleitner: >and wonder if maybe those synchronized methods were put in by mistake (especially given the nearby comments about it not being thread safe).

Meant to say put in by mistake on the `ObjectRange` class originally.

paulk-asert:
I removed the slightly flawed attempts at synchronization and adopted a more Java behavior when calling next by throwing a NoSuchElementException once a NumberRange iterator is exhausted.

jwagenleitner:
Looks really good, nice work.

Just my opinion, but I would not introduce the additional public API alternatives for  equals/hashCode until a need presents itself.  I think the doc in there stating instances of the class are not good candidates for hashed keys is good and sufficient.

jwagenleitner:
Forgot to ask if it would be worthwhile to add `@since` tags on the new IntRange method and on the NumberRange class.

paulk-asert:
Yes, those additional methods do kind of represent possible premature feature introduction. I'll remove them (or at least comment them out) and I'll add the @since tags too.

paulk-asert:
Proposed PR merged.

PR #367, opened 2016-07-19
GROOVY-7600: @Immutable support for Optional?

paulk-asert:


src/test/org/codehaus/groovy/transform/ImmutableTransformTest.groovy
```diff
@@ -18,12 +18,40 @@
...
+@RunWith(JUnit4)
 class ImmutableTransformTest extends GroovyShellTestCase {
 
+    @Rule public TestName nameRule = new TestName()
+
+    @Before
+    void setUp() {
+        super.setUp()
+        // check java version requirements
+        def m = System.getProperty("java.version") =~ /\d\.\d+/
```
jwagenleitner: This may fail on jdk9 with the new versioning scheme.  Maybe the following would work since (at the moment) this returns 9 on jdk9 ea's.

```
new BigDecimal(System.getProperty('java.specification.version')) >= 1.8
```

paulk-asert:
PR merged

PR #369, opened 2016-07-20
test cleanup: remove println calls

jwagenleitner:
In many cases asserts were already in place to ensure the expected
output.  In places were asserts did not already exist, the println
calls were replaced by asserts.


paulk-asert:
+1

PascalSchumacher:
nice work :+1:

PR #370, opened 2016-07-21
Add access to declared variables in the VariableScope

natgabb:
For easier parsing of symbols.

### DIFF
--- a/src/main/org/codehaus/groovy/ast/VariableScope.java
+++ b/src/main/org/codehaus/groovy/ast/VariableScope.java
@@ -178,4 +178,26 @@ public class VariableScope  {
     public Iterator<Variable> getReferencedClassVariablesIterator() {
         return getReferencedClassVariables().values().iterator();
     }
+
+    /**
+     * Gets a map containing the variables declared in this scope.
+     * This map cannot be modified.
+     * @return a map containing the declared variable references
+     */
+    public Map<String, Variable> getDeclaredVariables() {
+        if (declaredVariables == Collections.EMPTY_MAP) {
+            return declaredVariables;
+        } else {
+            return Collections.unmodifiableMap(declaredVariables);
+        }
+    }
+
+    /**
+     * Gets an iterator for the declared class variables. The remove
+     * operation is not supported.
+     * @return an iterator for the declared variables
+     */
+    public Iterator<Variable> getDeclaredVariablesIterator() {
+        return getDeclaredVariables().values().iterator();
+    }
 }


src/main/org/codehaus/groovy/ast/VariableScope.java
```diff
@@ -178,4 +178,26 @@ public Object removeReferencedClassVariable(String name) {
...
     }
-}
\ No newline at end of file
+
+    /**
+     * Gets a map containing the variables declared in this scope.
+     * This map cannot be modified.
+     * @return a map containing the declared variable references
+     */
+    public Map<String, Variable> getDeclaredVariables() {
```
baybatu: In javadoc, it says "This map cannot be modified" but this method can return two type of Map implementations: HashMap (see putDeclaredVariable method in the same file) and UnmodifiableMap stated in 'else' block.

blackdrag:
can you describe a little how you would use that instead of getting the variable directly? I am asking because I am not aware of such a use case outside of the core compiler and interested in what you want to do.

src/main/org/codehaus/groovy/ast/VariableScope.java
natgabb: Good point! I fixed it. This is also the case in `getReferencedClassVariables', would you like me to change it?

natgabb:
@blackdrag I want to be able to get a hold of all symbols within a `CompilationUnit`, including locally declared variables. I don't think there is currently a way of doing this? I am building a groovy [language server](https://github.com/Microsoft/language-server-protocol/), which implements features like search all workspace symbols, auto complete, goto definition, find all references, etc.

src/main/org/codehaus/groovy/ast/VariableScope.java
natgabb: Actually `Collections.EMPTY_MAP` is an immutable map. See: https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#EMPTY_MAP
This means the map will be immutable in both cases since `new HashMap<String, Integer>()` is not equal to the `Collections.EMPTY_MAP`. I've reset the last commit.

baybatu: Yes, you're right :) 👍

JiahuiJiang:
@baybatu is this PR good to go? :)

natgabb:
@paulk-asert could you please take a look? :)

blackdrag:
@natgabb 
I did not know you want this for something IDE-alike. You have to be aware, that there are many VariableScope instances, even inside a method. In total I think VariableScopeVisitor will do too much for you - and without VariableScoprVisitor, there will be no VariableScope to depend on. So if you find it does too much, we should further discuss what you need exactly and how that can be done. Which then should be a discussion on the dev-list ;) But till then I ok this PR

+1

src/main/org/codehaus/groovy/ast/VariableScope.java
blackdrag: please do not change it for getReferencedClassVariables, we have code directly mutating that map

paulk-asert:
PR merged, thanks! Just as a side comment, it is usually a good idea to create a Jira issue to track such changes (as well as the PR). I created one this time here: https://issues.apache.org/jira/browse/GROOVY-7894
Thanks again.

natgabb:
@blackdrag thanks a lot! :) Where can I find the dev-list?
@paulk-asert thanks :) I'll make sure to do that next time

natgabb:
@blackdrag @paulk-asert Are there dev/snapshot distributions of groovy available somewhere?

blackdrag:
you can find all lists and  to subscribe to them on http://groovy-lang.org/mailing-lists.html Using nabble is also possible, but has sometimes problems. It is better to do it directly

paulk-asert:
snapshot builds are on the teamcity CI server:
http://ci.groovy-lang.org/project.html?projectId=Groovy
just click on any of the builds with an "Artifacts" link.

natgabb:
@blackdrag Hi! After some more testing I realized that this doesn't really expose method declared variables. The map only contains method parameters. Any idea why this is the case? How can I access a method's Variable nodes? I create a Jira issue, which has more detail: https://issues.apache.org/jira/browse/GROOVY-7903

PR #373, opened 2016-07-26
Add tests for valid and invalid integer literals (shouldCompile/shouldNotCompile)

msgilligan:
Add tests for integer literals that `shouldCompile` and `shouldNotCompile`.

Some of these tests currently fail because of [GROOVY-7385](https://issues.apache.org/jira/browse/GROOVY-7385).

I'm not sure what the right way is to submit failing tests, so please let me know if I should change things up. This PR was submitted earlier, but was lost when GROOVY-7385 was closed -- before it was reopened.

### DIFF
--- a/src/spec/test/SyntaxTest.groovy
+++ b/src/spec/test/SyntaxTest.groovy
@@ -153,6 +153,33 @@ class SyntaxTest extends CompilableTestSupport {
         '''
     }
 
+    void testValidIntegerLiterals() {
+        shouldCompile '''
+             def a = 2147483647I
+             def b = -2147483648I
+             def c = -2147483647I
+             def d = 9223372036854775807L
+             def e = -9223372036854775808L
+             def f = -9223372036854775807L
+         '''
+    }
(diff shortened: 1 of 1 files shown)


paulk-asert:
For failing tests which are supposed to pass, the general pattern is to use `GroovyTestCase#notYetImplemented` method or the `@NotYetImplemented` annotation and leave `shouldCompile` in the test itself. This indicates we actually want it to pass but haven't implemented it yet. Also, it would be great to update the Numbers section of `core-syntax.adoc` and add in references to the tags you've added and potentially one or more lines from the passing test too.

msgilligan:
Actually, the unimplemented cases are cases that I want to fail, but don't. But I get your point.

jwagenleitner:
These tests pass in master, looks like this was fixed in commit a258a3d026ebefc411797b2f69915fe1860262db.  So I went ahead and merged as is so we have tests to cover the invalid ranges.  Thanks.

PR #375, opened 2016-07-29
GROOVY-7893: LexerFrame could be made more general purpose and made a…

paulk-asert:
…ccessible from GroovyConsole


paulk-asert:
I'd be interested in opinions as to whether the LexerFrame class should be moved to the groovy-console module?

glaforge:
It's only used in the console anyway, I believe, so it sounds a good idea to move it in the console module to me.

paulk-asert:
It has only been standalone so far - in fact I had forgotten it existed until I was reminded again just recently. This PR hooks it into the console.

PR #376, opened 2016-07-30
GROOVY-7774: Collection addAll fails CompileStatic type checking when…

paulk-asert:
… adding a collection of subtypes

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -5176,7 +5176,9 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
             G key = outer.getKey();
             List<Map.Entry<K, V>> entries = outer.getValue();
             Map<K, V> target = createSimilarMap(self);
-            putAll(target, entries);
+            for (Map.Entry<K, V> entry : entries) {
+                target.put(entry.getKey(), entry.getValue());
+            }
             answer.put(key, target);
         }
         return answer;
@@ -8903,8 +8905,8 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
      * @return the same map, after the items have been added to it.
      * @since 1.6.1
      */
-    public static <K, V> Map<K, V> putAll(Map<K, V> self, Collection<Map.Entry<K, V>> entries) {
-        for (Map.Entry<K, V> entry : entries) {
+    public static <K, V> Map<K, V> putAll(Map<K, V> self, Collection<Map.Entry<? extends K, ? extends V>> entries) {
+        for (Map.Entry<? extends K, ? extends V> entry : entries) {
             self.put(entry.getKey(), entry.getValue());
         }
         return self;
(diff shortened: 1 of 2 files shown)


jwagenleitner:
+1

paulk-asert:
I added in some Map methods which suffered the same problem

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -5169,12 +5170,13 @@ public static Map groupBy(Object[] self, List<Closure> closures) {
      * @return a new Map grouped by keys
      * @since 1.0
      */
-    public static <G, K, V> Map<G, Map<K, V>> groupBy(Map<K, V> self, @ClosureParams(MapEntryOrKeyValue.class) Closure<G> closure) {
-        final Map<G, List<Map.Entry<K, V>>> initial = groupEntriesBy(self, closure);
-        final Map<G, Map<K, V>> answer = new LinkedHashMap<G, Map<K, V>>();
-        for (Map.Entry<G, List<Map.Entry<K, V>>> outer : initial.entrySet()) {
+    public static <G, K, V> Map<G, Map<? extends  K, ? extends V>> groupBy(
+            Map<K, V> self, @ClosureParams(MapEntryOrKeyValue.class) Closure<G> closure) {
```
jwagenleitner: Should the wildcards be on the self parameter instead?  Just not used to seeing them on return types, but nested wildcards often confuse me :smile:

Would the closure parameter benefit from having `? extends G`?

jwagenleitner: Nevermind my comments on wildcards on self param, the more I think about it that doesn't make sense.

However, wont changing the return type to have wildcards extends mean that nothing can be put in the map that is mapped to the groupBy key which some may rely on being able to do?

paulk-asert: It isn't exactly what I would like but groupBy calls putAll and the other changes are to make Java happy. We might have to do some casting and suppress some warnings instead.

blackdrag: we should use the following guideline: no wildcards on self or returns, only on parameters we use as input. In mutating calls, even the (in/)out parameters should not have wildcards. Because wildcards make it very very difficult to use the container for any adding.

paulk-asert:
I removed the groupBy changes and inlined a few lines instead.

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -5176,7 +5176,9 @@ public static Map groupBy(Object[] self, List<Closure> closures) {
             G key = outer.getKey();
             List<Map.Entry<K, V>> entries = outer.getValue();
             Map<K, V> target = createSimilarMap(self);
-            putAll(target, entries);
```
blackdrag: you removed the putAll here because of generics?

paulk-asert: Yes, but feel free to play around with other options if you have a strong preference.

blackdrag: my suggestion would be to change putAll to `public static <K, V> Map<K, V> putAll(Map<K, V> self, Collection<? extends Map.Entry<? extends K, ? extends V>> entries) `
which more or less adds `? extends` before the Map.Entry part and to change back to use the putAll instead of the loop

paulk-asert: Nice suggestion. Done.

PR #378, opened 2016-08-05
GROOVY-7902: collate() infinite loop with zeroed step

fmamud:
Fixing infinite loop in `Iterable.collate` method when `step` parameter is equals to zero.

Example:
```groovy
[ 1, 2, 3 ].collate( 2, 0 )
```
The result is a infinite loop because the step parameter is used directly in the for increment: `pos += step` 

In the fixing  I assumed that the `step` parameter should be equals to `size` parameter, looks like in the `collate(self, size, size, keepRemainder)` overloaded method, with absence of `step` and I considered backward compatibility with negative steps.

Thks.
Felipe Mamud

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -3055,6 +3055,9 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
         if (size <= 0 || selfList.isEmpty()) {
             answer.add(selfList);
         } else {
+            if (step == 0) {
+                step = size;
+            }
             for (int pos = 0; pos < selfList.size() && pos > -1; pos += step) {
                 if (!keepRemainder && pos > selfList.size() - size) {
                     break ;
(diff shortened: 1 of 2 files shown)


paulk-asert:
I like your idea of avoiding the infinite loop but I am not sure of your proposed fix. I think we should either provide an iterator-backed implementation and "leave" the infinite loop which would allow:
`assert [ 1, 2, 3 ].collate( 2, 0 ).take(2) == [[1, 2], [1, 2]]`
or alternatively throw an `IllegalArgumentException` (or an extra flag to toggle the behaviors).

fmamud:
@paulk-asert I totally agree with you. I think we could create a `collateStream` new method that provides an iterator-backed implementation and I think we shouldn't mistake with `collate` method (for many reasons, p.e. backward compatibility). I created the implementation that uses `IllegalStateExeception` when occurs zeroed step. What do you think? 3ed1c5c

paulk-asert:
Looks good. I am unsure what we'd call an iterator-backed implementation, but it would be a method returning an iterator, not the current methods which do an eager calculation of the result, so what you have makes sense for the current ones.

paulk-asert:
PR merged, thanks!

fmamud:
@paulk-asert PR merged? I think this PR was closed. It's normal and this PR will merge afterwards
? 😃 

thks.

PR #381, opened 2016-08-10
GROOVY-7014: A Java compile time error results in a runtime exception…

paulk-asert:
… in Groovy

### DIFF
--- a/src/main/org/codehaus/groovy/ast/tools/ClassNodeUtils.java
+++ b/src/main/org/codehaus/groovy/ast/tools/ClassNodeUtils.java
@@ -23,11 +23,19 @@ package org.codehaus.groovy.ast.tools;
 import org.codehaus.groovy.ast.ClassHelper;
 import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.MethodNode;
+import org.codehaus.groovy.ast.Parameter;
+import org.codehaus.groovy.ast.PropertyNode;
+import org.codehaus.groovy.ast.expr.Expression;
+import org.codehaus.groovy.ast.expr.MapExpression;
+import org.codehaus.groovy.ast.expr.SpreadExpression;
+import org.codehaus.groovy.ast.expr.TupleExpression;
 
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
+import static org.codehaus.groovy.ast.ClassHelper.boolean_TYPE;
 
 public class ClassNodeUtils {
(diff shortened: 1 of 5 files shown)


paulk-asert:
Dealing with list spreading is tricky. This PR makes an improvement but is still not a complete solution, so the API has a boolean to try the partial fuzzy matching. Rather than exposing another API call within ClassNode which leaks out the fuzzy boolean, I'll likely move that logic into a util class.

src/main/org/codehaus/groovy/control/StaticImportVisitor.java
```diff
@@ -317,6 +337,10 @@ protected Expression transformMethodCallExpression(MethodCallExpression mce) {
         return result;
     }
 
+    private boolean isInnerClass(ClassNode currentClass) {
```
jwagenleitner: There was an identical method added to `StaticVerifier`, how about replacing both with a static method in `ClassNodeUtils`?

src/main/org/codehaus/groovy/ast/tools/ClassNodeUtils.java
```diff
@@ -65,4 +73,122 @@ public static void addDeclaredMethodMapsFromSuperInterfaces(ClassNode cn, Map<St
...
+     * Returns the property name, e.g. age, given an accessor name, e.g. getAge.
+     * Returns the original if a valid prefix cannot be removed.
+     *
+     * @param accessorName the accessor name of interest, e.g. getAge
+     * @return the property name, e.g. age, or original if not a valid property accessor name
+     */
+    public static String getPropNameForAccessor(String accessorName) {
+        int prefixLength = accessorName.startsWith("is") ? 2 : 3;
+        if (accessorName.length() < prefixLength + 1) return accessorName;
+        if (!isValidAccessorName(accessorName)) return accessorName;
```
jwagenleitner: Would it make sense to move the length check to the `isValidAccessorName` method below?  That would ensure `isValidAccessorName` would not return true for methods named `get()`/`is()`/`set()`.

jwagenleitner:
+1 (with just a few minor comments)

src/main/org/codehaus/groovy/ast/tools/ClassNodeUtils.java
paulk-asert: I refactored to support this. It means we calculate the prefix length an extra time in one case but the code reads better which I think is more important.

src/main/org/codehaus/groovy/control/StaticImportVisitor.java
paulk-asert: Done

PR #382, opened 2016-08-10
GROOVY-7880: Diamond Operator for own class causes NullPointerExcepti…

paulk-asert:
…on if Static Compilation is enabled


jwagenleitner:
+1

PR #385, opened 2016-08-17
GROOVY-7907: Cannot assign value of type java.lang.Object with vararg…

paulk-asert:
…s, perameterized method and @CompileStatic


jwagenleitner:
+1

PR #388, opened 2016-08-19
Calculating hashcode of IntRange by pairing function

upadhyayap:
https://issues.apache.org/jira/browse/GROOVY-7834?jql=project%20%3D%20GROOVY%20AND%20id%20%3D%20GROOVY-7834

### DIFF
--- a/src/main/groovy/lang/IntRange.java
+++ b/src/main/groovy/lang/IntRange.java
@@ -422,7 +422,7 @@ public class IntRange extends AbstractList<Integer> implements Range<Integer> {
         final int from = this.getFrom();
         final int to = this.getTo();
 
-        hashCode = new Integer(((from+to+1)*(from+to))/2).intValue()+to;
+        hashCode = ((from+to+1)*(from+to))/2+to;
         return hashCode;
     }
 }
--- a/src/main/groovy/lang/IntRange.java
+++ b/src/main/groovy/lang/IntRange.java
@@ -415,4 +415,14 @@ public class IntRange extends AbstractList<Integer> implements Range<Integer> {
         step(step, adapter);
         return adapter.asList();
     }
+
+    @Override
+    public int hashCode(){
+        int hashCode;
+        final int from = this.getFrom();
+        final int to = this.getTo();
+
+        hashCode = new Integer(((from+to+1)*(from+to))/2).intValue()+to;
+        return hashCode;
+    }
 }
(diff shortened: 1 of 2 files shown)


upadhyayap:
Now calculating hash code of IntRange by taking only upper and lower bounds into consideration. This means that two IntRanges with same lower and upper bounds have same hash code irrespective of the step. will that be a problem? should i also consider the step during calculation?

src/test/groovy/lang/IntRangeTest.groovy
```diff
@@ -138,4 +138,15 @@ class IntRangeTest extends GroovyTestCase {
         assert bs[-1..<-7].toString() == '{0, 5}'
         assert bs[20..<-8].toString() == '{2, 3}'
     }
+
+    void testHashCode(){
```
upadhyayap: best place to test hashcode is HashMap. Let know if need to check this up for more range of values.

upadhyayap:
This is way faster than iterating through whole collection.

src/main/groovy/lang/IntRange.java
```diff
@@ -415,4 +415,14 @@ public void step(int step, Closure closure) {
...
         return adapter.asList();
     }
+
+    @Override
+    public int hashCode(){
+        int hashCode;
+        final int from = this.getFrom();
+        final int to = this.getTo();
+
+        hashCode = new Integer(((from+to+1)*(from+to))/2).intValue()+to;
```
jwagenleitner: No need to create a new `Integer` here.

upadhyayap: Correct because hashCode is already an int.

upadhyayap: Changed and committed.

jwagenleitner:
I think this would work ok.  Since step is not used in `equals(IntRange)` I don't think it needs to be considered.  Not much harm in returning the same hashCode for non-equal objects, as long as the same hashCode is returned when two IntRange's are `equal(IntRange)` it should be ok.

paulk-asert:
Agreed that step isn't a problem. I would have expected that one of reverse or inclusive would have impacted something but looking into it, I don't think they do, so I think the change is all good. I'll have to think a bit more about test coverage but +1 from me in any case. If someone gets around to committing before I get some more time to think about what else might be required, I can always add more after the fact.

jwagenleitner:
PR merged, thanks!

PR #390, opened 2016-08-22
GROOVY-7888: Type checker infers wrong type on compound assignment (e…

paulk-asert:
….g. += for collection) for property


jwagenleitner:
+1

PR #392, opened 2016-08-23
GROOVY-7912: MissingPropertyException when referencing a static impor…

paulk-asert:
…t in a closure's optional parameters


jwagenleitner:
+1

PR #395, opened 2016-08-25
Document error fix

binwang-developer:
Fixed 2 errors of code in this page: [Design patterns in Groovy](http://groovy-lang.org/design-patterns.html)

### DIFF
--- a/src/spec/test/DesignPatternsTest.groovy
+++ b/src/spec/test/DesignPatternsTest.groovy
@@ -24,7 +24,7 @@ class DesignPatternsTest extends CompilableTestSupport {
         shouldCompile '''
             // tag::abstract_factory_example1[]
             class TwoupMessages {
-                def welcome = 'Welcome to the twoup game, you start with $1000'
+                def welcome = 'Welcome to the twoup game, you start with \$1000'
                 def done = 'Sorry, you have no money left, goodbye'
             }
 
--- a/src/spec/test/DesignPatternsTest.groovy
+++ b/src/spec/test/DesignPatternsTest.groovy
@@ -204,7 +204,7 @@ class DesignPatternsTest extends CompilableTestSupport {
             // tag::adapter_inheritance_usage[]
             def hole = new RoundHole(radius: 4.0)
             (4..7).each { w ->
-                def peg = new SquarePegAdapter(peg: new SquarePeg(width: w))
+                def peg = new SquarePegAdapter(width: w)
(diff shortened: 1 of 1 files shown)


src/spec/test/DesignPatternsTest.groovy
```diff
@@ -24,7 +24,7 @@ class DesignPatternsTest extends CompilableTestSupport {
         shouldCompile '''
             // tag::abstract_factory_example1[]
             class TwoupMessages {
-                def welcome = 'Welcome to the twoup game, you start with $1000'
+                def welcome = 'Welcome to the twoup game, you start with \$1000'
```
paulk-asert: I don't believe the slash is needed.

src/spec/test/DesignPatternsTest.groovy
```diff
@@ -204,7 +204,7 @@ class DesignPatternsTest extends CompilableTestSupport {
             // tag::adapter_inheritance_usage[]
             def hole = new RoundHole(radius: 4.0)
             (4..7).each { w ->
-                def peg = new SquarePegAdapter(peg: new SquarePeg(width: w))
+                def peg = new SquarePegAdapter(width: w)
```
paulk-asert: Nice catch. This is the  weakness of this example. It uses println statements instead of asserts so we just used shouldCompile rather than assertScript, otherwise we would have caught this.

src/spec/test/DesignPatternsTest.groovy
binwang-developer: I actually tested the cede as it was as and it threw out some exception: 
```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
E:\Learn\groovy\DesignPatterns\AbstractFactoryPattern\test.groovy: 2: illegal string body character after dollar sign;
   solution: either escape a literal dollar sign "\$5" or bracket the value expression "${5}" @ line 2, column 64.
    two-up game, you start with $1000"
                                 ^

1 error
```

The version of groovy that I am using is 2.4.7

paulk-asert: It looks like you have changed the single quoted string into a double quoted GString. The example uses single quotes but you would need the slash if you used double quotes.

binwang-developer: Well, you got me. Sorry about that~

binwang-developer:
One of the commits is unnecessary. So I am gonna close this PR and open a new one.

PR #397, opened 2016-08-26
GROOVY-7916: MissingPropertyException when accessing a static inner c…

paulk-asert:
…lass member from a sub class

### DIFF
--- a/src/main/org/codehaus/groovy/control/ResolveVisitor.java
+++ b/src/main/org/codehaus/groovy/control/ResolveVisitor.java
@@ -837,19 +837,37 @@ public class ResolveVisitor extends ClassCodeExpressionTransformer {
             }
         }
         if (objectExpression instanceof ClassExpression && pe.getPropertyAsString() != null) {
-            // possibly an inner class
+            // possibly an inner class (or inherited inner class)
             ClassExpression ce = (ClassExpression) objectExpression;
-            ClassNode type = new ConstructedNestedClass(ce.getType(), pe.getPropertyAsString());
-            if (resolve(type, false, false, false)) {
-                Expression ret = new ClassExpression(type);
-                ret.setSourcePosition(ce);
-                return ret;
+            ClassNode classNode = ce.getType();
+            while (classNode != null) {
+                ClassNode type = new ConstructedNestedClass(classNode, pe.getPropertyAsString());
+                if (resolve(type, false, false, false)) {
+                    if (classNode == ce.getType() || isVisibleNestedClass(type, ce.getType())) {
+                        Expression ret = new ClassExpression(type);
+                        ret.setSourcePosition(ce);
+                        return ret;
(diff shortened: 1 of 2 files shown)


src/main/org/codehaus/groovy/control/ResolveVisitor.java
```diff
@@ -852,6 +858,21 @@ protected Expression transformPropertyExpression(PropertyExpression pe) {
         return ret;
     }
 
+    private boolean isVisibleNestedClass(ClassNode type, ClassNode ceType) {
+        if (!type.isRedirectNode()) return false;
+        ClassNode redirect = type.redirect();
+        // public or private
+        if (Modifier.isPublic(redirect.getModifiers()) || Modifier.isProtected(redirect.getModifiers())) return true;
```
jwagenleitner: The comment is misleading (private -> protected), but the code reads very good so maybe no comment needed.

src/main/org/codehaus/groovy/control/ResolveVisitor.java
```diff
@@ -852,6 +858,21 @@ protected Expression transformPropertyExpression(PropertyExpression pe) {
...
+        // package local
+        PackageNode classPackage = ceType.getPackage();
+        PackageNode nestedPackage = redirect.getPackage();
+        if ((redirect.getModifiers() & (Modifier.PROTECTED | Modifier.PUBLIC | Modifier.PRIVATE)) == 0 &&
+                ((classPackage == null && nestedPackage == null) ||
+                        classPackage != null && nestedPackage != null && classPackage.equals(nestedPackage)))
+            return true;
+        return false;
```
jwagenleitner: Comment says package local but it could be private at this point.  If you added an early exit `if (Modifier.isPrivate(redirect.getModifiers()) return false;` above the comment this would be package-private and the modifier check could be eliminated.

If that's the case then the return could be simplified:

```
return (classPackage == null && nestedPackage == null) || 
         (classPackage != null && classPackage.equals(nestedPackage))
```

I notice that `PackageNode` doesn't implement `equals` so it will end up being an `==` check.  I'm not that familiar with how the AST is built and if the same `PackageNode` is used for a given name so just wanted to bring that up.

jwagenleitner:
+1  (just a few minor comments)

src/main/org/codehaus/groovy/control/ResolveVisitor.java
paulk-asert: Oops, comment is obviously wrong - I'll fix or remove it

src/main/org/codehaus/groovy/control/ResolveVisitor.java
paulk-asert: I should probably use getName() on the package. It won't be private or the binary and will pick that up.

PR #398, opened 2016-08-26
Document error fix: instantiation using wrong parameter

binwang-developer:
Fixed 1 error of code in this page: [Design patterns in Groovy](http://groovy-lang.org/design-patterns.html), under section 1.2.2.


paulk-asert:
PR merged. Thanks!

PR #399, opened 2016-08-27
GROOVY-7920: Simple getAt example produces a BUG exception

paulk-asert:


jwagenleitner:
+1

I think with this change it would miss a default `getAt` method on an interface.  I don't think it's a real issue since it would be very edge case-y and I don't think there's much support now for default/static interface methods, just mentioning it.

paulk-asert:
OK, I'll have a think about that possibility and check whether GROOVY-7037 and GROOVY-6182 are related too. My inclination is to treat those as separate issues if more work is required.

PR #400, opened 2016-08-28
GROOVY-7921: using "this" as the target of an assignment not working …

paulk-asert:
…with Category annotation


jwagenleitner:
+1

paulk-asert:
This PR fixes the mentioned problem and frees up this as the target of an assignment for the purposes of DSLs and @Category classes but I noticed that outside those use cases we treat assignment of this as a NOP:
{code}
this = ''
println this
{code}
With or without @CompileStatic. I'll raise that as another issue since I believe we should treat that as some kind of error.

paulk-asert:
Merged.

PR #402, opened 2016-08-28
Extra test cases for Collection .intersect() and .disjoint()

YouriAckx:
`Collection.intersect()` was partly broken (or at least utterly inconsistent) between versions `2.4.2` and `2.4.6`. Simple test case:

    def foo = [[0, 0], [0, 1]]
    def bar = [[0, 1], [1, 1]]
    assert [[0, 1]] == foo.intersect(bar)

would fail in the aforementioned versions as `[]` would be returned. This is fixed in `2.4.7` but to avoid future regressions, it is nice to have test covering this case, as suggested by @glaforge and @sbglasius on [Twitter](https://twitter.com/YouriAckx/status/769230038847918080?lang=en).

A basic search didn't reveal any similar test cases (some `Foo` intersection comparison can be found but it can be seen as a different test case).


jwagenleitner:
Thanks! More tests are always a good thing. :smile:

PR #403, opened 2016-08-28
GROOVY-7611: java.util.Optional should evaluate to false if empty

jwagenleitner:


src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -10563,13 +10563,21 @@ private void prepare() {
...
      * @param object the object to coerce
      * @return the boolean value
      * @since 1.7.0
      */
     public static boolean asBoolean(Object object) {
-        return object != null;
+        if (object == null) {
+            return false;
+        }
+        if ("java.util.Optional".equals(object.getClass().getName())) {
```
blackdrag: just a very small thing.. could you add a todo to move this into a jvm8 module once we decided to make that our minimum jdk for the build?

paulk-asert:
+1

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
jwagenleitner: Good suggestion, I've added the todo comment.

paulk-asert:
Cédric, do you have any stats from your experiments with gradle on how frequently asBoolean(Object) is called? Just curious from a "favor correctness ahead of premature optimizations" point of view.

jwagenleitner:
@melix, thanks for the feedback.  Fair point about performance there would definitely be some cost to this and I can image quite a few loops using `if (somevar)` in place of a null check.  I can implement in a `groovy-jdk8` sub-project but am not sure the issue warrants the added complexity to the build and raising the minimum build jdk to 8.  Would be interested to know what others think.  Might be best to defer this until a later time unless we are ready now to move to 8 for build.

jwagenleitner:
Will close this and wait until we have 8 for the min build version.

PR #404, opened 2016-08-31
GROOVY-6764: Problem Referring To Statically Imported Constants

paulk-asert:


jwagenleitner:
+1

I thought maybe there'd be a potential stackoverflow problem by not also passing the visited Set in the recursive calls, but seems as if interface cycles are detected before this point so looks good.

PR #405, opened 2016-08-31
GROOVY-7497: GroovyDoc reports only last method of a script

paulk-asert:


paulk-asert:
This PR should now do all methods and also @Field annotated variable definitions within scripts.

blackdrag:
+1

PR #409, opened 2016-09-02
GROOVY-4018: Make the Groovy truth value of NaN be false

jwagenleitner:
Just thought I'd take a stab at implementing this, maybe there are performance concerns with this.  It may be more efficient to use `FLOAD`/`DLOAD` before the `isNaN` call instead of the `DUP2` up front, but in the context of those methods I wasn't sure how to access the index of the variable on the stack.

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/asm/BytecodeHelper.java
+++ b/src/main/org/codehaus/groovy/classgen/asm/BytecodeHelper.java
@@ -669,4 +669,78 @@ public class BytecodeHelper implements Opcodes {
         }
         return h;
     }
+
+    /**
+     * Converts a primitive type to boolean.
+     *
+     * @param mv method visitor
+     * @param type primitive type to convert
+     */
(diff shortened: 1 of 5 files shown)


paulk-asert:
+1, looks okay to me but I'm not super familiar with how performance might be affected at this level

jwagenleitner:
Thanks for looking it over, I appreciate it.

I'll let it sit for a while and see if I can get some time to do a quick benchmark.  Not sure how often `floats/doubles` are used in tight loops and compared using the Groovy Truth, but probably wouldn't hurt to make sure it's not too costly of an extra check for `NaN`.

jwagenleitner:
Updated the PR to consolidate the primitive-to-boolean conversion for both prim opts and static compilation and changed since tags to `3.0.0`.

paulk-asert:
LGTM. How about we change the since version to 2.6.0 and merge into master and GROOVY_2_6_X?

jwagenleitner:
Thanks Paul, I updated the since tags and merged into the branches suggested.

PR #410, opened 2016-09-02
GroovyRowResult.entrySet should return a Set of Entries

AarjavP:
Currently it returns a `Set`, I would prefer it if it returned a `Set<Map.Entry>` instead.


paulk-asert:
Merged, thanks!

PR #412, opened 2016-09-04
throw NoSuchElementException instead of returning null

jt1796:
Looking for an easy place to get started.

### DIFF
--- a/src/main/groovy/lang/IntRange.java
+++ b/src/main/groovy/lang/IntRange.java
@@ -26,6 +26,7 @@ import java.util.AbstractList;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
+import java.util.NoSuchElementException;
 
 /**
  * Represents a list of Integer objects starting at a specified {@code from} value up (or down)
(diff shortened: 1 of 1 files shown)


jwagenleitner:
Thanks for the PR.  Would you mind creating a JIRA ticket for this change (https://issues.apache.org/jira/browse/GROOVY)?  It helps to have a issue filed since they help to build a changelog for releases.  Also, if you're up for it, the same problem exists in `ObjectRange` if you want to make the same fix there.

jwagenleitner:
I created [GROOVY-7960](https://issues.apache.org/jira/browse/GROOVY-7960) for this and will merge this PR soon unless there are objections.

jwagenleitner:
Ok.  I had opened GROOVY-7961 for ObjectRange when I opened the one for IntRange, so 7962 is a duplicate.

blackdrag:
I think we can merge that

paulk-asert:
+1 (I changed the issue to have a breaking label - this is slightly pedantic as it is an undesirable edge case that will break but it should make it easier to find if anyone is relying on that edge case - see my comments on the issue for further details)

jt1796:
All,
The matching PR is here:
https://github.com/apache/groovy/pull/440

jwagenleitner:
PR merged, thanks!

PR #414, opened 2016-09-05
GROOVY-6831: Access to private static fields from within inner classe…

paulk-asert:
…s is inconsistent

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/InnerClassCompletionVisitor.java
+++ b/src/main/org/codehaus/groovy/classgen/InnerClassCompletionVisitor.java
@@ -18,6 +18,7 @@
  */
 package org.codehaus.groovy.classgen;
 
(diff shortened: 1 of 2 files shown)


paulk-asert:
This is the same as the old pull request:
https://github.com/groovy/groovy-core/pull/431
Cédric was concerned that the CompileStatic "part of the issue" wouldn't be correct. It is true that this change alters a missingMethod hook that is still generated even with CompileStatic in place but I checked and the code generated with CompileStatic looks good and doesn't use the methodMissing hooks (and appears to work fine). I'll assume that some other code change fixed the CompileStatic issue.

jwagenleitner:
+1

paulk-asert:
I'll close this PR and open a new one. If `$static_propertyMissing` is plumbed in correctly, the need to have the extra outer levels goes away since it will propagate outwards.

PR #415, opened 2016-09-06
Remove outdated parser loading logic.

tychobrailleur:
It looks like a long time ago there were two Groovy parsers, a “Classic”
and Antlr.  The Antlr one was loaded using reflection, while the classic
one was simply instantiated.  The reflection logic was kept after
definitely switching to Antlr (cf. ebeff74), although it is still
normally instantiated if reflection fails.

This commit simply removes all the reflection stuff.

### DIFF
--- a/src/main/org/codehaus/groovy/control/CompilerConfiguration.java
+++ b/src/main/org/codehaus/groovy/control/CompilerConfiguration.java
@@ -698,7 +698,7 @@ public class CompilerConfiguration {
 
     public ParserPluginFactory getPluginFactory() {
         if (pluginFactory == null) {
-            pluginFactory = ParserPluginFactory.newInstance(true);
+            pluginFactory = ParserPluginFactory.newInstance();
         }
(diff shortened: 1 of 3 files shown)


jwagenleitner:
Change looks good to me.  Though I think it might be better to keep the `#newInstance(boolean)` and create a new no-args newInstance that it could delegate to just in case anyone is currently using the method.  That way it wouldn't be a breaking change (even though the `org.codehaus.groovy` package is considered somewhat internal) and would be easier to incorporate into the 2_4_X branch also.

tychobrailleur:
@jwagenleitner Fair enough!  I have restored the old method, and added comments to document the backward compatibility bit.  Thanks!

paulk-asert:
I would probably deprecate the old method too

jwagenleitner:
@tychobrailleur Thanks for the PR!

I went ahead and added the deprecated tags as Paul suggested.

PR #416, opened 2016-09-08
GROOVY-7926 - Don't use a CastExpression if the return type is void.

graemerocher:
As per https://issues.apache.org/jira/browse/GROOVY-7926 traits currently try to cast the void type which leads to mysterious exceptions on some JVMs (notably JRocket and the IBM JVM)

### DIFF
--- a/src/main/org/codehaus/groovy/transform/trait/TraitComposer.java
+++ b/src/main/org/codehaus/groovy/transform/trait/TraitComposer.java
@@ -293,7 +293,8 @@ public abstract class TraitComposer {
 
         ClassNode[] exceptionNodes = correctToGenericsSpecRecurse(genericsSpec, copyExceptions(helperMethod.getExceptions()));
         ClassNode fixedReturnType = correctToGenericsSpecRecurse(genericsSpec, helperMethod.getReturnType());
-        Expression forwardExpression = genericsSpec.isEmpty()?mce:new CastExpression(fixedReturnType,mce);
+        boolean noCastRequired = genericsSpec.isEmpty() || fixedReturnType.getName().equals(ClassHelper.VOID_TYPE.getName());
+        Expression forwardExpression = noCastRequired ? mce : new CastExpression(fixedReturnType,mce);
         int access = helperMethod.getModifiers();
         // we could rely on the first parameter name ($static$self) but that information is not
         // guaranteed to be always present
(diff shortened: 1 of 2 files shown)


src/test/org/codehaus/groovy/transform/traitx/Groovy7926Bug.groovy
```diff
@@ -0,0 +1,50 @@
+package org.codehaus.groovy.transform.traitx
+
+import org.codehaus.groovy.classgen.asm.AbstractBytecodeTestCase
+import org.codehaus.groovy.classgen.asm.InstructionSequence
+
+/**
+ * Created by graemerocher on 08/09/2016.
+ */
+class Groovy7926Bug extends AbstractBytecodeTestCase {
```
jwagenleitner: Would probably be good to remove the class doc and add the license header.

src/main/org/codehaus/groovy/transform/trait/TraitComposer.java
```diff
@@ -293,7 +293,8 @@ private static void createForwarderMethod(
 
         ClassNode[] exceptionNodes = correctToGenericsSpecRecurse(genericsSpec, copyExceptions(helperMethod.getExceptions()));
         ClassNode fixedReturnType = correctToGenericsSpecRecurse(genericsSpec, helperMethod.getReturnType());
-        Expression forwardExpression = genericsSpec.isEmpty()?mce:new CastExpression(fixedReturnType,mce);
+        boolean noCastRequired = genericsSpec.isEmpty() || fixedReturnType.getName().equals(ClassHelper.VOID_TYPE.getName());
```
jwagenleitner: Just curious if you saw problems using just `fixedReturnType.equals(ClassHelper.VOID_TYPE)` that neccessitated comparing by name?  Most places that I see that compare types don't see to use the getName so am just curious.

jwagenleitner:
+1

src/main/org/codehaus/groovy/transform/trait/TraitComposer.java
graemerocher: Just seemed more efficient. ClassNode.equals does this:

```groovy
    public boolean equals(Object o) {
        if (redirect!=null) return redirect().equals(o);
        if (!(o instanceof ClassNode)) return false;
        ClassNode cn = (ClassNode) o;
        return (cn.getText().equals(getText()));
    }
```

The `getText()` method calls `getName()` anyway

jwagenleitner: Thanks for the explanation.

paulk-asert:
Merged (with license header as per John's comment). Thanks!

PR #417, opened 2016-09-08
Documentation fixes to The Groocy Development Kit

tinne:
Documentation fixes to [The Groocy Development Kit](http://groovy-lang.org/groovy-dev-kit.html) section on Working with collections
- property access comment targets entry instead of key
- two introductions followed by two corresponding code sections


jwagenleitner:
Thanks!

PR #418, opened 2016-09-13
GROOVY-7937: CLONE - same linkedlist code different behavior between …

paulk-asert:
…groovy and java (fix priority of DGM methods vs actual methods on an object)


jwagenleitner:
+1

PR #419, opened 2016-09-13
GROOVY-6396: same linkedlist code different behavior between groovy a…

paulk-asert:
…nd java

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -8928,19 +8928,23 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     /**
-     * Removes the initial item from the List, similar to pop on a Stack.
+     * Removes the initial item from the List.
+     *
      * <pre class="groovyTestCase">
      * def list = ["a", false, 2]
      * assert list.pop() == 'a'
      * assert list == [false, 2]
      * </pre>
      *
+     * This is similar to pop on a Stack where the first item in the list
+     * represents the top of the stack.
+     *
      * Note: The behavior of this method changed in Groovy 2.5 to align with Java.
      * If you need the old behavior use 'removeLast'.
      *
      * @param self a List
      * @return the item removed from the List
-     * @throws NoSuchElementException if the list is empty and you try to pop() it.
+     * @throws NoSuchElementException if the list is empty
      * @since 1.0
      */
     public static <T> T pop(List<T> self) {
@@ -8951,22 +8955,25 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     /**
-     * Removes the last item from the List. Using add() and pop()
-     * is similar to push and pop on a Stack.
+     * Removes the last item from the List.
+     *
      * <pre class="groovyTestCase">
      * def list = ["a", false, 2]
      * assert list.removeLast() == 2
      * assert list == ["a", false]
      * </pre>
      *
+     * Using add() and removeLast() is similar to push and pop on a Stack
+     * where the last item in the list represents the top of the stack.
+     *
      * @param self a List
      * @return the item removed from the List
-     * @throws NoSuchElementException if the list is empty and you try to pop() it.
+     * @throws NoSuchElementException if the list is empty
      * @since 2.5.0
      */
     public static <T> T removeLast(List<T> self) {
         if (self.isEmpty()) {
-            throw new NoSuchElementException("Cannot pop() an empty List");
+            throw new NoSuchElementException("Cannot removeLast() an empty List");
         }
         return self.remove(self.size() - 1);
     }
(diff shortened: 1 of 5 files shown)


jwagenleitner:
Since there is already a DGM method `T last(List<T>)` that does what `popLast()` does and `List#add(T)` appends just like `pushLast(T)`, maybe if those were mentioned in the javadoc for `push/pop` as alternatives to use if location of the items is important the new methods wouldn't be needed?

jwagenleitner:
Nevermind my last comment about `last`, I should have noticed the get vs remove.

paulk-asert:
You are correct about `pushLast` vs `add` but since this is a breaking change I was thinking having the `pushLast` alias will make it easier to explain to people. If you want the old behavior, use xxxLast. It also aligns with the naming scheme for Deque, e.g. peekFirst, peekLast, etc. But if there is a general consensus that the extra alias isn't needed, I am happy to remove it.

jwagenleitner:
I definitely agree that it would be good to have some replacement for removing the last element from the list (the reverse of `add`).  My guess it that that is a use case for using `pop` today (I know I have code that does that).  One alternative may be to add a `removeLast()` in place of `popLast()` and leave out `pushLast` in favor or `add`.  Just a thought, but think the `xxxxLast` methods are ok.

paulk-asert:
Changed as per John's suggestions.

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -8913,11 +8913,15 @@ public int compare(Map.Entry<K, V> e1, Map.Entry<K, V> e2) {
     }
 
     /**
-     * Removes the last item from the List. Using add() and pop()
-     * is similar to push and pop on a Stack.
-     * <pre class="groovyTestCase">def list = ["a", false, 2]
-     * assert list.pop() == 2
-     * assert list == ["a", false]</pre>
+     * Removes the initial item from the List, similar to pop on a Stack.
```
jwagenleitner: I wonder if most associate removing items from the top of a Stack, so in this case would it make sense to say 
>..similar to pop on a Deque.

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -8928,6 +8932,27 @@ public int compare(Map.Entry<K, V> e1, Map.Entry<K, V> e2) {
         if (self.isEmpty()) {
             throw new NoSuchElementException("Cannot pop() an empty List");
         }
+        return self.remove(0);
+    }
+
+    /**
+     * Removes the last item from the List. Using add() and pop()
```
jwagenleitner: Should this be `removeLast()` instead of `pop()`

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -8928,6 +8932,27 @@ public int compare(Map.Entry<K, V> e1, Map.Entry<K, V> e2) {
...
+     * is similar to push and pop on a Stack.
+     * <pre class="groovyTestCase">
+     * def list = ["a", false, 2]
+     * assert list.removeLast() == 2
+     * assert list == ["a", false]
+     * </pre>
+     *
+     * @param self a List
+     * @return the item removed from the List
+     * @throws NoSuchElementException if the list is empty and you try to pop() it.
```
jwagenleitner: I think `if the list is empty` suffices here and don't think the rest of the sentence is needed (same for `pop` throws).

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -8928,6 +8932,27 @@ public int compare(Map.Entry<K, V> e1, Map.Entry<K, V> e2) {
...
+     * @since 2.5.0
+     */
+    public static <T> T removeLast(List<T> self) {
+        if (self.isEmpty()) {
+            throw new NoSuchElementException("Cannot pop() an empty List");
```
jwagenleitner: `pop()` -> `removeLast()`

src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -8966,20 +8991,24 @@ public int compare(Map.Entry<K, V> e1, Map.Entry<K, V> e2) {
     }
 
     /**
-     * Appends an item to the List. Synonym for add().
-     * <pre class="groovyTestCase">def list = [3, 4, 2]
+     * Prepends an item to the start of the List, similar to push on a stack.
```
jwagenleitner: Same comment as with pop, would using Deque instead of Stack make sense?

paulk-asert:
John, I incorporated your changes but tried to not explicitly reference `Deque`. `Deque`/`LinkedList` is obviously the class we are trying to align behavior with and it lets you `push`/`pop` off the head (start) or end (via `add`/`removeLast`) but I was trying to explain the behavior without the complication of necessarily wanting to have to consider having both capabilities. Let me know what you think.

jwagenleitner:
I think the docs do a good job of describing the behavior and agree that it's best to not mention specific interface/impl.  Everything looks good to me.

PR #420, opened 2016-09-15
GROOVY-7940: @Lazy not generating "is" property accessor for booleans

jwagenleitner:


blackdrag:
+1

src/main/org/codehaus/groovy/transform/LazyASTTransformation.java
```diff
@@ -156,8 +156,11 @@ private static void addNonThreadSafeBody(BlockStatement body, FieldNode fieldNod
     private static void addMethod(FieldNode fieldNode, BlockStatement body, ClassNode type) {
         int visibility = ACC_PUBLIC;
         if (fieldNode.isStatic()) visibility |= ACC_STATIC;
-        final String name = "get" + MetaClassHelper.capitalize(fieldNode.getName().substring(1));
-        fieldNode.getDeclaringClass().addMethod(name, visibility, type, Parameter.EMPTY_ARRAY, ClassNode.EMPTY_ARRAY, body);
+        String propName = MetaClassHelper.capitalize(fieldNode.getName().substring(1));
+        fieldNode.getDeclaringClass().addMethod("get" + propName, visibility, type, Parameter.EMPTY_ARRAY, ClassNode.EMPTY_ARRAY, body);
+        if (ClassHelper.boolean_TYPE.equals(type)) {
+            fieldNode.getDeclaringClass().addMethod("is" + propName, visibility, type, Parameter.EMPTY_ARRAY, ClassNode.EMPTY_ARRAY, body);
```
paulk-asert: I'd probably use 'stmt(callThisX("get" + propName))' instead of 'body'. That should give a slightly smaller class.

paulk-asert:
+1 (see inline comment)

src/main/org/codehaus/groovy/transform/LazyASTTransformation.java
jwagenleitner: Thanks, I've updated so it delegates to the getter as suggested.

PR #421, opened 2016-09-15
GROOVY-7938: inconsistent access of methods in outer class

paulk-asert:


graemerocher:
Will this work with `@CompileStatic`? I noticed the use of static method missing

paulk-asert:
This PR in its current form doesn't address any issues with @CompileStatic access for methods in outer classes - this is just handling the dynamic code cases. It shouldn't impact @CS classes but also won't fix any bugs with @CompileStatic for these scenarios.

jwagenleitner:
+1

PR #422, opened 2016-09-16
GROOVY-7922: Static type checking not strict enough in the presence o…

blindpirate:
…f ambiguous method matching

the issue has been resolved and some refactor are done

### DIFF
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -887,7 +887,7 @@ public abstract class StaticTypeCheckingSupport {
             // we want to add one, because there is an interface between
             // the interface we search for and the interface we are in.
             if (sub != -1) {
-                sub+=(i+1); // GROOVY-6970: Make sure we can choose between equivalent methods
+                sub+=1; // GROOVY-6970: Make sure we can choose between equivalent methods
             }
             // we are interested in the longest path only
             max = Math.max(max, sub);
@@ -971,7 +971,7 @@ public abstract class StaticTypeCheckingSupport {
         }
         List<MethodNode> bestChoices = new LinkedList<MethodNode>();
         int bestDist = Integer.MAX_VALUE;
-        Collection<MethodNode> choicesLeft = removeCovariants(methods);
+        Collection<MethodNode> choicesLeft = removeCovariantsAndInterfaceEquivalents(methods);
         for (MethodNode candidateNode : choicesLeft) {
             ClassNode declaringClassForDistance = candidateNode.getDeclaringClass();
             ClassNode actualReceiverForDistance = receiver != null ? receiver : candidateNode.getDeclaringClass();
@@ -1106,7 +1106,7 @@ public abstract class StaticTypeCheckingSupport {
         return raw;
     }
 
-    private static Collection<MethodNode> removeCovariants(Collection<MethodNode> collection) {
+    private static Collection<MethodNode> removeCovariantsAndInterfaceEquivalents(Collection<MethodNode> collection) {
         if (collection.size()<=1) return collection;
         List<MethodNode> toBeRemoved = new LinkedList<MethodNode>();
         List<MethodNode> list = new LinkedList<MethodNode>(new HashSet<MethodNode>(collection));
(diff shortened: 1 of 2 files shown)


blindpirate:
When I was trying to figure out why [GROOVY-7922](https://issues.apache.org/jira/browse/GROOVY-7922) occured, I found that it was because of a mistake in fixing [GROOVY-6970](https://issues.apache.org/jira/browse/GROOVY-6970). In that solution, the developer was letting an interface's super interfaces (at same level) have different "distances" to it.

For example,

```
interface A{ void foo(); }
interface B{ void foo(); }
interface C extends A, B{ void foo(); }
```

When in static compilation, a specific version of method is chosen by comparing "distance". In this case, the distance of A and B to C is apparently equal, which will result in GROOVY-6970. In `org.codehaus.groovy.transform.stc.StaticTypeCheckingSupport`, line 890, we can find the original solution was that let A and B have different "distance" by assigning an additional `i`。I think the solution to GROOVY-6970 is not correct because it will result in GROOVY-7922, so I revert it and add a branch to re-fix GROOVY-6970.

src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1157,6 +1141,56 @@ private static ClassNode makeRawType(final ClassNode receiver) {
...
+        } else if (twoRT.isDerivedFrom(oneRT) || twoRT.implementsInterface(oneRT)) {
+            toBeRemoved.add(one);
+        }
+    }
+
+    private static boolean areOverloadMethodsInSameClass(MethodNode one, MethodNode two){
+        return one.getName().equals(two.getName()) && one.getDeclaringClass()==two.getDeclaringClass();
+    }
+
+    private static boolean allParameterTypesAreSame(Parameter[] onePars, Parameter[] twoPars) {
```
jwagenleitner: Might be able to use [`ParameterUtils#parametersEqual(Parameter[],Parameter[])`] (https://github.com/apache/groovy/blob/14266ad5b7c265b55334c029e87b1d79ebd2210e/src/main/org/codehaus/groovy/ast/tools/ParameterUtils.java#L26-L38) instead of this new method.

blindpirate: Oh, you're right. I will use this method instead.

blackdrag: I think I can +1 it after the change John suggested

src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1157,6 +1142,45 @@ private static ClassNode makeRawType(final ClassNode receiver) {
...
+        ClassNode oneDC=one.getDeclaringClass();
+        ClassNode twoDC=two.getDeclaringClass();
+        if(oneDC.implementsInterface(twoDC)){
+            toBeRemoved.add(two);
+        }else{
+            toBeRemoved.add(one);
+        }
+    }
+
+    private static boolean areEquivalentInterfaceMethods(MethodNode one, MethodNode two, Parameter[] onePars, Parameter[] twoPars) {
```
jwagenleitner: I think it would simplify things if this just took 2 MethodNode's, the parameters can be gotten from the nodes so no need to pass them in I think?

src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1116,39 +1117,23 @@ private static ClassNode makeRawType(final ClassNode receiver) {
...
-                            if (oneRT.isDerivedFrom(twoRT) || oneRT.implementsInterface(twoRT)) {
-                                toBeRemoved.add(two);
-                            } else if (twoRT.isDerivedFrom(oneRT) || twoRT.implementsInterface(oneRT)) {
-                                toBeRemoved.add(one);
-                            }
+                Parameter[] onePars = one.getParameters();
+                Parameter[] twoPars = two.getParameters();
+                if (onePars.length == twoPars.length) {
+                    if (areOverloadMethodsInSameClass(one,two)) {
+                        if (ParameterUtils.parametersEqual(onePars, twoPars)){
```
jwagenleitner: If the `Parameter[]` parameters are removed from `areEquivalentInterfaceMethods` method I think these arguments can be replaced with `one.getParameters(), two.getParameters()` and then the local variables wouldn't be needed.  Just a suggestion.

src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
blindpirate: Yes, good suggestion. I will do that.

jwagenleitner:
PR merged, thanks!

blindpirate:
Thank you!

PR #423, opened 2016-09-18
GROOVY-7942: @Lazy transform creates unneeded getter/setter for backi…

paulk-asert:
…ng store field


jwagenleitner:
+1, nice catch.

PR #424, opened 2016-09-18
GROOVY-7933: fix method selection for boolean and char primitives

jwagenleitner:
boolean and char were not factored into the primitive distance tables which
would cause methods with Object to be selected in preference to those
with boolean/char and their respective Wrapper.

Inserted `boolean` and `char` (and their wrappers) in the distance table and kept all distances the same for the existing classes in the table (adding boolean and char as the farthest dist from the existing classes).

Had to fix one test because `HashCodeHelper` has overloaded methods and prior to this change it was selecting the method with `Object` instead of `boolean`.


blackdrag:
+1

paulk-asert:
I'm +1 for this change. But I am keen that we don't forget about the inconsistencies highlighted in the discussion.
```
import groovy.transform.*

class Foo {
    private static final int val = 2
    private static final Integer val2 = 3
    int a(int a) { 0 }
    int a(Integer a) { 1 }

    @CompileStatic
    def csMethod() {
        assert a(val) == 0
        assert a(val2) == 1
        def x = a(val)
        assert x == 0
        x = a(val2)
        assert x == 1
    }

    def plainMethod() {
        assert a(val) == 1
        assert a(val2) == 1
        def x = a(val)
        assert x == 0
        x = a(val2)
        assert x == 1
    }
}

new Foo().csMethod()
new Foo().plainMethod()
```
We should either list the differences as bugs or "features" and, if a feature, have some doco somewhere that recommends that if both primitive and wrapper versions of a method are available that they produce the same result so that @CS or primopts won't change the result.

paulk-asert:
Actually, wrt my last comment, perhaps this issue is a good place to track those differences:
https://issues.apache.org/jira/browse/GROOVY-7850

PR #425, opened 2016-09-19
GROOVY-6831: Access to static fields/properties from within inner cla…

paulk-asert:
…sses is inconsistent


paulk-asert:
Proposed PR merged.

PR #427, opened 2016-09-19
Compatibility in 2.4.x with Gradle's classloader cleanup

jwagenleitner:
Gradle has a cleanup mechanism in place that reflectively accesses
the ClassInfo.klazz field.  This change prevents an exception being
thrown by the cleanup method.  This workaround also means that the
cleanup strategy will no longer work, the fix to replace klazz with a
WeakReference should eliminate the need for the explicit cleanup.

For more details, see dev mailing list thread:
https://mail-archives.apache.org/mod_mbox/incubator-groovy-dev/201609.mbox/%3CCAHPL-JkQ%2BU8PfaxyVhtE%3DvGV%2BsXXmCzxOsCR8UbBww6P2vskPg%40mail.gmail.com%3E

### DIFF
--- a/src/main/org/codehaus/groovy/reflection/ClassInfo.java
+++ b/src/main/org/codehaus/groovy/reflection/ClassInfo.java
@@ -49,7 +49,19 @@ public class ClassInfo implements Finalizable {
     private final LazyClassLoaderRef artifactClassLoader;
     private final LockableObject lock = new LockableObject();
     public final int hash = -1;
-    private final WeakReference<Class<?>> klazz;
+    private final WeakReference<Class<?>> classRef;
+
+    // TODO: should be able to remove the klazz field once 2.5 becomes the mainline release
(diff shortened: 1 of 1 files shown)


paulk-asert:
+1. I would that this is worth doing in the meantime anyway. If we get a Gradle fix later, we can always remove the field.

blackdrag:
as step in between, yes +1

oehme:
We upgraded to 2.5.3 and the problem persists, so now I had to adjust our cleanup code to use `classRef` instead of `klazz`.

It would be great if the root cause of this would be fixed, i.e. Groovy shouldn't leak strong references to classes. You've replaced the `klazz` reference with a weak one, but the `dgmMetaMethods` field will still transitively hold strong references to any classes used as method return types etc. Therefor their classloaders will also be leaked, creating giant metaspace leaks.

PR #430, opened 2016-09-26
GROOVY-7909 Calling parents method from trait using Parent.super.method() fail depending on trait declaration order

blindpirate:


blindpirate:
I made the commit with wrong name, and I will open a PR again.

PR #432, opened 2016-09-26
GROOVY-7909 Calling parents method from trait using Parent.super.method() fail depending on trait declaration order 

blindpirate:


src/test/groovy/bugs/Groovy7909Bug.groovy
```diff
@@ -0,0 +1,57 @@
+package groovy.bugs
```
jwagenleitner: Could you add the license header to this file, should be able to grab it from another file in the same package.

blindpirate: It has been added.

src/main/org/codehaus/groovy/transform/trait/SuperCallTraitTransformer.java
```diff
@@ -36,13 +38,19 @@
...
+import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
+
 /**
  * This transformer is used to transform calls to <code>SomeTrait.super.foo()</code> into the appropriate trait call.
  *
  * @author Cédric Champeau
  * @since 2.3.0
  */
 class SuperCallTraitTransformer extends ClassCodeExpressionTransformer {
+    public static final String UNRESOLVED_HELPER_CLASS = "UNRESOLVED_HELPER_CLASS";
```
jwagenleitner: This is only used in the same package, I think package-private would be better than public.

blindpirate: Yes, I haven't realize it. Done as you say.

blackdrag:
+1

paulk-asert:
Merged, thanks!

PR #436, opened 2016-10-04
GROOVY-7952: Property expressions for extension methods starting with…

shils:
… 'is' fail STC

* Add support for 'is' getter method variants of GROOVY-5580

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java
+++ b/src/main/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java
@@ -158,19 +158,24 @@ public class StaticTypesCallSiteWriter extends CallSiteWriter implements Opcodes
 
         // GROOVY-5580, it is still possible that we're calling a superinterface property
         String getterName = "get" + MetaClassHelper.capitalize(methodName);
+        String altGetterName = "is" + MetaClassHelper.capitalize(methodName);
         if (receiverType.isInterface()) {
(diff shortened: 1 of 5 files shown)


blackdrag:
should we limit the lookup of is-property-accessor-methods to boolean returning ones?

shils:
Agreed, the lookups should match groovy's generation of synthetic methods.

paulk-asert:
+1 with the check for boolean

shils:
Amended to reflect @blackdrag 's comment

src/main/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java
```diff
@@ -158,19 +158,24 @@ public void makeGetPropertySite(Expression receiver, final String methodName, fi
...
+                if (getterMethod == null) getterMethod = anInterface.getGetterMethod(altGetterName);
+                if (getterMethod != null) break;
             }
             // GROOVY-5585
-            if (getterMethod==null) {
+            if (getterMethod == null) {
                 getterMethod = OBJECT_TYPE.getGetterMethod(getterName);
             }
+            if (getterMethod == null) {
+                getterMethod = OBJECT_TYPE.getGetterMethod(altGetterName);
```
paulk-asert: There are no isXxx methods in Object? Is this needed?

shils: I suppose not, removed.

PR #438, opened 2016-10-05
GROOVY-7948: fix completion of static imports in groovysh

heenenee:
Fixed completion of static imports; updated unit tests; did some refactorings in ImportsSyntaxCompletor; added comment explaining how imports are stored in Groovysh.  Also did some manual testing of completion functionality using groovysh.bat in a cleanly built binary distribution.

### DIFF
--- a/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/completion/ImportsSyntaxCompletor.groovy
+++ b/subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/completion/ImportsSyntaxCompletor.groovy
@@ -31,8 +31,10 @@ class ImportsSyntaxCompletor implements IdentifierCompletor {
     // cache for all preimported classes
     List<String> preimportedClassNames
     // cache for all manually imported classes
-    final Map<String, Collection<String>> cachedImports = new HashMap<>().withDefault {String key ->
-        collectImportedSymbols(key)
+    final Map<String, Collection<String>> cachedImports = new HashMap<String, Collection<String>>().withDefault {String key ->
+        Collection<String> matchingImports = new TreeSet<String>()
+        collectImportedSymbols(key, matchingImports)
+        matchingImports
     }
 
     ImportsSyntaxCompletor(final Groovysh shell) {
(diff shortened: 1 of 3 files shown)


blackdrag:
This change is probably to be considered as a breaking change. If John is ok with this I am ok with having this on master for 2.5.0

jwagenleitner:
Thanks for the PR.

I haven't gone through it in detail, but wonder if we didn't use the diamond operator on `cachedImports` and retained the method signature of `collectImportedSymbols` if this fix would be eligible for 2_4_X?

heenenee:
I could get rid of the diamond operator and preserve the method signatures,
and incorporate whatever other feedback if it helps get the fix released
sooner.

jwagenleitner:
I forgot to mention that while the changes to the test (removing `import`) seems to be a significant change, I think this was a problem with the tests not matching the actual behavior of Groovysh since it stores imports with the `import` as mentioned by the added comment.

In addition to fixing completion of static imports, I also found this fixes the completion of star imports.  Prior to this fix `import java.text.*` and typing `new SimpleDa<tab>` would not return `SimpleDateFormat`, after this fix it does.

subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/completion/ImportsSyntaxCompletor.groovy
```diff
@@ -87,51 +78,49 @@ class ImportsSyntaxCompletor implements IdentifierCompletor {
...
      * finds matching imported classes or static methods
-     * @param prefix
-     * @param importSpec
-     * @param matches
-     * @return
+     * @param importSpec an import statement without the leading 'import ' or trailing semicolon
+     * @return all names matching the importSpec
      */
-    void collectImportedSymbols(final String importSpec, final Collection<String> matches) {
+    SortedSet<String> collectImportedSymbols(final String importSpec) {
```
blackdrag: we decided to be strict about such things in a bugfix release and 2.4.8 is one. So yeah, unless the method signature is made compatible no 2.4.x

heenenee:
I removed the diamond operator and preserved the method signature and parameter names of `collectImportedSymbols` to make the fix available to 2.4.x.

jwagenleitner:
PR merged for 2.4.x and 2.5, thanks!

PR #440, opened 2016-10-06
Groovy-7961 : NoSuchElementException

jt1796:
https://issues.apache.org/jira/browse/GROOVY-7961

This is a similar PR to 
https://github.com/apache/groovy/pull/412


src/main/groovy/lang/ObjectRange.java
```diff
@@ -450,6 +451,9 @@ public Comparable next() {
                 value = peek();
                 nextFetched = true;
             }
+            if (value == null) {
+                throw new NoSuchElementException();
+            }
```
jwagenleitner: I think these first 2 `if` statements could be replaced by the following and would make it cleaner:
```
if (!hasNext()) {
    throw new NoSuchElementException();
}
```

jt1796: Thanks, much better.

jwagenleitner:
PR merged, thanks!

PR #441, opened 2016-10-07
GROOVY-7951: MethodCallExpression.transformExpression does not copy generic types

jwagenleitner:
PR created from patch supplied in [GROOVY-7951](https://issues.apache.org/jira/browse/GROOVY-7951).  It seems like a reasonable change to make, though I did have trouble thinking up a way to test it that would be meaningful.  If anyone can nudge me in the right direction I can try to write up a test if you think it's warranted in this case.

I considered whether `genericsTypes.clone()` should be used, but it didn't seem to be common and most uses of the array seem to be read only.

### DIFF
--- a/src/main/org/codehaus/groovy/ast/expr/MethodCallExpression.java
+++ b/src/main/org/codehaus/groovy/ast/expr/MethodCallExpression.java
@@ -72,6 +72,7 @@ public class MethodCallExpression extends Expression implements MethodCall {
         answer.setSafe(safe);
         answer.setSpreadSafe(spreadSafe);
         answer.setImplicitThis(implicitThis);
+        answer.setGenericsTypes(genericsTypes);
         answer.setSourcePosition(this);
         answer.setMethodTarget(target);
         answer.copyNodeMetaData(this);
--- a/src/test/groovy/bugs/Groovy7951Bug.groovy
+++ b/src/test/groovy/bugs/Groovy7951Bug.groovy
@@ -1,0 +1,75 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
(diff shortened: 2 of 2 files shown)


paulk-asert:
What about something like:
```
import org.codehaus.groovy.ast.ClassNode
import org.codehaus.groovy.ast.ClassCodeVisitorSupport
import org.codehaus.groovy.ast.expr.MethodCallExpression
import org.codehaus.groovy.control.CompilerConfiguration
import org.codehaus.groovy.control.SourceUnit
import org.codehaus.groovy.control.customizers.CompilationCustomizer
import org.codehaus.groovy.classgen.GeneratorContext
import static org.codehaus.groovy.control.CompilePhase.CANONICALIZATION

def config = new CompilerConfiguration()
def checker = new CompilationCustomizer(CANONICALIZATION) {
  void call(SourceUnit src, GeneratorContext ctxt, ClassNode cn) {
    new ClassCodeVisitorSupport() {
      void visitMethodCallExpression(MethodCallExpression mce) {
        if (mce.objectExpression.text == 'java.util.Collections' &&
            mce.method.text == 'emptyList') {
            assert mce.genericsTypes != null && mce.genericsTypes*.name == ['Date']
        }
        super.visitMethodCallExpression(mce)
      }
      SourceUnit getSourceUnit() { src }
    }.visitClass(cn)
  }
}
config.addCompilationCustomizers(checker)

new GroovyShell(config).parse '''
    Collections.<Date>emptyList()
'''
```
Currently the macro transform (so 2.5 only) is stealing the generics information away but after the fix it stays put.

jwagenleitner:
Thanks for the test case @paulk-asert, that was helpful.  I added the `ClassCodeExpressionTransformer` transformer so that the test would break in 2_4_X (without relying on the macro transform being in place) since this fix seems like something that could go in that branch as well as master, but would be interested to know what you think.

paulk-asert:
Nice. +1 for both branches.

PR #444, opened 2016-10-10
GROOVY-7646: allow classes to be removed from the ClassInfo cache

jochenberger:
This is the first part of what @jwagenleitner originally did for #325.


jwagenleitner:
Thanks!  I merged this in master and 2_4_X so it should be in 2.4.8 when released.

PR #445, opened 2016-10-10
GROOVY-7646: remove classes via InvokerHelper when closing the GroovyClassLoader or flushing its cache

jochenberger:
See #325, this ensures that classes are properly cleaned up after when their ClassLoader or its cache are discarded. Together with #444, this should allow classes to be GC'ed as soon as they are not used anymore.

### DIFF
--- a/src/main/groovy/lang/GroovyClassLoader.java
+++ b/src/main/groovy/lang/GroovyClassLoader.java
@@ -983,9 +983,7 @@ public class GroovyClassLoader extends URLClassLoader {
     @Override
     public void close() throws IOException {
         super.close();
-        for (Class cl : classCache.values()) {
-            InvokerHelper.removeClass(cl);
-        }
+        clearCache();
     }
 
     private static class TimestampAdder extends CompilationUnit.PrimaryClassNodeOperation implements Opcodes {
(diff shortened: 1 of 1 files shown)


src/main/groovy/lang/GroovyClassLoader.java
```diff
@@ -969,13 +970,24 @@ public Void run() {
...
         }
         synchronized (sourceCache) {
             sourceCache.clear();
         }
     }
 
+    @Override
+    public void close() throws IOException {
+        super.close();
+        for (Class cl : classCache.values()) {
```
jochenberger: We could also just call `clearCache()` from here.

blackdrag: +1 for clear cache... especially since that does proper synchronization

jochenberger: I thought that synchronization might not be an issue since we are `close`-ing the ClassLoader anyway and I wanted to avoid the additional cost of synchronizing over and clearing `sourceCache`. But I guess it's better safe than sorry, so I made the change.

jwagenleitner:
Thanks!  I only merged this into master (2.5.0) since `URLClassLoader` implements `Closeable` as of Java 7.  The 2.4.x branch still supports Java 6.

PR #446, opened 2016-10-10
GROOVY-7291 Declaration of double variable without value assignment referrenced in closure

blindpirate:
According to [GROOVY-5570](https://issues.apache.org/jira/browse/GROOVY-5570), a variable referenced by closure must be non-primitive type, and it results in [GROOVY-7291](https://issues.apache.org/jira/browse/GROOVY-7291). In this case, I replace the EmptyExpression to a ConstantExpression with value 0.


jwagenleitner:
PR merged, thanks!

PR #449, opened 2016-10-15
GROOVY-7932: generate bridge methods for private constructors during …

shils:
…static compilation


src/main/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
```diff
@@ -284,17 +286,36 @@ private static void addPrivateBridgeMethods(final ClassNode node) {
...
-                MethodNode bridge = node.addMethod(
-                        "access$"+i, access,
-                        correctToGenericsSpecRecurse(genericsSpec, method.getReturnType(), methodSpecificGenerics),
-                        newParams,
-                        method.getExceptions(),
-                        returnStatement);
+
+                MethodNode bridge;
+                if (method instanceof ConstructorNode) {
+                    // create constructor with an nested class as the first parameter, creating one if necessary
```
paulk-asert: trivial typo: an nested => a nested

paulk-asert:
+1, nice work

src/main/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
shils: good catch, fixed

jwagenleitner:
+1, I tested and this also fixes [GROOVY-6468](https://issues.apache.org/jira/browse/GROOVY-6468).

PR #453, opened 2016-10-29
[GROOVY-7981] Set access level for constructors created by groovy.tra…

alien11689:
…nsform.Immutable

### DIFF
--- a/src/main/groovy/transform/AccessModifier.java
+++ b/src/main/groovy/transform/AccessModifier.java
@@ -1,0 +1,25 @@
+package groovy.transform;
+
+import java.lang.reflect.Modifier;
+
+/**
+ * This enumeration can be used in {@link Immutable} annotation for setting access modifier for constructors
+ *
+ * @author Dominik Przybysz
+ */
+public enum AccessModifier {
+    PUBLIC(Modifier.PUBLIC),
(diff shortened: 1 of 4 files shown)


blackdrag:
Would be nice to see Paul comment on this, but I think we I can a +1 on this

paulk-asert:
This is probably okay.  Some notes below:

`AccessModifier.PROTECTED` doesn't make a lot of sense since `Immutable` classes are final. Another option would be to also drop package private and switch to just a Boolean `privateConstructors` but I think having a general purpose enum is probably going to be more consistent in the long run if we need this capability in other places later, so I'd probably just leave it.

Currently you can use `@Immutable` in conjunction with `@Builder`, e.g.:
```
@Immutable @Builder(builderStrategy=InitializerStrategy)
class Person { String name }
def john = new Person(Person.createInitializer().name('John'))
```
I'd just note that the current PR doesn't change the visibility of the constructor supplied by `@Builder`.

paulk-asert:
Your PR inspired me to add a @VisibilityOptions annotation to the recently added @NamedVariant transform (targeted for 2.5.0). I am going to add support for it into TupleConstructor and MapConstructor transforms (and perhaps builders) soon. Keep up the great suggestions.

PR #454, opened 2016-10-31
GROOVY-7976: Sort methods that accept a comparator should accept Comp…

paulk-asert:
…arator<? super T>

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -8300,7 +8300,7 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
      * @return the sorted map
      * @since 1.7.2
      */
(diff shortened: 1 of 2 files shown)


mauromol:
Hi Paul, I hope I'm not doing a mistake by writing here. Anyway I had a look to the commit and your new test method is `testSortMethodsWithComparatorAcceptingSubclass` and it tests for a `Comparator<? extends CharSequence>`, IMHO, however, the point of the bug report is that those methods should accept a Comparator accepting SUPERclass, so that you can sort, say, an `Iterable<String>` with a `Comparator<Object>` (which, as an example, may sort based on the object `toString()` output,).

paulk-asert:
Hi Mauro, writing here is fine! I misunderstood what you were after - adding a failing test to your issue is a good way to remove ambiguity! :-) OK, so there are two issues. Firstly, the DGM signatures needed changing which is what I have added so far. But there is also a bug in the type checker at the moment which now needs fixing. It still reports an error but shouldn't for the case you are after even after the signature changes.

PR #455, opened 2016-11-02
GROOVY-7797 Private trait method called from within a closure has the wrong "this" context

blindpirate:
Use helper class as method receiver instead of wrong "this" when private trait method in a closure is called.


blackdrag:
First of all I advice doing much smaller pull requests and split a bug change and a refactoring into separate ones. This will very much help to reduce the time needed till somebody does a review. 

But besides that +1

blindpirate:
@blackdrag Yes, you're right.

paulk-asert:
Proposed PR merged with a few minor tweaks - other minor changes caused the merge to fail. Our bad for taking so long I guess. But all good now. Thanks!

PR #457, opened 2016-11-05
If a converter is found, further process the result

jameskleeh:


jameskleeh:
Should this line be using the generator instead?
 https://github.com/apache/groovy/blob/master/subprojects/groovy-json/src/main/java/groovy/json/StreamingJsonBuilder.java#L731

jwagenleitner:
Yes I think that method was added fairly recently and was missed, the generator should replace all calls to `JsonOutput.toJson`.

jwagenleitner:
Would like to know what you think about removing the `Object convert(Object)` method and just leave the `Object convert(Object,String)` method on the Converter, doesn't seem like both are needed as long as it documented that the key may be null.

jameskleeh:
@jwagenleitner Makes sense to me

jwagenleitner:
PR merged, thanks!

PR #458, opened 2016-11-07
Fix test case and documentation

kno10:
`Foo.i` is a field; only `Foo.i.class` could be `int.class`


paulk-asert:
PR merged, thanks!

PR #459, opened 2016-11-10
GROOVY-7994: Anonymous inner class believes protected method in paren…

paulk-asert:
…t's superclass returns Object


jwagenleitner:
+1

PR #460, opened 2016-11-10
GROOVY-7995: Short syntax of closure call invokes wrong closure if wrapped in another closure and @CompileStatic is applied

blindpirate:


blindpirate:
It seems that some errors occurred because of 503 in installing jdk. [https://travis-ci.org/apache/groovy/jobs/174813683](https://travis-ci.org/apache/groovy/jobs/174813683)

image

jwagenleitner:
Might just be a temporary problem, if you squash the commits and force push it should kick off the TravisCI again.

blindpirate:
I'm afraid I can't squash the commits because they are across a merge - I should have synced my master with upstream instead of merging my own branched into it. 

But it doesn't matter, I just closed and reopened this PR to trigger travis build.

blackdrag:
After looking at this pullrequest I am not sure if this does not go a bit over board. It means all the calls to call() and call(Object) will then go to call(Object[]) instead. The problem with this now is, that call(Object[]) will use a dynamic method call to call the appropriate doCall method, while in the presented cases call() and call(Object) do direct method calls. In other words we would loose a lot of performance here. I also think that part of the problem gets "covered up" by now using a different invocation path in the StaticMethodcallWriter. The key element should be isCallOnClosure, as has bee correctly identified. But if we applied for example this patch here [pr460.diff.txt](https://github.com/apache/groovy/files/736549/pr460.diff.txt) it will make the supplied test case work... while not giving up on the call semantics

jwagenleitner:
Thanks for the PR, merged parts of it along with @blackdrag 's suggested changes.

blindpirate:
Thanks @jwagenleitner

PR #461, opened 2016-11-16
@TailRecursive doc should read "Method annotation ..."

leoger:
Minor text fixes to TailRecursive annotation's documentation:
 - it is a method annotation, not a class annotation
 - since I was already here, added an 's' to make pluralization of "recursive calls" consistent


jwagenleitner:
Thanks!

PR #462, opened 2016-11-22
GROOVY-6175: invoking Closure property like method fails because of doCall

jwagenleitner:
Per the Closure class docs, to be able to use a Closure in short form, e.g., c(), in subclasses, you need to provide a doCall method with any signature you want to. This ensures that getMaximumNumberOfParameters() and getParameterTypes() will work too without any additional code. If no doCall method is provided a closure must be used in its long form, e.g., c.call().

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/memoize/Memoize.java
+++ b/src/main/org/codehaus/groovy/runtime/memoize/Memoize.java
@@ -137,6 +137,10 @@ public abstract class Memoize {
             }
             return result == MEMOIZE_NULL ? null : (V) result;
         }
+
+        public V doCall(final Object... args) {
+            return call(args);
+        }
     }
     
     private static class SoftReferenceMemoizeFunction<V> extends MemoizeFunction<V> {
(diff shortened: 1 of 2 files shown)


blackdrag:
not sure this will do. You do want the parameter types and number of parameters depending on the closure you memoized. What happens if you keep those as in the original but do the call/doCall change?

jwagenleitner:
Thanks.  I added back setting the parameters and added a test for param count/types.  Also made the same doCall change to the `SoftReferenceMemoizeFunction` to make sure it'll pick up the right method.

jwagenleitner:
Because of [GROOVY-4482](https://issues.apache.org/jira/browse/GROOVY-4482) I wonder if it would be safer to leave the `call(Object...)` method in place and have `doCall` delegate to it in order to avoid the `invokeMethod` that `Closure.call` uses.  Not sure how to test against GAE.

jwagenleitner:
I went ahead and left the original `call(...)` and just added a `doCall(...)` that delegates to call.

paulk-asert:
+1

PR #463, opened 2016-11-23
GROOVY-7584: transient fields in trait are not transient in implement…

paulk-asert:
…ing class


paulk-asert:
I have to rework one test before committing due to different field orderings between jdk 7 and 8 which are affecting the toString() value used in a test.

PR #472, opened 2016-12-29
Macro methods

bsideup:
Introduce macro methods and re-implement macro as a macro method


subprojects/groovy-macro/src/main/groovy/org/codehaus/groovy/macro/transform/MacroTransformation.java
```diff
@@ -18,42 +18,113 @@
...
+
+                for (int i = 0; i < callArguments.size(); i++) {
+                    argumentsList[i] = ClassHelper.make(callArguments.get(i).getClass());
+                }
+
+                methods = StaticTypeCheckingSupport.chooseBestMethod(MACRO_CONTEXT_CLASS_NODE, methods, argumentsList);
+
+                for (MethodNode macroMethodNode : methods) {
+                    if (!(macroMethodNode instanceof ExtensionMethodNode)) {
+                        // TODO is it even possible?
```
melix: If you don't know if it's possible, it's probably better to throw an error. Then if the case appears we know we need to fix it.

subprojects/groovy-macro/src/main/groovy/org/codehaus/groovy/macro/transform/MacroTransformation.java
```diff
@@ -18,42 +18,113 @@
...
                 return sourceUnit;
             }
 
             @Override
-            public Expression transform(final Expression exp) {
-                if (exp instanceof ConstructorCallExpression) {
-                    MethodCallExpression call = exp.getNodeMetaData(MacroTransformation.class);
-                    if (call!=null) {
-                        return call;
+            public void visitMethodCallExpression(MethodCallExpression call) {
```
melix: This method is very long. Could you split it into smaller focused pieces?

subprojects/groovy-macro/src/main/groovy/org/codehaus/groovy/macro/transform/MacroMethodsCache.java
```diff
@@ -0,0 +1,144 @@
...
+
+import java.util.*;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * TODO share some code with {@link org.codehaus.groovy.transform.stc.StaticTypeCheckingSupport.ExtensionMethodCache}
+ * @author Sergei Egorov <bsideup@gmail.com>
+ * @since 2.5.0
+ */
+class MacroMethodsCache {
```
danielsun1106: Maybe you can reuse LRU cache in the parrot branch: https://issues.apache.org/jira/browse/GROOVY-7977

bsideup: We could once parrot is merged to master :)

paulk-asert:
Proposed PR merged. Thanks Sergei!

PR #474, opened 2017-01-02
Fixing broken bold style

opporancisis:
Double asterisk should be used. Otherwise single asterisk conflicts with *:m1 on the same line.


jwagenleitner:
Thanks!

PR #481, opened 2017-01-26
Fix misprint in javadoc for href attribute

Alykoff:
See RFC-2368 (https://tools.ietf.org/html/rfc2368).
The right scheme for  mailtoURL  is  "mailto:" [ to ] [ headers ]

### DIFF
--- a/src/main/org/codehaus/groovy/ast/ASTNode.java
+++ b/src/main/org/codehaus/groovy/ast/ASTNode.java
@@ -44,7 +44,7 @@ import java.util.Map;
  * </ul>
  *
  * @author <a href="mailto:james@coredevelopers.net">James Strachan</a>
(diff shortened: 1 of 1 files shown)


paulk-asert:
Thanks. Just as a general note, Apache encourages projects to not include @author tags in their source code. Apart from reducing maintenance, it is meant to increase the community feeling of ownership of the whole codebase. After all, you can find out who wrote which lines from git, so it isn't a huge loss of information. So, a PR that deleted the authors would have been just as good - we try hard to make sure that anyone who has contributed anything of any consequence is mentioned in the pom - and the two authors in this file are.

In any case, we haven't adopted that policy globally yet, just slowly heading in that direction - so I merged your PR, since it is better to not have the error.

PR #484, opened 2017-01-29
GROOVY-8067: Possible deadlock when creating new ClassInfo entries in the cache

jwagenleitner:
While I have been able to replicate the deadlock between `GroovyClassValuePreJava7$Segment` and the `GlobalClassSet#add`, I have not directly observed one between the `modifiedExpandos` and the `GlobalClassSet`, but think that it would be good to isolate their reference processing too since both lock in their operations.

### DIFF
--- a/src/main/org/codehaus/groovy/reflection/ClassInfo.java
+++ b/src/main/org/codehaus/groovy/reflection/ClassInfo.java
@@ -456,9 +456,10 @@ public class ClassInfo implements Finalizable {
     }
 
     private static class GlobalClassSet {
-    	
-    	private final ManagedLinkedList<ClassInfo> items = new ManagedLinkedList<ClassInfo>(weakBundle);
-    	
+        // New ReferenceBundle is used to avoid possible deadlocks with modifiedExpandos
+    	private final ManagedLinkedList<ClassInfo> items =
+                new ManagedLinkedList<ClassInfo>(ReferenceBundle.newWeakBundle());
+
     	public int size(){
 		return values().size();
     	}
--- a/src/main/org/codehaus/groovy/reflection/GroovyClassValuePreJava7.java
+++ b/src/main/org/codehaus/groovy/reflection/GroovyClassValuePreJava7.java
@@ -29,7 +29,14 @@ import org.codehaus.groovy.util.ReferenceBundle;
  * @param <T>
  */
 class GroovyClassValuePreJava7<T> implements GroovyClassValue<T> {
-	private static final ReferenceBundle weakBundle = ReferenceBundle.getWeakBundle();
+
+	// Computing values for this cache occurs while holding a segment lock and part of the
+	// computation requires a lock on GlobalClassSet.add, backed by a ManagedLinkedList.
+	// If objects from this cache are dequeued by the add operation then it could attempt
+	// to remove an entry from a segment of this cache for which another thread holds a lock.
+	// If that other thread is also computing a new value it will be blocked on the add operation.
+	// To avoid possible deadlocks a different ReferenceQueue is required for this cache.
+	private static final ReferenceBundle weakBundle = ReferenceBundle.newWeakBundle();
 
 	private class EntryWithValue extends ManagedConcurrentMap.EntryWithValue<Class<?>,T>{
 
--- a/src/main/org/codehaus/groovy/util/ReferenceBundle.java
+++ b/src/main/org/codehaus/groovy/util/ReferenceBundle.java
@@ -21,6 +21,9 @@ package org.codehaus.groovy.util;
 import java.lang.ref.ReferenceQueue;
 
 public class ReferenceBundle{
+
+    private static final int DEFAULT_IDLE_THRESHOLD = 500;
+
     private final ReferenceManager manager;
     private final ReferenceType type;
     public ReferenceBundle(ReferenceManager manager, ReferenceType type){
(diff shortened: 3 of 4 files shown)


paulk-asert:
Wrt ClassInfoDeadlockTest.java#L58: should we change that comment, or make it a manual test?

jwagenleitner:
Since this test can take minutes to run, requires several threads and performs GC I don't think it should be mixed in with the normal unit tests.  But it would be nice to have a way to include tests like these (spin up threads, GC, etc).  Is there a certain way manual tests are implemented, would it be the benchmark or performance tests?

blackdrag:
maybe we could add this to the performance tests part, yes

But frankly I am surprised if your change solves the problem, since it looks to me as if it is the same reference queue, that is being used when the error happens. Also in the reference queue itself there is no visible locking, the locks that seem to cause the problem stem from our code itself. How about changing ManagedLinkedList to use a ConcurrentLinkedQueue inside? Then we will not have to synchronize on it anymore.

jwagenleitner:
>maybe we could add this to the performance tests part, yes

I will take a closer look at that sub-project and see if I can work in this test.

>But frankly I am surprised if your change solves the problem, since it looks to me as if it is the same reference queue, that is being used when the error happens

The new `ReferenceBundle#newWeakBundle` method creates a new `ReferenceQueue` on each call, so now the queue for the `GroovyClassValuePreJava7` and `GlobalClassSet` would be separate.  No longer would adding an element to the set possibly trigger a remove from the map.  Given that do you still believe this will not solve the problem?

>How about changing ManagedLinkedList to use a ConcurrentLinkedQueue inside? Then we will not have to synchronize on it anymore.

That sounds like a good idea, how about I add a todo comment for now and I can work on submitting a PR for that soon?  I would want to put some tests around it and that may take some time.  And I think it would be good to make sure this issue gets fixed for 2.4.9.

Only concern I could see is that removal time may be more.  Now the Element is a ManagedReference and it can just un-link itself, with the CLQ it would have to walk the list to find the ManagedReference (used to wrap the real value) to remove.  And the time it takes to remove elements can impact the time it takes to add elements when there are references to clean up.  Overall probably not significant given the better locking in CLQ for adds over having to synchronize as is done now.

blackdrag:
>>     But frankly I am surprised if your change solves the problem, since it looks to me as if it is the same reference queue, that is being used when the error happens
> 
> The new ReferenceBundle#newWeakBundle method creates a new ReferenceQueue on each call, so now the queue for the GroovyClassValuePreJava7 and GlobalClassSet would be separate. No longer would adding an element to the set possibly trigger a remove from the map. Given that do you still believe this will not solve the problem?

Even if you have one queue or GroovyClassValuePreJava7 and GlobalClassSet, you still have a ManagedLinkedList, which you add and remove elements from. And each time you add, you potentially remove. The only thing different to before is that the ReferenceQueue will not be shared with another ManagedLinkedList. But the lock in ClassInfo is on the list and the other lock is on GroovyClassValuePreJava7Segment. Those will stay. The only thing changing is that the remove is done with a different timing in my opinion. You would now for example need at least 500 ClassInfos before we start checking for orphaned values.
 
>     How about changing ManagedLinkedList to use a ConcurrentLinkedQueue inside? Then we will not have to synchronize on it anymore.
> 
> That sounds like a good idea, how about I add a todo comment for now and I can work on submitting a PR for that soon? I would want to put some tests around it and that may take some time. And I think it would be good to make sure this issue gets fixed for 2.4.9.

sure
 
> Only concern I could see is that removal time may be more. Now the Element is a ManagedReference and it can just un-link itself, with the CLQ it would have to walk the list to find the ManagedReference (used to wrap the real value) to remove. And the time it takes to remove elements can impact the time it takes to add elements when there are references to clean up. Overall probably not significant given the better locking in CLQ for adds over having to synchronize as is done now.

hmm... he alternative would be to change ManagedLinkedList to use a CAS style approach. I think that is not all that difficult to do for a linked list... but is of course source for new errors too.

jwagenleitner:
>The only thing different to before is that the ReferenceQueue will not be shared with another ManagedLinkedList. But the lock in ClassInfo is on the list and the other lock is on GroovyClassValuePreJava7Segment.

I believe the key difference is that the `ManagedLinkedList` used by `GlobalClassSet` no longer shares a `ReferenceQueue` with the `ManagedConcurrentMap` (subclass `GroovyClassValuePreJava7`).  It is ok for the list (which is synchronized) to remove it's own entries on add.  The problem occurred because by sharing a `ReferenceQueue` the list could also try to remove entries from the map.

>You would now for example need at least 500 ClassInfos before we start checking for orphaned values.

That is a good point and I hadn't considered the threshold setting.  Running `println org.codehaus.groovy.reflection.ClassInfo.size()` from the command line returns ~200, it jumps to ~900 if I run the same command in `GroovyConsole`.  Maybe a value of 200 would be a better threshold setting for `newWeakBundle`.

Though maybe none of it is needed if `ManagedLinkedList` is changed to a non-blocking CLQ.  I'm just not sure what new problems that might introduce and if the weakly consistent iterator would work ok for the `modifiedExpandos` `ManagedLinkedList` (I suspect it wouldn't be a problem).

jwagenleitner:
As suggested, I opened an alternative PR #489 that uses a `ManagedConcurrentLinkedQueue` to avoid the need to use synchronized blocks.

jwagenleitner:
Closing this in favor of PR #489

PR #485, opened 2017-01-30
Parrot for java7

jespersm:
I'm sorry for vandalising the source - but here's how it could work with Java 7, so we could get it into Groovy 2.x

### DIFF
--- a/subprojects/groovy-parser-antlr4/src/main/java/org/apache/groovy/parser/antlr4/TryWithResourcesASTTransformation.java
+++ b/subprojects/groovy-parser-antlr4/src/main/java/org/apache/groovy/parser/antlr4/TryWithResourcesASTTransformation.java
@@ -7,6 +7,7 @@ import org.codehaus.groovy.ast.expr.BinaryExpression;
 import org.codehaus.groovy.ast.expr.BooleanExpression;
 import org.codehaus.groovy.ast.expr.ConstantExpression;
 import org.codehaus.groovy.ast.expr.DeclarationExpression;
+import org.codehaus.groovy.ast.expr.Expression;
 import org.codehaus.groovy.ast.expr.MethodCallExpression;
 import org.codehaus.groovy.ast.expr.VariableExpression;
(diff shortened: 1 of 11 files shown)


subprojects/groovy-parser-antlr4/src/main/java/org/apache/groovy/parser/antlr4/TryWithResourcesASTTransformation.java
```diff
@@ -342,7 +348,7 @@ private ExpressionStatement createAddSuppressedStatement(String primaryExcName,
                 new MethodCallExpression(
                         new VariableExpression(primaryExcName),
                         "addSuppressed",
-                        new ArgumentListExpression(Collections.singletonList(new VariableExpression(suppressedExcName))));
+                        new ArgumentListExpression(Collections.singletonList((Expression) new VariableExpression(suppressedExcName))));
```
danielsun1106: Casting the type to "Expression" seems unnecessary ;)

jespersm: True! But the type inference on Java 7 needs this, or you'll get a type mismatch, since a List<VariableExpression> is not the same as List<Expression>. On Java 8, the type inference can solve this by itself, by checking the supertypes of the generic argument to singletonList.

I found 5-6 of these kind of required change.

danielsun1106: Gotcha.
P.S.  I deleted the comment "unnecessary casting type" by accident... it's hard to edit comments on my cell phone.

danielsun1106:
@blackdrag @paulk-asert 
I've reviewed the PR. +1 from me to merge it as soon as possible :)

paulk-asert:
Merged into the GROOVY_2_6_X branch. Thanks! Great work!

PR #487, opened 2017-02-02
GROOVY-8069: Annotation based logging of complex expression in closur…

shils:
…e throws GroovyCastException with SC.


paulk-asert:
+1

PR #488, opened 2017-02-03
GROOVY-8030: An unexpected error has occurred when using 'kkk = kkk +…

paulk-asert:
… 12'


blackdrag:
first of all I think I can give this a +1, but the code makes me now wonder if we really do the right thing with flow typing here. I would not have expected the information if kkk is int or String on the accessed variable, which should be the definition. I remember I had an unsatisfying discussion about this with Cedric in the past already. But changing that is out of scope for this PR in my opinion and maybe it is just me not understanding something too

paulk-asert:
Hi Jochen, yep agree that other parts need to be further investigated to make sure we have the correct behavior - I think it is wrong in some cases too. But my attempts to change things in the StaticTypeCheckingVisitor seemed to be impacting too many things so this was the minimal change to fix the current issue.

PR #489, opened 2017-02-05
GROOVY-8067: Possible deadlock when creating new ClassInfo entries in the cache

jwagenleitner:
As suggested in PR #484 removed the locking on the `ManagedLinkedList` by creating a new `ManagedConcurrentLinkedQueue`.

Also added a `stress` subproject for tests that employ many threads, need GC, or just in general try to break things and take a long time.  These require a special property to be set in order to run, otherwise they will be skipped.  I tried to work it out in the `performance` subproject, but that seems to be very specialized for the compiler tests.  Open to suggestions on a better way to handle these types of tests.

### DIFF
--- a/src/main/org/codehaus/groovy/reflection/ClassInfo.java
+++ b/src/main/org/codehaus/groovy/reflection/ClassInfo.java
@@ -74,7 +74,8 @@ public class ClassInfo implements Finalizable {
     private static final ReferenceBundle softBundle = ReferenceBundle.getSoftBundle();
     private static final ReferenceBundle weakBundle = ReferenceBundle.getWeakBundle();
     
-    private static final ManagedLinkedList<ClassInfo> modifiedExpandos = new ManagedLinkedList<ClassInfo>(weakBundle);
+    private static final ManagedConcurrentLinkedQueue<ClassInfo> modifiedExpandos =
+            new ManagedConcurrentLinkedQueue<ClassInfo>(weakBundle);
 
     private static final GroovyClassValue<ClassInfo> globalClassValue = GroovyClassValueFactory.createGroovyClassValue(new ComputeValue<ClassInfo>(){
 		@Override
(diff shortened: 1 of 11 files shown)


src/main/org/codehaus/groovy/reflection/ClassInfo.java
```diff
@@ -186,30 +185,20 @@ public void setStrongMetaClass(MetaClass answer) {
...
-          ((ExpandoMetaClass)answer).inRegistry = true;
-          synchronized(modifiedExpandos){
-            for (Iterator<ClassInfo> it = modifiedExpandos.iterator(); it.hasNext(); ) {
-              ClassInfo info = it.next();
-                if(info == this){
-                  it.remove();
-                }
-             }
-             modifiedExpandos.add(this);
-          }
```
jwagenleitner: Removed because it appeared to be a duplicate of the logic above.  Only if the previous previous `strongMetaClass` value was an expando would it have been added to the map, so the iteration/removal above should have taken care to remove this `ClassInfo` from the `modifiedExpandos` and there's no reason to iterate again.

blackdrag:
that looks very promising to me +1

src/main/org/codehaus/groovy/util/ManagedConcurrentLinkedQueue.java
```diff
@@ -0,0 +1,187 @@
...
+
+        @Override
+        public void finalizeReference() {
+            queue.remove(this);
+            super.finalizeReference();
+        }
+
+    }
+
+    private final class Iter implements Iterator<T> {
```
jwagenleitner: think this can be made a static nested class

subprojects/stress/src/test/java/org/codehaus/groovy/reflection/ClassInfoDeadlockStressTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+    private void attemptDeadlock(final Class<?> cls) throws Exception {
+        for (int i = 0; i < THREAD_COUNT; i++) {
+            Runnable runnable = new Runnable() {
+                @Override
+                public void run() {
+                    Class<?> newClass = (cls == null) ? createRandomClass() : cls;
+                    try {
+                        startLatch.await();
+                    } catch (InterruptedException ie) {
```
jwagenleitner: can use `ThreadUtils.awaite()`

jwagenleitner:
@blackdrag @paulk-asert do you have any reservations about adding a new sub-project for the manual/stress tests?  I tried to work them into `performance` but it currently doesn't include the root project in the test dependencies.  It also depends on Java 1.8 and I think it would be good to be able to run these tests in the min supported version 1.7.

blackdrag:
no reservations on my side

PR #491, opened 2017-02-07
GROOVY-8068: improper logging in groovy.sql.Sql

paulk-asert:


subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
```diff
@@ -578,17 +578,26 @@ public static Sql newInstance(Map<String, Object> args) throws SQLException, Cla
...
+        LOG.fine("url = " + url);
         if (props != null) {
-            System.err.println("url = " + url);
-            System.err.println("props = " + props);
-            connection = DriverManager.getConnection(url.toString(), new Properties(props));
+            Properties propsCopy = new Properties(props);
+            connection = DriverManager.getConnection(url.toString(), propsCopy);
+            if (propsCopy.containsKey("password")) {
+                // don't log the password
+                propsCopy = new Properties(propsCopy);
```
jwagenleitner: I don't think this is needed since it's already been copied?

jwagenleitner:
+1 with just one minor comment

subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
paulk-asert: It is certainly conservative code. I didn't want the password appearing in logging and I didn't know whether any drivers out there might hold onto the properties object and re-use it when e.g. growing a connection pool or something.

jwagenleitner: That makes sense, I missed that the copy was passed to the DriverManager.  Sorry for the noise.

PR #494, opened 2017-02-09
Safe number parsing methods for more convenient XML parsing with GPat…

mortensoby:
The following results in af NumberFormatException:
xml.'**'.find{it.name()` == 'someBigDecimal'}.toBigDecimal()
if the element content is empty which means the following construct is necessary:

```
xml.'**'.find{it.name() == 'someEmptyBigDecimal' && it.text()}?.toBigDecimal()
```

This pull request ensures that null is returned if the element is empty for the following methods:
```
toBigInteger()
toBigDecimal()
toDouble()
toBigLong()
toBigFloat()
```


danielsun1106:
Merged. Thanks!

PR #500, opened 2017-02-19
GROOVY-8056: GroovyCodeSource(URL) can leak a file handler

jwagenleitner:
URLConnect.getContentEncoding returns the Content-Encoding
HTTP Header [1] which is not a charset.  Since this method would
have either returned null or an invalid charset, the code path
specifying the encoding would normally not have been executed.
The charset may be contained in the Content-Type header, but
rather than attempt to parse that string which would require
closing the connection, this fix avoids opening the connection
and relies on the default charset.

[1] https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.11

### DIFF
--- a/src/main/groovy/lang/GroovyCodeSource.java
+++ b/src/main/groovy/lang/GroovyCodeSource.java
@@ -173,12 +173,7 @@ public class GroovyCodeSource {
         this.name = url.toExternalForm();
         this.codeSource = new CodeSource(url, (java.security.cert.Certificate[]) null);
         try {
-            String contentEncoding = url.openConnection().getContentEncoding();
-            if (contentEncoding != null) {
-                this.scriptText = ResourceGroovyMethods.getText(url, contentEncoding);
-            } else {
-                this.scriptText = ResourceGroovyMethods.getText(url); // falls-back on default encoding
-            }
+            this.scriptText = ResourceGroovyMethods.getText(url); // default encoding
         } catch (IOException e) {
             throw new RuntimeException("Impossible to read the text content from " + name, e);
         }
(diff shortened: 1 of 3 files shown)


jwagenleitner:
The rationale behind not attempting to actually get the charset (if available) from the URLConnection is because it hasn't been an issue to date.  So rather than introduce the complexity of parsing the `Content-Type` header it seems reasonable to defer until someone actually reports an issue.  The original issue that introduced the code [GROOVY-3934](https://issues.apache.org/jira/browse/GROOVY-3934) was not an issue about charset.

jwagenleitner:
Travis is failing on the JDK8 build which I think is unrelated to this PR and seems to have started recently when the gradle version was bumped to 3.3.  Tests pass locally for me.

danielsun1106:
@jwagenleitner Java 8 requires more memory, we can request sudo to gain more system resource of travis-ci instance: https://github.com/apache/groovy/commit/c79ef79f87f68ea29b3f32b456f1b9ac286bb8c8

jwagenleitner:
I rebased to pick up the latest travisci change.

jwagenleitner:
An example to show contentEncoding vs contentType

```groovy
url = 'https://raw.githubusercontent.com/jwagenleitner/groovy/a0e3335ab1d46fbe7ca7a0391fdf864c12d15f01/src/test/groovy/util/GroovyScriptEngineReloadingTest.groovy'

uc = new URL(url).openConnection()
uc.requestMethod = 'GET'

assert uc.contentEncoding == null
assert uc.contentType.toLowerCase().contains('utf-8')

uc = new URL(url).openConnection()
uc.requestMethod = 'GET'
uc.setRequestProperty('accept-encoding', 'gzip')

assert uc.contentEncoding == 'gzip'
assert uc.contentType.toLowerCase().contains('utf-8')
```

jwagenleitner:
@blackdrag @paulk-asert if either of you have a chance I would be curious to know your opinion about this proposed change/fix.

jwagenleitner:
Closing this in favor of PR #557 just in case someone is relying on the old behavior.

PR #502, opened 2017-02-22
GROOVY-8096 setScriptBaseClass with Java class breaks @Field Binding init due to call to super() instead of super(Binding)

kreiger:
This test fails because `ClassNode.getSuperClass().getDeclaredConstructor(SCRIPT_CONTEXT_CTOR)` doesn't see the constructors in the Java base class.

`ModuleNode.setScriptBaseClassFromConfig(ClassNode)`
calls `.setSuperClass(ClassHelper.make(baseClassName))` on the `scriptDummy` `ClassNode`.

 The `ClassNode` created for this script's base class has `.lazyInitDone = true` and `.constructors = null` so when `.getSuperClass().getDeclaredConstructor(SCRIPT_CONTEXT_CTOR)` is called by `ModuleNode.createStatementsClass()`, then `ClassNode.constructors` is set to an empty ArrayList in `ClassNode.getDeclaredConstructors()`

The script constructor is then generated as

         Constructor(Binding context) {
             super();                   // Fields are initialized after the call to super()
                                        // Fields are initialized here
             setBinding(context);       // Fields are initialized before the call to setBinding(context)
         }

instead of

         Constructor(Binding context) {
             super(context);            // Fields are initialized after the call to super(context)
         }

Fields are initialized between the call to super() and the setBinding(context)
which means Field initializers don't have access to the Binding context.

This leads to `MissingPropertyException` because we're trying to look up variables from the `new Binding()` created in the default constructor, instead of the binding we passed in.

### DIFF
--- a/src/test/groovy/lang/BindingScript.java
+++ b/src/test/groovy/lang/BindingScript.java
@@ -1,0 +1,32 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package groovy.lang;
+
+/**
+ * A Script which requires a Binding passed in the constructor and disallows calling the default constructor.
+ */
+public abstract class BindingScript extends Script {
+    protected BindingScript() {
+        throw new UnsupportedOperationException("\n\t*******\n\tBindingScript() should not be called! Should be calling BindingScript(Binding)!\n\t*******");
+    }
+
+    protected BindingScript(Binding binding) {
+        super(binding);
+    }
+}
--- a/src/test/groovy/lang/GroovyShellTest2.groovy
+++ b/src/test/groovy/lang/GroovyShellTest2.groovy
@@ -43,6 +43,60 @@ class GroovyShellTest2 extends GroovyTestCase {
         assert result == arg0
     }
 
+    void testBindingsInFieldInitializersWithAnnotatedJavaBaseScript() {
+        def shell = new GroovyShell();
+        def scriptText = '''
+        @groovy.transform.BaseScript groovy.lang.BindingScript baseScript
+        @groovy.transform.Field def script_args = getProperty('args')
+
+        assert script_args[0] == 'Hello Groovy'
+        script_args[0]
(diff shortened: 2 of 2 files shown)


src/test/groovy/lang/BindingScript.java
```diff
@@ -0,0 +1,32 @@
...
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package groovy.lang;
+
+/**
+ * A Script which requires a Binding passed in the constructor and disallows calling the default constructor.
+ */
+public abstract class BindingScript extends Script {
+    protected BindingScript() {
```
danielsun1106: How about make the default constructor `private`? Developers can get compilation error other than runtime error, then the program will be more robust IMO.

kreiger: Unfortunately the exact same `super()` call is generated, because `classNode.getSuperClass().getDeclaredConstructor(SCRIPT_CONTEXT_CTOR)` still returns null in `ModuleNode.createStatementsClass()` even though that constructor exists.

If i make the default constructor `private` i get `IllegalAccessError`:

    java.lang.reflect.InvocationTargetException
        at TestBindingsInFieldInitializersWithConfigJavaBaseScript.main(TestBindingsInFieldInitializersWithConfigJavaBaseScript.groovy)
    Caused by: java.lang.IllegalAccessError: tried to access method groovy.lang.BindingScript.<init>()V from class TestBindingsInFieldInitializersWithConfigJavaBaseScript
	    at TestBindingsInFieldInitializersWithConfigJavaBaseScript.<init>(TestBindingsInFieldInitializersWithConfigJavaBaseScript.groovy)

If i remove the default constructor i get `NoSuchMethodError`:

    java.lang.reflect.InvocationTargetException
        at TestBindingsInFieldInitializersWithConfigJavaBaseScript.main(TestBindingsInFieldInitializersWithConfigJavaBaseScript.groovy)
    Caused by: java.lang.NoSuchMethodError: groovy.lang.BindingScript: method <init>()V not found
	    at TestBindingsInFieldInitializersWithConfigJavaBaseScript.<init>(TestBindingsInFieldInitializersWithConfigJavaBaseScript.groovy)

kreiger:
Created JIRA ticket at https://issues.apache.org/jira/browse/GROOVY-8096

src/test/groovy/lang/BindingScript.java
kreiger: If i remove both constructors the `super()` call just goes to the default constructor in groovy.lang.Script, giving `MissingPropertyException` on trying to access the wrong `Binding context` on field initalization:

    groovy.lang.MissingPropertyException: No such property: args for class: TestBindingsInFieldInitializersWithConfigJavaBaseScript
 	    at TestBindingsInFieldInitializersWithConfigJavaBaseScript.<init>(TestBindingsInFieldInitializersWithConfigJavaBaseScript.groovy:2)
	    at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	    at groovy.lang.GroovyShell.runScriptOrMainOrTestOrRunnable(GroovyShell.java:271)

paulk-asert:
Just to clarify - this isn't a PR designed to fix a problem (although BindingScript might prove useful in a solution) but to illustrate a problem. Correct?

kreiger:
@paulk-asert  Correct, this just adds a couple of unit tests to illustrate the problem. I hope you agree that it's better than just describing it. :)

I was able to make a patch myself where i use Class.forName() to look up the class, but i highly doubt it is correct.

PR #503, opened 2017-02-22
Provide logging for bytecode generation

danielsun1106:
## Background
Debugging bytecode issues is a hard work, one of reasons is lack of logging.

## Solution
The PR provides LoggableClassVisitor and LoggableTextifier to make debugging easy. We can debug **step by step** and investigate bytecode generation **line by line**. Logging bytecode generation can be enabled by `-Dgroovy.log.classgen=true`

### DIFF
--- a/src/main/org/codehaus/groovy/classgen/AsmClassGenerator.java
+++ b/src/main/org/codehaus/groovy/classgen/AsmClassGenerator.java
@@ -95,7 +95,7 @@ import java.util.Map;
  */
 public class AsmClassGenerator extends ClassGenerator {
 
-    private final ClassVisitor cv;
+    private ClassVisitor cv;
     private final GeneratorContext context;
     private final String sourceFile;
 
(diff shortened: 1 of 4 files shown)


danielsun1106:
Here is the JIRA issue to track the PR: https://issues.apache.org/jira/browse/GROOVY-8095

danielsun1106:
@paulk-asert Could you review the PR?

danielsun1106:
@paulk-asert ping :-)

danielsun1106:
@blackdrag Jochen, could you set aside some time to review the PR?

paulk-asert:
I am leaning towards +1 but the two things I was planning to look into are:
1. Do we open ourselves up (any worse than now) to any security vulnerabilities if an AST transform could somehow inject a malicious controller. Possibly nothing to worry about, but I just haven't thought much about it yet.
1. Would it make sense to use the `LoggableTextifier` in the GroovyConsole. The answer might be no, I just haven't thought much about it yet.

blackdrag:
@paulk-asert for (1) I think no; for (2) I would say that depends on our working process. I rarely investigate a bytecode issue in the groovyConsole, thus it would get me nothing there. But maybe Daniel does?

@danielsun1106 looks basically ok to me, but in the past I had only limited use for something like this. In some cases I did change the visitors to print the bytecode to a file, in other cases I added also LDC/POP combinations to add markers in the bytecode based on the part of ACG I was in at that moment. In other cases I was simply working with bytecode test cases or simple classes I compile.

danielsun1106:
https://github.com/apache/groovy/commit/ab5defa3c1bfbf0c993394d11bb0f1e655283acc

PR #507, opened 2017-03-02
Fix wildcard arg handling in startGroovy.bat

AarjavP:
Full disclosure: I don't fully understand this fix but it seems to be working better for me.
From answer here: http://stackoverflow.com/a/12772969/5130921
The problem I had was similar, I was passing `-cp` `somedir\*` as arguments, and I was getting the same message `The syntax of the command is incorrect`

Also not sure if this is related but I also cannot pass in these arguments without them being changed: 
`-Dlog4j.configuration="file:simple/path/to.properties"`
However, `-Dlog4j.configuration=file:simple/path/to.properties` works fine (just quotes removed)
The proposed change does not fix the above error, just the `*` thing.

### DIFF
--- a/src/bin/startGroovy.bat
+++ b/src/bin/startGroovy.bat
@@ -178,7 +178,7 @@ goto process_arg
 
 :get_arg
 rem remove quotes around first arg
-for %%i in (%1) do set _ARG=%_ARG% %%~i
+for /f %%i in (%1) do set _ARG=%_ARG% %%~i
 rem set the remaining args
 set _ARGS=%2
 rem remove the leading space we'll add the first time


paulk-asert:
I believe ages ago windows "globbing" worked to some degree. If you did:
```groovy -e "println args" *.groovy```
then windows would replace *.groovy with all the groovy filenames and pass all of them to your script (which in this case would just print out the names as a list).

This saved you having to worry about doing the expansion yourself - which at the time wasn't particularly easy outside of `ant`. I think that capability broke a long time ago and now only the first file found is returned or the error `The syntax of the command is incorrect` is returned if no matching file is found. I believe adding the `/F` will pass through the asterisk. This is a breaking change but in reality is probably what we want to do. We just need to document it and have some tests around it.

Similar suggestions have been made before but we haven't applied them because we never got the tests and doco in place. It would be great if you think you have time to flesh out either or both of those things but otherwise we'll add it to the long todo list.

The -D handling is also known about - our general strategy is that we want to reduce any processing happening in the bat file and have more handled in Groovy - but again we want some tests in place before making drastic changes.

We also need to get CI agents running windows that will then test any created tests - but that is a separate activity. If you do think you have time to do any additional work, I am happy to provide assistance.

paulk-asert:
This was merged with an identical commit. Thanks.

PR #513, opened 2017-03-14
GROOVY-8046: ClassFormatError void field

paulk-asert:


jwagenleitner:
Should there be a similar checks for local variable declarations and parameters?

paulk-asert:
Good point. I'll add those.

PR #514, opened 2017-03-19
Adds documentation around 'remove' method Integer/Object ambiguity

VoltiSubito:
I'm decently new to Groovy and was recently caught off-guard by attempting to use the `remove` method on a list of Integers, so I thought I'd try to update the document to add `removeAt` and `removeElement` to that section.


jwagenleitner:
Merged, thanks.

PR #515, opened 2017-03-22
Fix running out of memory ("GC overhead limit exceeded") build proble…

kyegupov:
…m that happens during findBugs analysis (I've encountered that personally - findBugs was definitely exhausting 1GB heap)


blackdrag:
I think it is save to do this: +1

jwagenleitner:
Merged, thanks.

PR #516, opened 2017-03-22
More useful error message in NodeList.replaceNode method

kyegupov:
I ran into this unhelpful message when using one of plugins for Gradle that patched config files. Knowing the actual number of nodes would help to identify the issue.


blackdrag:
+1

jwagenleitner:
PR merged, thanks.

PR #517, opened 2017-03-23
Differentiate between XML breadthFirst() traversal and * navigation

manouti:
The analogy between '*' and breadthFirst() traversal seems wrong. The former only traverses one level, while the latter searches through the next levels as well. See http://stackoverflow.com/questions/42985716/groovy-xml-tree-traversal-breadthfirst-method-using-as-syntactic-sugar:

    def books = '''\
    <response>
       <books>
          <book available="20" id="1">
             <title>foo</title>
             <author id="1">foo author</author>
          </book>
          <book available="14" id="2">
             <title>bar</title>
             <author id="2">bar author</author>
          </book>
       </books>
    </response>'''

    def response = new XmlSlurper().parseText(books)
    def bk = response.'*'.find { node ->
       node.name() == 'book' && node['@id'].toInteger() == 2
    }
    assert bk.empty

> whereas using breadthFirst() explicitly does what I expect both to do which is to do breadth-first traversal:

    def books = '''\
    <response>
       <books>
          <book available="20" id="1">
             <title>foo</title>
             <author id="1">foo author</author>
          </book>
          <book available="14" id="2">
             <title>bar</title>
             <author id="2">bar author</author>
          </book>
       </books>
    </response>'''

    def response = new XmlSlurper().parseText(books)
    def bk = response.breadthFirst().find { node ->
       node.name() == 'book' && node['@id'].toInteger() == 2
    }
    assert bk.title == 'bar' // bk is no longer an empty list of children

### DIFF
--- a/subprojects/groovy-xml/src/spec/doc/xml-userguide.adoc
+++ b/subprojects/groovy-xml/src/spec/doc/xml-userguide.adoc
@@ -219,47 +219,43 @@ the
 
 Both of them are equally valid.
 
-=== Speed things up with breadthFirst and depthFirst
+=== Speed things up with * and ** navigation
 
 If you ever have used XPath you may have used expressions like
 
 * `//` : Look everywhere
 * `/following-sibling::othernode` : Look for a node "othernode" in the same level
 
-More or less we have their counterparts in `GPath` with the methods
-`breadthFirst()` and `depthFirst()`.
+More or less we have their counterparts in `GPath` with the shortcuts `*` and `**`.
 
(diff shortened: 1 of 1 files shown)


paulk-asert:
The shortcut for `depthFirst()` is '**' while the shortcut for `children()` is '*'. Obviously what is there is not quite right. I'll check shortly whether I think your changes capture everything that we need.

paulk-asert:
I merged your changes with minor tweaks. Thanks, much appreciated!

PR #521, opened 2017-04-06
GROOVY-7879 Groovy calls wrong method if there is a static method on …

dkhamitov:
Hi guys, can we consider a small kind of tactical fix (imho) to GROOVY-7879 or is it a bad idea/solution? It fixes the situation when we have:
```java
interface Foo {
    static String foo() {
        return "I'm Foo";
    }
}

class Bar implements Foo {
    static String foo() {
        return "I'm Bar";
    }
}
```
but assert fails:
```java
assert Bar.foo() == "I'm Bar" //Output: I'm Foo
```
This pull request is for the parrot branch because static methods on interfaces were introduced in Java 8


paulk-asert:
At first glance it seems okay - I'll look in more detail when I am back from holidays - unless someone beats me to it. I would however suggest test coverage for @CompileStatic.

dkhamitov:
Ok, I'll do it

danielsun1106:
@dkhamitov
FYI, `apache:parrot` is for experimenting new grammar, so it's better to submit your PR merging into other branch(e.g. `GROOVY_2_5_X`). Thanks for your contribution :-)

dkhamitov:
Yes, it would be great if we had this in GROOVY_2_5_X but this branch compiles for Java 7:
```gradle
sourceCompatibility = 1.7
targetCompatibility = 1.7
```
It works if I take this version and use it with Java 8 and static methods on interfaces. But how can we deal with unit tests in this branch? Interfaces with static methods just don’t compile in Java 7. Any idea?

dkhamitov:
@paulk-asert @danielsun1106 Tests for \@CompileStatic added. I also noticed interesting behaviour:
```java
interface Foo {
  static void foo() {
    System.out.println("static void foo() called");
  }
}

class Bar implements Foo { }

Bar.foo() \\works in Groovy (dynamic) but fails in Java and Groovy (@CompileStatic) 
```
as per JLS8 - 8.4.8 Inheritance, Overriding, and Hiding

paulk-asert:
Merged, thanks! I wasn't sure about your last comment. Did you mean you get an error in an all Java 8 scenario? What error message?

dkhamitov:
@paulk-asert Cool. About my previous comment. I've rechecked that case when a class doesn't hide static methods on the implemented interface and we try to call those methods on the class (not on the interface). In this situation Java just doesn't compile:
```java
interface Foo {
  static void foo() {
    System.out.println("static void foo() called");
  }
}

class Bar implements Foo { }

public class FooBarTest {
        Bar.foo();
    }
}
```
```bash
$ javac FooBarTest.java
FooBarTest.java:3: error: cannot find symbol
        Bar.foo();
           ^
  symbol:   method foo()
  location: class Bar
1 error
```
But Groovy does compile, both dynamic and \@CompileStatic. So I was wrong about \@CompileStatic point and here is a couple of questions:
1) Shouldn't \@CompileStatic behave like Java in this case?
2) Should we fix this? I could investigate how it can be done

Just in case, here is a snippet from *JLS8*:
**8.4.8 Inheritance, Overriding, and Hiding**
A class C inherits from its direct superclass all concrete methods m (both **static** and instance) of the superclass for which all of the following are true:
...
A class does not inherit **static** methods from its superinterfaces.

paulk-asert:
I suspect that should be treated as an error. I think it's worth another
Jira.

dkhamitov:
@paulk-asert ok, I'll raise a jira and look into it

dkhamitov:
https://issues.apache.org/jira/browse/GROOVY-8164 raised. I'm looking into it

PR #523, opened 2017-04-12
cleanup now that jdk7 is baseline

jwagenleitner:
Avoids reflective checks/constructions where it is no longer needed since 7 is now baseline in 2_5_X and master.

### DIFF
--- a/gradle/indy.gradle
+++ b/gradle/indy.gradle
@@ -21,14 +21,7 @@
 // with regards to invoke dynamic support (indy)
 
 rootProject.ext.indyCapable = {
-    boolean capable = true
-    try {
-        Class.forName('java.lang.invoke.MethodHandle')
-    } catch (e) {
-        capable = false
-    }
-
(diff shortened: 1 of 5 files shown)


src/main/org/codehaus/groovy/vmplugin/VMPluginFactory.java
```diff
@@ -26,39 +28,10 @@
...
-            plugin = createPlugin(JDK5_CLASSNAME_CHECK, JDK5_PLUGIN_NAME);
-        }
-    }
+    private static final VMPlugin plugin = new Java7();
 
     public static VMPlugin getPlugin() {
         return plugin;
     }
 
-    private static VMPlugin createPlugin(String classNameCheck, String pluginName) {
```
blackdrag: we will still need this for the JDK9 plugin

paulk-asert:
+1 for GROOVY_2_5_X and master

PR #526, opened 2017-04-21
cleanup now that jdk7 is baseline

PascalSchumacher:
for master and for the GROOVY_2_5_X and GROOVY_2_6_X branches


jwagenleitner:
+1

PR #527, opened 2017-04-23
remove skipping of tests for builds.apache.org as we no longer use it

PascalSchumacher:
even if groovy decides to use builds.apache.org again, these workarounds may not be necessary today


jwagenleitner:
+1

paulk-asert:
I am +0 on this change. I think we want to get builds.apache.org working. It's just that no-one has had time to get it working.

PascalSchumacher:
I Paul I understand that was the goal, but it did not happen during the last two years.

When I added these workarounds one and a half year ago is was working most of the time, but it was unreliable enough to be useless.

If/when groovy use build.apache.org again almost everything (jdks, groovy codebase, vms/servers, operating systems) will have changed even more, so I doubt that the same workarounds are necessary.

PR #532, opened 2017-05-01
Prevent CachedField and CachedMethod from leaking access permissions …

dpolivaev:
…to scripts

https://issues.apache.org/jira/browse/GROOVY-8163

### DIFF
--- a/src/main/groovy/lang/MetaClassImpl.java
+++ b/src/main/groovy/lang/MetaClassImpl.java
@@ -1832,6 +1832,9 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
             } catch (IllegalArgumentException e) {
                 // can't access the field directly but there may be a getter
                 mp = null;
+            } catch (CacheAccessControlException e) {
+                // can't access the field directly but there may be a getter
+                mp = null;
             }
         }
 
--- a/src/main/org/codehaus/groovy/reflection/AccessPermissionChecker.java
+++ b/src/main/org/codehaus/groovy/reflection/AccessPermissionChecker.java
@@ -1,0 +1,83 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.codehaus.groovy.reflection;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.ReflectPermission;
+import java.security.AccessControlException;
+
+import groovy.lang.GroovyObject;
+
+class AccessPermissionChecker {
+
+    private static final ReflectPermission REFLECT_PERMISSION = new ReflectPermission("suppressAccessChecks");
+
+    private static void checkAccessPermission(Class<?> declaringClass, final int modifiers, boolean isAccessible) {
+        final SecurityManager securityManager = System.getSecurityManager();
+        if (securityManager != null && isAccessible) {
+                if (((modifiers & Modifier.PRIVATE) != 0
+                        || ((modifiers & (Modifier.PUBLIC | Modifier.PROTECTED)) == 0
+                             && packageCanNotBeAddedAnotherClass(declaringClass)))
+                        && !GroovyObject.class.isAssignableFrom(declaringClass)) {
+                        securityManager.checkPermission(REFLECT_PERMISSION);
+                }
+                else if ((modifiers & (Modifier.PROTECTED)) != 0
+                    && declaringClass.equals(ClassLoader.class)){
+                    securityManager.checkCreateClassLoader();
+                }
+        }
+    }
+
+    private static boolean packageCanNotBeAddedAnotherClass(Class<?> declaringClass) {
+        return declaringClass.getName().startsWith("java.");
+    }
+
+    static void checkAccessPermission(Method method) {
+        try {
+            checkAccessPermission(method.getDeclaringClass(), method.getModifiers(), method.isAccessible());
+        } catch (AccessControlException e) {
+            throw createCacheAccessControlExceptionOf(method, e);
+        }
+    }
+
+    private static CacheAccessControlException createCacheAccessControlExceptionOf(Method method, AccessControlException e) {
+        return new CacheAccessControlException(
+                "Groovy object can not access method " + method.getName()  + " cacheAccessControlExceptionOf class " + method.getDeclaringClass().getName()
+                        + " with modifiers \"" + Modifier.toString(method.getModifiers()) + "\"", e);
+    }
+
+    static void checkAccessPermission(Field field) {
+        try {
+            checkAccessPermission(field.getDeclaringClass(), field.getModifiers(), field.isAccessible());
+        } catch (AccessControlException e) {
+            throw createCacheAccessControlExceptionOf(field, e);
+        }
+    }
+
+    private static CacheAccessControlException createCacheAccessControlExceptionOf(Field field, AccessControlException e) {
+        return new CacheAccessControlException(
+                "Groovy object can not access field " + field.getName()  + " cacheAccessControlExceptionOf class " + field.getDeclaringClass().getName()
+                        + " with modifiers \"" + Modifier.toString(field.getModifiers()) + "\"", e);
+    }
+
+    private AccessPermissionChecker() {}
+
+}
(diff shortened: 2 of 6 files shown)


src/main/org/codehaus/groovy/reflection/AccessPermissionChecker.java
```diff
@@ -0,0 +1,62 @@
...
+package org.codehaus.groovy.reflection;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.lang.reflect.ReflectPermission;
+
+import groovy.lang.GroovyObject;
+
+class AccessPermissionChecker {
```
dpolivaev: done

src/main/org/codehaus/groovy/reflection/AccessPermissionChecker.java
```diff
@@ -0,0 +1,62 @@
...
+
+	private static final ReflectPermission REFLECT_PERMISSION = new ReflectPermission("suppressAccessChecks");
+
+	static private void checkAccessPermission(Class<?> declaringClass, final int modifiers, boolean isAccessible) {
+		final SecurityManager securityManager = System.getSecurityManager();
```
dpolivaev: Signature changed. The check is performed only if (securityManager != null && isAccessible)

src/main/org/codehaus/groovy/reflection/AccessPermissionChecker.java
```diff
@@ -0,0 +1,62 @@
...
+					securityManager.checkCreateClassLoader();
+				}
+		}
+	}
+
+	private static boolean packageCanNotBeAddedAnotherClass(Class<?> declaringClass) {
+		return declaringClass.getName().startsWith("java.");
+	}
+
+	static public void checkAccessPermission(Method method) {
```
dpolivaev: fixed

src/main/org/codehaus/groovy/reflection/AccessPermissionChecker.java
```diff
@@ -0,0 +1,62 @@
...
+		checkAccessPermission(method.getDeclaringClass(), method.getModifiers(), method.isAccessible()
+		);
+	}
+
+	public static void checkAccessPermission(Field field) {
```
dpolivaev: fixed

src/main/org/codehaus/groovy/reflection/CachedField.java
```diff
@@ -51,6 +52,12 @@ public int getModifiers() {
      */
     public Object getProperty(final Object object) {
         try {
+            AccessPermissionChecker.checkAccessPermission(field);
+        }
+        catch (AccessControlException ex) {
+            throw new IllegalArgumentException("Illegal access to field" + " " + field.getName());
```
dpolivaev: To use GroovyRuntimeException we need to patch also MetaClassImpl.getProperty . It was not possible with byte buddy, but as a groovy patch it is possible. I shall do it in a separate commit.

src/main/org/codehaus/groovy/reflection/CachedField.java
```diff
@@ -65,6 +72,12 @@ public Object getProperty(final Object object) {
      * @throws RuntimeException if the property could not be set
      */
     public void setProperty(final Object object, Object newValue) {
+        try {
+            AccessPermissionChecker.checkAccessPermission(field);
+        }
+        catch (AccessControlException ex) {
+            throw new IllegalArgumentException("Illegal access to field" + " " + field.getName());
```
dpolivaev: As above

src/main/org/codehaus/groovy/reflection/CachedMethod.java
```diff
@@ -90,6 +91,12 @@ public CachedClass getDeclaringClass() {
 
     public final Object invoke(Object object, Object[] arguments) {
         try {
+            AccessPermissionChecker.checkAccessPermission(cachedMethod);
+        }
+        catch (AccessControlException ex) {
+            throw new InvokerInvocationException(new IllegalArgumentException("Illegal access to method" + cachedMethod.getName()));
```
dpolivaev: As above

src/main/org/codehaus/groovy/reflection/CachedMethod.java
```diff
@@ -124,6 +131,12 @@ public String getSignature() {
     }
 
     public final Method setAccessible() {
+        try {
+            AccessPermissionChecker.checkAccessPermission(cachedMethod);
+        }
+        catch (AccessControlException ex) {
+            throw new IllegalArgumentException("Illegal access to method" + cachedMethod.getName());
```
dpolivaev: As above

src/main/org/codehaus/groovy/reflection/CachedMethod.java
```diff
@@ -324,6 +337,12 @@ else if (o2 instanceof CachedMethod)
     }
 
     public Method getCachedMethod() {
+        try {
+            AccessPermissionChecker.checkAccessPermission(cachedMethod);
+        }
+        catch (AccessControlException ex) {
+            throw new IllegalArgumentException("Illegal access to method" + cachedMethod.getName());
```
dpolivaev: As above

dpolivaev:
Changes suggested in the review are implemented.
See also https://issues.apache.org/jira/browse/GROOVY-8163?focusedCommentId=16009695&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16009695

src/main/groovy/lang/MetaClassImpl.java
```diff
@@ -1832,6 +1832,9 @@ public Object getProperty(Class sender, Object object, String name, boolean useS
             } catch (IllegalArgumentException e) {
                 // can't access the field directly but there may be a getter
                 mp = null;
+            } catch (GroovyRuntimeException e) {
+                // can't access the field directly but there may be a getter
+                mp = null;
```
jwagenleitner: I don't understand why this was added.  Was there a test that failed because of the other changes that required this.  I would have assumed an `AccessControlException` from `CachedField.getProperty` would be treated similar to the `IllegalAccessException` in the same method.

dpolivaev: I do not have unit test to explain this catch block. I do have the integration test https://issues.apache.org/jira/browse/GROOVY-8163?focusedCommentId=16009695&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16009695. The problem is that for some classes C with private member called "name" class property C.class.name which corresponds to java calls C.class.getName() does not work unless this catch block is added.

src/main/groovy/lang/MetaClassImpl.java
```diff
@@ -1832,6 +1832,9 @@ public Object getProperty(Class sender, Object object, String name, boolean useS
             } catch (IllegalArgumentException e) {
                 // can't access the field directly but there may be a getter
                 mp = null;
+            } catch (GroovyRuntimeException e) {
```
blackdrag: At this point mp should be a CachedField. Since AccessPermissionChecker.checkAccessPermission(field); may now throw a GroovyRuntimeException, we have here the requirement to catch it. Problem is, that mp might not be a CachedField. If we will call a user method here instead, this would lead to a wrong flow. Thus I think the GroovyRuntimeException usage is a bad idea. Better make a new Exception, that extends GroovyRuntimeException and throw that in AccessPermissionChecker, which we then can catch

src/main/org/codehaus/groovy/reflection/CachedMethod.java
```diff
@@ -90,6 +92,12 @@ public CachedClass getDeclaringClass() {
 
     public final Object invoke(Object object, Object[] arguments) {
         try {
+            AccessPermissionChecker.checkAccessPermission(cachedMethod);
+        }
+        catch (AccessControlException ex) {
+            throw new InvokerInvocationException(new GroovyRuntimeException("Illegal access to method" + cachedMethod.getName()));
+        }
+        try {
             return cachedMethod.invoke(object, arguments);
```
blackdrag: can we move thos try-catch into checkAccessPermission ?

src/main/groovy/lang/MetaClassImpl.java
dpolivaev: I think AccessPermissionChecker should only throw exceptions extending from AccessControlException. because it specifically checks access permissions. 

So I suggest to introduce `public class CacheAccessControlException extends AccessControlException`

src/main/org/codehaus/groovy/reflection/CachedMethod.java
dpolivaev: I think that InvokerInvocationException is specific to call of invoke() and it should be not thrown  by AccessPermissionChecker directly.

src/main/groovy/lang/MetaClassImpl.java
dpolivaev: On the other side if it does not work because AccessControlExceptions are not handled by the calling code properly `public class CacheAccessControlException extends GroovyRuntimeException` could also be taken as a hack.

dpolivaev: How do you see it?

dpolivaev: Done

dpolivaev:
Could you please review the changes I implemented on your behalf?

blackdrag:
+1

jwagenleitner:
@dpolivaev can you squash up the commits and change the tabs in `AccessPermissionChecker` to spaces? If not, just let me know and I can take care of it.  In the future, it's helpful to create a separate branch for the PR to make it easier to rebase and avoid merge commits.

@blackdrag I would guess this is a candidate for 2_5_X and above, but maybe too much of a change for 2.4.12, though maybe not `breaking` in a code sense it could affect those running with security managers.  Do you agree?

blackdrag:
agreed

dpolivaev:
Done

jwagenleitner:
Thanks.

PR #533, opened 2017-05-03
GROOVY-7840: Verifier#makeDescriptorWithoutReturnType uses ClassNode#…

paulk-asert:
…toString with generics

### DIFF
--- a/src/main/org/apache/groovy/ast/tools/MethodNodeUtils.java
+++ b/src/main/org/apache/groovy/ast/tools/MethodNodeUtils.java
@@ -1,0 +1,70 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.groovy.ast.tools;
+
+import org.codehaus.groovy.ast.MethodNode;
+import org.codehaus.groovy.ast.Parameter;
+
+/**
+ * Utility class for working with MethodNodes
+ */
+public class MethodNodeUtils {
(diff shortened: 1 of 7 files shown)


src/main/org/apache/groovy/ast/tools/MethodNodeUtils.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class MethodNodeUtils {
+    /**
+     * Return the method node's descriptor including its
+     * name and parameter types without generics.
+     *
+     * @param mNode the method node
+     * @return the method node's abbreviated descriptor excluding the return type
+     */
+    public static String methodDescriptorWithoutReturnType(MethodNode mNode) {
```
jwagenleitner: This seems like it could be an instance method such as `MethodNode#getSignature()` ([JLS 8.4.2](http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.4.2)) or `MethodNode#getTypeDescriptorWithoutReturnType()`.

paulk-asert: I kept it distinct because I wasn't really sure that we should tie the particular formatting we have chosen to the more general method that some might expect to find in MethodNode. If you look at:
org.objectweb.asm.Type#getMethodDescriptor(java.lang.reflect.Method)
It would produce different output.

For example, for this method:
`int[] method(List arg1, String arg2)`
ASM would produce the familiar:
`(Ljava/util/List;Ljava/lang/String;)[I`
whereas we produce:
`[I method(java.util.List, java.lang.String)`
What we produce is suitable for our purposes (and we need the name) but I wonder what people would expect in MethodNode.

src/main/org/apache/groovy/ast/tools/MethodNodeUtils.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Return the method node's descriptor which includes its return type,
+     * name and parameter types without generics.
+     *
+     * @param mNode the method node
+     * @return the method node's descriptor
+     */
+    public static String methodDescriptor(MethodNode mNode) {
```
paulk-asert: See above.

paulk-asert:
I added an inline comment. It isn't clear-cut though, so let me know what you think. I have numerous other methods that would go in the respective util classes in another spike. But your right, if they stay a private constructor would be good.

src/main/org/apache/groovy/ast/tools/MethodNodeUtils.java
jwagenleitner: That is what is sort of confusing, the method descriptor for bytecode vs the method signature used to detect duplicate signatures.  Agree that the formatting is particular to it's use in the `Verifier` and having in a separate utils class is probably the best approach.

jwagenleitner:
Thanks for the reply Paul.  I see the point about the formatting and not wanting to make it part of the regular `MethodNode` api, so the utility classes are probably the best approach in this case.

PR #537, opened 2017-05-06
GROOVY-8157: Flow typing doesn't work with assignment to a parameter

shils:


shils:
@paulk-asert Good point, I'm not sure why I skipped that here. Amended.

PR #538, opened 2017-05-06
GROOVY-8135: SecureASTCustomizer whitelist does not work

armsargis:
For arrays we should get componentType instead of type


jwagenleitner:
Thanks!

PR #539, opened 2017-05-07
GROOVY-8074: Statically compiled dot property accesses instance prope…

shils:
…rty instead of Map property


src/test/org/codehaus/groovy/classgen/asm/sc/ArraysAndCollectionsStaticCompileTest.groovy
```diff
@@ -128,6 +128,18 @@ class ArraysAndCollectionsStaticCompileTest extends ArraysAndCollectionsSTCTest
         }
     }
 
+    void testMapSubclassPropertyStyleAccess() {
```
paulk-asert: maybe add in:
`// GROOVY-8074`

shils: Added, thanks for the reminder.

PR #540, opened 2017-05-14
GROOVY-7535: Groovy category throwing MissingMethodException and Miss…

jwagenleitner:
…ingPropertyException when using multiple threads

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/GroovyCategorySupport.java
+++ b/src/main/org/codehaus/groovy/runtime/GroovyCategorySupport.java
@@ -40,7 +40,6 @@ import java.util.concurrent.atomic.AtomicInteger;
 public class GroovyCategorySupport {
 
     private static int categoriesInUse = 0;
-    private static final AtomicInteger atomicCategoryUsageCounter = new AtomicInteger();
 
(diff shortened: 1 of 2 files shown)


jwagenleitner:
I believe the following scenario accurately describes the race condition that causes the problem:

1. T1 in endScope calls `atomicCategoryUsageCounter.decrementAndGet()` and receives `0`
2. T2 in newScope calls `atomicCategoryUsageCounter.incrementAndGet()` and receives `1`
3. T2 assigns `categoriesInUse` a value of `1`
4. T1 assigns `categoriesInUse` a value of `0`
5. Callsite called from T2 calls `hasCategoryInCurrentThread` and return is `false`

Another way to fix would be to remove the int counter and just use the existing AtomicInteger. However, my hunch is that the int counter was retained because `hasCategoryInCurrentThread()` is probably called frequently so there might have been some performance concern over requiring a method call and a volatile read that the AtomicInteger would incur in order to implement the quick exit check for that method (`if (categoriesInUse == 0) return false`).

There is also a race between scopes in calls to `DefaultMetaClassInfo.setCategoryUsed`, so I think it would also be good to add those calls to the synchronized blocks if they are added.

PR #541, opened 2017-05-16
Improve naming and documentation for String.tokenize

mnonnenmacher:
Calling the delimiter argument token is confusing as according to the
StringTokenizer documentation delimiters are NOT treated as tokens, so call
it delimiter instead. Also make clear that every character in the
CharSequence argument is treated as a separate delimiter.


paulk-asert:
Good suggestion - merged - thanks!

PR #544, opened 2017-05-20
refactor: type safety and formatting

jwagenleitner:


src/main/org/codehaus/groovy/transform/trait/Traits.java
```diff
@@ -355,19 +355,19 @@ static String getSuperTraitMethodName(ClassNode trait, String method) {
...
-    public static @interface Implemented {}
+    public @interface Implemented {}
 
     /**
      * Internal annotation used to indicate that a method is a bridge method to a trait
      * default implementation.
      */
     @Retention(RetentionPolicy.RUNTIME)
     @Target(ElementType.METHOD)
-     public static @interface TraitBridge {
```
graemerocher: This is a breaking change, some libraries like GORM needs to inspect this information in order to know whether a property or fields i provided by a trait. See https://github.com/grails/grails-data-mapping/blob/master/grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/FieldEntityAccess.java#L227

src/main/org/codehaus/groovy/transform/trait/Traits.java
```diff
@@ -355,19 +355,19 @@ static String getSuperTraitMethodName(ClassNode trait, String method) {
      */
     @Retention(RetentionPolicy.RUNTIME)
     @Target(ElementType.METHOD)
-    public static @interface Implemented {}
+    public @interface Implemented {}
```
graemerocher: Also breaking. See https://github.com/grails/grails-data-mapping/blob/master/grails-datastore-core/src/main/groovy/org/grails/datastore/mapping/reflect/FieldEntityAccess.java#L504

src/main/org/codehaus/groovy/transform/trait/Traits.java
jwagenleitner: The `static` modifier is redundant on inner interfaces. The bytecode will remain unchanged with or without the modifier so the code referenced should continue to work.

src/main/org/codehaus/groovy/transform/trait/Traits.java
graemerocher: Ah of course.. silly me

paulk-asert:
+1

PR #545, opened 2017-05-21
GROOVY-7611: java.util.Optional should evaluate to false if empty (Java8 VMPlugin)

jwagenleitner:
Target for this would be `2_5_X` and above.  For `2_5_X` and `2_6_X` it would require that the release process build with JDK 8 so would need to backport some of the build checks for Java version to those branches.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -268,12 +268,18 @@ sourceSets {
                     'src/main',
                     "$generatedDirectory/src/main"
             ]
+            if (!JavaVersion.current().isJava8Compatible()) {
+                exclude '**/vm8/*'
+            }
         }
         groovy {
             srcDirs = [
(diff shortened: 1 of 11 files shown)


blackdrag:
+1

danielsun1106:
nice!

jwagenleitner:
Note to self, need to test GDK docs for new method on Optional and may need to add to https://github.com/apache/groovy/blob/7ab4a7176841f35f919892032d1e3628f1213aa3/gradle/docs.gradle#L157

jwagenleitner:
This is missing a static compilation test, it looks like `@CompileStatic` does pick up plugin default methods.  This change does work with `@TypeChecked` though.

PR #546, opened 2017-05-21
GROOVY-7654: Iterable as List and Iterable.asList() have different semantics

armsargis:
Introduce DefaultGroovyMethods.asType(Iterable, Class) to handle correctly 'asType' to iterables

### DIFF
--- a/src/test/org/codehaus/groovy/runtime/DefaultGroovyMethodsTest.groovy
+++ b/src/test/org/codehaus/groovy/runtime/DefaultGroovyMethodsTest.groovy
@@ -236,7 +236,28 @@ public class DefaultGroovyMethodsTest extends GroovyTestCase {
         assertEquals(3, list.get(2));
     }
 
+    // GROOVY-7654
+    public void testIterableAsList() {
+        def list = [1, 2, 3]
+        def iterable = new IterableWrapper(delegate: list)
+
+        def iterableAsList = iterable.asList()
+        def iterableAsType = iterable as List
+        
+        assertEquals(iterableAsList, iterableAsType)
+        assertEquals(1, iterableAsList[0])
+        assertEquals(1, iterableAsType[0])
+    }
+
     private static class MyList extends ArrayList {
         public MyList() {}
+    }
+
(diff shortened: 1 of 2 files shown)


src/test/org/codehaus/groovy/runtime/DefaultGroovyMethodsTest.groovy
```diff
@@ -236,7 +236,28 @@ public class DefaultGroovyMethodsTest extends GroovyTestCase {
...
     }
+
+    private static class IterableWrapper implements Iterable {
+        Iterable delegate
+
+        Iterator iterator() {
+            delegate.iterator()
+        }
+    }
 }
```
blackdrag: I am missing a test 

def iterableAsIterable = iterable as Iterable
assert iterable.AsIterable.is(iterable)
def iterableAsIterableWrapper = iterable as IterableWrapper
assert iterable.AsIterable.is(iterable)

armsargis: Actually, test I got from jira issue and meaning I think is to create concrete implementation of Iterable which is not extending from collection. And for concrete implementation we need return some iterator.

Also I thinking to change code like:

```
public static <T> T asType(Iterable iterable, Class<T> clazz) {
        if (Collections.class.isAssignableFrom(clazz)) {
            return asType((Collection) toList(iterable), clazz);
        }

        return asType((Object) toList(iterable), clazz);
    }
```

armsargis: I will add missing test cases

blackdrag: If the iterable is an instance of Collection, then why isn´t the Collection taking method called directly?

blackdrag: interesting choice to use Collection.class.isAssignableFrom(clazz). But sure, why not... +1

PR #547, opened 2017-05-23
GROOVY-8186: Builder ExternalStrategy constructors have trouble with …

paulk-asert:
…private fields


paulk-asert:
Propose PR merged.

PR #549, opened 2017-05-25
GROOVY-8114: Include correct time unit in timeout error message

paulk-asert:
PR for patch as supplied by Jochen Kemnade. Planned for 2.5+.


jwagenleitner:
+1

PR #551, opened 2017-05-25
GROOVY-8204: @Delegate on arrays causes NPE during compilation

paulk-asert:


jwagenleitner:
+1

src/main/org/codehaus/groovy/ast/ClassHelper.java
```diff
@@ -440,12 +440,15 @@ private static boolean hasUsableImplementation(ClassNode c, MethodNode m) {
     /**
      * Returns a super class or interface for a given class depending on a given target.
      * If the target is no super class or interface, then null will be returned.
+     * For a non=primitive array type, returns an array of the componentType's superclass.
```
jwagenleitner: non-primitive

paulk-asert: fixed the typo, thanks

PR #552, opened 2017-05-26
GROOVY-7979: Prevent JsonSlurper issue on a single minus

JamesLaverack:
Add bound checking to JSON parser code to account for a single minus character.


blackdrag:
+1

jwagenleitner:
Thanks!

PR #555, opened 2017-05-31
GROOVY-8200 - Shorthand |= results in NPE

noamt:
Fixing the NPE with an explicit test and assignment of the right-hand boolean operator.
Assigning a primitive boolean variable with null, like `boolean x = null`, will end up as `false`.
I think it only makes sense to apply the same behavior to the right hand operator.

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -15760,8 +15760,7 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
      * @since 1.0
      */
     public static Boolean and(Boolean left, Boolean right) {
-        right = Boolean.TRUE.equals(right);
-        return left && right;
+        return left && Boolean.TRUE.equals(right);
     }
 
     /**
@@ -15773,8 +15772,7 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
      * @since 1.0
      */
     public static Boolean or(Boolean left, Boolean right) {
-        right = Boolean.TRUE.equals(right);
-        return left || right;
+        return left || Boolean.TRUE.equals(right);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


dlehammer:
@noamt; looks good to me functionality- and coverage-wise :)
I'm by no means an expert on the Groovy documentation-convention, but since the internal behavior for the right-argument has changed - shouldn't the Groovydoc match the new behavior?

Best regards,

Daniel

noamt:
Care to elaborate?

dlehammer:
Excerpt from _src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java_

```
/**
 * Logical conjunction of two boolean operators.
 *
 * @param left left operator
 * @param right right operator
 * @return result of logical conjunction
 * @since 1.0
 */
public static Boolean and(Boolean left, Boolean right) {
	if (right == null) {
		right = Boolean.FALSE;
	}
	return left && right;
}
```

My point is that the Groovydoc for both arguments is identical, but the right-argument now explicitly adheres to the Groovy truth-convention, the documentation doesn't reflect that.

Best regards :)

dlehammer:
@noamt; I've mulled over the implementation, and the right-argument evaluation and assignment could be replaced with a single null-safe expression `Boolean.TRUE.equals(right)` in the return statement :) #ZeroSideEffect

noamt:
Great catch. Will do

noamt:
Thanks @dlehammer , I'll apply the changes

noamt:
@blackdrag 
Where do you think would be a fitting place for the shorthand test cases? Doesn't feel like they should be `DefaultGroovyMethodsTest`

dlehammer:
@noamt; if I remember correctly, @blackdrag suggested [OperatorsTest](https://github.com/apache/groovy/blob/183b8fbf0248a2dceffbba684e50c3c2c060e46c/src/spec/test/OperatorsTest.groovy) for the shorthand test-coverage.

jwagenleitner:
This looks ok to me but @blackdrag 's [comment](https://issues.apache.org/jira/browse/GROOVY-8200?focusedCommentId=16023385&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16023385) does point out the asymmetry this may cause with other implementations of the methods on other objects as well as the already defined default methods (for example, `BitSet` and `Number`) which currently also throw an NPE, for example.

```groovy
Number n1 = 4
Number n2 = null

n1 |= n2 // NPE
// should it be equivalent of n1 | (n2 ?: 0)
```

But offhand I don't see a reason to not support this on `Boolean`, but would be curious to know what @blackdrag thinks.

PR #556, opened 2017-06-01
GROOVY-8209 simple http client without extra dependecies

dlukyanov:
The idea is to create a simple http client without extra dependencies that will cover rest client needs.

Code example:
```groovy
def r = HTTP.get(
  url : 'https://www.googleapis.com/customsearch/v1',
  query: [ q: 'hello world' ],
  headers: [
    'Accept':'application/json',
    //'Authorization' : ...,
  ],
  //optional - ssl context 
  ssl : getNaiveSSLContext(),
  //optional - receiver
  receiver : { stream,context -> new JsonParser().parse(stream) },
  //optional - body
)
assert r.response.code == 200
println r.response.body

```


dlukyanov:
Hello groovy team, I'd like to understand would it be interesting to integrate this class into groovy language.
In attachment there is a generated javadoc and asciidoc documents.
Waiting for review or any comments.

[groovydoc_http.zip](https://github.com/apache/groovy/files/1068069/groovydoc_http.zip)

paulk-asert:
Thanks for your submission. I am hoping to get time to review shortly.

graemerocher:
-1 from me unless it can be made compatible with the upcoming Java 9 native client http://download.java.net/java/jdk9/docs/api/java/net/http/HttpClient.html

jwagenleitner:
My opinion is that something like this would be best provided as an external library and not included (from the start anyway) as part of Groovy itself or one of its subprojects.  There are already some nice http libraries such as [http-builder-ng](https://github.com/http-builder-ng/http-builder-ng) and [OkHttp](http://square.github.io/okhttp/) that are more complete and widely used.  Groovy has added some nice enhancements to the `URL` class for very basic fetching content, any more in my opinion is really best left to an outside library.

dlukyanov:
i see. ok, let's close this request.

eric-milles:
Isn't the JDK providing [HttpClient](https://docs.oracle.com/en/java/javase/11/docs/api/java.net.http/java/net/http/HttpClient.html), HttpRequest and HttpResponse as part of Java 11?

dlukyanov:
The idea is to simplify code required to perform put/post/... Class created before java11.

blackdrag:
closing as per https://github.com/apache/groovy/pull/556#issuecomment-315630564

PR #557, opened 2017-06-03
GROOVY-8056: GroovyCodeSource(URL) can leak a file handler

jwagenleitner:
A safer fix in terms of compatibility compared to PR #500.


jwagenleitner:
PR merged.

PR #558, opened 2017-06-04
GROOVY-8218 @Sortable allows reversed natural ordering

Leward:
See [GROOVY-8218](https://issues.apache.org/jira/browse/GROOVY-8218)

### DIFF
--- a/src/main/groovy/transform/Sortable.java
+++ b/src/main/groovy/transform/Sortable.java
@@ -45,6 +45,7 @@ import org.codehaus.groovy.transform.GroovyASTTransformationClass;
  *     priority will be according to the order given in the includes list)</li>
  *     <li>have three {@code Comparator} methods named {@code comparatorByFirst},
  *     {@code comparatorByLast} and {@code comparatorByBorn}</li>
+ *     <li>sort by natural order by default, reversed natural order can be specified</li>
  * </ul>
  * The properties within the class must themselves be {@code Comparable} or {@code @Sortable}.
  * <p>More examples:</p>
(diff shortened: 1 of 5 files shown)


src/main/org/codehaus/groovy/ast/tools/GeneralUtils.java
```diff
@@ -219,10 +219,38 @@ public static ClosureExpression closureX(Statement code) {
...
 
+    /**
+     * Build a binary expression that compares two values
+     * @param lhv expression for the value to compare from
+     * @param rhv expression for the value value to compare to
+     * @param reversed whether to use natural ordering or reversed natural ordering
+     * @return the expression comparing two values
+     */
+    public static BinaryExpression cmpX(Expression lhv, Expression rhv, boolean reversed) {
+        return (reversed) ? cmpXReversed(lhv, rhv) : cmpX(lhv, rhv);
```
jwagenleitner: What about `return (reversed) ? cmpX(rhv, lhv) : cmpX(lhv, rhv);`?  Could eliminate the need for the new `cmpXReversed` method.

src/main/org/codehaus/groovy/transform/SortableASTTransformation.java
```diff
@@ -112,7 +113,7 @@ private static void implementComparable(ClassNode classNode) {
         }
     }
 
-    private static Statement createCompareToMethodBody(List<PropertyNode> properties) {
+    private static Statement createCompareToMethodBody(List<PropertyNode> properties, Boolean reversed) {
```
jwagenleitner: `boolean` as the parameter type would be consistent with the other methods below.

src/main/groovy/transform/Sortable.java
```diff
@@ -145,4 +170,9 @@
      * Must not be used if 'includes' is used.
      */
     String[] excludes() default {};
+
+    /**
+     * Set to true so that comparator uses reversed natural order.
+     */
```
jwagenleitner: Would be good to add `@since 2.5.0`

src/main/org/codehaus/groovy/transform/AbstractASTTransformation.java
```diff
@@ -115,6 +115,14 @@ public int getMemberIntValue(AnnotationNode node, String name) {
         return 0;
     }
 
+    public boolean getMemberBoolValue(AnnotationNode node, String name) {
```
jwagenleitner: I think it's common to use `memberHasValue` for evaluating boolean attributes, so this new method is probably not needed.

src/main/org/codehaus/groovy/transform/SortableASTTransformation.java
```diff
@@ -82,6 +82,7 @@ public void visit(ASTNode[] nodes, SourceUnit source) {
     private void createSortable(AnnotationNode annotation, ClassNode classNode) {
         List<String> includes = getMemberStringList(annotation, "includes");
         List<String> excludes = getMemberStringList(annotation, "excludes");
+        boolean reversed = getMemberBoolValue(annotation, "reversed");
```
jwagenleitner: `boolean reversed = memberHasValue(annotation, "reversed", true);`

paulk-asert:
John, your suggestions look good. I'll be back from hols next week and can
check further then.

Leward:
Thank you @jwagenleitner for the feedback, I applied your comments.

jwagenleitner:
Thanks for the PR, much appreciated.

Just a note, I relocated the newly added `GeneralUtils#cmpX(Expression,Expression,boolean)` method into the `SortableASTTransformation` class as a private helper method.  Given its limited use it didn't seem to warrant exposing it as a public method.  If/when other uses come up then it may make sense to move it back to a public method like `GeneralUtils`, but I think it's always good to err on the side of less visibility whenever possible.

PR #559, opened 2017-06-11
GROOVY-8222: Setting Source Position in newly created PropertyExpression

ArsenBabakhanyan:


jwagenleitner:
Thanks for the PR.  While I think it's generally good practice to make variable names more expressive, in this case I think `mv` and to a lesser degree `pexp` are commonly used in that particular class and in this case the renames make it a little more difficult to see what is actually changing.  Would you mind changing the var names back?

ArsenBabakhanyan:
@jwagenleitner 
I have updated PR.
I am thinking about cleaning up this class up, i.e. doing some refactoring and updating names and creating PR

src/main/org/codehaus/groovy/classgen/AsmClassGenerator.java
```diff
@@ -1251,11 +1251,13 @@ private void processClassVariable(String name) {
             mv.visitInsn(DUP);
 
             loadThisOrOwner();
-            mv.visitLdcInsn(name);
+            mv.visitLdcInsn(expression.getName());
 
             mv.visitMethodInsn(INVOKESPECIAL, "org/codehaus/groovy/runtime/ScriptReference", "<init>", "(Lgroovy/lang/Script;Ljava/lang/String;)V", false);
         } else {
-            PropertyExpression pexp = new PropertyExpression(new VariableExpression("this"), name);
+            PropertyExpression pexp = new PropertyExpression(new VariableExpression("this"), expression.getName());
```
jwagenleitner: The patch provided in the JIRA creates a new `VariableExpression` (see below), I think that needs to be done so the source position is not set on the constant `VariableExpress.THIS_EXPRESSION` via the `getObjectExpression().setSourcePosition(..)` call.

```java
PropertyExpression pexp = new PropertyExpression(new VariableExpression("this", ClassHelper.DYNAMIC_TYPE), name);
```

jwagenleitner: Scratch that comment, I was looking at the wrong source.

ArsenBabakhanyan: ok

src/main/org/codehaus/groovy/classgen/AsmClassGenerator.java
```diff
@@ -1218,7 +1218,7 @@ public void visitVariableExpression(VariableExpression expression) {
 
         BytecodeVariable variable = controller.getCompileStack().getVariable(variableName, false);
         if (variable == null) {
-            processClassVariable(variableName);
+            processClassVariable(expression);
```
blackdrag: I am feeling a bit partial about the change from variableName to expression. I know you need it to set the source position later on. But should we start doing some funky stuff where variableName or expression are changed, the this will cause a problem we may not even notice correctly. So I would suggest something I really really do not often... make the two final

ArsenBabakhanyan: this PR is closed, as I have mentioned before i am planning to do some refactoring and creating a new PR, I will make expression final there

jwagenleitner: Thanks for the PR.  I merged before seeing the comment from @blackdrag, so another refactor PR would be good.

PR #560, opened 2017-06-12
GROOVY-8226: JSR308 initial plumbing tweaks

paulk-asert:


src/main/org/codehaus/groovy/vmplugin/vm8/Java8.java
```diff
@@ -51,4 +53,13 @@ public int getVersion() {
         return 8;
     }
 
+    protected int getElementCode(ElementType value) {
```
jwagenleitner: might be good to have `@Override`

paulk-asert: done

paulk-asert:
Actually, I am going to refine the Jira issues a little further. The current proposal doesn't push the relevant annotations to all the expected places for all cases that currently pass the grammar. This change may stand once I see what else needs to be done for the extra issues that I'll create or some aspects may move into or out of this issue.

danielsun1106:
+1

PR #563, opened 2017-06-21
vmplugin

jwagenleitner:
Alternative to PR #562.  I tried to make incremental commits to make it easier to review.

Still outstanding are the indy related classes in `org.codehaus.groovy.vmplugin.v7`.  Just not sure what a good package name for those would be, maybe `org.apache.groovy.internal.runtime.indy`?


jwagenleitner:
The `leftShift` default method in the Java5 extension class was handled differently due to GROOVY-2987 so that is why the change was made to the existing `StringGroovyMethods.leftShift(StringBuilder..)` method.

PR #564, opened 2017-06-24
GROOVY-8197: Make JUnit3/4 GroovyRunners

jwagenleitner:


jwagenleitner:
Just a couple of things of note:

1. Since `GroovySystem.RUNNER_REGISTRY` is part of the public API the `Map` interface was retained for compatibility.
1. This PR also changes how the `TestNgRunner` is registered no longer requiring it to `Grab`'d in order to use it from the distribution since it will automatically be picked up from the classpath (see comments in [GROOVY-8196](https://issues.apache.org/jira/browse/GROOVY-8196).

jwagenleitner:
Thanks for taking a look Paul.  As suggested I moved the `TestNgRunner` into a `org.apache.groovy.plugin.testng` package deprecating the old runner.  As I was doing that it also seemed like it would be a good idea to move the junit runners their own package so I moved them to `org.apache.groovy.plugin.junit`.  Also rebased to fix some merge conflicts.

jwagenleitner:
Since the Junit runners didn't need to be public I "hid" them in a `DefaultRunners` class and since they are part of core am forgoing the IO to load them as services and adding them directly to the registry. If one or more of them is moved out in a future release they can go in a `org.apache.groovy.plugin.junit[3,4,?]` package similar to testng.

PR #565, opened 2017-06-25
GROOVY-8236: Report more meaningful error for versions of Groovy not …

paulk-asert:
…supporting @Repeatable


paulk-asert:
This is intended for 2_4_X only. See also GROOVY-8234.

paulk-asert:
Merged onto GROOVY_2_4_X

PR #567, opened 2017-06-26
GROOVY-8234: Add @Repeatable java8 annotation support

paulk-asert:


paulk-asert:
This is intended for 2_5_X+. See also GROOVY-8236.

graemerocher:
Nice!!

paulk-asert:
If you get a chance to try it out (in particular with Grails) let me know. I haven't done as much testing as I would like. I'll check comments when I return from holidays.

PR #568, opened 2017-06-27
GROOVY-8042 groovyConsole gets confused with a slashy string with an escaped forward slash

dkhamitov:
Hi, guys! Here is quite a simple fix for the case when a slashy string renders in yellow just up to the escaped slash, not the closing one (https://issues.apache.org/jira/browse/GROOVY-8042). If we can have more tests on this code, please let me know how we can achieve this. The tests which I've added are kind of regression tests.


jwagenleitner:
Thank you.

PR #570, opened 2017-06-29
GROOVY-8240: Compilation error in @CompileStatic

avafanasiev:
I have no idea why should generated methods should have receiver as declared class. It causes bug https://issues.apache.org/jira/browse/GROOVY-8240


jwagenleitner:
Thanks.

PR #572, opened 2017-07-09
GROOVY-8251: Implement withCloseable on AutoCloseable

jwagenleitner:
`NioGroovyMethods.withAutoCloseable` was added for 2.5.0 as part of [GROOVY-7572](https://issues.apache.org/jira/browse/GROOVY-7572).  It appears it was originally placed there because at the time Java 6 was still used, but now 2.5.0 is 7+.

First commit 91220015319 proposes moving the method (as-is) into the core `IOGroovyMethods` class.  This is where the `withCloseable(java.io.Closeable...)` method is located.  This change also moves the tests for both methods from NIO to core.  Since 2.5.0 has not been officially released there may not be a need to deprecate or worry about binary compatibility for the `withAutoCloseable` method?

Second commit 261bc174980d proposes to rename the new but unreleased method from `withAutoCloseable` to `withCloseable` to be consistent with the existing method on `java.io.Closeable`.  See discussion on [GROOVY-8251](https://issues.apache.org/jira/browse/GROOVY-8251).


jwagenleitner:
Updated PR to also include [GROOVY-8259 adding suppressed exceptions](https://issues.apache.org/jira/browse/GROOVY-8259).

PascalSchumacher:
>It appears it was originally placed there because at the time Java 6 was still used, but now 2.5.0 is 7+.

That was the reason (I think I merged the pull request.).

>Since 2.5.0 has not been officially released there may not be a need to deprecate or worry about binary compatibility for the withAutoCloseable method?

+1 no need to keep binary compatibility to a beta release.

paulk-asert:
+1 no need to keep binary compatibility with beta-1

PR #576, opened 2017-07-22
GROOVY-7995: @CS short syntax closure call from closure

jwagenleitner:
Short syntax of closure call invokes wrong closure if wrapped in another
closure. This fix includes a combination of the contributed commit from
PR #460 along with the patch (see PR comments) provided by Jochen.


paulk-asert:
+1

PR #578, opened 2017-07-27
Call iterators hasNext as last when simpler checks can be done first

MartyNeal:
Addresses https://issues.apache.org/jira/browse/GROOVY-8271


paulk-asert:
Proposed PR merged. Thanks.

PR #585, opened 2017-08-14
[Sql] Factorizes batchCount incrementation and thus executeBatch() execution

akapps:
Code duplication.

We could restrain "batchCount" and "batchSize" visibility, but this would be a possible API breakage.
Also `incrementBatchCount()` method could be package-private, so that it is not published as part of the inherited API, but I couldn't find a good reason why. It could be final, but as `processResult(int[])` isn't, I sticked to the same logic.


paulk-asert:
LGTM

paulk-asert:
merged, thanks!

PR #586, opened 2017-08-15
GROOVY-8288 executeBatch() should not be called with batchCount == 0

akapps:
This is a first proposition of resolution for [GROOVY-8288](https://issues.apache.org/jira/browse/GROOVY-8288).
There still are open questions (see my comments on changes).

Thank you for your feedback.

### DIFF
--- a/subprojects/groovy-sql/src/main/java/groovy/sql/BatchingStatementWrapper.java
+++ b/subprojects/groovy-sql/src/main/java/groovy/sql/BatchingStatementWrapper.java
@@ -83,14 +83,27 @@ public class BatchingStatementWrapper extends GroovyObjectSupport {
     }
 
     public int[] executeBatch() throws SQLException {
-        int[] lastResult = delegate.executeBatch();
-        processResult(lastResult);
+        if (shouldCallDelegate()) {
+            int[] lastResult = delegate.executeBatch();
+            processResult(lastResult);
+        }
         int[] result = new int[results.size()];
         for (int i = 0; i < results.size(); i++) {
             result[i] = results.get(i);
         }
         reset();
         return result;
+    }
+
+    private boolean shouldCallDelegate() {
+        if (batchCount > 0) {
+            return true;
+        } else if (results.isEmpty()) {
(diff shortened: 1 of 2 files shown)


subprojects/groovy-sql/src/main/java/groovy/sql/BatchingStatementWrapper.java
```diff
@@ -83,8 +83,15 @@ public void clearBatch() throws SQLException {
...
-        processResult(lastResult);
+        if (batchCount > 0) {
+            int[] lastResult = delegate.executeBatch();
+            processResult(lastResult);
+        }
+        else if (results.isEmpty()) {
+            log.warning("Nothing has been added to batch");
+            // we let the JDBC provider decide how to handle an empty batch execution
+            return delegate.executeBatch();
+        }
```
akapps: This "else if" is not really elegant for such a particular case...
I could have done `if (batchCount > 0 || results.isEmpty())` in the first condition, but I thought it made the intention unreadable.

subprojects/groovy-sql/src/test/groovy/groovy/sql/SqlBatchTest.groovy
```diff
@@ -119,6 +121,44 @@ class SqlBatchTest extends GroovyTestCase {
...
+    }
+
+    void testWithBatchNoPreparedStatementNothingSpecified() {
+        def numRows = sql.rows("SELECT * FROM PERSON").size()
+        assert numRows == 3
+
+        def result = sql.withBatch { ps ->
+            // Do nothing - not a good practice at all...
+        }
+        assert result == [] as int[]
```
akapps: It is quite surprising at first sight that HSQL driver behaves differently if `executeBatch()` is called with nothing added to the batch, whether a preparedStatement is used or not.

I felt like it's not Groovy responsibility to "suppress" the exception for the preparedStatement case, and I wanted to test-specify those 2 behaviours as they are the current ones.

danielsun1106:
Merged. Thanks!

PR #588, opened 2017-08-19
Return 404 for non-existing paths in `TemplateServlet` when running from .war

tom-mi:
When `TemplateServlet` was used in application servers that don't unpack a war (i.e. when the real path of resources could not be retrieved), an internal server error was returned upon trying to retrieve a non-existing template file. I had this problem on payara micro.
With this fix, a 404 is returned instead.
When running from a unpacked war, this had already worked before.


jwagenleitner:
Thanks for the PR.  It would be helpful to have an issue we could associate this with, would you mind creating a JIRA ticket for this fix?  https://issues.apache.org/jira/projects/GROOVY

tom-mi:
Opened https://issues.apache.org/jira/browse/GROOVY-8294
Please let me know if you need additional information.

jwagenleitner:
Thanks!

PR #589, opened 2017-08-23
Groovyc Resource Leaks

Persi:
Groovyc did not release resources held by GroovyClassLoader correctly.


Persi:
Travis failed because the tests took too long. Can someone tell me what to do? As I am not the first one with this problem if I take a look at the build history.

danielsun1106:
Try again or increase the timeout by modifying the .travis.yml, in which you will find a number 60.
BTW, the latter is not recommended because 60 min is long enough for normal build.

Persi:
How can I try again? Increasing the timeout wouldn't be my first choice either, but I can't find a button in Travis to tell him to rerun.

danielsun1106:
Just push a new commit ;-)

.travis.yml
```diff
@@ -22,7 +22,7 @@ jdk:
 
 install: true
 
-script: travis_wait 60 ./gradlew test
+script: travis_wait 59 ./gradlew test
```
Persi: Just changed this to have changes to commit. This commit would be reverted if Travis does not fail with  test timeout.

paulk-asert:
Was there a Jira or email discussion about this leak? On face value the change seems OK but no test case is supplied so if we can't think of a suitable test we'd need to do some manual testing.

Persi:
Hello Paul,

because of the API change I wasn't sure if merging in 2.4 branch is the right way.
I'll close this PR and merge my change in 2.5 branch, keep the old 'makeCompileUnit()' method and mark it as deprecated. I'll try to create a test for that, but that's difficult because with building on ant project you don't recognize the resource leak. It is just a problem if you build multiple ant projects in a row within the same JVM (that's what we do actually).

Thanks for your advice and ... I'll be back ;)

Marcus

PR #590, opened 2017-08-23
s/Homer Simson/Homer Simpson/

holgerschroeder:
Hi,

perhaps Homer in this example is named after the character in the "The Simpsons".
If yes, let's get his last name spelling correct...

kind regards, Holger


jwagenleitner:
Thanks!

PR #591, opened 2017-08-25
Resource leaks in Groovyc

Persi:
Building multiple ant projects in a row within the same JVM sometimes fails on windows with locked jars. This is e.g. the case if two build projects has circle dependencies.

### DIFF
--- a/subprojects/groovy-ant/src/main/java/org/codehaus/groovy/ant/Groovyc.java
+++ b/subprojects/groovy-ant/src/main/java/org/codehaus/groovy/ant/Groovyc.java
@@ -291,7 +291,7 @@ public class Groovyc extends MatchingTask {
      * @param version the bytecode compatibility mode
      */
     public void setTargetBytecode(String version) {
-        
+
         for (String allowedJdk : CompilerConfiguration.ALLOWED_JDKS) {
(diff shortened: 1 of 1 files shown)


jwagenleitner:
Does https://issues.apache.org/jira/browse/GROOVY-7573 look related?  Also, just for future reference it's probably best to do the PR against master.

Persi:
Hello @jwagenleitner, thanks for your link. This sounds absolutely like our error case.
I know that pushing against master is preferred, but in this case I wanted to bring the change live a little bit earlier, as Groovy 2.6 will release way later than 2.5.

Thanks for your advice.

danielsun1106:
@Persi 
FYI, normally we merge PRs into master and back-port them to other branches ;)

Persi:
@danielsun1106 so do I have to merge the changes to 2.6 Branch and create a new PR or is it OK to merge to 2.5? Because I don't want to break your rules ;)

danielsun1106:
@Persi Thanks for your contribution. Would you mind creating a new PR to merge into master and closing this one?

Persi:
Merging in Master is preferred, so I'll merge to master and create new PR.

PR #595, opened 2017-08-29
Groovyc Resource Leaks

Persi:
Building multiple ant projects in a row within the same JVM sometimes fails on windows with locked jars. This is e.g. the case if two build projects has circle dependencies.

Relates to https://issues.apache.org/jira/browse/GROOVY-7573


danielsun1106:
Merged. Thanks!

PR #597, opened 2017-08-30
GROOVY-8305: Fix default Ivy settings file for local m2 path

apupier:
- worth being tested on Linux
- not sure how to write a test for that

### DIFF
--- a/src/resources/groovy/grape/defaultGrapeConfig.xml
+++ b/src/resources/groovy/grape/defaultGrapeConfig.xml
@@ -26,7 +26,7 @@
         <ivy pattern="${user.home}/.groovy/grapes/[organisation]/[module]/ivy-[revision].xml"/>
         <artifact pattern="${user.home}/.groovy/grapes/[organisation]/[module]/[type]s/[artifact]-[revision](-[classifier]).[ext]"/>
       </filesystem>
-      <ibiblio name="localm2" root="file:${user.home}/.m2/repository/" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true"/>
+      <ibiblio name="localm2" root="file:/${user.home}/.m2/repository/" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true"/>
       <!-- todo add 'endorsed groovy extensions' resolver here -->
       <ibiblio name="jcenter" root="https://jcenter.bintray.com/" m2compatible="true"/>
       <ibiblio name="ibiblio" m2compatible="true"/>


src/resources/groovy/grape/defaultGrapeConfig.xml
```diff
@@ -26,7 +26,7 @@
         <ivy pattern="${user.home}/.groovy/grapes/[organisation]/[module]/ivy-[revision].xml"/>
         <artifact pattern="${user.home}/.groovy/grapes/[organisation]/[module]/[type]s/[artifact]-[revision](-[classifier]).[ext]"/>
       </filesystem>
-      <ibiblio name="localm2" root="file:${user.home}/.m2/repository/" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true"/>
+      <ibiblio name="localm2" root="file:/${user.home}/.m2/repository/" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true"/>
```
jaikiran: Actually, as noted in the RFC[1] and the simplified version[2], this should be:

```
file:///${user.home}/.m2/repository/
```
That should work fine both on Windows and *nix

[1] https://tools.ietf.org/html/rfc8089
[2] https://en.wikipedia.org/wiki/File_URI_scheme

apupier: I tried with file:/// and it is not working:

cachedGrapes: resource not reachable for org.apache.camel#camel-catalog-provider-karaf;2.19.2: res=C:\Users\Aurelien Pupier\.groovy\grapes\org.apache.camel\camel-catalog-provider-karaf\jars\camel-catalog-provider-karaf-2.19.2.jar
	 trying file:///C:\Users\Aurelien Pupier/.m2/repository/org/apache/camel/camel-catalog-provider-karaf/2.19.2/camel-catalog-provider-karaf-2.19.2.jar
		tried file:///C:\Users\Aurelien Pupier/.m2/repository/org/apache/camel/camel-catalog-provider-karaf/2.19.2/camel-catalog-provider-karaf-2.19.2.jar

apupier: it is working with file://

lhein: file:///${user.home}/.m2/repository/ 

is most likely not correct. You will end up with 4 slashes on Linux/Unix.

It should be file:// afaik

lhein: besides that the change works on my fedora machine.

lhein: just tested with file:// and it fails for me.

apupier: to sum up:
- "file:/"
-- Windows OK
-- Linux OK
- "file://"
-- Windows OK
-- Linux KO
- "file:///"
-- Windows KO
-- Linux KO

and the winner is: "file:/"

apupier:
What can we do to ensure that this PR is reviewed/merged as soon as possible?

danielsun1106:
+1

@paulk-asert  What do you think about the PR?

paulk-asert:
I get different results to the summary above. Perhaps I am running a different test. I don't have a space in my user.home directory either. For me I get:
{noformat}
file: Windows OK, Ubuntu OK (status quo)
file:/ Windows OK, Ubuntu KO
file:// Windows KO, Ubuntu OK
file:/// Windows OK, Ubuntu OK
{noformat}

apupier:
with the provided script on Windows I have the same results:

> file: Windows OK --> this one was not working in the reported case but working with provided script
> file:/ Windows OK
> file:// Windows KO
> file:/// Windows OK

apupier:
I tried also with this script to try to be as close as possible to intiial usecase but still cannot reproduce the problem with only file:

> import groovy.grape.Grape;
> import groovy.lang.GroovyClassLoader;
> Grape.grab([classLoader : new GroovyClassLoader(),group :'org.apache.camel', module : 'camel-catalog-provider-karaf', version : '2.19.2B'])

the code calling Grape: https://github.com/apache/camel/blob/camel-2.19.2/platforms/camel-catalog-maven/src/main/java/org/apache/camel/catalog/maven/MavenVersionManager.java

I'm wondering what is causing the issue, any ideas on what can be the possible root cause in order that i can try to reproduce are very welcome :s

jaikiran:
@apupier, In the IVY JIRA you filed originally, can you add the steps that I can follow to reproduce this issue?

apupier:
> @apupier, In the IVY JIRA you filed originally, can you add the steps that I can follow to reproduce this issue?

Just to be sure that you noticed it, I provided the steps few days ago: https://issues.apache.org/jira/browse/GROOVY-8305?focusedCommentId=16167801&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16167801

jaikiran:
@apupier, I wasn't following that JIRA so hadn't noticed your update which included the steps to reproduce. Thanks for letting me know. I'll give it a try in the next few days and see how it goes.

apupier:
@jaikiran i'm wondering if the openjdk and oracle jdk are not behaving differently for this "file:" stuff, on a server we had a lot of errors of ivy cache in log, I changed from openjdk to oracle jdk and there are no more errors. I think that it might explain the difference of result that we have.

src/resources/groovy/grape/defaultGrapeConfig.xml
nickboldt: This seems logical/obvious since ${user.home} includes a prefix slash, and for a valid URI you need two slashes. Thus when in doubt, file:/${user.home}/path/to/things. This is true for p2 repo URIs too.

apupier:
> @apupier, I wasn't following that JIRA so hadn't noticed your update which included the steps to reproduce. Thanks for letting me know. I'll give it a try in the next few days and see how it goes.

@jaikiran Had you have time to have a look?

jaikiran:
@apupier, Yes I did - in fact I tried it out on one of the Windows system that I borrowed from a friend. My testing didn't reproduce the original issue. However, I did notice a similar issue that had been reported before (in a different context)[1]. That one too is hard to reproduce and this all seems to point to a JRE specific behaviour.

Given all this, I don't have a solution nor an explanation on what specific code/component contributes to this, yet.

[1] https://bz.apache.org/bugzilla/show_bug.cgi?id=59104

danielsun1106:
Apache Ivy 2.5.0-rc1 is out. Maybe we can bump Ivy to 2.5.0-rc1 and see whether the issue exists.
http://ant.apache.org/ivy/

danielsun1106:
Groovy is using Ivy 2.5.0-rc1 on master and 2.6.x branches. Feel free to give it a try.

eric-milles:
When I have used a filesystem-based ibiblio in the past on Windows, the format looked like
```
<ibiblio name="Local" m2compatible="true" root="file:///C:/Users/..."
```
As stated above, `${user.home}` is going to resolve to "C:\..." on Windows and "/home/..." on Linux.  So, I don't think you can unconditionally set a file url prefix that will work for both.  Is there a variable that can be used to get a file: url for the user's directory?

eric-milles:
Another approach would be to set the "ivy.ibiblio.default.artifact.root" property from the code and remove the root attribute from the settings file.  I thinnk this defaults to Maven Central's URL.  So you would need to add a root attribute to the ibiblio below jcenter, which is intended to point to maven central.

https://ant.apache.org/ivy/history/master/resolver/ibiblio.html

eric-milles:
Last line below is proposed addition to `GrapeIvy.groovy` that would allow you to use "${user.home.url}" in the default settings xml.
```groovy
    GrapeIvy() {
        // if we are already initialized, quit
        if (enableGrapes) return

        // start ivy
        Message.defaultLogger = new DefaultMessageLogger(System.getProperty('ivy.message.logger.level', '-1') as int)
        settings = new IvySettings()
        settings.setVariable('user.home.url', new File(System.getProperty('user.home')).toURI().toURL() as String)
```

PR #599, opened 2017-09-13
GROOVY-7721: Static type checking fails when compiling against a Java interface call

avafanasiev:
Arrays is also covariants in java, so they should be also compared. 
Can't find exact documentation, but  JLS 15.12.2.5 says:
– Otherwise, if all the maximally specific methods are abstract or default, and
the signatures of all of the maximally specific methods have the same erasure
(§4.6), then the most specific method is chosen arbitrarily among the subset
of the maximally specific methods that have the most specific return type.

### DIFF
--- a/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -1149,11 +1149,18 @@ public abstract class StaticTypeCheckingSupport {
     private static void removeMethodWithSuperReturnType(List<MethodNode> toBeRemoved, MethodNode one, MethodNode two) {
         ClassNode oneRT = one.getReturnType();
         ClassNode twoRT = two.getReturnType();
-        if (oneRT.isDerivedFrom(twoRT) || oneRT.implementsInterface(twoRT)) {
+        if (isCovariant(oneRT, twoRT)) {
             toBeRemoved.add(two);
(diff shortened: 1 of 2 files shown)


danielsun1106:
It has been merged manually, thanks!
This is my first time to merge PR. Sorry for failing to preserve your information because of my wrong operation...

@paulk-asert Could you tell me what the correct operation should be?
```
Merging isn't done through the github interface - that's an Apache workflow "feature". Instead just manually merge and amend (one of) the commit(s) to include wording like: "(closes #374)".
```

danielsun1106:
`GROOVY-7721: Static type checking fails when compiling against a Java interface call (closes #599)` https://github.com/apache/groovy/commit/34df6cf461a5ecce66f2c2c0f14489cf96feaf34

danielsun1106:
I'm trying to keep your contributor information, so I reverted the commit

danielsun1106:
Merged. Thanks again!
https://github.com/apache/groovy/commit/8ad6e770ab2a6c6968ae02ddc5d49d7d6e6f79f3

danielsun1106:
https://github.com/apache/groovy/commit/8ad6e770ab2a6c6968ae02ddc5d49d7d6e6f79f3

avafanasiev:
Thanks to you! It would be nice, if you merge it to 2.5.0. Can you?

danielsun1106:
https://github.com/apache/groovy/commit/849a275bc8895b48879c3aacabe2aeae1462d1ed

As you can see, it has already been merged into 2.5.0 too ;)

avafanasiev:
Thanks!

PR #602, opened 2017-09-15
GROOVY-8319: Improve smart type on list expressions

avafanasiev:
I hope it should work like expected.


danielsun1106:
Merged. Thanks!

PR #605, opened 2017-09-23
GROOVY-8220: SC GroovyCastException on parameter flow typing

jwagenleitner:
GROOVY-8157 introduced flow typing for parameters and this fix is
required in order to track their assignments in `if` branches for
temporary type assignments.


paulk-asert:
Interesting solution. I also did something similar but then backed it out and had a parallel `ifElseForWhileAssignmentTrackerParams`. But your approach is simpler. I'll have a think about whether I see any cases that might be handled differently.

paulk-asert:
I will probably merge this PR in preference to mine (PR#606). Just wondering whether wrapping is needed in the case of a closure shared variable? I haven't had time to think it through myself yet.

PR #606, opened 2017-09-24
GROOVY-8220: GroovyCastException with CompileStatic

paulk-asert:


paulk-asert:
I will probably close those PR in preference to PR#605 but just making it available for easy reference later if needed.

PR #607, opened 2017-09-26
GROOVY-8330: Wrong 'Inconvertible types' error on casting interface

avafanasiev:
Make cast error a little bit like java.


danielsun1106:
Merged. Thanks!

PR #610, opened 2017-10-01
GROOVY-8326: @Override should not copied onto methods generated by ap…

jwagenleitner:
…plying @Memoize


paulk-asert:
+1. It begs the question about whether other annotations should have special treatment, e.g. should @SafeVarargs or @SuppressWarnings be transferred only onto the new internal method and be removed from the original? But I see no reason why the approach being used couldn't be extended over time to handle such cases - it will all be "invisible" to the user.

erdi:
Thanks for picking this up so quickly, @jwagenleitner.

paulk-asert:
Merged!

PR #611, opened 2017-10-01
GROOVY-8336: Static compilation requires casting inside instanceof ch…

paulk-asert:
…eck (additional cases)


paulk-asert:
I tried numerous "simpler" solutions but the INFERRED_TYPE metadata is used in many places and changing that info earlier led to problems like call-site writing with too specific types, so I stashed away the instanceof info under the INFERRED_RETURN_TYPE key and pull it out later when needed. Happy to hear of simpler ways of doing this if anyone can see an alternative.

PR #612, opened 2017-10-02
GROOVY-8252: AIOOBE in combination of ncurry and rcurry

paulk-asert:
Because of the way method selection works with overloaded methods, the position of params when applying `rcurry` is calculated lazily. E.g. for a method closure for the `foo` method, if there are two foo variants `foo(String, String)` and `foo(String, String, String)`, then right currying a single string parameter and supplying a single string argument will result in the `foo(String, String)` method being called but supplying two string arguments will call the 3 param variant to be called.

It doesn't make sense to do this for `ncurry` since a specific index is being supplied. I have clarified the documentation and implementation to capture those aspects of the current design and introduced some checking to avoid the AIOOBE when using `ncurry` when those assumptions are violated and instead throw an `IllegalArgumentException`.

### DIFF
--- a/src/main/groovy/lang/Closure.java
+++ b/src/main/groovy/lang/Closure.java
@@ -556,6 +556,10 @@ public abstract class Closure<V> extends GroovyObjectSupport implements Cloneabl
      * assert halver(8) == 4
      * </pre>
      *
+     * The position of the curried parameters will be calculated lazily, for example,
+     * if two overloaded doCall methods are available, the supplied arguments plus the
+     * curried arguments will be concatenated and the result used for method selection.
+     *
      * @param arguments the arguments to bind
      * @return the new closure with its arguments bound
      * @see #curry(Object...)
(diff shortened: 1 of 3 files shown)


paulk-asert:
PLEASE NOTE: **_don't commit as is_**. Needs further work but might help with discussions.

Notes about this implementation: I avoided relying solely on the parameter types for `doCall` to trigger a MME in the case of missing arguments. This was to produce a less confusing error message in the following case:
```
def cl = { a, b, c = 'x' -> a + b + c }
assert cl.rcurry('b', 'c')()
```
since some users might expect the default param to be taken into consideration and there will be a `doCall(String, String)` and we have supplied `(String, String)`. Basically, I think we can't support both rcurry and default args and should document this. But there are also issues with the alternative I chose. See the two TODO comments in the test file. One also shows that we can't use `ncurry` either with default args. The other shows an issue with method closures in combination with the above implementation. A test that used to work would now fail:
```
groovy.ClosureCurryTest > testCurryWithMethodClosure FAILED
    java.lang.IllegalArgumentException: When currying expected at least 2 argument(s) to be supplied before known curried arguments but found 1
        at org.codehaus.groovy.runtime.CurriedClosure.getUncurriedArguments(CurriedClosure.java:101)
        at groovy.ClosureCurryTest.testCurryWithMethodClosure(ClosureCurryTest.groovy:265)
```

PR #613, opened 2017-10-07
@AutoFinal: Added support to auto-finalize closure parameters

mgroovy:
1. Extended AutoFinalASTTransformation with support to handle Closure|s in class nodes
2. Added several black box tests (see AutoFinalTransformBlackBoxTest) that check whether applying @AutoFinal to a class leads to the expected Groovy compiler errors ("The parameter [$paramName] is declared final but is reassigned") in methods/closures
3. Each test also checks that the exact same code without @AutoFinal compiles without errors
4. Extended the AutoFinal description to include the new feature
5. Refactored the description to make the intended usage more clear


paulk-asert:
This looks good at first glance. Nice work! I'll have a bit more of an in-depth look during the week.

paulk-asert:
I created an issue to track this: GROOVY-8347

paulk-asert:
Merged! Thanks! I have some minor tweaks but I'll do them after the fact.

mgroovy:
Great G-)

paulk-asert:
I made some minor modifications:

- added support for annotating a field or local variable assigned a closure value (only needed for fine-grained application of this annotation)
- automatically processed inner classes if annotated at the class level
- added guards to avoid processing twice (if say at both the class and method level - it's an idempotent change but more efficient)
- if annotated at the method level, automatically covers closures within that method

If we do ever add configuration options via annotation attributes, the processing changes above should put us in a good place, e.g. if you use configscript to turn on for all classes, you could turn off for one class with something like one of either of these:
```
@AutoFinal(enabled=false)
@AutoFinal(methods=false, constructors=false, innerClasses=false, closures=false)
```

Let me know what you think. I haven't completed the tests yet.

mgroovy:
* Looks all good to me G-)
* Even though I am generally all in favor of giving people options, still convinced disabling @AutoFinal locally makes no sense in practice, and it might even be a bad idea to support that (especially at the class level - least suprise an' all...)
* Could ask in Groovy users group, though, if someone can come up with an example where e.g. @AutoFinal(enabled=false) at the method level would make sense...

paulk-asert:
"Least surprise" is possibly like beauty ... in the eye of the beholder. If I see an `AutoFinal(disabled=true)`, I might be quite surprised until I realised a configscript is in place - which might be even more surprising if I didn't know that ahead of time. :-)

mgroovy:
* "Least surprise" in this case means I can trust that all parameters everywhere are final. One cannot apply "least surprise" to global settings, because they are, well, global, so whoever enables the global setting is responsible for making everyone involved aware of the setting (so that `AutoFinal(disabled=true)` does not confuse anyone).
* The alternative is to put AutoFinal explcitely on every class, and enforce this with a check-in-code-rule.

paulk-asert:
I think we are in agreement. I made changes to allow disabling. Let me know what you think.

I haven't finished extra tests yet. If you have time to work on some of those, that would help me out a lot; otherwise, I'll do them later this week.

PR #615, opened 2017-10-10
@AutoFinal: Added @AutoFinal(enabled=false) tests

mgroovy:
Added assertAutoFinalTestScriptWithDisabledAnnotation to AutoFinalTransformBlackBoxTest, which checks for all test scripts that they compile in Groovy if the test script class is annotated with @AutoFinal(enabled=false) (i.e. the same as if no AutoFinal annotation was present)


paulk-asert:
Merged, thanks. For future reference, you might want to rebase against master just before creating your PR. I just merged the two commits instead of the PR this time. Git might have been smart enough to deal with the noise but I wasn't going to muck around and find out.

PR #617, opened 2017-10-11
GROOVY-8352

aalmiray:
See https://issues.apache.org/jira/browse/GROOVY-8352

Retention policy is set to `RUNTIME`in order to allow frameworks and libraries to have access to the new metadata.

There's a breaking change in `Verifier`, a protected method now returns `MethodNode` instead of `void`. This change was reviewed with @blackdrag and deemed the better solution among alternatives (such as duplicate method & deprecate old).

The JaCoCo team has done their part to prepare their project for this new annotation (see https://github.com/jacoco/jacoco/pull/610).

### DIFF
--- a/src/main/groovy/transform/Generated.java
+++ b/src/main/groovy/transform/Generated.java
@@ -25,10 +25,6 @@ import java.lang.annotation.Target;
 
 /**
  * The Generated annotation is used to mark members that have been generated.
- *
- * @author Andres Almiray
- * @author Jochen Theodorou
- * @author Mark Hoffmann
  */
 @Target({ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.TYPE, ElementType.FIELD})
 @Retention(RetentionPolicy.RUNTIME)
--- a/src/main/groovy/transform/Generated.java
+++ b/src/main/groovy/transform/Generated.java
@@ -1,0 +1,36 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package groovy.transform;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The Generated annotation is used to mark members that have been generated.
+ *
+ * @author Andres Almiray
+ * @author Jochen Theodorou
+ * @author Mark Hoffmann
+ */
+@Target({ElementType.METHOD, ElementType.CONSTRUCTOR, ElementType.TYPE, ElementType.FIELD})
+@Retention(RetentionPolicy.RUNTIME)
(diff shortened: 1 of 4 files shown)


src/main/groovy/transform/Generated.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * The Generated annotation is used to mark members that have been generated.
+ *
+ * @author Andres Almiray
+ * @author Jochen Theodorou
+ * @author Mark Hoffmann
```
marchof: It's "Marc Hoffmann" 😉

paulk-asert: Actually, we are mostly trying not to use @author tags these days - though we haven't made it mandatory.

Apache encourage removal of @author tags. It's not a mandated rule but they err on the side of wanting the whole community to own the whole codebase rather than people being put off by not wanting to touch a file "owned" by someone else. Git will tell us who the contributor was on a more fine-grained and accurate level than such tags anyway but obviously won't capture multiple authors committing under one of those names. Make sure anyone who has made a significant contribution is added into the `pomconfigurer.gradle` file.

marchof: @paulk-asert I'm completely o.k. with removing my name here at all.

paulk-asert: @marchof the pomconfigurer.gradle file is a good option too Marc! :-)

aalmiray: updated PR by removing @author tags

paulk-asert:
I get the same failures locally as travis-ci. I haven't had time to fully investigate but it seems like we can get away without setting a source unit when we create little snippets by hand (in our test code for instance) so long as we don't have any annotations in those snippets. While I wonder whether checking for version >= jdk5 with `isAnnotationCompatible` is even needed anymore it might just be best to not add the `Generated` annotation if we don't find a source unit? What do you think?

aalmiray:
The test for `@Generated` was adapted from the one for `@Immutable` hence the check. @paulk-assert if you think it's worth removing then sure, I can remove it 😄  What other changes should be made to the test case?

src/main/org/codehaus/groovy/classgen/Verifier.java
```diff
@@ -384,6 +387,7 @@ public void visit(MethodVisitor mv) {
     protected void addGroovyObjectInterfaceAndMethods(ClassNode node, final String classInternalName) {
         if (!node.isDerivedFromGroovyObject()) node.addInterface(ClassHelper.make(GroovyObject.class));
         FieldNode metaClassField = getMetaClassField(node);
+        AnnotationNode generatedAnnotation = new AnnotationNode(ClassHelper.make(GENERATED_ANNOTATION));
```
paulk-asert: I needed a guard on adding the annotation to keep the build happy:
```
boolean shouldAnnotate = classNode.getModule().getContext() != null;
```

aalmiray: 2 small fixes added 😄

marchof:
FYI: The filter for the new annotation will be part of JaCoCo 0.8.0 release soon.

blackdrag:
I pushed the branch. I do agree we should have a guideline for when to use it and when not. But I think we should first start using it and then decide what we want to advice

cavdhut:
I see it is not adding the @Generated annotation to get/set methods generated by groovy for private variables.  As a result code coverage reports are still not that accurate as code coverage reports still show get/set methods not covered.

blackdrag:
You mean for example in class X{def y} the getY and setY method?

cavdhut:
Yes..

cavdhut:
Any thoughts on above request to add @Generated annotation for Groovy generated get/set methods for private variables?

michelzanini:
Hi @aalmiray / @blackdrag ,

I wonder why the auto-generated setters and getters, as well as equals / hashcode are not also marked as `@Generated`.

I have the following class:
screen shot 2018-07-16 at 17 13 52

I am getting this coverage report:
screen shot 2018-07-16 at 17 13 43

I am using Groovy 2.5.1 and JaCoCo 0.8.1.

This class was supposed to have 100% code coverage.

Also something I get coverage saying I have not covered all branches of a line, but the line has just one branch:

screen shot 2018-07-16 at 17 19 09

What do you guys think?

PR #618, opened 2017-10-14
Condense org.codehaus.groovy.runtime.ArrayUtil

Nektro:
use varargs instead of manual 250-length Array construction


jwagenleitner:
Hi @Nektro thanks for the PR. The purpose of that `ArrayUtils` class is to ease bytecode generation and avoid needing to write bytecodes to generate an array as is needed for the varargs. The class comments show an example and explains why the individual methods are needed and they are statically linked to in generated bytecode by the `CallsiteWriter`.

Nektro:
@jwagenleitner sorry about that didn't see that it wasn't used in your actual program. Interesting JIT quirk you all found. Feel free to close :)

PR #619, opened 2017-10-20
GROOVY-8360: Enums that are inner classes should have the static clas…

shils:
…s modifier

### DIFF
--- a/src/test/gls/enums/EnumTest.groovy
+++ b/src/test/gls/enums/EnumTest.groovy
@@ -603,6 +603,80 @@ class EnumTest extends CompilableTestSupport {
             assert Foo.getAnnotations().size() == 1
           '''
     }
+
+    void testNestedEnumHasStaticModifier_GROOVY_8360() {
+        assertScript '''
+            class Foo {
+                enum Bar {
+                    X('x'), Y
+                    String s
+                    Bar(String s) { this.s = s }
+                    Bar() {}
+                }
+            }            
+            assert java.lang.reflect.Modifier.isStatic(Foo.Bar.modifiers)
+            assert Foo.Bar.X.s == 'x'
+        '''
+    }
+
+    void testEnumWithinInnerClassHasStaticModifier_GROOVY_8360() {
+        assertScript '''
+            class Foo {
+                class Baz {
+                    enum Bar {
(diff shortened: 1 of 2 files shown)


paulk-asert:
On the one hand, the JLS[1] says: "_A nested enum type is implicitly_ `static`", so your proposed change looks good. However it goes on to say that "_This implies that it is impossible to declare an enum type in the body of an inner class_". With some simple experimenting, it seems Groovy doesn't have this same restriction. I think we just need to do a little more checking that such cases aren't adversely affected (or understand if they are). I'll try to play around a little more myself but have other things on the TODO list first.

[1] [8.9. Enum Types](http://docs.oracle.com/javase/specs/jls/se8/html/jls-8.html#jls-8.9)

shils:
@paulk-asert That's an interesting case. I added a few variations of my original test to cover that along with static compilation. Looking at the byte code, enums nested within inner classes don't seem to have any references to instances of their enclosing classes, so I'd *think* this would be ok. I'll try to test this with some ecosystem projects when I have time.

src/test/gls/enums/EnumTest.groovy
```diff
@@ -592,6 +592,80 @@ class EnumTest extends CompilableTestSupport {
...
+                enum Bar {
+                    X('x'), Y
+                    String s
+                    Bar(String s) { this.s = s }
+                    Bar() {}
+                }
+            }          
+            @groovy.transform.CompileStatic
+            void test() {
+                assert java.lang.reflect.Modifier.isStatic(Foo.Bar.getModifiers())
```
shils: Unrelated to this issue - property style getter calls on Enum classes throw NoSuchFieldErrors at runtime when statically compiled.

paulk-asert: I created GROOVY-8369 for this. I can't at first glance see a reason why we'd not support this with @CS.

paulk-asert:
+1 for this to be merged to at least 2_5_X. I suspect there are further cases which differ from Java which need to be clarified but I think this takes us in the right direction and it will be further locking down to rule out cases that don't really make sense that Groovy appears to support but doesn't really and can be done outside the scope of this PR.

PR #620, opened 2017-10-27
Adding spacing around equals operator for consistency

russellsanborn:
I noticed this while looking at the documentation for operators:
http://groovy-lang.org/operators.html


paulk-asert:
Merged, thanks!

PR #621, opened 2017-10-28
Replaced deprecated padLeft call in BaseDuration

russellsanborn:
`StringGroovyMethods.padLeft(String, Number, String)` is deprecated. I converted this to use the recommended `StringGroovyMethods.padLeft(CharSequence, Number, CharSequence)`.

Added `this` to `millis` and `seconds` in the `toString()` method to be consistent.

Added additional tests for different `toString()` cases to cover missing features.
* The default of 0 is returned when there is 0 time passes in
* `year`, `month`, and `seconds` are tested now
* missing both `milliseconds` and `milliseconds`
* having `seconds` but missing `milliseconds`
* negative `milliseconds` with 0 `seconds`
* `seconds` and `milliseconds` both present
* `milliseconds` greater than 1000


paulk-asert:
Merged, thanks!

PR #625, opened 2017-10-31
GROOVY-8369: Statically compiled property access on enum class throws…

shils:
… NoSuchFieldError

### DIFF
--- a/src/test/org/codehaus/groovy/classgen/asm/sc/FieldsAndPropertiesStaticCompileTest.groovy
+++ b/src/test/org/codehaus/groovy/classgen/asm/sc/FieldsAndPropertiesStaticCompileTest.groovy
@@ -696,4 +696,21 @@ import org.codehaus.groovy.transform.sc.ListOfExpressionsExpression
             assert astTrees['A$_closure1'][1].contains('INVOKESTATIC A.pfaccess$02 (LA;Ljava/lang/String;)Ljava/lang/String;')
         }
     }
+
+    //GROOVY-8369
+    void testPropertyAccessOnEnumClass() {
+        try {
+            assertScript '''
+                enum Foo {}
+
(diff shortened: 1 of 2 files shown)


src/test/org/codehaus/groovy/classgen/asm/sc/FieldsAndPropertiesStaticCompileTest.groovy
```diff
@@ -696,4 +696,21 @@ import org.codehaus.groovy.transform.sc.ListOfExpressionsExpression
...
+                def test() {
+                    println Foo.getModifiers() // => 16401 // ENUM | FINAL | PUBLIC (see GROOVY_8360 wrt STATIC)
+                    println Foo.modifiers      // java.lang.NoSuchFieldError: modifiers
+                }    
+                test()
+            '''
+        } finally {
+            //println astTrees
+        }
+    }
```
paulk-asert: Perhaps a slight improvement to my original test would be comparing property access to the getter result and also confirming that enum constants take preference over property access (which worked before but I couldn't see covered in existing tests).
```
@Deprecated
enum Foo {
  annotations
}

def test() {
  assert 'annotations' == Foo.annotations.toString()
  assert Foo.getAnnotations().size() == 1
  assert Foo.getModifiers() == Foo.modifiers
}

test()
```

paulk-asert: Actually, I'll merge now and tweak the test as per above.

paulk-asert: Oops, I just noticed you beat me to it. I'll just make the tweak.

PR #627, opened 2017-11-03
Made the DefaultGroovyMethodsTest file more groovy like.

russellsanborn:
* Removed unnecessary semicolons and `public` modifiers.
* replaced java style for loops with groovy `eachWithIndex` closures
* instantiating maps and lists with `[]` and `[:]` instead of using `new ArrayList`, `Arrays.asList`, or `new LinkedHashMap`

### DIFF
--- a/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -15952,7 +15952,7 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
      * @since 1.0
      */
     public static Boolean and(Boolean left, Boolean right) {
-        return left && Boolean.TRUE.equals(right);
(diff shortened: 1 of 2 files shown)


paulk-asert:
The first three bullet points are non-controversial changes - the file was originally a Java file and wasn't refactored much when changed to Groovy. The null on the LHS probably requires more discussion on the mailing list. There are places where we try to hide NPEs and other places we don't. We value consistency across all cases ahead of commutativity (consistency between LHS and RHS).

Splitting into two PRs or at least two commits would make it easier to progress the non-controversial changes.

russellsanborn:
@paulk-asert I updated this PR to only have the non-controversial changes and updated the bullet points.

I opened a new PR with the changes to `DefaultGroovyMethods` boolean operations:
https://github.com/apache/groovy/pull/636

paulk-asert:
Merged, thanks.

PR #629, opened 2017-11-07
GROOVY-8295: Groovyc ant task can overflow Windows command line if cl…

paulk-asert:
…asspath is large. PR based on Eric's patch in the related issue.


eric-milles:
Thanks for moving this forward.  One concern I had is on Linux/Unix/MacOSX is accidental replacement of a common path.  For example if your project working directory was "/home" or "/" (heaven forbid), you could replace "/home/users/name/home/project/xyz" with "./users/name./project/xyz".

This was not a concern for my Windows testing, since a drive name always roots the absolute path.

paulk-asert:
Good point. I adjusted the patch accordingly.

PR #630, opened 2017-11-07
GROOVY-8289: STC and default value in ctor is causing debugging error

paulk-asert:
PR based on Eric's suggested fix in the issue. Still needs a test.


paulk-asert:
Test added.

PR #633, opened 2017-11-12
GROOVY-8377 Equals and hash code ast performance

erdi:
See [GROOVY-8377](https://issues.apache.org/jira/browse/GROOVY-8377) for the rationale behind this PR.

Given how simple the fix is and its impact (from 3k ops/ms to 30k ops/ms on my machine) I'd suggest considering backporting of it to 2.4.x, 2.5.x and 2.6.x lines.

`<rant>`
Working on this was harder then it should be down to:
- the build takes a minute(!) to configure, I believe it is down to [this code](https://github.com/apache/groovy/blob/1c42f421edd91f31aa35a3e7b7113910718b35d5/build.gradle#L93)
- the jmh benchmarks for Groovy code are broken and [needed fixing as part of my PR](https://github.com/erdi/groovy/commit/fc1d085a120bd7a542edeb7c1ff4b06816b776b2)

I understand that I worked against master, but the experience was highly unwelcoming.
`</rant>`

### DIFF
--- a/src/main/org/codehaus/groovy/ast/tools/GeneralUtils.java
+++ b/src/main/org/codehaus/groovy/ast/tools/GeneralUtils.java
@@ -72,6 +72,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+import static org.codehaus.groovy.syntax.Types.COMPARE_NOT_IDENTICAL;
+
 /**
  * Handy methods when working with the Groovy AST
  */
@@ -79,6 +81,7 @@ public class GeneralUtils {
     public static final Token ASSIGN = Token.newSymbol(Types.ASSIGN, -1, -1);
     public static final Token EQ = Token.newSymbol(Types.COMPARE_EQUAL, -1, -1);
     public static final Token NE = Token.newSymbol(Types.COMPARE_NOT_EQUAL, -1, -1);
+    public static final Token NOT_IDENTICAL = Token.newSymbol(COMPARE_NOT_IDENTICAL, -1, -1);
     public static final Token LT = Token.newSymbol(Types.COMPARE_LESS_THAN, -1, -1);
     public static final Token AND = Token.newSymbol(Types.LOGICAL_AND, -1, -1);
     public static final Token OR = Token.newSymbol(Types.LOGICAL_OR, -1, -1);
@@ -563,6 +566,10 @@ public class GeneralUtils {
 
     public static BinaryExpression ltX(Expression lhv, Expression rhv) {
         return new BinaryExpression(lhv, LT, rhv);
+    }
+
+    public static BinaryExpression notIdenticalX(Expression lhv, Expression rhv) {
+        return new BinaryExpression(lhv, NOT_IDENTICAL, rhv);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


blackdrag:
+1

jwagenleitner:
With the original code (pre-PR) I get 70k indy and 3k non-indy. With the PR as is I get 30k for both indy and non-indy. Testing with commented out `or` change I get 70k for indy and 30k for non-indy. In other words, the commented out code has the same performance for non-indy and doesn't reduce indy perf (based on my benchmark runs).

Was there a reason to leave the commented out statement or was it the intent to use that in the PR instead of the current statement?

erdi:
Sorry, it did not cross my mind to run the benchmark using indy. Is there an easy way to do so?

The commented out code, which I was not intending to commit, was my initial attempt at fixing the issue. My worry was that using the `or` condition introduces an additional call to the getter which might be significant for non trivial getter implementations. I could evaluate the getter once and store it in a variable but that would lead to a more complex solution. Switching receiver and argument of the `is()` call felt like a simpler solution. I was not aware that it will have negative impact on indy performance.

Should we go with the null check solution whoch does not degrade indy performance and ignore the additional evaluation of the getter it introduces for non null property values? I’m more than happy to ammend the PR in the morning.

jwagenleitner:
If you add `-Pindy=true` argument when executing the `jmh` task it should compile the test groovy code using `indy`. For example, I ran:

`./gradlew -PbenchInclude=GeneratedHashCode -Pindy=true :perf:jmh`

I'd be curious if you also see similar numbers. As you say, adding the null check may have a perf hit for non-trivial uses so maybe the decrease in indy perf is not as bad as it look.

blackdrag:
I am actually wondering... isn't `notX(sameX(fieldExpr, varX("this"))` The same as  `!this.@x.is(this)`,with x being a field? I assume we have this to avoid stack overflows... I have two comments to this actually:
1) I think we should not support that case
2) If we support it, it is better to create a binary expression with the token `COMPARE_NOT_IDENTICAL` or `!==`. Even if the language does not support it directly, the compiler does afaik since at least groovy++. That way there would be no even a dynamic call.

erdi:
These are the results I'm getting. I also included an implementation using `COMPARE_NOT_IDENTICAL` as suggested by Jochen. Let me know what you want me to do. :)

pre-PR, null property/field
non-indy 3k
indy 100K

pre-PR, non-null property/field
non-indy 25k
indy 27k

`is()` receiver and arg switched, null property/field
non-indy 31k
indy 33k

`is()` receiver and arg switched, non-null property/field
non-indy 27k
indy 28k

`or`, null property/field
non-indy 39k
indy 99k

`or`, non-null property/field
non-indy 23k
indy 28k

`COMPARE_NOT_IDENTICAL`, null property/field
non-indy 40k
indy 98k

`COMPARE_NOT_IDENTICAL`, non-null property/field
non-indy 32k
indy 86k

blackdrag:
welll... for me that speaks clearly for COMPARE_NOT_IDENTICAL

paulk-asert:
+1 for COMPARE_NOT_IDENTICAL

erdi:
I updated the fix to use `COMPARE_NOT_IDENTICAL` and force pushed.

erdi:
That's a bummer but it is what it is. :)

PR #634, opened 2017-11-12
GROOVY-7315: CompileStatic/TypeChecked cannot create non-static neste…

paulk-asert:
…d inner class using named-arg short-hand syntax


paulk-asert:
We could potentially optimise this a little further by using node metadata instead of redoing some of the checks for the inner class case but I'll likely leave that as a future activity.

PR #638, opened 2017-11-17
Remove deprecated StringGroovyMethods.capitalize call

russellsanborn:
I also noticed that propName was being capitalized three times in the createComparatorFor() method, so I only call it once at the beginning.

I followed the same style of passing in the CharSequence like it was defined in the deprecated method:
```groovy
@Deprecated
public static String capitalize(String self) {
    return capitalize((CharSequence) self);
}
````

If you would rather have something like the following, or something else let me know:
```groovy
CharSequence propName = property.getName();
String capitalizedPropName = StringGroovyMethods.capitalize(propName);

```


paulk-asert:
Merged, thanks.

PR #642, opened 2017-12-05
Fix documentation to say that private fields are turned into properties

alexstajos:
I'm new to Groovy, and this section of the documentation was confusing to me and seemed to contradict what is in the [Properties](http://docs.groovy-lang.org/latest/html/documentation/#properties) section. I think this is change would be correct.


paulk-asert:
Actually, private isn't right but what is there isn't as clear as it should be either. I'll adjust both places to make it clearer.

paulk-asert:
I made the documentation clearer. Let me know what you think. It won't appear on the website until the next release but you can check the source for now.

alexstajos:
Looks good, thanks for taking the time to update!

PR #643, opened 2017-12-11
GROOVY-8241 SAM parameter type inference for explicit parameter

avafanasiev:
It should fix GROOVY-8241, GROOVY-7061, GROOVY-8317.

### DIFF
--- a/src/test/groovy/transform/stc/MethodCallsSTCTest.groovy
+++ b/src/test/groovy/transform/stc/MethodCallsSTCTest.groovy
@@ -364,6 +364,29 @@ class MethodCallsSTCTest extends StaticTypeCheckingTestCase {
             '''
     }
 
+    void testShouldNotFailThanksToInstanceOfChecksAndWithoutExplicitCasts2() {
+        assertScript '''
+                static String foo(String s) {
+                    'String'
+                }
+                static String foo(Integer s) {
+                    'Integer'
+                }
+                static String foo(Boolean s) {
+                    'Boolean'
+                }
+                ['foo',123,true].each { argument ->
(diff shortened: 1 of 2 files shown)


danielsun1106:
+1
@paulk-asert any thoughts on the PR?

danielsun1106:
Merged. Thanks :-)

avafanasiev:
Thanks for merge!

src/test/groovy/transform/stc/MethodCallsSTCTest.groovy
```diff
@@ -383,6 +406,94 @@ class MethodCallsSTCTest extends StaticTypeCheckingTestCase {
...
+        assertScript '''
+            import java.util.function.Predicate
+            
+            static boolean foo(Predicate<? super String> p) {
+                p.test("foo")
+            }
+            
+            static def testPredicate() {
+                foo { it ->
+                    it.toUpperCase()
```
daniel-huss: I'm confused :) This works because `foo` happens to always pass a String to `p::test`, but what if it passed `new Object()` instead? If testPredicate were  `@CompileStatic` I would expect a compile error here. Is this disparity intentional?

avafanasiev: Actually you can't pass anything  except String to Predicate<? super String> p . Cause it should be able to be assigned to captured type '? super String' . This behavior matches java's lambda behavior.

daniel-huss: Oh, you're right. I forgot Groovy's interpretation of <? super X> is simply <X> :)

PR #645, opened 2017-12-18
DefaultGroovyMethods bug fix: Set.intersect(Iterable) throwing if the iterable is bigger than the set

hlerebours:


subprojects/groovy-groovysh/src/main/groovy/org/codehaus/groovy/tools/shell/completion/ImportsSyntaxCompletor.groovy
```diff
@@ -105,12 +105,12 @@ class ImportsSyntaxCompletor implements IdentifierCompletor {
...
                     List<String> clazzSymbols = ReflectionCompletor.getPublicFieldsAndMethods(clazz, '')*.value
-                    List<String> importedSymbols;
+                    Collection<String> importedSymbols
                     if (symbolName == '*') {
-                        importedSymbols = clazzSymbols;
+                        importedSymbols = clazzSymbols
                     } else {
                         Set<String> acceptableMatches = [symbolName, symbolName + '(', symbolName + '()']
-                        importedSymbols = acceptableMatches.intersect(clazzSymbols)
+                        importedSymbols = (acceptableMatches as Collection).intersect(clazzSymbols)
```
paulk-asert: Was changing this file needed?

hlerebours: Yes, because intersect() being called on a Set, it returns a Set now, not a List. So without this change, the assignment to importedSymbols throws a ClassCastException.

About the cast of the Set to Collection, it was just for the call to intersect() to not be ambiguous, for there were two candidates: intersect(Set, Iterable) and intersect(Collection, Collection), the first one just calling the latter.

paulk-asert:
Merged, thanks!

PR #648, opened 2017-12-24
GROOVY-8423: Do not adjust access modifiers for inner class table ent…

shils:
…ries

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/AsmClassGenerator.java
+++ b/src/main/java/org/codehaus/groovy/classgen/AsmClassGenerator.java
@@ -305,11 +305,6 @@ public class AsmClassGenerator extends ClassGenerator {
         }
         int mods = adjustedClassModifiersForInnerClassTable(cn);
 
-
-        if (Modifier.isPrivate(mods)) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/classgen/AsmClassGenerator.java
```diff
@@ -305,11 +305,6 @@ private void makeInnerClassEntry(ClassNode cn) {
         }
         int mods = adjustedClassModifiersForInnerClassTable(cn);
 
-
-        if (Modifier.isPrivate(mods)) {
```
shils: I'm unsure about removing this block as the commit message suggests it was necessary for android support. Does anyone know the history behind this? @melix

paulk-asert:
+1 after android support feedback received. I didn't work on that aspect.

melix:
I'm sorry I was sick the last few days I missed the ping. I can clearly remember that without this change the class format wasn't correct for Android. But it's far away and I haven't worked on this for years now. I'm afraid I'm no help.

shils:
Thanks for reviewing, and hope you're feeling better. I'll come back to this once I've done more in-depth testing.

PR #654, opened 2018-01-20
Support native lambda in static compilation mode

danielsun1106:
Current implementation has supported:

* common scenarios of Java's lambda
* modifying non-final local variable
* callable lambda

Here are the tests:
https://github.com/apache/groovy/blob/native-lambda/src/test/groovy/transform/stc/LambdaTest.groovy


blackdrag:
Do we really need Lambda in groovy.lang? We do have Closure and Lambda extends Closure, why should we have Lambda in groovy.lang as well? There is no need to have it as parameter in any API imho. Instead I would go directly for GeneratedLambda only. Also we should consider making the new classes in here use the apache package structure instead of the codehaus one. Then about FunctionalInterface... I thought this is a documentation only marking of interfaces. Imho there is no requirement to have this annotation to use lambdas, not in Groovy and not in Java. So I would remove that kind of logic. You added getClassVisitor to ACG. I think that is available through the controller, so no need to add this. Then I do not understand the change you did with CorrectAccessedVariableVisitor. Ignoring that you made it an explicit inner class, why did you change it to extend ClassCodeVisitorSupport instead of CodeVisitorSupport? The usage is the same. Giving this visitor more capabilties that are not used is not good, a refactoring is bound to make something out of this, that it is not. Also you should undo the method visibility of removeInitialValues back to private. You are not using the method outside of the class. Then your triple DUP...what is this for? Later I see a lot of POP, which you are not doing. This hints me, that the DUPs are not required and that you do not clean the operand stack up, instead you leave it to the compiler infrastructure, which can do this to some etend, but you should not depend on it. So please add the POPs yourself. or avoid the surplus looking DUPs.

danielsun1106:
Hi Jochen( @blackdrag  ),

     Thanks a lot for your reviewing the code.

> Do we really need Lambda in groovy.lang? We do have Closure and Lambda extends Closure, why should we have Lambda in groovy.lang as well?

I agree with you, I've removed `Lambda`

> Then about FunctionalInterface... I thought this is a documentation only marking of interfaces. Imho there is no requirement to have this annotation to use lambdas, not in Groovy and not in Java. So I would remove that kind of logic.

You're right, I've remove the check.

> Then I do not understand the change you did with CorrectAccessedVariableVisitor. 

I want to reuse its code. As you can see, `TransformationVisitor` extends `CorrectAccessedVariableVisitor` 
https://github.com/apache/groovy/blob/dbcac60f64f921cd5e8ee02c4af126c9a6d8b883/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesLambdaWriter.java#L413

>  why did you change it to extend ClassCodeVisitorSupport instead of CodeVisitorSupport?

I need to `visitMethod`, which does not exist in `CodeVisitorSupport`. Maybe I revert it to `CodeVisitorSupport` and delegate `visitVariableExpression` to a `CodeVisitorSupport` instance
https://github.com/apache/groovy/blob/dbcac60f64f921cd5e8ee02c4af126c9a6d8b883/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesLambdaWriter.java#L334

> Also you should undo the method visibility of removeInitialValues back to private. You are not using the method outside of the class.

I use `removeInitialValues` in class `StaticTypesLambdaWriter`, so we have to make it `protected`...
https://github.com/apache/groovy/blob/dbcac60f64f921cd5e8ee02c4af126c9a6d8b883/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesLambdaWriter.java#L352

> Then your triple DUP...what is this for? Later I see a lot of POP, which you are not doing. This hints me, that the DUPs are not required and that you do not clean the operand stack up, instead you leave it to the compiler infrastructure, which can do this to some etend, but you should not depend on it. So please add the POPs yourself. or avoid the surplus looking DUPs.

Jochen, I tried to pop just now, but failed.. the stack seems to be empty already(see `java.lang.ArrayIndexOutOfBoundsException: size==0 `). Any further hints for me?
https://github.com/apache/groovy/blob/552e01039aaad874a1409f4029df132817c7b0bb/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesLambdaWriter.java#L163-L177

PR #661, opened 2018-02-08
Documentation: fix style-guide range example

jja:
e674d77 added more information about ranges
but included erroneous code for `range.size()`.

The style guide does not seem to have tests associated like some of the other
docs. If you want me to create a test, please give me some guidance.

I don't think code sample from e674d77 was tested. It was probably just
an ad hoc example written on the fly.


jja:
...an alternative fix would be to change the ranges `from` values to `0`

danielsun1106:
Merged. Thanks!

paulk-asert:
Thanks John for the fix and Daniel for merging. I moved the relevant example into a test - so it should remain working from now on. Thanks again!

PR #662, opened 2018-02-13
GROOVY-8477: @Immutable-related transformations should be more config…

paulk-asert:
…urable

### DIFF
--- a/src/main/groovy/groovy/transform/Immutable.groovy
+++ b/src/main/groovy/groovy/transform/Immutable.groovy
@@ -18,6 +18,8 @@
  */
 package groovy.transform
 
+import groovy.transform.construction.ImmutablePropertyHandler
+
 /**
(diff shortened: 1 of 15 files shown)


paulk-asert:
The visibility of various methods needs to be looked at but with the proposed PR and a locally hacked Guava property handler, I get the following result:
{code}
import groovy.transform.Immutable
import groovy.transform.construction.GuavaImmutablePropertyHandler
@Immutable(propertyHandler=GuavaImmutablePropertyHandler)
class Person {
  List names = ['John', 'Smith']
  List books = ['GinA', 'ReGinA']
}

['names', 'books'].each {
  println new Person()."$it".dump()
}
//<com.google.common.collect.RegularImmutableList@90b9bd9 array=[John, Smith]>
//<com.google.common.collect.RegularImmutableList@95b86f34 array=[GinA, ReGinA]>
{code}

PR #663, opened 2018-02-14
GROOVY-8471: Contents of META-INF/services/org.codehaus.groovy.source…

paulk-asert:
….Extensions conflict with Maven/Jisaw


ceki:
Looks easy enough.

PR #668, opened 2018-02-28
Fix XmlParser default constructor Javadoc

sergeyh:
Currently, Javadoc says it creates "non-validating and non-namespace-aware" parser, but in fact, it _is_ namespace aware.


paulk-asert:
Merged, thanks!

PR #669, opened 2018-03-02
Fix ArrayIndexOutOfBoundsException in CallSiteWriter.getCreateArraySignature()

Linary:
It would throw `ArrayIndexOutOfBoundsException` if the `numberOfArguments` exceed 255,
but there is no check for array index and it's hard to get the root cause to users.

We should throw a more readable exception here.


danielsun1106:
Failed to compile, here is the error message:

```
/home/travis/build/apache/groovy/src/main/java/org/codehaus/groovy/classgen/asm/CallSiteWriter.java:76: error: ')' expected
                      numberOfArguments);
                                        ^
1 error
:compileJava FAILED
```

danielsun1106:
Merged. Thanks!

PR #671, opened 2018-03-05
GROOVY-8422: Incorrect properties copy in Sql.newInstance

jwagenleitner:
The provided Properties should be passed to the DriverManager as-is.
A copy is only needed when changes are made to the provided Properties
in order to mask sensitive information for logging purposes.


paulk-asert:
+1

PR #672, opened 2018-03-06
GROOVY-7624: Collection asImmutable() methods aren't immutable

paulk-asert:


paulk-asert:
The proposed PR provides `asUnmodifiable()` aliases for the current behavior. The `asImmutable()` variants are then using a copy of the original, making them effectively immutable.

PR #676, opened 2018-03-22
Add DGM `md5`

danielsun1106:


src/test/org/codehaus/groovy/runtime/DefaultGroovyMethodsTest.groovy
```diff
@@ -310,4 +312,18 @@ class DefaultGroovyMethodsTest extends GroovyTestCase {
         assertTrue(DefaultGroovyMethods.implies(false, null))
         assertFalse(DefaultGroovyMethods.implies(true, null))
     }
+
+    void testMd5() {
+        try {
+            Assert.assertEquals("e99a18c428cb38d5f260853678922e03", StringGroovyMethods.md5("abc123"))
```
paulk-asert: StringGroovyMethods should be DefaultGroovyMethods or the DGM method should move? In any case, perhaps EncodingGroovyMethods might be better?

paulk-asert:
Needs a Jira ticket

src/test/org/codehaus/groovy/runtime/DefaultGroovyMethodsTest.groovy
danielsun1106: My bad... we should use `DefaultGroovyMethods`

danielsun1106:
OK. I'll create a Jira ticket later

PR #678, opened 2018-03-30
GROOVY-8525: Binary compatibility issue for GroovyClassLoader between…

paulk-asert:
… 2.4 vs later branches


paulk-asert:
see #679

PR #679, opened 2018-03-30
GROOVY-8525: Binary compatibility issue for GroovyClassLoader between…

paulk-asert:
… 2.4 vs later branches (alternative to PR#678)

What I haven't yet checked is whether we need to make some of the delegated method calls just throw UnsupportedOperationException in order to keep the claims like @Threadsafe.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
+++ b/src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
@@ -31,7 +31,6 @@ import java.util.concurrent.locks.StampedLock;
  * but it is not reentrant, in other words, <b>it may cause deadlock</b> if {@link #getAndPut(K, ValueProvider)} OR {@link #getAndPut(K, ValueProvider, boolean)} is called recursively:
  * readlock -> upgrade to writelock -> readlock(fails to get and wait forever)
  *
- *
  * @param <K> type of the keys
  * @param <V> type of the values
  * @since 3.0.0
@@ -95,7 +94,7 @@ public class StampedCommonCache<K, V> implements EvictableCache<K, V>, ValueConv
      * {@inheritDoc}
      */
     @Override
-    public V get(final K key) {
+    public V get(final Object key) {
         return doWithReadLock(c -> c.get(key));
     }
 
@@ -175,6 +174,11 @@ public class StampedCommonCache<K, V> implements EvictableCache<K, V>, ValueConv
         return doWithReadLock(c -> c.values());
     }
 
+    @Override
+    public Set<Entry<K, V>> entrySet() {
+        return doWithReadLock(c -> c.entrySet());
+    }
+
     /**
      * {@inheritDoc}
      */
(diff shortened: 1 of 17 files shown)


src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
```diff
@@ -175,6 +174,11 @@ private V compute(K key, ValueProvider<? super K, ? extends V> valueProvider, bo
         return doWithReadLock(c -> c.values());
     }
 
+    @Override
+    public Set<Entry<K, V>> entrySet() {
+        return commonCache.entrySet();
```
danielsun1106: The code should be protected by `doWithReadLock`, we should do same thing for `isEmpty` `containsValue`, etc.

src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
```diff
@@ -199,20 +208,35 @@ public int size() {
...
      */
     @Override
-    public V remove(final K key) {
+    public V remove(final Object key) {
         return doWithWriteLock(c -> c.remove(key));
     }
 
+    @Override
+    public void putAll(Map<? extends K, ? extends V> m) {
+
```
danielsun1106: The implementation is missing and should be protected by `doWithWriteLock`

src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
paulk-asert: true, but I think we have to throw UnsupportedOperationException for entrySet and keySet because they return a view which might then be modified under the covers without locks?

src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
paulk-asert: Oops, I was wondering whether to lock for the whole putAll or loop and lock per entry and then forgot to supply either! I'll do the former for now.

src/main/java/org/codehaus/groovy/runtime/memoize/StampedCommonCache.java
danielsun1106: Implementing an interface and throwing `UnsupportedOperationException` does not look elegant...

paulk-asert: Inelegant but the Java collections way! ;-) Better to look inelegant rather than provide an apparently good looking but buggy implementation! In any case, I have implemented entrySet() in the same way as keys() and values() are implemented. The concern though is that if you look at ConcurrentHashMap, keySet(), entrySet() and values() are backed by views using iterators and spliterators that guarantee "weak consistency". It doesn't seem like we are making any such considerations for StampedCommonCache? I will leave that as a future exercise - we can always create such views or throw UnsupportedOperationException if we deem there is no other way to be safe.

danielsun1106: I will set aside some time to complete the exercise 😉

PR #686, opened 2018-04-17
@Newify table name pattern support

mgroovy:
Added class name pattern parameter support to `@Newify` annotation (e.g. `@Newify(pattern=/[A-Z].*/)`) that allows users to define the classes with newify support through the pattern in addition to the existing class list parameter.


src/main/java/org/codehaus/groovy/transform/NewifyASTTransformation.java
```diff
@@ -196,39 +322,10 @@ private boolean shouldTransform(DeclarationExpression exp) {
     }
 
     private boolean hasClassesToNewify() {
-        return classesToNewify != null && !classesToNewify.getExpressions().isEmpty();
+        return (classesToNewify != null && !classesToNewify.getExpressions().isEmpty()) || (classNamePattern != null); // 2018-04-07 TODO: Check if classNamePattern != null can be made more "early out"
```
danielsun1106: I see `TODO` here, are you going to refine it later?

src/main/java/org/codehaus/groovy/transform/NewifyASTTransformation.java
```diff
@@ -286,31 +372,106 @@ private static boolean isNewMethodStyle(MethodCallExpression mce) {
...
         }
+        
         if (classType != null) {
-            return new ConstructorCallExpression(classType, args);
+            Expression argsToUse = argsExp;
+            //if(classType.getOuterClass() != null && !classType.isStaticClass()) {
+            if(classType.getOuterClass() != null && ((classType.getModifiers() & org.objectweb.asm.Opcodes.ACC_STATIC) == 0)) {
+                if(!(argsExp instanceof ArgumentListExpression)) {
+                    addError("Non-static inner constructor arguments must be an argument list expression; pass 'this' pointer explicitely as first constructor argument otherwise.", mce);
+                    return mce; // TODO: Expected behavior to return  the untransformed method call here ?
```
danielsun1106: Similarly, I see `TODO` here, are you going to refine it later? ;-)

src/main/java/org/codehaus/groovy/transform/NewifyASTTransformation.java
mgroovy: This TODO refers to e.g. an edge case optimization to defer the class name Pattern compilation to the point where is actually accessed. I don't expect this to make sense (also I am now checking against the class name pattern before checking against the class name list, since I expect this to be faster in practice), so I have not implemented it. Should probaly not be a TODO any more, but at most a regular comment, I can change this and redo the PR if you want.

src/main/java/org/codehaus/groovy/transform/NewifyASTTransformation.java
mgroovy: This is the first Groovy extension I am doing on my own, and I took this code part from existing Groovy code, and wanted to make sure somebody more experienced had a look at this. If you say this code is fine, then the TODO can be removed (see above comment).

danielsun1106:
I failed to merge the PR. Could you squash the commits of the PR and create a new PR?
P.S. this is a mirror, we can not merge PR via github

mgroovy:
Hi Daniel, will do (I also added two more tests (for the import alias case) and removed the TODOs).

danielsun1106:
As the PR #689 supersedes the PR. Please close it. Thanks.
P.S. I failed to close it by myself via commit comment `closes #686`...

mgroovy:
(obsolete since superseded by #689 => closing)

PR #691, opened 2018-04-26
GROOVY-8561: Support java-like array

danielsun1106:
## Background
As we all know, Java array is one of features widely applied in Java projects. In order to improve the compatibility with Java(Copy & Paste), the PR on the new Parrot parser will make Groovy support java-like array. 
### One-Dimensional array
```groovy
String[] names = {'Jochen', 'Paul', 'Daniel'}
assert ['Jochen', 'Paul', 'Daniel'].toArray() == names
```
```groovy
String[] names = {'Daniel'} // Note: this is an array, not a closure
assert ['Daniel'].toArray() == names
```
```groovy
String[] names = {} // Note: this is an array, not a closure
assert 0 == names.length
```
### Two-Dimensional array
```groovy
int[][] data = {
    {1, 2, 3},
    {4, 5, 6},
    {7, 8, 9},
    new int[] { 10, 11, 12 },
    {13, 14, 15}
}
assert [1, 2, 3].toArray() == data[0]
assert [4, 5, 6].toArray() == data[1]
assert [7, 8, 9].toArray() == data[2]
assert [10, 11, 12].toArray() == data[3]
assert [13, 14, 15].toArray() == data[4]
```
```groovy
int[][] data = new int[][] {
    {1, 2, 3},
    {4, 5, 6},
    {7, 8, 9},
    new int[] { 10, 11, 12 },
    {13, 14, 15}
}
assert [1, 2, 3].toArray() == data[0]
assert [4, 5, 6].toArray() == data[1]
assert [7, 8, 9].toArray() == data[2]
assert [10, 11, 12].toArray() == data[3]
assert [13, 14, 15].toArray() == data[4]
```
```groovy
int[][] data = {
    {3},
    {6},
    {9},
    new int[] {12},
    {15},
    {}
}
assert [3].toArray() == data[0]
assert [6].toArray() == data[1]
assert [9].toArray() == data[2]
assert [12].toArray() == data[3]
assert [15].toArray() == data[4]
assert 0 == data[5].length
```
### Annotation array
```groovy
import java.lang.annotation.RetentionPolicy
import java.lang.annotation.Retention

public @interface PropertySource {
    String value();
}

public @interface PropertySources {
    PropertySource[] value();
}

@PropertySources({
    @PropertySource("classpath:1.properties"),
    @PropertySource("file:2.properties")
})
public class Controller {}

def propertySources = Controller.getAnnotation(PropertySources).value()
assert 2 == propertySources.length
def values = propertySources.collect { it.value() }
assert ['classpath:1.properties', 'file:2.properties'] == values
```
### More array examples
Here are some tests[[1]](https://github.com/danielsun1106/groovy/blob/f20deb8bd6a1a90a4bfee8458e13438f7adf3732/subprojects/parser-antlr4/src/test/resources/core/Array_02x.groovy) [[2]](https://github.com/danielsun1106/groovy/blob/c78b95888a176d05b3fd410f45866b2ceed8344c/subprojects/parser-antlr4/src/test/resources/core/Annotation_11x.groovy) for reference.
## Limitation
The array type must be specified when declaring array, e.g.
```groovy
int[][] a = {{1, 2, 3}, {4, 5, 6}}  // valid
def a = {{1, 2, 3}, {4, 5, 6}}   // invalid
```

## Known breaking changes
@paulk-asert  added some more details about breaking changes and suggested alternatives in the jira issue: https://issues.apache.org/jira/browse/GROOVY-8561
#### Before
  ```groovy
Closure[] y = { {-> 1 + 1 } }
assert y[0].call().call() == 2
  ```
#### After
  ```groovy
Closure[] y = { {-> 1 + 1 } }
assert y[0].call() == 2
  ```
#### Before
  ```groovy
String[] a = {}
assert 1 == a.length
assert a[0].contains('closure')
  ```
#### After
  ```groovy
String[] a = {}
assert 0 == a.length
  ```
## Provisos(by @paulk-asert )
* we need to update the breaking changes section of 3.0.0 release notes on the website (site/src/site/releasenotes/groovy-3.0.adoc) to include the cases listed in GROOVY-8561
* we need to update the "Array initializers" section of core-differences-java.adoc (we need to explain what cases remain - obviously if no array type is specified is one case - I haven't checked whether Java allows arrays of lambdas or something similar which we may not support - I don't believe so but haven't researched into it)
* we need to update the "Arrays" section of core-syntax.adoc (I think we need to indicate that square brackets is the idiomatic approach and also explicitly say that closure is the default interpretation of curly braces unless the array target type can be inferred from the context.)


paulk-asert:
I believe this change deserves wide feedback. Before that I think we need to expand the description above just a little more. We should highlight a bit more what things are consistent/inconsistent before/after the change. I'll try to have a go shortly - just have a few other things on the todo list before then - but feel free to expand if you have time.

pditommaso:
Will not this clash with the closure syntax? For example the following snippet will be interpreted as a closure or one item array? 

```
def names = {'Daniel'} 
```

danielsun1106:
closure as the array type is not specified.

PR #693, opened 2018-05-03
Allow methodMissing/propertyMissing to be defined through category

ecerulm:
Resolves GROOVY-3867

### DIFF
--- a/src/main/groovy/groovy/lang/MetaClassImpl.java
+++ b/src/main/groovy/groovy/lang/MetaClassImpl.java
@@ -950,7 +950,6 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
             if (method == null && GroovyCategorySupport.hasCategoryInCurrentThread()) {
                 method = getCategoryMethodMissing(instanceKlazz);
                 if (method != null) {
-                    //TODO add onMethodMissingFound(method)
                     return method.invoke(instance, new Object[]{methodName, arguments});
                 }
             }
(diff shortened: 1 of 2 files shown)


src/main/groovy/groovy/lang/MetaClassImpl.java
```diff
@@ -945,6 +945,15 @@ private Object invokeMissingMethod(Object instance, String methodName, Object[]
                 onInvokeMethodFoundInHierarchy(method);
                 return method.invoke(instance, invokeMethodArgs);
             }
+
+            // last resort look in the category
+            if (method == null && GroovyCategorySupport.hasCategoryInCurrentThread()) {
+                method = getCategoryMethodMissing(instanceKlazz);
+                if (method != null) {
+                    //TODO add onMethodMissingFound(method)
```
danielsun1106: I see a `TODO` here. Are you going to refine it later?

ecerulm: No, I think I should just remove that TODO. Currently there is no onMethodMissingFound() for the "regular" methodMissing() (not from a category). As far as I understand the on*Found() hooks are only used by the ExpandoMetaClass to store/cache resolved methods, and I guess the missingMethod from a category should not be remembered (it should disappear after the category is no longer in effect).

danielsun1106:
Merged. Thanks!
https://github.com/apache/groovy/commit/ad664b181249c4b8e8821003d7e750632943ce36

PR #701, opened 2018-05-10
GROOVY-8520 fixes (attempt 3)

remkop:
Rebased on master, replaces PR #696


paulk-asert:
Merged. I took 3 of the 4 commits. I reworked the groovy.util.CliBuilder to delegate to the commons-cli version. It seems better not to pollute the picocli version for the sake of providing functionality to the groovy.util version which we are possibly going to delete in 2.6/3. I think most people are going to see that the groovy,util version is deprecated and try to move off it quickly anyway.

PR #705, opened 2018-05-12
GROOVY-7883  Static compiler prefers private constructor over public …

danielsun1106:
…if private matches be

### DIFF
--- a/subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
+++ b/subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
@@ -3973,7 +3973,7 @@ public class Sql {
      * @return the resulting list of rows
      * @throws SQLException if a database error occurs
      */
-    protected List<GroovyRowResult> asList(String sql, ResultSet rs,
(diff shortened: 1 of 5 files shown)


subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
```diff
@@ -3973,7 +3973,7 @@ protected final ResultSet executePreparedQuery(String sql, List<Object> params)
      * @return the resulting list of rows
      * @throws SQLException if a database error occurs
      */
-    protected List<GroovyRowResult> asList(String sql, ResultSet rs,
+    public List<GroovyRowResult> asList(String sql, ResultSet rs,
```
danielsun1106: The test accesses the `protected` method `asList`, which is not allowed. so I change its visibility to `public`.
groovy.sql.SqlSTCTest#testAsList
```
    void testAsList() {
        shell.evaluate '''
            def test(groovy.sql.Sql sql, java.sql.ResultSet rs) { 
                sql.asList('SELECT * FROM FOO', rs) { println it.columnCount } 
            }
        '''
    }
```

PR #706, opened 2018-05-15
GROOVY-8355: Instanceof inference does not work on field assigning

danielsun1106:


jwagenleitner:
According to the Cédric's comment on this issue, `instanceof` is a special case. If the decision is to support it based on the type assigned then the docs may need some updating as well.

http://docs.groovy-lang.org/next/html/documentation/core-semantics.html#type-inference

> At compile time, we can’t make any guarantee about the type of a field. Any thread can access any field at any time and between the moment a field is assigned a variable of some type in a method and the time is is used the line after, another thread may have changed the contents of the field. This is not the case for local variables: we know if they "escape" or not, so we can make sure that the type of a variable is constant (or not) over time. Note that even if a field is final, the JVM makes no guarantee about it, so the type checker doesn’t behave differently if a field is final or not.

/cc @melix @paulk-asert

paulk-asert:
Yes, the pre-existing behavior is by design. While I think there are interesting discussions to be had about the intuitiveness of either alternative behavior, I think the existing behavior is the more conservative for similar reasons to why Java didn't allow var for fields/method types:
https://blog.codefx.org/java/java-10-var-type-inference/#Why-Can8217t-Field-And-Method-Types-Be-Inferred
I would certainly expect such a change to be discussed on the mailing list and a [VOTE][LAZY] if needed,

danielsun1106:
OK. I will revert it for the time being.

melix:
Agreed. Type inference is the hardest thing in the type checker, because every decision we make has to be carefully discussed. As Paul said, the reasons we don't do type inference on fields is the same reason why `var` isn't allowed on fields for Java (which makes me think we made a good decision because those people are way smarter than I am).

danielsun1106:
Yep. Type inference is really hard. Fixing some issues may break existing test cases easily... so some of them even cost my hours... but type inference is quite useful and interesting 😉

P.S. I have reverted the PR

PR #707, opened 2018-05-16
GROOVY-8509: SC error call to protected method from same package

jwagenleitner:
Because they were closely related, this PR also includes the following:

1. A change to the fix introduced by GROOVY-7883 (b1d1232770aa) to prevent filtering protected methods in the `StaticTypeCheckingVisitor` if caller is in the same package. From the same commit, removes the test `Groovy7883Bug#test3` because it should compile and the new test introduced for GROOVY-8509 in `MethodCallsStaticCompilation` takes it place.

1. A change to revert groovy.sql.Sql#asList (b1d1232770aa) back to `protected`. Fix for GROOVY-7883 enforced visibility for method calls in `TypedChecked` and `CompileStatic` modes. Test was originally put in place to verify the added ClosureParams, so changed test to access method via subclass so that the access modifier for the method can remain protected. I think the spirit of the test is preserved in this case without having to open access to the method under test.

### DIFF
--- a/src/main/java/org/apache/groovy/ast/tools/ClassNodeUtils.java
+++ b/src/main/java/org/apache/groovy/ast/tools/ClassNodeUtils.java
@@ -310,4 +310,20 @@ public class ClassNodeUtils {
         }
         return false;
     }
+
+    /**
+     * Determine if the given ClassNode values have the same package name.
+     *
+     * @param first a ClassNode
+     * @param second a ClassNode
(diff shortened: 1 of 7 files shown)


danielsun1106:
+1

paulk-asert:
+1, John, let me know if you'll get a chance to commit in the next day. If not, I'll merge before producing the rc-3 candidate.

jwagenleitner:
Thank you both for the review.

PR #708, opened 2018-05-17
GROOVY-6668: Static compiler doesn't coerce GString for getAt() call

danielsun1106:


src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticInvocationWriter.java
```diff
@@ -468,7 +469,11 @@ protected void loadArguments(List<Expression> argumentList, Parameter[] para) {
                 expression.putNodeMetaData(PARAMETER_TYPE, para[i].getType());
                 expression.visit(acg);
                 if (!isNullConstant(expression)) {
-                    operandStack.doGroovyCast(para[i].getType());
+                    if (expression instanceof GStringExpression) {
+                        operandStack.doGroovyCast(ClassHelper.STRING_TYPE);
```
blackdrag: what if the method has a GString parameter? Converting to string is then wrong. You should add a test for this

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -4101,13 +4101,17 @@ protected static ClassNode getGroupOperationResultType(ClassNode a, ClassNode b)
...
+    }
+
+    protected ClassNode inferComponentType(final ClassNode containerType, ClassNode indexType) {
         final ClassNode componentType = containerType.getComponentType();
         if (componentType == null) {
             // GROOVY-5521
             // try to identify a getAt method
             typeCheckingContext.pushErrorCollector();
-            MethodCallExpression vcall = callX(varX("_hash_", containerType), "getAt", varX("_index_", indexType));
+            MethodCallExpression vcall = callX(varX("_hash_", containerType), "getAt", varX("_index_", convertGStringToStringType(indexType)));
```
blackdrag: what if I add a DGM getAt(GString) for Map? This code looks like I would no longer be able to call it

src/test/groovy/bugs/Groovy6668Bug.groovy
```diff
@@ -0,0 +1,65 @@
...
+        @groovy.transform.CompileStatic
+        class OtherThing {
+            OtherThing() {
+                Map<String, String> m = [:]
+                def k = "foo"
+                m["$k"].toUpperCase()
+            }
+        }
+        
+        OtherThing
```
blackdrag: new OtherThing()?

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
danielsun1106: Good idea! 
```groovy
    /**
     * Get the value via the string value of the key
     * See GROOVY-6668 and the PR#708 (https://github.com/apache/groovy/pull/708)
     *
     * @param self a map
     * @param key a key
     * @param <V> value type
     * @return the value
     */
    public static <V> V getAt(Map<?, V> self, GString key) {
        if (null == key) {
            return null;
        }

        return self.get(key.toString());
    }
```

but I wonder why we should not treat GString as String?
As a groovy user, I won't write code like `GString str = "${'someString'}"`, and I just write `String str = "${'someString'}"` or  `def str = "${'someString'}"` then treat `str` as a `String`...

src/test/groovy/bugs/Groovy6668Bug.groovy
danielsun1106: I just keep the code of the issue as it is. the `new OtherThing()` version is `testGroovy6668WithData` ;-)

paulk-asert: why not just shouldCompile then?

danielsun1106: I've fixed that according your suggestion :-)

melix:
@danielsun1106 If you need reviews from me, especially "within 72h", please make sure to mention me on GitHub, I get a ton of emails from JIRA, there's almost no chance I notice a mention there.

danielsun1106:
@melix OK. I will mention you in time.
BTW, I wonder why we don't treat all `GString` as ` String`. For me, `GString` is just an internal class of Groovy, I never use its API. In addition, as we all know, `String` is widely used in Java world, so I believe few user is willing to use `GString` but just treat `GString` as a powerful version of `String`.

melix:
If `GString` had been implemented _eager_, it could have been transparent to the user, and never exposed as a public API. But that's not how it is, so we cannot treat it as such. Let me give examples:

```
def map = [:]
map[123] = "Hello $world"
```

then pass it to Java code. What should it see in the map ? A `String` ? A `GString` ? If it's a string, then it needs to be coerced as soon as it is put into the map. If it's a GString, then you're exposing it as a public API, and it means that it needs not to be coerced when you get it from the map. If so, why would you coerce the key but not the value? What if I write this:

```
def map = [:]
map[123] = "some $thing"
println(map[123].strings) // 1
def x = map[123] // 2
String x = map[123] // 3
GString x = map[123] // 4
```
1 should always work, because what we have put in the map is a GString, so we shouldn't coerce.
2 should also return a GString
3 uses explicit coercion, so you're asking for the evaluation of the GString
4 is explicitly typed, and nothing happens

It's all valid scenarios. So handling coercion for methods is one thing (coerce to a String when you have a method that takes a String), but handling them as a single object everywhere is another. This issue is about handling the coercion of a special "hidden" method call (`getAt(String)`) and is as such a bit special, I agree, but it doesn't mean we can do coercion everywhere.

danielsun1106:
@melix `"Hello $world"` seems to be evaluated eagerly. As `world` is not defined, error occurred.
```groovy
def map = [:]
map[123] = "Hello $world"
```

```
groovy.lang.MissingPropertyException: No such property: world for class: ConsoleScript3
```

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
blackdrag: we can treat GString as a String if we know the target is a String. Plus the rules for assignment and for method calls are not identical. But my comment was not about adding getAt(GString) for Map. My comment was that this change looks like the following will no longer work:
  
      class X {
        int getAt(GString gs) {1}
      }
      assert new X()["${42}"] == 1

blackdrag:
as there is some confusion about the eagerness of GString:
* the referenced variables are bound eager
* toString on GString evaluates the GString, this includes toString for the values
In this example: 
```groovy
def x = 1
def gstring = "$x"
assert gstring == "1"
x = 2
assert gstring == "1"
```
we can see that gstring is bound to what is referenced by x. It seems like a complete eager evaluation, but that is hidden by the fact that x is not mutated here:
```groovy
class X {
  int i = 0
  String toString(){Integer.toString(i++)}
}
def x = new X()
def gstring = "$x"
assert gstring == "0"
assert gstring == "1"
```
as can be seen here, each conversion to String will reevaluate the GString and since the toString of X is for every call different, the GString will be evaluated to a different String each time.

blackdrag:
also I should add that I think now that most users actually would want to have the eager variant

melix:
@danielsun1106 of course it fails, I didn't expect you to _run_ the code, it was an illustration that `GString` and `String` are different. It doesn't matter if it's eagerly evaluated or not in the examples, what matters is that they are different objects.

danielsun1106:
```groovy
class X {
  int i = 0
  String toString(){Integer.toString(i++)}
}
def x = new X()
def gstring = "$x"
assert gstring == "0"
assert gstring == "1"
```

The feature looks weird... I expect `gstring` is always `0`...

blackdrag:
@danielsun1106  you have to consider the story behind it, and that is reusable templates. But I can repeat mysefl "I think now that most users actually would want to have the eager variant"

danielsun1106:
@blackdrag We can consider to change the weird behavior since Groovy 3 :-)

danielsun1106:
Another issue is related to the PR:  

PR #711, opened 2018-05-18
GROOVY-8589 bugfix: should not show error about missing mandatory args when --help or --version is specified

remkop:
GROOVY-8589 bugfix: should not show error about missing mandatory args when --help or --version is specified

The second commit in this PR is to upgrade to picocli 3.0.1.
This allows us to remove some code since the parent command name is now automatically included in the usage help for subcommands.


remkop:
See also https://github.com/remkop/picocli/releases/tag/v3.0.1 for more details.

PR #712, opened 2018-05-19
GROOVY-8171: Escaped dollar slashy difference between old and parrot …

jwagenleitner:
…parsers


danielsun1106:
+1

paulk-asert:
+1, good catch

PR #713, opened 2018-05-20
GROOVY-8590: STC incorrectly infers type of nested method call used i…

jwagenleitner:
…n a return stmt


danielsun1106:
+1

paulk-asert:
+1 since what is there is broken but wrt to change this is fixing, I wonder whether there is an issue with one of the assumptions of the type checker, see "Closures vs methods" a little way down from here:
http://docs.groovy-lang.org/latest/html/documentation/#_return_type_inference

PR #714, opened 2018-05-21
GROOVY-8586: Support Java 9 ARM blocks

danielsun1106:


danielsun1106:
72 hours have gone. I will merge it soon if no one has objections.

PR #717, opened 2018-05-23
Groovy 8562 wrong closure delegation for property access and compile static

MeneDev:
Roundup:

The StaticTypeCheckingVisitor would not look consider FieldNode to potentially delegate to a different Field. I think this could only occur for name collisions, i.e. the owner has a field with the same name as the delegate.
Calling tryVariableExpressionAsProperty solved this, however this broke other tests. To solve this I cheked IMPLICIT_RECEIVER, if unset and PROPERTY_OWNER is present, checkOrMarkPrivateAccess must be called to ensure the accessors are generated.

The VariableExpressionTransformer ignored delegation metadata when a private field accessor was present. I inversed that logic to give delegation metadata precedence over private field accessors.

Build passes https://travis-ci.org/MeneDev/groovy/builds/382561549

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
@@ -100,6 +100,7 @@ import org.codehaus.groovy.syntax.Token;
 import org.codehaus.groovy.syntax.TokenUtil;
 import org.codehaus.groovy.syntax.Types;
 import org.codehaus.groovy.transform.StaticTypesTransformation;
+import org.codehaus.groovy.transform.sc.StaticCompilationMetadataKeys;
 import org.codehaus.groovy.transform.trait.Traits;
 import org.codehaus.groovy.util.ListHashMap;
 import org.objectweb.asm.Opcodes;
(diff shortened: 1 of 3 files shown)


danielsun1106:
@paulk-asert It looks good for me. What do you think about the PR?

paulk-asert:
on my TODO list to review - at first glance I am optimistic

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -596,6 +597,26 @@ public void visitVariableExpression(VariableExpression vexp) {
...
+                if (tryVariableExpressionAsProperty(vexp, vexp.getName())) {
+                    // IMPLICIT_RECEIVER is handled elsewhere
+                    // however other access needs to be fixed for private access
+                    if (vexp.getNodeMetaData(StaticTypesMarker.IMPLICIT_RECEIVER) == null) {
+                        boolean lhsOfEnclosingAssignment = isLHSOfEnclosingAssignment(vexp);
+                        ClassNode owner = (ClassNode) vexp.getNodeMetaData(StaticCompilationMetadataKeys.PROPERTY_OWNER);
+                        if (owner != null) {
+                            fieldNode = owner.getField(vexp.getName());
+                            vexp.setAccessedVariable(fieldNode);
+                            checkOrMarkPrivateAccess(vexp, fieldNode, lhsOfEnclosingAssignment);
```
danielsun1106: `lhsOfEnclosingAssignment` is only used here(in the `if` block)

paulk-asert:
Merged. Thanks!

PR #724, opened 2018-05-25
Release 2.5.0 related fixes

jwagenleitner:
Some recent commits seem to cause problems with the Nextflow 2.5.0 snapshot builds on the CI server. One can lead to a NPE (commit 90486ae1075d14a62d14452abfea0c27485a67b5) because fieldNode can be reassigned `null` after the `instanceof` check. Another change (commit 57cfd2a3e4d985b3248569ea1d92b02c59627703) seems to have caused an issue with assigning parameterized types.

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -62,6 +62,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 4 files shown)


danielsun1106:
The PR looks great for me. Out of curiosity, why does the  helper method cause the issue(`GenericsUtils.extractPlaceholders(GenericsUtils.findParameterizedType(declaringClassForDistance, actualReceiverForDistance));
 `)? I will set aside some time to investigate.

jwagenleitner:
It was extracting the actual type argument of `String` rather than the type `Trie`.

paulk-asert:
+1, you actually fixed something on my TODO list to look at today! :-)

danielsun1106:
@paulk-asert it's better to add your tests too

PR #735, opened 2018-05-29
Fix file descriptor leaks when using URL connections

oehme:
The JVM keeps a cache of URL connections, which keeps
the underlying files open for the lifetime of the VM
unless the `setUseCaches(false)` method is used.

This change makes it possible to reuse the Groovy compiler,
e.g. embedded in the Gradle daemon without preventing files
from being deleted on Windows.

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
@@ -34,6 +34,7 @@ import org.codehaus.groovy.control.Phases;
 import org.codehaus.groovy.control.SourceUnit;
 import org.codehaus.groovy.control.messages.SimpleMessage;
 import org.codehaus.groovy.control.messages.WarningMessage;
+import org.codehaus.groovy.util.URLStreams;
 
 import java.io.BufferedReader;
 import java.io.IOException;
@@ -221,7 +222,7 @@ public final class ASTTransformationVisitor extends ClassCodeVisitorSupport {
                 String className;
                 BufferedReader svcIn = null;
                 try {
-                    svcIn = new BufferedReader(new InputStreamReader(service.openStream(), "UTF-8"));
+                    svcIn = new BufferedReader(new InputStreamReader(URLStreams.openUncachedStream(service), "UTF-8"));
                     try {
                         className = svcIn.readLine();
                     } catch (IOException ioe) {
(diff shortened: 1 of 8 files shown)


src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
```diff
@@ -221,7 +222,9 @@ private static void doAddGlobalTransforms(ASTTransformationsContext context, boo
                 String className;
                 BufferedReader svcIn = null;
                 try {
-                    svcIn = new BufferedReader(new InputStreamReader(service.openStream(), "UTF-8"));
+                    URLConnection urlConnection = service.openConnection();
+                    urlConnection.setUseCaches(false);
+                    svcIn = new BufferedReader(new InputStreamReader(urlConnection.getInputStream(), "UTF-8"));
```
danielsun1106: Readers will close their underlying Readers(e.g. `BufferedReader`(i.e. `svcIn`) --closes--> `InputStreamReader` --closes--> the stream instance created by `service.openStream()`), and `svcIn` will be closed in the `finally` block, could you please explain how file descriptor leaks?

oehme: As explained in the commit message, the jar will be kept open by the cache in the jvm unless you deactivate that cache. This is meant as a performance optimization and unfortunately it's on by default.

paulk-asert:
It would be good to help fix some of the windows locking issues. This change looks reasonable to me on first glance. We should do a quick check if there are any negative performance impacts by not having that caching in place.

danielsun1106:
It seems that similar issues exist in other classes, we should `setUseCaches(false)` too?, e.g.
```java
                URLConnection conn = source.openConnection();
                lastMod = conn.getLastModified();
                conn.getInputStream().close();
```
( https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java#L335-L337 )

oehme:
Yes, ideally we'd fix all of those. I only found this particular one because it broke one of our integration tests that was testing AST transforms. I will do a pass over the code base and fix the other places I find.

oehme:
This is ready for another review.

danielsun1106:
Merged. Thanks!

P.S. Please close the PR, I forgot to add `closes #735` in the commit log...

https://github.com/apache/groovy/commit/3453072a0210990c8c9563557af12de3da4bbf4a

oehme:
Thanks!

PR #736, opened 2018-05-31
Fix for the GROOVY-8421 issue.

h2gr:
Avoiding Gradle deprecation warnings about registering invalid task inputs by not adding the checkstyle HTML report generation task in subprojects with no Java source files.


paulk-asert:
merged, thanks!

PR #740, opened 2018-06-01
GROOVY-8619 Update docs for Groovy Shell

remkop:
see https://issues.apache.org/jira/browse/GROOVY-8619


danielsun1106:
Merged. Thanks!

PR #741, opened 2018-06-01
GROOVY-8618 Update docs for GroovyConsole

remkop:
see https://issues.apache.org/jira/browse/GROOVY-8618


danielsun1106:
Merged. Thanks!

PR #742, opened 2018-06-01
GROOVY-8620 Update docs for Grape dependency manager

remkop:
see https://issues.apache.org/jira/browse/GROOVY-8620


danielsun1106:
Merged. Thanks.

PR #753, opened 2018-06-09
GROOVY-8632: Groovy 2.5.0 fails to compile Google Java Client sample code

jwagenleitner:
Class files can contain INNERCLASS references to other
classes inner classes whose name may be the same name as a
contained inner class. By storing modifiers in a map keyed
by short class name there is a possibility for the wrong
modifiers to be stored.

Since generated class files for inner classes contain an INNERCLASS
self reference, the logic can be simplified to look for a matching
name and storing those access modifiers. This eliminates the need
to search the outer class for the INNERCLASS reference.


danielsun1106:
+1

paulk-asert:
Nice work John!

PR #758, opened 2018-06-16
GROOVY-8637: @NamedVariant processing error

paulk-asert:


paulk-asert:
In DelegateASTTransformation, we skip methods in GroovyObject. Here I chose to look for ones annotated with @Generated. At the moment, I believe this is identical behavior but if we start making more use of @Generated I think the approach chosen here will be the correct approach. We should consider retrofitting this approach for @Delegate.

PR #760, opened 2018-06-19
GROOVY-8224: Stubs for classes implementing traits with fields don't …

paulk-asert:
…include getters/setters


jwagenleitner:
+1

PR #762, opened 2018-06-27
BigDecimalMath.MAX_DIVISION_SCALE is hard-coded to 10 digits

pierrepinon:
The "scale" for BigDecimal division is hard-coded to 10 (BigDecimalMath.MAX_DIVISION_SCALE), which is fewer than the roughly equivalent 16-17 digits for double (64-bit) division. It seems like the default scale should be at least as accurate as double division. Also it is probably desirable for this setting to be configurable in some way, maybe more than one. Maybe a global Java system property setting, and then some form of context related override (per-thread, per-script, etc.).

I use this fix for 2 years without side effects on a software for a big company. For comparison, Excel use a round with 15 digit after decimal


danielsun1106:
At first, thanks for your contribution. 
I wonder whether you have tested the performance?

danielsun1106:
Merged. Thanks!

PR #770, opened 2018-07-06
GROOVY-8690 bump picocli to v3.2.0 from 3.1.0

remkop:
Details in https://issues.apache.org/jira/browse/GROOVY-8690


paulk-asert:
+1 for master and GROOVY_2_5_X

PR #774, opened 2018-07-12
GROOVY-8282: Error while popping argument from operand stack tracker …

paulk-asert:
…in class ...$Trait$Helper

Bare named argument parameters are captured in a TupleExpression rather than an ArgumentListExpression.
While it would be nice to retrofit APP to have the correct type, that would impact many more places.


paulk-asert:
Proposed PR merged

PR #775, opened 2018-07-13
GROOVY-8639: @Sortable annotation is not able to use accessible paren…

paulk-asert:
…t properties


paulk-asert:
Proposed PR merged

PR #777, opened 2018-08-01
add support for JDK 10 and JDK 11 targets

keeganwitt:


paulk-asert:
I think adding JDK10/11 is a good thing. I am not sure making them "consistent" when Java actually made a change helps us though. Once we also add a `isPostJDK10(bytecodeVersion)` method it will return true for JDK9 which returns '9' for `System.getProperty('java.specification.version')`.

keeganwitt:
We could still change the config option so that args to groovyc are similar to each other, but I guess keeping it consistent with java.specification.version makes sense.  `isPostJDK9()` is only used by tests it looked like at the moment, but it keeps the logic simpler there to make it match what Java returns.

PR #781, opened 2018-08-09
GROOVY-8732: @CompileStatic refers to private field in parent class

paulk-asert:
Currently JavaBean style getters aren't recognised
before some preliminary provisions are made for the
case of accessing the private field in places where that
is allowed. The change is just not to proceed with those
provisions if the JavaBean getter is detected.


paulk-asert:
No test added just yet but the bytecode looks good with visual inspection.

lptr:
Looks good!

PR #782, opened 2018-08-10
GROOVY-8741: Make single character GString behave like similar String in '=='

vmj:
If there's interest to fix [GROOVY-8741](https://issues.apache.org/jira/browse/GROOVY-8741), this might be it.


danielsun1106:
Thanks for your contribution!
LGTM, but it's a breaking change, so it should be good for master(3.0.0+) only.
@paulk-asert What do you think about it?

paulk-asert:
+1

danielsun1106:
Merged. Thanks

PR #790, opened 2018-08-30
GROOVY-8766: Add manifest capability entries for OSGi spifly

graben:


paulk-asert:
Looks good. So just to confirm. It would never be needed for the core jar if it was using service loader? Just the fragment jars?

Also, it would be good to tweak to rule out when only org.codehaus.groovy.ast.builder.AstBuilderTransformation is found in the META-INF services directory. We should really move that to META-INF/groovy since defining a global transform isn't really a service. It only affects groovy-macro in your list. Let me know if you want to tweak your PR or otherwise I can make that change.

paulk-asert:
That's correct but let's just leave your PR as is. I'll make another issue to fix here we place global transformation meta-inf information.

paulk-asert:
I created GROOVY-8771 to track the other issue.

graben:
Okay, I'll keep the PR as is.

paulk-asert:
Okay, I have a PR for GROOVY-8771 ready, though I doubt it will impact your testing. I'll hold off merging this PR until I head further. Perhaps add the new stacktrace with woven classes into the Jira issue.

graben:
Unfortunately it's the same exception. After a bit of debugging it seems that woving the class fail or could not be done but without any obvious reason (error). I think that need further investigation.

paulk-asert:
There was some push back on moving the global transforms META-INF files, so I'll leave them where they are for now. If you find a different solution for the other problems you were facing, I can help with the tweaks for the build to account for those files if you need me to.

graben:
After some investigation it now works. But I have to remove the classloader argument for my testing groovy bundle fragment (groovy-json) to get it be woven by spifly. This change is not added in this PR since this really might be problematic. I also have to add the spifly consumer capability together to the provider one.

@paulk-asert: How should we move forward?

graben:
Woven means that the spifly mediator adds his classloader switching functionality to bring consumer and provider together. This is done by adding code via asm to the groovy code itself. My test at the end is quite simple. I'm loading an osgi container (felix) together with spifly (+dependencies) + groovy + groovy-json and a simple bundle doing a really primitive groovy script:
`println new groovy.json.JsonSlurper().parseText('{\"2\":2}')`
Of cause it is very important to have the right order of bundles being deployed! Sometimes the order was wrong and I got other errors and didn't see the cause in the first moment. But this can be handled by example using karaf features or by hard coded order.
The manifest changes are simple the usage of what the OSGi standard says you should do.
Clear enough? :-)

paulk-asert:
I believe so. Just to make sure I understand, I'll try to restate. Let me know if this sounds right.

If we put your change in (give or take the changes wrt `AstBuilderTransformation` which I can look at), then it should fix the underlying problem. Folks who get the order right will have no problems. If someone gets the order wrong (from an OSGi point of view) then they might still get errors but it won't be due to a Groovy problem.

graben:
That's not 100% correct. As mentioned before I had to remove the classloader switch in groovy-json which is not part of this PR since I can not estimate if this will cause problems in normal (non-OSGi) usage. (https://github.com/apache/groovy/blob/GROOVY_2_5_X/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/FastStringUtils.java#L40) I didn't look if this is done in any other groovy subproject, too. The rest of your statement is correct.

paulk-asert:
You removed the classloader parameter? Okay, that is different. What happens if you instead try adding a context classloader earlier? Thread.currentThread().setContextClassLoader(loader)? Not sure what loader will be for OSGi. It's been too long, I'd need to look that up again. Otherwise, is there any benefit to making the MANIFEST change?

graben:
Set tccl to tccl? I'm not sure what the classloader is really for in this context. In OSGi the tccl stuff is handled by weaving. Without removing the MANIFEST changes are useless!

paulk-asert:
I was suggesting something like what is shown here:
https://stackoverflow.com/questions/2198928/better-handling-of-thread-context-classloader-in-osgi

graben:
The actual code does use the tccl in groovy-json. No tccl switch as shown in your example. The question is if this really necessary?

graben:
@paulk-asert: Filter for `org.codehaus.groovy.transform.ASTTransformation` added

graben:
@paulk-asert: No response for a while. Would be nice to get this into 2.5.3.

paulk-asert:
I haven't had time to refresh my memory on the requirements for the classloader parameter. I believe it's necessary for numerous scenarios, e.g. when using @Grab. Without that change, aren't the other changes worthless?

graben:
I don't now how @grab is working but the same thing this code does (get tccl for service loading)  already does the JDK itself. IMHO there is no reason to do this "twice". Yes, spifly only instrument usage of ServiceLoader.load without explict classloader because it does the tccl classloader weaving itself. (classic OSGi try/catch/finally block to temporary set tccl to a specific bundleclassloader)

graben:
@paulk-asert : Did you already had time to refresh your memory. I still think it should be no problem to remove the classloader line to get current one.

graben:
https://issues.apache.org/jira/browse/ARIES-1854 adds a new feature to spi-fly supporting ServiceLoader.load(Class, ClassLoader). With this the actual hack to groovy code seems to be obsolete! I'll give it a try soon.

graben:
@paulk-asert : Work's for me now even without changing groovy code.

graben:
@paulk-asert : Chance to get this patch into 2.5.4?

paulk-asert:
Merged! Thanks for reporting the issue and your patience while we got it merged.

PR #791, opened 2018-08-31
GROOVY-8771: global AST transformation declarations should move to ME…

paulk-asert:
…TA-INF/groovy instead of META-INF/services


paulk-asert:
Leaving files where they are for now. We can come back and re-look at this PR if additional impacts are found down the track.

PR #792, opened 2018-09-08
GROOVY-8778: Cast short-hand breaks for empty map

paulk-asert:


subprojects/parser-antlr4/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -2337,14 +2337,11 @@ public Expression visitPathElement(PathElementContext ctx) {
...
-                MapEntryExpression mapEntryExpression = mapEntryExpressionList.get(0);
-
-                if (mapEntryExpression.getKeyExpression() instanceof SpreadMapExpression) {
-                    right = mapEntryExpression.getKeyExpression();
-                } else {
-                    right = mapEntryExpression;
-                }
+            if (mapEntryExpressionList.size() == 0) {
+                // expecting list of MapEntryExpressions later so use SpreadMap to smuggle empty MapExpression to later stages
+                right = new SpreadMapExpression(configureAST(new MapExpression(), ctx.namedPropertyArgs()));
```
danielsun1106: It's better to `configureAST` the instance of `SpreadMapExpression` too, or its node position will be missing.

paulk-asert: I had that originally but we throw that expression away in resolve visitor.

danielsun1106: The implementation of parser should not decouple with back end(i.e. not rely on the implementation of back end). 
If the implementation of resolve visitor changes in the future, the expression is not thrown away, the node position is missing.

paulk-asert: Yes, you are right - if we ever do support getAt(Map) more generally, we would need to process that differently, so best it is correct to start with.

PR #793, opened 2018-09-08
GROOVY-8780: Add `sha256` DGM

danielsun1106:


src/main/java/org/codehaus/groovy/runtime/EncodingGroovyMethods.java
```diff
@@ -389,6 +390,26 @@ public static String md5(byte[] self) throws NoSuchAlgorithmException {
         return digest(self, MD5);
     }
 
+    /**
+     * Calculate SHA-256 of the CharSequence instance
+     * @return SHA-256 value
+     * @throws NoSuchAlgorithmException if SHA-256 algorithm not found
+     * @since 2.5.0
```
paulk-asert: 2.5.3?

src/main/java/org/codehaus/groovy/runtime/EncodingGroovyMethods.java
```diff
@@ -389,6 +390,26 @@ public static String md5(byte[] self) throws NoSuchAlgorithmException {
...
+     */
+    public static String sha256(CharSequence self) throws NoSuchAlgorithmException {
+        return digest(self, SHA_256);
+    }
+
+    /**
+     * Calculate SHA-256 of the byte array
+     * @return SHA-256 value
+     * @throws NoSuchAlgorithmException if SHA-256 algorithm not found
+     * @since 2.5.0
```
paulk-asert: ditto

danielsun1106: OK. I'll fix it. Thanks ;-)

PR #796, opened 2018-09-13
GROOVY-8775: Bug joint compilation in ant task groovyc: classpath not set

paulk-asert:
This doesn't fix the issue but enables a workaround until we fix fully


paulk-asert:
Ignore this proposal, we can just use classpath already.

PR #799, opened 2018-09-21
GROOVY-8805 reloading a script under GroovyScriptEngine should succee…

gclayburg:
…d when dependent class is deleted


danielsun1106:
LGTM. Thanks for your contribution.

danielsun1106:
Merged. Thanks!

danielsun1106:
As the PR is not binary compatible, so it will be merged into master, i.e. groovy 3

gclayburg:
ok cool.  I wasn't really sure which branch I should be working with.  I just chose the one I did based on the version I was using, i.e 2.5.x

PR #806, opened 2018-10-06
GROOVY-8831: Add instance methods from `String` to `GString`

danielsun1106:


paulk-asert:
With respoect to this commit, there are some issues and other points that deserve discussion on the mailing list:
* we can't just cut and paste the comments from Java source code due to the different licence
* for the methods from Comparable and CharSequence we can use "@inheritDoc" (that will work when doing javadoc and we should use that approach for groovydoc too - even though we need to fix usage of that feature for groovydoc - there is an existing issue)
* for other methods when can either paraphrase in our own wording and/or point to the related String methods using "@see"
* it's a bit confusing for the methods overlapping with those from StringGroovyMethods, e.g. `contains`, was there a reason to replace them - if so we should have that commented somewhere
* for many other methods, Groovy's runtime and static compiler already handle the GString to String conversion, was there a reason to provide implementations since we used to get them for free

danielsun1106:
@paulk-asert  I've reverted the commit:
https://github.com/apache/groovy/commit/5ff63b1cdad87450893c5809f82b962c87688e23

adithyank:
I have a point here. Not sure whether I can  post....

- Using the `ExtensionModule` feature, we have implemented multiple methods for easy `String` handling. Ex. `println 'some text here'.md5`
- After this, intuitively I started coding as `println "some $var".md5`. But, it fails with error `No such property: md5 for class: org.codehaus.groovy.runtime.GStringImpl`
- So, any method that is available in the `String` class in groovy environment due to a any of the below reasons, they should be made available in `GString` also. 
     * `String`'s native methods
     * Using groovy extension module facilty
     * custom meta class
     * methods in StringGroovyMethods
- Yes. There will be can be more than one method impls with same signature in these places. Groovy runtime can resolve the calls with some order. With some proper documentation, it can be educated to developers or script writers.

`StringGroovyMethods` has lot of overloadings as `GString` variant to make the users feel transparant between `String` and `GString`. The idea is right. But, the way the methods are given to `GString` is very manual and can be made automatic using some ways...

danielsun1106:
FYI, `'abc'.md5()`

adithyank:
> FYI, `'abc'.md5()`

```bash
adithyan@adithyan:~$ groovy -e 'println "abc".md5()'
Caught: groovy.lang.MissingMethodException: No signature of method: java.lang.String.md5() is applicable for argument types: () values: []
Possible solutions: any(), is(java.lang.Object), any(groovy.lang.Closure), use([Ljava.lang.Object;), use(java.util.List, groovy.lang.Closure), use(java.lang.Class, groovy.lang.Closure)
groovy.lang.MissingMethodException: No signature of method: java.lang.String.md5() is applicable for argument types: () values: []
Possible solutions: any(), is(java.lang.Object), any(groovy.lang.Closure), use([Ljava.lang.Object;), use(java.util.List, groovy.lang.Closure), use(java.lang.Class, groovy.lang.Closure)
        at script_from_command_line.run(script_from_command_line:1)
```

adithyank:
Also, my point is not only for this one method. It is about the availability of all the methods in `String` in groovy environment in `GString` also.. Thanks

danielsun1106:
Groovy 2.5.4 is required to use `md5` DGM.

PR #810, opened 2018-10-13
fixed for NullPointerException in MissingMethodException

qxo:
fixed for NullPointerException in MissingMethodException (when type is null)

```
Caused by: java.lang.NullPointerException
    at groovy.lang.MissingMethodException.getMessage (MissingMethodException.java:60)
    at java.lang.Throwable.getLocalizedMessage (Throwable.java:391)
    at java.lang.Throwable.toString (Throwable.java:480)
    at java.lang.String.valueOf (String.java:2994)
    at org.fusesource.jansi.FilterPrintStream.println (FilterPrintStream.java:238)
    at java.lang.Throwable$WrappedPrintStream.println (Throwable.java:748)
    at java.lang.Throwable.printStackTrace (Throwable.java:655)
    at java.lang.Throwable.printStackTrace (Throwable.java:643)
    at java_lang_Throwable$printStackTrace.call (Unknown Source)
    at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall (CallSiteArray.java:47)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call (AbstractCallSite.java:116)
    at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call (AbstractCallSite.java:128)
```

### DIFF
--- a/src/main/groovy/groovy/lang/MissingMethodException.java
+++ b/src/main/groovy/groovy/lang/MissingMethodException.java
@@ -54,16 +54,22 @@ public class MissingMethodException extends GroovyRuntimeException {
     }
 
     public String getMessage() {
-        return "No signature of method: "
-                + (isStatic ? "static " : "")
-                + type.getName()
-                + "."
-                + method
+         String s = "No signature of method: ";
+         if(isStatic) {
+             s += "static ";
+         }
+         if( type != null  ) {
+             s +=  type.getName() + ".";
+         }
+         s += method
                 + "() is applicable for argument types: ("
                 + InvokerHelper.toTypeString(arguments, 60)
                 + ") values: "
-                + InvokerHelper.toArrayString(arguments, 60, true)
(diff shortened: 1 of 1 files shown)


danielsun1106:
LGTM.
Would you mind to create a JIRA ticket to track the issue? ( https://issues.apache.org/jira/projects/GROOVY/summary )

paulk-asert:
While your suggestion is possibly fine, it would be good to dive into this a bit more to understand what is going on. It would seem strange that type is null at that point. What seems a little weird to me is that your stack trace has:
{code}
java.lang.Throwable.getLocalizedMessage (Throwable.java:391)
{code}
which should be making a call to {{Throwable#getMessage()}}.
Instead we see in your stacktrace a call to {{MissingMethodException#getMessage()}}.
It would be good to know that the runtime didn't pick the wrong method at that point.

Do you have a simple standalone example that can reproduce the error?

PR #811, opened 2018-10-15
GROOVY-8843: Fix illegal reflective access within o.c.g.vmplugin.v7.J…

paulk-asert:
…ava7

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v9/Java9.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v9/Java9.java
@@ -18,14 +18,78 @@
  */
 package org.codehaus.groovy.vmplugin.v9;
 
+import org.codehaus.groovy.GroovyBugError;
 import org.codehaus.groovy.vmplugin.v8.Java8;
 
+import java.lang.invoke.MethodHandles;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+
 /**
- * Java 9 based functions will be added here if needed.
+ * Additional Java 9 based functions will be added here as needed.
  */
 public class Java9 extends Java8 {
+
+    private static class LookupHolder {
+        private static final Method PRIVATE_LOOKUP;
+        private static final Constructor<MethodHandles.Lookup> LOOKUP_Constructor;
+        static {
+            Constructor<MethodHandles.Lookup> lookup = null;
+            Method privateLookup = null;
+            try { // java 9
+                privateLookup = MethodHandles.class.getMethod("privateLookupIn", Class.class, MethodHandles.Lookup.class);
+            } catch (final NoSuchMethodException | RuntimeException e) { // java 8 or fallback if anything else goes wrong
+                try {
+                    lookup = MethodHandles.Lookup.class.getDeclaredConstructor(Class.class, Integer.TYPE);
+                    if (!lookup.isAccessible()) {
+                        lookup.setAccessible(true);
+                    }
+                } catch (final NoSuchMethodException ex) {
+                    throw new IllegalStateException("Incompatible JVM", e);
+                }
+            }
+            PRIVATE_LOOKUP = privateLookup;
+            LOOKUP_Constructor = lookup;
+        }
+    }
+
+    private static Constructor<MethodHandles.Lookup> getLookupConstructor() {
+        return LookupHolder.LOOKUP_Constructor;
+    }
+
+    private static Method getPrivateLookup() {
+        return LookupHolder.PRIVATE_LOOKUP;
+    }
+
+    public static MethodHandles.Lookup of(final Class<?> declaringClass) {
+        try {
+            if (getPrivateLookup() != null) {
+                return MethodHandles.Lookup.class.cast(getPrivateLookup().invoke(null, declaringClass, MethodHandles.lookup()));
+            }
+            return getLookupConstructor().newInstance(declaringClass, MethodHandles.Lookup.PRIVATE).in(declaringClass);
+        } catch (final IllegalAccessException | InstantiationException e) {
+            throw new IllegalArgumentException(e);
+        } catch (final InvocationTargetException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
     @Override
     public int getVersion() {
         return 9;
+    }
+
+    @Override
+    public Object getInvokeSpecialHandle(Method method, Object receiver) {
+        if (getLookupConstructor() != null) {
+            Class declaringClass = method.getDeclaringClass();
+            try {
+                return of(declaringClass).unreflectSpecial(method, receiver.getClass()).bindTo(receiver);
+            } catch (ReflectiveOperationException e) {
+                throw new GroovyBugError(e);
+            }
+        }
+        return super.getInvokeSpecialHandle(method, receiver);
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -414,11 +414,11 @@ task checkCompatibility {
     }
 }
 
-if (!JavaVersion.current().java8Compatible) {
+if (!JavaVersion.current().java9Compatible) {
     logger.lifecycle '''
     **************************************** WARNING ********************************************
-    ******   You are running the build with an older JDK. NEVER try to release with 1.7.   ******
-    ******   You must use a JDK 1.8+ in order to compile all features of the language.     ******
+    ******   You are running the build with an older JDK. NEVER try to release with 1.8.   ******
+    ******   You must use a JDK 1.9+ in order to compile all features of the language.     ******
     *********************************************************************************************
 '''
 }
--- a/src/main/java/org/codehaus/groovy/vmplugin/v7/Java7.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v7/Java7.java
@@ -29,40 +29,44 @@ import java.security.AccessController;
 import java.security.PrivilegedAction;
 
 /**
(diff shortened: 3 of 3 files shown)


src/main/java/org/codehaus/groovy/vmplugin/v9/Java9.java
```diff
@@ -18,14 +18,78 @@
...
+        private static final Constructor<MethodHandles.Lookup> LOOKUP_Constructor;
+        static {
+            Constructor<MethodHandles.Lookup> lookup = null;
+            Method privateLookup = null;
+            try { // java 9
+                privateLookup = MethodHandles.class.getMethod("privateLookupIn", Class.class, MethodHandles.Lookup.class);
+            } catch (final NoSuchMethodException e) { // java 8
+                try {
+                    lookup = MethodHandles.Lookup.class.getDeclaredConstructor(Class.class, Integer.TYPE);
+                    if (!lookup.isAccessible()) {
```
blackdrag: we can actually not rely on this being accessible... I am missing the IllegalAccessException and InaccessibleObjectException handling here...

src/main/java/org/codehaus/groovy/vmplugin/v9/Java9.java
```diff
@@ -18,14 +18,78 @@
...
+    private static Method getPrivateLookup() {
+        return LookupHolder.PRIVATE_LOOKUP;
+    }
+
+    public static MethodHandles.Lookup of(final Class<?> declaringClass) {
+        try {
+            if (getPrivateLookup() != null) {
+                return MethodHandles.Lookup.class.cast(getPrivateLookup().invoke(null, declaringClass, MethodHandles.lookup()));
+            }
+            return getLookupConstructor().newInstance(declaringClass, MethodHandles.Lookup.PRIVATE).in(declaringClass);
```
blackdrag: If PRIVATE_LOOKUP lookup never fails as I think, then LOOKUP_Constructor will be always the same. Not sure if "in" (https://docs.oracle.com/javase/9/docs/api/java/lang/invoke/MethodHandles.Lookup.html#in-java.lang.Class-) then can still do something useful.

src/main/java/org/codehaus/groovy/vmplugin/v9/Java9.java
```diff
@@ -18,14 +18,78 @@
...
 public class Java9 extends Java8 {
+
+    private static class LookupHolder {
+        private static final Method PRIVATE_LOOKUP;
+        private static final Constructor<MethodHandles.Lookup> LOOKUP_Constructor;
+        static {
+            Constructor<MethodHandles.Lookup> lookup = null;
+            Method privateLookup = null;
+            try { // java 9
+                privateLookup = MethodHandles.class.getMethod("privateLookupIn", Class.class, MethodHandles.Lookup.class);
```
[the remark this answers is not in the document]
paulk-asert: We don't eventually, I was just letting the build be built from Java 8 (with current warnings) until we get more things in place.

danielsun1106:
I wish the PR could be refined and merged soon, massive warnings drive me crazy 😖

paulk-asert:
I will add it soon and it gets rid of one set of warnings but only to
reveal the next set which we haven't progressed yet.

emre-yildirim:
Hello, I'm trying to implement the API test automation using rest assured in a project where Java 11 have been used.
and I'm having the following warning when I execute the code

> WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.vmplugin.v7.Java7$1 (file:/C:/Users/myildirim/.m2/repository/org/codehaus/groovy/groovy/2.5.6/groovy-2.5.6.jar) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class,int)
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.vmplugin.v7.Java7$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

The code is simple as follows

> package com.fieldcode.absenceservice;
> import org.junit.Test;
> import static io.restassured.RestAssured.given;
> public class AbsenceApiTEST {
>   @Test
>   public void absencesAvailability(){
>     given()
>         .when()
>         .head("http://localhost:8082/rest/absences/")
>         .then()
>         .assertThat()
>         .statusCode(200);}}

related dependencies
>     <dependency>
>       <groupId>io.rest-assured</groupId>
>       <artifactId>rest-assured</artifactId>
>       <version>4.0.0</version>
>       <scope>test</scope>
>     </dependency>
>     <dependency>
>       <groupId>io.rest-assured</groupId>
>       <artifactId>json-schema-validator</artifactId>
>       <version>4.0.0</version>
>       <scope>test</scope>
>     </dependency>
>     <dependency>
>       <groupId>junit</groupId>
>       <artifactId>junit</artifactId>
>       <version>4.12</version>
>       <scope>test</scope>
>     </dependency>

danielsun1106:
@emre-yildirim @paulk-asert 
Current solution still tries to access `private Lookup(Class<?> lookupClass, int allowedModes)`, so the illegal access warnings will still appear. I suggest to forbidden illegal access when using indy since Groovy 3.0.0.

paulk-asert:
It also looks like you are using 2.5.6 which won't have most of the warnings turned off. You can safely ignore them or if using the command line tools, you might want to use the GROOVY_TURN_OFF_JAVA_WARNINGS environment switch.

plittlefield:
This does not work for me... I have set the environment variable and still get the warnings...

```
$ env | grep GROOVY
GROOVY_TURN_OFF_JAVA_WARNINGS=true
```
...and when I run FileBot I get...

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.vmplugin.v7.Java7$1 (file:/usr/share/filebot/jar/groovy.jar) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class,int)
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.vmplugin.v7.Java7$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
Here is my system information...

```
FileBot 4.9.0 (r6899)
JNA Native: 4.0.1
MediaInfo: 19.09.20200314
p7zip: p7zip Version 9.20 (locale=en_GB.UTF-8,Utf16=on,HugeFiles=on,4 CPUs)
unrar: UNRAR 5.30 beta 2 freeware
Extended Attributes: OK
Unicode Filesystem: OK
Script Bundle: 2020-03-02 (r625)
Groovy: 2.5.8
JRE: Java(TM) SE Runtime Environment 13.0.2
JVM: 64-bit Java HotSpot(TM) 64-Bit Server VM
```

Thanks,

Paully

danielsun1106:
Please try 3.0.2

plittlefield:
How do I do that for FileBot which requests the groovy amc plugin remotely?

https://github.com/filebot/scripts

Thanks,

Paully

paulk-asert:
The general answers to your question are:

- use JDK8
- move to a later Groovy version
- ignore the warnings (they don't affect any operation)
- alter the startup scripts to turn off warnings (FileBot seems to have some of this in place already, maybe it could be extended)

I don't really know much about FileBot. You could contact the authors and ask them to upgrade Groovy. But just having a quick look, it does look like a FileBot install (I looked at portable version) has a `jar` directory. You could simply replace the 9 groovy jars in that directory. You could try with 2.5.10 first and if that doesn't work, try 3.0.2. Make sure to back up your install before tweaking it.

plittlefield:
Thanks for your suggestions, I tried both of those versions and neither helped.

I also updated FileBot to the very latest version which now uses Groovy 3.0.2 by default and that still had the WARNINGS when using Oracle Java 12.

So, I have gone back to OpenJDK 8 for the time being and all is fine now.

Paully

danielsun1106:
When you used Groovy 3.0.2, what warnings did you see?
Could you submit a JIRA issue to track it?

plittlefield:
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.vmplugin.v7.Java7$1 (file:/usr/share/filebot/jar/groovy.jar) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class,int)
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.vmplugin.v7.Java7$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```
How do I submit a JIRA issue?

Newbie :-)

danielsun1106:
> How do I submit a JIRA issue?

Here is the link to JIRA:

Please provide a sample code to reproduce the issue. Thanks!

plittlefield:
> Please provide a sample code to reproduce the issue. Thanks!

Do you mean tell the how to use FileBot ?!

danielsun1106:
The target of the sample code is to reproduce the issue

paulk-asert:
The GROOVY_TURN_OFF_JAVA_WARNINGS environment flag would have only worked if FileBot called into the Groovy startup scripts which it doesn't. You'd need to change possibly one of the ini files in FileBot to try to do similar things to what Groovy's startup scripts do.

With respect to steps to reproduce, yes it would be helpful if you could give us some simple steps we can follow to reproduce the issue, e.g.:

1. install FileBot (link to instructions)
2. run this command: 'some command here'

I don't think any of us has used FileBot before and this would speed up our ability to reproduce locally so we can debug further.

PR #812, opened 2018-10-16
GROOVY-8848: Named parameters documentation missing explanation for mixing multiple parameters use case

wololock:
This pull request fixes https://issues.apache.org/jira/browse/GROOVY-8848

The problem explained in updated documentation "Named arguments" section was reported on Stack Overflow today - https://stackoverflow.com/q/52832382/2194470


paulk-asert:
Merged (with minor tweaks), thanks!

PR #815, opened 2018-10-23
GROOVY-8855: Matcher.asBoolean() does not rely on matchers internal search index anymore

wololock:
Fixes https://issues.apache.org/jira/browse/GROOVY-8855

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/StringGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/StringGroovyMethods.java
@@ -104,7 +104,7 @@ public class StringGroovyMethods extends DefaultGroovyMethodsSupport {
         }
 
         RegexSupport.setLastMatcher(matcher);
-        return matcher.find();
+        return matcher.find(0);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


wololock:
I have found one test case that was causing infinite loop after introducing this change. There was a single test that expected this side effect here:

eric-milles:
I think the `==~` operator is available for this purpose.  Slightly different semantic between `find()` and `matches()`, but find keeps searching for next match which is an expected/intended behavior.

wololock:
@eric-milles, I wouldn't say that's the same use case. The JIRA issue focuses on a side effect that is caused by calling:

```
matcher as Boolean
```

Usually you don't expect that casting an object to some other type causes internal changes in the casting object. I agree that using `==~` operator is easier and safer to use, and this is what the guy who asked question on Stack Overflow eventually did. However, the question is still open - should `Matcher.asBoolean()` change the internal state of the object? It shouldn't in my opinion, and that is why this pull request was created. But it's up to you to decide if this makes sense or not.

eric-milles:
I'm a +1 for this change as long as there is some way to communicate that the pattern of using the Matcher's Groovy truth value in a loop will require the addition of ".find()" in Groovy 3.

PR #816, opened 2018-10-27
GROOVY-8858: Refine GCL to avoid occupying Permanent Area/Metaspace repeatedly for same source code

danielsun1106:


drosenbauer:
Yay!

PR #818, opened 2018-11-01
GROOVY-8866: Implement `GProperties` to handle properties file smartly

danielsun1106:


paulk-asert:
I like this idea. Seems worth an email to the dev list. I also wonder whether we need a way to escape curly braces if I want a line that isn't interpolated - one that might actually match an existing property name.

danielsun1106:
We should support escaping curly braces.
In addition, I plan to support importing multiple properties files and accessing properties in groovy script

danielsun1106:
We can escape curly braces with `{{...}}``, e.g. {{abc}}` will be translated as `{abc}`, and `${{abc}}` will be translated as `${abc}`

PR #819, opened 2018-11-04
GROOVY-7975/GROOVY-3278/GROOVY-7854: improved accessing of constants …

paulk-asert:
…for annotation attributes

### DIFF
--- a/src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
+++ b/src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
@@ -18,20 +18,291 @@
  */
 package org.apache.groovy.ast.tools;
 
+import org.codehaus.groovy.ast.ClassHelper;
 import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.FieldNode;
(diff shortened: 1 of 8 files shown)


paulk-asert:
I have only implemented string concatenation and +,-,/,* for ints and doubles. If folks can test these out I can expand to more types.

src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
```diff
@@ -18,20 +18,205 @@
...
+                            break;
+                    }
+                    if (newVal != null) {
+                        ConstantExpression newExp = new ConstantExpression(newVal, true);
+                        newExp.setSourcePosition(be);
+                        return newExp;
+                    }
+                }
+            }
+        } else if (isTypeOrArrayOfType(targetType, ClassHelper.Double_TYPE, false) || isTypeOrArrayOfType(targetType, ClassHelper.double_TYPE, false)) {
```
danielsun1106: looks like some template code, which is duplicated now. It's better to refactor it IMO

paulk-asert: Already refactored - thanks for the suggestion.

PR #826, opened 2018-11-19
GROOVY-8894: Improve performance of DGM `leftShift(OutputStream, Inpu…

danielsun1106:
…tStream)`


paulk-asert:
If you had a multi-threaded Swing app that was displaying tail of a file, wouldn't Thread.yield be relevant?

danielsun1106:
The inputstream is not thread-safe. 
1) If it is shared by multi-threads, it should be synchronized, so it will be read by only one thread at the same time. Once it is read, the count is -1 or greater than 0;
2) If each thread has its own inputstream, i.e. not shared by other threads, the private inputstream will be read util end of stream reached;

To sum up, as you can see, the scenario that the count is 0 will not happen.

paulk-asert:
Your analysis seems correct. I tried to go back in the commit history and looks like it is lost why that line was added. It's over 10 years old - perhaps very old JDK versions had different behavior. In any case, I think we can ignore it and go with your change.

PR #830, opened 2018-11-23
GROOVY-7647: Incorrect line information for debug

paulk-asert:


paulk-asert:
I changed one existing test since the bytecode now produced looks closer to what Javac produces for that case.

PR #835, opened 2018-12-03
GROOVY-8908: Bump gradle to 5.0

danielsun1106:


subprojects/parser-antlr4/build.gradle
```diff
@@ -30,9 +30,11 @@ def srcBase = "subprojects/parser-antlr4/src"
 def srcMain = "$srcBase/main"
 def srcTest = "$srcBase/test"
 
-final PARSER_PACKAGE_NAME = 'org.apache.groovy.parser.antlr4'
 generateGrammarSource {
+    maxHeapSize = '1g'
```
melix: From my tests it seems we can reduce to `512m`, otherwise LGTM :+1:

danielsun1106: https://travis-ci.org/danielsun1106/groovy/jobs/460259483
```
> Task :generateGrammarSource FAILED
* What went wrong:
Execution failed for task ':generateGrammarSource'.
> Java heap space
```

Groovy's grammar is quite complicated, so more memory will be consumed. As you can see, the `generateGrammarSource` task fails if the heap size is a bit small, e.g. `512m`. Also, you can try to clone apache/groovy and set the heap size to `maxHeapSize = '512m'` then run `gradlew clean generateGrammarSource --no-build-cache` locally, I believe OOM can be reproduced.

PR #837, opened 2018-12-10
GROOVY-8917: Failed to infer parameter type of some SAM, e.g. BinaryO…

danielsun1106:
…perator

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
+++ b/src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
@@ -20,6 +20,7 @@ package org.codehaus.groovy.ast.tools;
 
 import antlr.RecognitionException;
 import antlr.TokenStreamException;
+import groovy.lang.Tuple2;
 import groovy.transform.stc.IncorrectTypeHintException;
 import org.apache.groovy.util.SystemUtil;
 import org.codehaus.groovy.GroovyBugError;
@@ -53,7 +54,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Predicate;
 
+import static groovy.lang.Tuple.tuple;
 import static org.codehaus.groovy.ast.GenericsType.GenericsTypeName;
 import static org.codehaus.groovy.transform.stc.StaticTypeCheckingSupport.getCorrectedClassNode;
 import static org.codehaus.groovy.transform.stc.StaticTypeCheckingSupport.implementsInterfaceOrIsSubclassOf;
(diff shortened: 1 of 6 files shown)


src/main/java/org/codehaus/groovy/classgen/asm/util/TypeUtil.java
```diff
@@ -213,4 +220,15 @@ private static String makeRefDescription(String name) {
...
+    private static final Map<ClassNode, ClassNode> PRIMITIVE_CLASSNODE_TO_WRAPPED_CLASSNODE_MAP = Maps.of(
+            ClassHelper.byte_TYPE, ClassHelper.Byte_TYPE,
+            ClassHelper.boolean_TYPE, ClassHelper.Boolean_TYPE,
+            ClassHelper.char_TYPE, ClassHelper.Character_TYPE,
+            ClassHelper.double_TYPE, ClassHelper.Double_TYPE,
+            ClassHelper.float_TYPE, ClassHelper.Float_TYPE,
+            ClassHelper.int_TYPE, ClassHelper.Integer_TYPE,
+            ClassHelper.long_TYPE, ClassHelper.Long_TYPE,
+            ClassHelper.short_TYPE, ClassHelper.Short_TYPE
+    );
```
paulk-asert: Isn't this already in ClassHelper#PRIMITIVE_TYPE_TO_WRAPPER_TYPE_MAP?

danielsun1106: @paulk-asert As we can see, the existing type mapping is Class to Class,  but we need map ClassNode to ClassNode. Though we can convert ClassNode to Class and get the mapping result then convert the result to ClassNode, it is less efficient.

https://github.com/apache/groovy/blob/9cea59b95367c1d563e719ebeddcdecb58199750/src/main/java/org/codehaus/groovy/classgen/asm/util/TypeUtil.java#L213

paulk-asert: Both appear to be Map<ClassNode, ClassNode> to me?

paulk-asert: I was referring to `ClassHelper#PRIMITIVE_TYPE_TO_WRAPPER_TYPE_MAP` not `TypeUtil#PRIMITIVE_TYPE_TO_WRAPPED_CLASS_MAP`.

paulk-asert: And I guess also `TypeUtil#PRIMITIVE_TYPE_TO_WRAPPED_CLASS_MAP` would be better named `PRIMITIVE_CLASS_TO_WRAPPED_CLASS_MAP`.

danielsun1106: Gotcha. I will use the map you suggest instead.

danielsun1106: @paulk-asert I've refined the PR according to your suggestion ;-)

PR #842, opened 2018-12-16
Fix typo

jslopezgithub:
The *export and the '""' in the documentation seems wrong, I believe this was the original intent.


paulk-asert:
Merged, thanks!

PR #843, opened 2018-12-17
GROOVY-7233: Configurable Access Modifier for Log AST Transformations

paulk-asert:
All of the logging AST transformations now support @VisibilityOptions.
This allows the log field to be public, private (the default), protected,
or package-private.  There is a new LoggingStrategyV2 which logging
strategies which want to support visibility should support.


danielsun1106:
+1

PR #844, opened 2018-12-20
MissingPropertyExceptionTest.testNullMessage

jglick:
Otherwise you get errors like

```
groovy.lang.MissingPropertyExceptionTest > testNullMessage FAILED
    java.lang.NullPointerException
        at groovy.lang.MissingPropertyException.getMessageWithoutLocationText(MissingPropertyException.java:66)
        at groovy.lang.GroovyRuntimeException.getMessage(GroovyRuntimeException.java:63)
        at java.lang.Throwable.getLocalizedMessage(Throwable.java:391)
        at java.lang.Throwable.toString(Throwable.java:480)
        at junit.framework.Assert.assertNull(Assert.java:268)
        at junit.framework.TestCase.assertNull(TestCase.java:438)
        at groovy.lang.MissingPropertyExceptionTest.testNullMessage(MissingPropertyExceptionTest.java:29)
```

it is unclear what null values for `property` and `type` would mean, but there is a constructor explicitly allowing that, and it is illegal for `getMessage`, `toString`, etc. to throw an undeclared NPE.

(A `MissingPropertyException` producing this error was originally observed as a build failure in a Jenkins project, though I do not have the details to know the root cause there.)

### DIFF
--- a/src/main/groovy/groovy/lang/GroovyRuntimeException.java
+++ b/src/main/groovy/groovy/lang/GroovyRuntimeException.java
@@ -60,7 +60,12 @@ public class GroovyRuntimeException extends RuntimeException {
     }
 
     public String getMessage() {
-        return getMessageWithoutLocationText() + getLocationText();
+        String messageWithoutLocationText = getMessageWithoutLocationText();
+        String locationText = getLocationText();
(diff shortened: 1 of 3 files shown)


danielsun1106:
Merged. Thanks for your contribution!

PR #848, opened 2019-01-01
Run dependency management call as privileged action

dpolivaev:
Because we run scripts with restricted permissions it would be nice if grape could run within a privileged action. Currently we use a metaclass based work around https://github.com/freeplane/freeplane/blob/b11727f6297c7b2beb75786dcdd177c3d087fae3/freeplane_plugin_script/src/main/java/groovy/runtime/metaclass/groovy/grape/GrapeMetaClass.java ,  however it relies on implementation details of Grape and may be broken one day.

### DIFF
--- a/src/main/groovy/groovy/grape/Grape.java
+++ b/src/main/groovy/groovy/grape/Grape.java
@@ -19,6 +19,8 @@
 package groovy.grape;
 
 import java.net.URI;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


danielsun1106:
+1

It is better to create a JIRA ticket to track the improvement:

dpolivaev:
https://issues.apache.org/jira/projects/GROOVY/issues/GROOVY-8949

danielsun1106:
Merged: https://github.com/apache/groovy/commit/9018229f2adf703e9ac51423b8934aafe62413de
Thanks for your contribution!

PR #851, opened 2019-01-08
GROOVY-8951: Traits defining getter conflicts with generated getter (…

paulk-asert:
…improvements for pre-compiled case)


danielsun1106:
+1

PR #857, opened 2019-01-21
GROOVY-7996: Using with method with a closure that references a prote…

paulk-asert:
…cted property produces ClassCastException


eric-milles:
This change breaks down if an additional level of closure is inserted.  The implicit this is transformed into a path expression in `StaticTypeCheckingVisitor.existsProperty`.

PR #858, opened 2019-01-21
GROOVY-4287: CLONE - import of static nested classes in external groo…

danielsun1106:
…vy files is broken


danielsun1106:
As 72 hours have gone and no one rejects the PR, I'll merge it later.

PR #860, opened 2019-01-26
Fix non-atomic operation on volatile field 'count'

danielsun1106:


blackdrag:
My personal analysis on this is the following.... as long as count is only changed within a lock/unlock sequence volatile should be good enough. But the rehash methods do not guarantee that. But then again the implementation would be broken if they do not lock on rehash... actually I wonder who is actually calling that method.

My conclusion: Get rid of these classes. Java's concurrent hash map provides much better functionality, much better tested and afaik with a better performance profile.

danielsun1106:
@melix I run static analysis from time to time, the potential issue was found by the tool. It is related to thread safe, so it's not easy to create a reproducable test...

danielsun1106:
@blackdrag Thanks for your analysis 😄 and your advice is always helpful.

paulk-asert:
It would certainly be good to do a quick benchmark check with those classes replaced before deprecating them if performance isn't impacted.

paulk-asert:
Well, I wasn't going to test performance in any generic sense - just checking before and after build completion times (and potentially stress/performance subprojects) was all I had in mind. That won't ensure that all users aren't impacted but would give us a quick indication of whether things were about the same or if something was particularly bad.

PR #861, opened 2019-01-26
Trivial refactoring: handle `Sql` with try-with-resources for simplicity

danielsun1106:


paulk-asert:
I merged this to get it into 2.5.6

PR #863, opened 2019-01-26
Trivial refactoring: replace explicit type with `<>` for simplicity

daniellansun:


paulk-asert:
+1. As further background, we have delayed doing such changes earlier since having such differences between branches makes applying patches more difficult - often leads to manual merge fixes which although simple, can be a little painful for large patches. Since 2.5 and 3.0 both now have JDK 7 min (8 for 3.0) I think it is safe to do this change. If we ever want to backport some fixes and re-release another 2.4 release, we'll have to be a little careful.

danielsun1106:
Understood. Keep it as it is or apply the patch for all branches.

paulk-asert:
> Understood. Keep it as it is or apply the patch for all branches.

Yep, all active branches. Of course new code within a branch (e.g. Parrot parser files) wouldn't have that restriction and potentially other areas like plugin code for later JDK versions for instance (e.g. Java9 file only gets compiled on JDK9+ regardless of branch minimum JDK).

PR #864, opened 2019-01-26
GROOVY-8970: Fix typo in MANIFEST.MF: Main-class

danielsun1106:


paulk-asert:
+1. My understanding is that header names are case insensitive but we should use the typical convention of using uppercase for both parts.

paulk-asert:
Thanks Daniel, I applied to make it into 2.5.6.

PR #868, opened 2019-01-30
wrong author of don xijote

jagedn:
the author of Don Xijote was Miguel, not Manuel


paulk-asert:
Thanks for that. Merged!

PR #870, opened 2019-02-04
GROOVY-8977: Collect default imported class info for better resolving…

danielsun1106:
… performance

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/DefaultImportedClassCollectorHelper.java
+++ b/src/main/java/org/codehaus/groovy/control/DefaultImportedClassCollectorHelper.java
@@ -1,0 +1,226 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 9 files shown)


src/main/java/org/codehaus/groovy/control/DefaultImportedClassCollectorHelper.java
```diff
@@ -0,0 +1,226 @@
...
+        private static final File CLASSES_INFO_FILE = new File(new File(GroovySystem.getGroovyRoot(), "caches"), "default_imported_classes");
+        private static final String JAVA_VERSION = System.getProperty("java.version");
+        private static final String GROOVY_VERSION = getGroovyVersion();
+
+        private static String getGroovyVersion() {
+            String version = GroovySystem.getVersion();
+            version = (null == version || version.trim().isEmpty()) ? SNAPSHOT : version;
+
+            // if using SNAPSHOT version(e.g. 3.0.0-SNAPSHOT), cache for only 1 hour because default imported classes may change when developing
+            return version.contains(SNAPSHOT) ? version + new SimpleDateFormat("-yyyyMMdd.HH").format(new Date()) : version;
```
melix: :x: I think this is problematic. We shouldn't bake in a rule like this. Ideally this should be a hash of the Groovy runtime.

danielsun1106: "Ideally this should be a hash of the Groovy runtime." -- could you provide me more details about Groovy runtime?

danielsun1106: @melix  "I would also avoid adding another third party dependency just for this: this is core groovy." -- Yep, I agree with you. so the third party dependency is optional, as you can see, the dependency is `compileOnly`

danielsun1106:
@melix "I would also avoid adding another third party dependency just for this: this is core groovy." -- Yep, I agree with you. so the third party dependency is optional, as you can see, the dependency is `compileOnly` and it will not be included in groovy distributions.
```groovy
compileOnly "io.github.classgraph:classgraph:$classgraphVersion" // for compilation only
testRuntime "io.github.classgraph:classgraph:$classgraphVersion" // for running tests
```

PR #871, opened 2019-02-04
Documentation: add mandatory tag for the groovy-all artifact

energister:
without `<type>pom</type>` tag maven breaks build with 

> Failure to find org.codehaus.groovy:groovy-all:jar:2.5.5 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced

 message.

The tag is already added to the http://groovy-lang.org/download.html but was forgotten here.


paulk-asert:
Merged, thanks!

PR #886, opened 2019-03-03
Groovy 8180 log to file

tttao:
- Add option in Preferences dialog to append output to file
- Modified Console to output to file additionnally to classic output

Jira: https://issues.apache.org/jira/browse/GROOVY-8180

### DIFF
--- a/subprojects/groovy-console/src/main/groovy/groovy/ui/Console.groovy
+++ b/subprojects/groovy-console/src/main/groovy/groovy/ui/Console.groovy
@@ -66,10 +66,12 @@ import javax.swing.event.HyperlinkEvent
 import javax.swing.event.HyperlinkListener
 import javax.swing.filechooser.FileFilter
 import javax.swing.text.AttributeSet
+import javax.swing.text.Document
 import javax.swing.text.Element
 import javax.swing.text.SimpleAttributeSet
 import javax.swing.text.Style
 import javax.swing.text.StyleConstants
+import javax.swing.text.StyledDocument
 import javax.swing.text.html.HTML
 import java.awt.BorderLayout
 import java.awt.Component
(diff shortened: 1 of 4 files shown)


danielsun1106:
LGTM

subprojects/groovy-console/src/main/groovy/groovy/ui/ConsolePreferences.groovy
```diff
@@ -63,26 +64,36 @@ class ConsolePreferences {
...
                 hbox {
                     button T['prefs.reset.defaults'], id: 'resetPrefsButton', actionPerformed: this.&onReset
                     hglue()
                     button T['prefs.close'], id: 'closePrefsButton', actionPerformed: this.&onClose
                 }
+
             }
         }
 
+        //console.swing.outputFileName.enabled = false
```
paulk-asert: should this be deleted?

subprojects/groovy-console/src/main/groovy/groovy/ui/Console.groovy
```diff
@@ -66,10 +66,12 @@ import javax.swing.event.HyperlinkEvent
 import javax.swing.event.HyperlinkListener
 import javax.swing.filechooser.FileFilter
 import javax.swing.text.AttributeSet
+import javax.swing.text.Document
 import javax.swing.text.Element
 import javax.swing.text.SimpleAttributeSet
 import javax.swing.text.Style
 import javax.swing.text.StyleConstants
+import javax.swing.text.StyledDocument
```
paulk-asert: Is this needed?

paulk-asert:
I merged with a few tweaks - thanks!

If you feel like polishing further, I noticed what looked like some extra newlines/whitespace in the file log that I wasn't expecting (perhaps OS specific) - might be worth investigating.

Also, I wonder whether folks would like the checkbox and filename saved under the OS preferences. Might be worth an email to the mailing list to see if folks have a strong preference. At the moment you need to set them each time you start the console.

tttao:
@paulk-asert Thanks for the feedback ! I will rop a mail to the mailing list regarding saving the checkbox+filename under the OS preferences - knowing that it might not be convenient to redirect output for every Console instances to the same log file.

PR #890, opened 2019-03-06
typo at junit and missing imports

jagedn:
the geb-junit is wrong (or at least currently is not present at Maven Central) and maybe is more clear to add the two missings import

by other hand the example is a little old and it requires a deep review. I'll try to do it in these days


paulk-asert:
Merged, thanks!

PR #891, opened 2019-03-08
GROOVY-9022: Added option under View menu of whether or not to have v…

adamldavis:
GROOVY-9022: Added option under View menu of whether or not to have vertical orientation (unchecking changes it to horizontal; value stored in Preferences)
default to true, so users will not see a change unless they want to

### DIFF
--- a/subprojects/groovy-console/src/main/groovy/groovy/ui/Console.groovy
+++ b/subprojects/groovy-console/src/main/groovy/groovy/ui/Console.groovy
@@ -638,11 +638,10 @@ class Console implements CaretListener, HyperlinkListener, ComponentListener, Fo
         if (oldValue != orientationVertical) {
             if (orientationVertical) {
                 splitPane.setOrientation(JSplitPane.VERTICAL_SPLIT)
-                splitPane.resizeWeight = 0.5
             } else {
                 splitPane.setOrientation(JSplitPane.HORIZONTAL_SPLIT)
-                splitPane.resizeWeight = 0.5
             }
+            splitPane.resizeWeight = detachedOutput ? 1.0 : 0.5
         }
     }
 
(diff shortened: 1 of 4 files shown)


danielsun1106:
+1

paulk-asert:
This looks very close to me. The only thing I noticed was a small glitch when switching from horizontal to vertical orientation while in detached mode. It corrects itself when you "reattach" but it would be nice if that glitch didn't show up in the first place. Let me know if you have time to look at that.

I also wondered whether to consider allowing the output window to be above or to the left but I guess we can consider that if we have subsequent requests for such functionality.

adamldavis:
Sure I can fix that

subprojects/groovy-console/src/main/groovy/groovy/ui/Console.groovy
```diff
@@ -638,11 +638,10 @@ class Console implements CaretListener, HyperlinkListener, ComponentListener, Fo
         if (oldValue != orientationVertical) {
             if (orientationVertical) {
                 splitPane.setOrientation(JSplitPane.VERTICAL_SPLIT)
-                splitPane.resizeWeight = 0.5
             } else {
                 splitPane.setOrientation(JSplitPane.HORIZONTAL_SPLIT)
-                splitPane.resizeWeight = 0.5
             }
+            splitPane.resizeWeight = detachedOutput ? 1.0 : 0.5
```
paulk-asert: still didn't seem to work for me

paulk-asert:
2019-03-13 22_16_44-GroovyConsoleGlitch2
2019-03-13 22_14_01-GroovyConsoleGlitch1

paulk-asert:
Looks like a call to `resetToPreferredSizes()` fixes those glitches above. I'll merge your request and add in the call.

paulk-asert:
Merged, thanks!

adamldavis:
Thanks so much! Feels great to contribute to Groovy in a small way.

PR #893, opened 2019-03-09
Added example in sql module for calling stored function with in out parameter

mawasak:
This is additional example for calling stored database function where parameter is "in out"


paulk-asert:
Hi, I merged this but tweaked it to use Hsqldb not Oracle. Our build assumes Hsqldb for the tests unless driver details are overridden with properties. We don't have Oracle available in our current CI environments. Thanks for the suggestion.

PR #895, opened 2019-03-12
GROOVY-7954 Fix equality check in special case

nguillaumin:
Fix equality check with two instances A & B when:
- A and B implement the same interface but are not of the same class
- The common interface implements `Comparable`

`DefaultTypeTransformation.compareToWithEqualityCheck(...)` checks for
assignable classes and use `compareTo()` when assignable on the class
level, but it doesn't check it on the interface level.

As a result `compareTo()` is not called at the interface level and `-1`
is returned.  The fix performs an actual equality check instead of
returning -1.

I added a unit test, but I'm not completely sure of the right structure
for unit tests so let me know if I should do it differently.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
+++ b/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
@@ -600,7 +600,7 @@ public class DefaultTypeTransformation {
         }
 
         if (equalityCheckOnly) {
-            return -1; // anything other than 0
+            return left.equals(right) ? 0 : -1;
         }
         String message = MessageFormat.format("Cannot compare {0} with value ''{1}'' and {2} with value ''{3}''",
                 left.getClass().getName(), left, right.getClass().getName(), right);
--- a/src/test/groovy/runtime/typehandling/EqualityTest.groovy
+++ b/src/test/groovy/runtime/typehandling/EqualityTest.groovy
@@ -1,0 +1,12 @@
+package groovy.runtime.typehandling
+
+class EqualityTest extends GroovyTestCase {
+
+    void testEquality() {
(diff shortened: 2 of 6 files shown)


janArb:
I could verify this merge fixes https://issues.apache.org/jira/browse/GROOVY-7954

danielsun1106:
LGTM.

@paulk-asert 
What do you think about the PR? Any chance to catch the train of Groovy 3.0.0-beta-1?

paulk-asert:
I'll take a look today

paulk-asert:
Merged with a minor tweak. I took a more conservative approach in the alternative logic since that was all that was required to make the described scenario work. We can always expand the logic further in line with your suggestion if further examples indicate the need. I also just moved your tests to an existing subdirectory for consistency.

nguillaumin:
Thanks!

MykolaGolubyev:
Thank you for working on this. I just hit an example where `lhs` implements `compareTo` in a way that `rhs` doesn't need to be `Comparable`. It is in the context of a testing framework and DSL surrounding it.

https://github.com/twosigma/webtau/blob/e1d20cdaf12df6250c1a86034a3451bb9dc14423/webtau-http-groovy/src/test/groovy/com/twosigma/webtau/http/GroovyDataNodeTest.groovy#L107

```
@Test
void "compare to complex values"() {
    def node = new GroovyDataNode(DataNodeBuilder.fromValue(new DataNodeId('test'), [1, 2, 3]))

    if (node == [1, 2, 3]) {
        // TODO
        // https://issues.apache.org/jira/browse/GROOVY-7954
        // this case is still not covered: left side is Comparable and its implementation would return 0 in this case
        // but the compareTo code is not being triggered
        throw new UnsupportedOperationException("once this is fixed in groovy we should update examples and docs")
    }
}
```

PR #896, opened 2019-03-12
GROOVY-9018: fix shellcheck warnings for bin scripts

calid:


paulk-asert:
I changed the .gitignore separately, so you can remove that commit.

Also, the completion script are generated by picocli, so we wouldn't apply that here since it would be overwritten the next time they are generated. Instead, if you have the time, it would be great if you approached that project and had them update their templates/code which are used when generating those files.

But the bootstrap changes look good to me. I'll see if anyone else comments.

calid:
ok sounds good, updated the PR accordingly

paulk-asert:
Merged, thanks!

PR #902, opened 2019-03-24
GROOVY-9056: Rename method `visit` of `ASTNode` to `accept` and add a…

danielsun1106:
…n alias method `visit`

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ASTNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ASTNode.java
@@ -50,8 +50,18 @@ public class ASTNode implements NodeMetaDataHandler {
     private int lastColumnNumber = -1;
     private Map metaDataMap = null;
 
+    public void accept(GroovyCodeVisitor visitor) {
+        throw new RuntimeException("No accept() method implemented for class: " + getClass().getName());
+    }
+
+    /**
+     * An alias method for {@link org.codehaus.groovy.ast.ASTNode#accept(GroovyCodeVisitor)}
+     * Note: the method will be removed in a future version, e.g. Groovy 4 or Groovy 5
+     *
+     * @param visitor the visitor
+     */
     public void visit(GroovyCodeVisitor visitor) {
-        throw new RuntimeException("No visit() method implemented for class: " + getClass().getName());
+        this.accept(visitor);
     }
 
     public String getText() {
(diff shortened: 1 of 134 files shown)


src/main/java/org/codehaus/groovy/ast/ASTNode.java
```diff
@@ -50,8 +50,18 @@
     private int lastColumnNumber = -1;
     private Map metaDataMap = null;
 
+    public void accept(GroovyCodeVisitor visitor) {
+        throw new RuntimeException("No accept() method implemented for class: " + getClass().getName());
+    }
+
+    /**
+     * An alias method for {@link org.codehaus.groovy.ast.ASTNode#accept(GroovyCodeVisitor)}
+     * Note: the method will be removed in a future version, e.g. Groovy 4 or Groovy 5
```
pzygielo: Why not @`Deprecate` it then?

danielsun1106: In order to not disturb groovy users, we even do not add `@Deprecated` to the alias method `visit`.

pzygielo: I was asking about `visit` method only, the one planned for removal. 
Wouldn't missing method (without prior warning) be more disturbing?

danielsun1106: As @paulk-asert said, "Once we have Java 9 as our minimum, we can use the `since` and `forRemoval` Deprecated attributes
and that will allow us to make such deprecations a little clearer. "

We won't remove it without any prior warnings, no worries ;-)
BTW, I believe the alias method `visit` will be reserved for a very long time.

eric-milles:
If someone has an existing class that extends {{ASTNode}} and provides custom visitation, it will be extending {{visit}}, not {{accept}}.  Your changes have {{accept}} being called throughout the codebase.  So for backwards compatibility, the default {{accept}} needs to call to the potentially overloaded {{visit}} until such time as {{visit}} is removed.

eric-milles:
Another problem this change will introduce: Any class that extends one of the `ASTNode` subclasses and intends to override the behavior of a parent's `visit` or even call to the parent's `visit` could be getting something unexpected.  Would a call `super.visit(visitor)` reliably map to the parent's `accept` method?

And if any class already has a method `accept(GroovyCodeVisitor)`, it will suddenly be an override.  (Probably less likely, but a consideration to make.)

paulk-asert:
Eric makes some good points. Maybe making this change is more trouble than it is worth. The visit/accept names are after all just a convention. Some languages use dispatch instead of visit by convention for instance. I agree that using visit/visit as our convention is probably unique if not a little strange but most of us have become familiar with it. An alternative might be to just rename the parameter from visitor -> acceptedVisitor. And perhaps add some Groovydoc: "... accepts a visitor .." or similar.

danielsun1106:
Since we can not keep backawards compatiblity via adding an alias `visit` method, I will close the PR and the related JIRA ticket.

P.S.  right/meaningful names can help us think well

PR #903, opened 2019-03-30
GROOVY-9060: Make the initial capacity of LinkedHashMap be power of 2…

danielsun1106:
… for better performance

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/InvokerHelper.java
+++ b/src/main/java/org/codehaus/groovy/runtime/InvokerHelper.java
@@ -70,6 +70,11 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import static java.lang.Math.ceil;
+import static java.lang.Math.log;
+import static java.lang.Math.max;
+import static java.lang.Math.pow;
+
 /**
  * A static helper class to make bytecode generation easier and act as a facade over the Invoker
  */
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/runtime/InvokerHelper.java
```diff
@@ -396,10 +402,33 @@ public static List createList(Object[] values) {
...
+     * 7: 8
+     * 8: 8
+     * 9: 16
+     * ...
+     *
+     * @param initialEntryCnt the initial entry count
+     * @return the initial capacity
+     */
+    public static int initialCapacity(int initialEntryCnt) {
+        return (int) pow(2, ceil(log(initialEntryCnt) / LN2));
```
blackdrag: How about  return 2<<Integer.highestOneBit​(initialEntryCnt)+1; ?

danielsun1106: `2<<Integer.highestOneBit​(initialEntryCnt)+1; ` increases its result too fast, e.g. 
When `initialEntryCnt` is 16, the result has already been 262144 😲

Thanks for your suggestion all the same.

blackdrag: ups, I did misread the documentation. Integer.highestOneBit​(initialEntryCnt)*2 then

danielsun1106: I will tweak the current implementation according to suggestion, which will be more efficient 😄

eric-milles:
Doesn't this just move the math from inside of HashMap to InvokerHelper.initialCapacity?  Where is the performance gain?  Is there any benchmark for this?

Did the constant {{LN2}} get used anywhere?  It appears to be added but unused.

danielsun1106:
See  https://github.com/apache/groovy/commit/5c8f08633a320be8836a249d577a549fe5966b76

PR #904, opened 2019-04-05
GROOVY-9067: Compile Java stubs with standard API

danielsun1106:


danielsun1106:
@paulk-asert @blackdrag 
Please set aside some time to review the PR.

blackdrag:
you are doing more than undoing the change for the in-memory compilation? In general I am partial to in-memory compilation, but it must be done using the normal official API or else it will stop working at one point. Is that not possible?

danielsun1106:
@blackdrag 
Jochen, in-memory compilation is still supported, but in-file compilation is changed to rely on the same API, e.g. `javax.tools.JavaCompiler`. As a result, we can compile all source code in the same way and it will be easier for us to maintain.

danielsun1106:
@blackdrag 
As far as I know, `javax.tools.JavaCompiler` is a normal official API and will not stop working suddenly, isn't it?  :wink:
https://docs.oracle.com/javase/7/docs/api/javax/tools/JavaCompiler.html

danielsun1106:
@blackdrag @paulk-asert 

To sum up, the current implementation supports compiling Java stubs(in-memory and in-file) in the same way via the normal standard API `javax.tools.JavaCompiler`, which can eliminate Jochen's worries. 

If you do not object, I will merge the PR this weekend.

paulk-asert:
I was planning to double check that the ant task wasn't affected by this change but haven't had time yet:
subprojects/groovy-ant/src/main/java/org/codehaus/groovy/ant/GenerateStubsTask.java

danielsun1106:
OK. I will wait you, Paul 😉

danielsun1106:
@paulk-asert 
I find `GenerateStubsTask` is not used any more:
https://github.com/apache/groovy/blob/master/subprojects/groovy-ant/src/main/resources/org/codehaus/groovy/antlib.xml#L26

danielsun1106:
@paulk-asert 
I'm verifying the case you mentioned, i.e. generating stubs with `GenerateStubsTask`:
https://travis-ci.org/danielsun1106/groovy/builds/528064320?utm_source=github_status&utm_medium=notification

danielsun1106:
@paulk-asert 
All tests pass :-)

PR #905, opened 2019-04-07
GROOVY-8339: Fix warning "An illegal reflective access operation has …

danielsun1106:
…occurred"

### DIFF
--- a/src/main/java/org/codehaus/groovy/reflection/AccessPermissionChecker.java
+++ b/src/main/java/org/codehaus/groovy/reflection/AccessPermissionChecker.java
@@ -20,6 +20,7 @@ package org.codehaus.groovy.reflection;
 
 import groovy.lang.GroovyObject;
 
+import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
@@ -59,11 +60,26 @@ final class AccessPermissionChecker {
         }
     }
 
+    static void checkAccessPermission(Constructor constructor) {
+        try {
+            checkAccessPermission(constructor.getDeclaringClass(), constructor.getModifiers(), constructor.isAccessible());
+        } catch (AccessControlException e) {
+            throw createCacheAccessControlExceptionOf(constructor, e);
+        }
+    }
+
     private static CacheAccessControlException createCacheAccessControlExceptionOf(Method method, AccessControlException e) {
         return new CacheAccessControlException(
                 "Groovy object can not access method " + method.getName()
                         + " cacheAccessControlExceptionOf class " + method.getDeclaringClass().getName()
                         + " with modifiers \"" + Modifier.toString(method.getModifiers()) + "\"", e);
+    }
+
+    private static CacheAccessControlException createCacheAccessControlExceptionOf(Constructor constructor, AccessControlException e) {
+        return new CacheAccessControlException(
+                "Groovy object can not access constructor " + constructor.getName()
+                        + " cacheAccessControlExceptionOf class " + constructor.getDeclaringClass().getName()
+                        + " with modifiers \"" + Modifier.toString(constructor.getModifiers()) + "\"", e);
     }
 
     static void checkAccessPermission(Field field) {
--- a/build.gradle
+++ b/build.gradle
@@ -236,12 +236,20 @@ sourceSets {
                 exclude '**/v8/*'
                 exclude '**/vm8/*'
             }
+            if (!JavaVersion.current().isJava9Compatible()) {
+                exclude '**/v9/*'
+                exclude '**/vm9/*'
+            }
         }
         groovy {
             srcDir "$generatedDirectory/antlr2/src/main"
             if (!JavaVersion.current().isJava8Compatible()) {
                 exclude '**/v8/*'
                 exclude '**/vm8/*'
+            }
+            if (!JavaVersion.current().isJava9Compatible()) {
+                exclude '**/v9/*'
+                exclude '**/vm9/*'
             }
         }
         resources {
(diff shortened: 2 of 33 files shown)


danielsun1106:
@paulk-asert @blackdrag 
The PR is ready to review :-)

src/main/java/org/codehaus/groovy/reflection/AccessPermissionChecker.java
```diff
@@ -59,13 +60,28 @@ static void checkAccessPermission(Method method) {
...
+
     private static CacheAccessControlException createCacheAccessControlExceptionOf(Method method, AccessControlException e) {
         return new CacheAccessControlException(
                 "Groovy object can not access method " + method.getName()
                         + " cacheAccessControlExceptionOf class " + method.getDeclaringClass().getName()
                         + " with modifiers \"" + Modifier.toString(method.getModifiers()) + "\"", e);
     }
 
+    private static CacheAccessControlException createCacheAccessControlExceptionOf(Constructor constructor, AccessControlException e) {
+        return new CacheAccessControlException(
```
melix: I feel like using an exception here is likely to have a huge impact on performance, given that you might throw an exception for each method. The problem is that it's going to fill the stack trace everytime here. Would be good to do some measurements.

danielsun1106: Thanks for your reviewing the PR.
I understand your worries ;-)  Actually the exception will be only thrown if illegal access happens when the accessible object is accessed.

glaforge: CacheAccessControlException could perhaps also override `fillInStackTrace()` to avoid filling in the stack trace, which is an expensive operation?

melix: Understood, but the whole point of this PR is that _it happens_ ;)

paulk-asert:
+1 with the following comments:
* I haven't done a performance check yet but will try to do that soon
* I think some of the functionality in ReflectionUtils might be better in VMPlugin? Since we already have some similar functionality there?
* If we do need to backport to 2_5_X, some of the lambda code will need to be back-ported

danielsun1106:
> I think some of the functionality in ReflectionUtils might be better in VMPlugin? Since we already have some similar functionality there?

`ReflectionUtils` provides some facade API for `VMPlugin` , which will make our life easy IMO ;-)

src/main/java/org/codehaus/groovy/reflection/AccessPermissionChecker.java
danielsun1106: > `CacheAccessControlException` could perhaps also override `fillInStackTrace()` to avoid filling in the stack trace, which is an expensive operation?

@glaforge Thanks for your suggestion. We should be able to override the `fillInStackTrace()`, but I am not sure whether we do not really need the stack trace information. 


```java
    /**
     * Fills in the execution stack trace. This method records within this
     * {@code Throwable} object information about the current state of
     * the stack frames for the current thread.
     *
     * <p>If the stack trace of this {@code Throwable} {@linkplain
     * Throwable#Throwable(String, Throwable, boolean, boolean) is not
     * writable}, calling this method has no effect.
     *
     * @return  a reference to this {@code Throwable} instance.
     * @see     java.lang.Throwable#printStackTrace()
     */
    public synchronized Throwable fillInStackTrace() {
        if (stackTrace != null ||
            backtrace != null /* Out of protocol state */ ) {
            fillInStackTrace(0);
            stackTrace = UNASSIGNED_STACK;
        }
        return this;
    }
```

As you know, `CacheAccessControlException` is an existing exception class, which has been used in Groovy implementation(shown as follows), so @melix's worries also exists in the Groovy for a long time ;-)

* https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/reflection/AccessPermissionChecker.java#L63
* https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/reflection/AccessPermissionChecker.java#L78

paulk-asert:
> > I think some of the functionality in ReflectionUtils might be better in VMPlugin? Since we already have some similar functionality there?
> 
> `ReflectionUtils` provides some facade API for `VMPlugin` , which will make our life easy IMO ;-)

For me there are two code smells but I don't think either would be hard to fix:
* Given that VMPlugin is designed to run the correct method for the version of the JVM we are using, we shouldn't need to also check the version
* We have some functionality in ReflectionUtils and some similar functionality in VMPlugin and it's not clear to me that they aren't all at the same level - if that is the case we'd need to document it that way but I suspect we can consolidate into one level

danielsun1106:
> Given that VMPlugin is designed to run the correct method for the version of the JVM we are using, we shouldn't need to also check the version

Agreed.

> We have some functionality in ReflectionUtils and some similar functionality in VMPlugin and it's not clear to me that they aren't all at the same level - if that is the case we'd need to document it that way but I suspect we can consolidate into one level

What I know about vm plugins is just to provide the plug-in functions for different JVMs. But vm plugins have little abstraction, just some concrete implementations. It's better to rely on abstraction if we want some stable API. (My 2 cents)
How about moving the implementation of the following method to VMPlugin(e.g. Java5-Java9 plugins), then add a method in `ReflectionUtils` with same signatures delegating to the method in VMPlugin.
`public static boolean trySetAccessible(AccessibleObject ao)` 
https://github.com/apache/groovy/blob/2a8effa0880d245f1a01417974c812aac9410bd4/src/main/java/org/codehaus/groovy/reflection/ReflectionUtils.java#L143

paulk-asert:
Something like that might be close to what I had in mind! ;-)

danielsun1106:
I'll change the code tonight.

danielsun1106:
@paulk-asert 
Minor refactoring is done. Please review again.

paulk-asert:
Merged. Great work Daniel!

PR #906, opened 2019-04-08
GROOVY-9073: added compiler option to enable preview features (JEP 12)

keeganwitt:


danielsun1106:
+1

keeganwitt:
I'm not sure why checks were failing.  It'd be kinda nice to have some tests for it too, but I'm not sure how best to do that.

danielsun1106:
Make some changes and push, which will trigger the CI build, at last we can see the build result.

JornVernee:
All of them seem to be failing with an NPE at `org.codehaus.groovy.classgen.asm.sc.StaticTypesMethodReferenceExpressionWriter.createParametersWithExactType(StaticTypesMethodReferenceExpressionWriter.java:207)`, e.g. : https://travis-ci.org/apache/groovy/jobs/517407421#L2283 just searched for 'caused by' in the travis log and it's all the same source. But, according to github, that line is a comment: https://github.com/apache/groovy/blob/d744dfbbdc0741750e3af0b233c46af298a1e724/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesMethodReferenceExpressionWriter.java#L204-L212

If it's an infrastructure problem you could try closing and re-opening the PR, which should trigger travis to run again.

keeganwitt:
Well, I should change groovyc Ant stuff anyway, and that will make it re-kick.

paulk-asert:
Okay, there was a missing character in one adoc file and a couple of tiny formatting fixes needed for two tests. So, I have a clean build - but I think we might need to do a little bit more work first at least documenting what this feature does. I'll comment further on the issue.

keeganwitt:
Sorry, I'm having some troubles running tests on my local machine at the moment.  Thanks for taking a look @paulk-asert.

paulk-asert:
I merged this. I did some tweaks so that the --enable-preview flag was passed to the JVM too. In fact it is now doubled up in some sense but we should be able to refactor later and remove that duplication. I just wasn't sure I had all of the scenarios where even just using the groovy command might write bytecode, so I played safe and left it in both places.

PR #907, opened 2019-04-09
Fix minor mismatch between text and code

mkutz:
In the test below the code sample March 1st, but the code creates a
March 2nd.


paulk-asert:
+1 Nice catch. The current test still passes but not according to how the text describes.

paulk-asert:
Merged.

PR #910, opened 2019-04-15
GROOVY-9083: groovyConsole should be updated to understand method ref…

paulk-asert:
…erences and lambda expressions


danielsun1106:
+1

mvmn:
Why doesn't this work in Groovy Shell though?

```
> [1,2,3].stream().map(String::valueOf).collect(java.util.stream.Collectors.toList())
groovysh_parse: 2: unexpected token: : @ line 2, column 29.
   [1,2,3].stream().map(String::valueOf).collect(java.util.stream.Collectors.toList())
                               ^
> [1,2,3].stream().map(v->String.valueOf(v)).collect(java.util.stream.Collectors.toList())
groovysh_parse: 2: unexpected token: -> @ line 2, column 23.
   [1,2,3].stream().map(v->String.valueOf(v)).collect(java.util.stream.Collectors.toList())
                         ^
```

danielsun1106:
@mvmn groovysh does not support new language features because it still bases on the old parser.

paulk-asert:
Our goal is to get as many of the tools ported across to the new parser before 3.0 GA but some may have to wait until a subsequent (e.g. 3.0.x or 3.1) release.

PR #913, opened 2019-04-26
GROOVY-9095: Default retention policy not implemented for annotation …

paulk-asert:
…(also covers GROOVY-9096)

[WIP] still requires tests. It's a little tricky to test since we can't use normal reflection for the CLASS case.

I also wonder whether we should do some more validation on the source retention flags. We would allow multiple to be set currently. Our code doesn't do this but we aren't stopping AST code from making such changes. It would make more sense to be an enum but I haven't changed that for backwards compatibility reasons.


paulk-asert:
Merged.

PR #917, opened 2019-05-01
GROOVY-9086: @CompileStatic of closure calls with OWNER_FIRST fail at…

paulk-asert:
… runtime with ClassCastException


paulk-asert:
Proposed PR merged.

PR #923, opened 2019-05-07
[GROOVY-9109] GroovyShell: don't override classloader if parent is al…

je-ik:
…ready GroovyClassLoader

### DIFF
--- a/src/main/java/groovy/lang/GroovyClassLoader.java
+++ b/src/main/java/groovy/lang/GroovyClassLoader.java
@@ -225,11 +225,14 @@ public class GroovyClassLoader extends URLClassLoader {
     }
 
     /**
-     * Retrieve configuration used by this class loader.
-     * @return CompilerConfiguration used by this classloader
+     * Check if this class loader has compatible {@link CompilerConfiguration}
+     * with the provided one.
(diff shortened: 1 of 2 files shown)


paulk-asert:
The build failure is highlighting the case where the config of the existing groovyclassloader is different to the supplied config.

je-ik:
Oops, thanks @paulk-asert. I didn't notice the fail.

paulk-asert:
That certainly fixed the breaking test. CompilerConfiguration is in general mutable - so being able to get the config allows mutation more readily which isn't ideal. I'll have a ponder whether there might be a better alternative. We have made the default configuration immutable recently, so maybe not as bad an issue as previously. If you passed in a config and retained a reference to it, you could change the config in any case.

je-ik:
@paulk-asert If I understand the problem correctly, you don't want to leak CompilerConfiguration from GroovyClassLoader, right? Does the last commit solve that?

je-ik:
@paulk-asert @danielsun1106 can we get this into upcoming 3.0.0-beta-3 release, please?

danielsun1106:
Merged. Thanks!

je-ik:
@danielsun1106 thanks! :-)

PR #932, opened 2019-05-23
Update eclipse gradle scripts

eric-milles:
@paulk-asert  This should fix the source folders and the project dependencies.  It does not fix the build cycle errors by itself.  However, I was able to add the following `user.gradle` contents to the base directory and the eclipse cycles are broken.

```gradle
project.configurations.testCompile.dependencies.removeAll { it.name =~ 'groovy-' }

ext {
    groovyVersion = '3.0.0-beta-1'
}

dependencies {
    testCompile "org.codehaus.groovy:groovy-ant:$groovyVersion", {
        exclude group: 'org.codehaus.groovy', module: 'groovy'
    }
    testCompile "org.codehaus.groovy:groovy-dateutil:$groovyVersion", {
        exclude group: 'org.codehaus.groovy', module: 'groovy'
    }
    testCompile "org.codehaus.groovy:groovy-macro:$groovyVersion", {
        exclude group: 'org.codehaus.groovy', module: 'groovy'
    }
    testCompile "org.codehaus.groovy:groovy-test:$groovyVersion", {
        exclude group: 'org.codehaus.groovy', module: 'groovy'
    }
    testCompile "org.codehaus.groovy:groovy-xml:$groovyVersion", {
        exclude group: 'org.codehaus.groovy', module: 'groovy'
    }
}
```

I will search for eclipse bugs on build cycle errors between main and test dependencies.  The test stuff is a recent addition to eclipse.

As it stands, you cannot build the projects in eclipse.  I still need to add the antlr grammar processing.  However you should be able to browse, search and edit files.  (After I fix some parser issues that are causing runaway indexing threads.)

I will also follow up with tasks to delete the files added to the `.settings` directory.


paulk-asert:
Merged, thanks.

PR #933, opened 2019-05-23
GROOVY-9141: add null check for abstract method with body

eric-milles:


danielsun1106:
As you are trying to modify the old parser, the test code is parsed by the new Parrot parser by default, so I think the test code can not cover the code changes of the old parser.

paulk-asert:
Merged with a few tweaks

PR #934, opened 2019-05-24
GROOVY-9103: CLONE - CLONE - Fix warning "An illegal reflective acces…

danielsun1106:
…s operation has occurred"

### DIFF
--- a/src/main/java/groovy/lang/MetaClassImpl.java
+++ b/src/main/java/groovy/lang/MetaClassImpl.java
@@ -142,6 +142,7 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
     };
     private static final MetaMethod[] EMPTY = MetaMethod.EMPTY_ARRAY;
     private static final MetaMethod AMBIGUOUS_LISTENER_METHOD = new DummyMetaMethod();
+    private static final boolean PERMISSIVE_PROPERTY_ACCESS = SystemUtil.getBooleanSafe("groovy.permissive.property.access");
 
     protected final Class theClass;
     protected final CachedClass theCachedClass;
@@ -173,6 +174,7 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
     private MetaMethod propertyMissingSet;
     private MetaMethod methodMissing;
     private MetaMethodIndex.Header mainClassMethodHeader;
+    private boolean permissivePropertyAccess = PERMISSIVE_PROPERTY_ACCESS;
 
      /**
       * Constructor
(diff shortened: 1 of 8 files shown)


src/main/java/groovy/lang/MetaClassImpl.java
```diff
@@ -2207,11 +2209,28 @@ private static MetaMethod getCategoryMethodSetter(Class sender, String name, boo
 //                    element = new MetaBeanProperty(mp.getName(), mp.getType(), null, mp.getSetter());
 //                }
             }
+            if (element instanceof MetaBeanProperty) {
+                MetaBeanProperty mbp = (MetaBeanProperty) element;
+                boolean getterAccessible = canAccess(mbp.getGetter());
+                boolean setterAccessible = canAccess(mbp.getSetter());
+
+                if (!(getterAccessible && setterAccessible)) continue;
```
danielsun1106: @paulk-asert Could you review the logic of fitering? 
getter and setter of property should be access legally, or the property will be filtered for now.
Also, filtering non-properties is a breaking change. Should we provide an option to allow the illegal access?(Suggestion: forbiddening illegal access by default)

danielsun1106:
JVM option `groovy.allow.illegal.access.properties` is added to enable illegal access properties for backward compatibility(Note: `false` is default value)

danielsun1106:
Search "WARNING: Illegal reflective access by" in the Java11 build log( https://api.travis-ci.org/v3/job/537214690/log.txt ), only 5 illegal access warning remains. They are all truely illegal access, we have to fix them by changing our code.

danielsun1106:
@paulk-asert The PR is ready to review :-)

danielsun1106:
@paulk-asert  ping  ;)

src/main/java/groovy/lang/MetaClassImpl.java
```diff
@@ -2207,11 +2210,34 @@ private static MetaMethod getCategoryMethodSetter(Class sender, String name, boo
...
+                    if (!(getterAccessible && setterAccessible)) continue;
+                }
+            }
+
             ret.add(element);
         }
         return ret;
     }
 
+    private static final boolean ALLOW_ILLEGAL_ACCESS_PROPERTIES = SystemUtil.getBooleanSafe("groovy.allow.illegal.access.properties");
```
paulk-asert: This all looks good - my only suspicion is that some folks will want to turn this feature on very selectively.
We could make an instance property which is set by the system property but could be toggled on an individual metaclass basis. I am also wondering whether "groovy.permissive.property.access" would be a better name. In future JDK versions the access might not be allowed at all - for us it is indicating that we will attempt to access the property.

danielsun1106: Thanks for you reviewing the PR, Paul :)

I wish I had caught the points:
1) Rename `groovy.allow.illegal.access.properties`  to `groovy.permissive.property.access`
2) Add an instance property whose value is initialized by the system property `groovy.permissive.property.access`, we use the instance property to control the illegal access to properties

danielsun1106:
@paulk-asert I've refined the switch according to your suggestion, pls review the PR again :)

https://github.com/apache/groovy/pull/934/commits/8d61efba824f80fea546c16ee8646921455e2b48

PR #936, opened 2019-05-26
GROOVY-9127: check for subclass access to field for this prop expression

eric-milles:


src/test/groovy/bugs/Groovy9141.groovy
```diff
@@ -18,41 +18,39 @@
...
-        assert err =~ / You cannot define an abstract method\[meth\] in the script. Try removing the 'abstract' /
-    }
+final class Groovy9141 extends GroovyTestCase {
 
     void testAbstractMethodWithBodyInClass() {
-        def err = shouldNotCompile """
-            class Main {
-                $METHOD_DEF
+        def err = shouldFail(CompilationFailedException, '''\
+            abstract class Main {
```
paulk-asert: Wouldn't it just be easier to add abstract in front of class on the one line and change the error message and leave the rest of the script as is?

eric-milles: I added a test case, refactored the style to match my other current and forthcoming tests and replaced use of CompilableTestSupport with the basic GroovyTestCase, now that I have seen how to check for compiler errors using {{shouldFai}}l.  As for the string substitution, I'd rather see the entire source being compiled and have a bit of repeat over having to mentally do variable substitutions when reading the tests.

If this style is acceptable, I may change other existing tests over time in this manner.  In the long term, the test cases could all look this way.

As for parserVersion, I mentioned in https://issues.apache.org/jira/browse/GROOVY-8547 that I'd prefer to see it eliminated.  And I planned on creating a pull request for it this weekend but gide side-tracked with the @CompileStatic and @PackageScope open items.

danielsun1106:
@eric-milles Would you mind to resolve the conflicts in file "Groovy9141.groovy"?

danielsun1106:
Merged. Thanks!

PR #938, opened 2019-05-27
Update tests for GROOVY-9043, GROOVY-9093, GROOVY-9106, et al.

eric-milles:
Fix class loading of p.One so that package-private use is as expected.  Prepped test cases for 9093 and 9106.


src/test/org/codehaus/groovy/transform/packageScope/PackageScopeTransformTest.groovy
```diff
@@ -18,20 +18,22 @@
...
 import java.lang.reflect.Modifier
 
-class PackageScopeTransformTest extends GroovyShellTestCase {
+@CompileStatic
+final class PackageScopeTransformTest extends GroovyShellTestCase {
     // GROOVY-9043
-    void testPackagePrivateAccessFromInnerClassCS() {
-        assertScript '''
+    void testStaticFieldAccessFromInnerClassCS() {
+        assertScript('''\
```
paulk-asert: The most common style within many of our tests is to avoid unnecessary characters like the open bracket (when not needed) and the backslash in this case. It no doubt affects the line count by one but hasn't seemed to worry anyone to date. Are there additional reasons why your style is to put them in from an Eclipse point of view?

eric-milles: I'm fine with that style for scripts and printlns.  But I find it much more natural to have the parens for method calls within classes.  I've seen the parens added back as soon as there is just one more param to pass, even though that form could have been written without parens.

The backslash is to escape the first newline; it is most useful if using `stripIndent()` on the string literal.  I would have provided that as well, but it was stripped off in a recent commit.

paulk-asert: The stripIndent would make it harder to align column information in error messages with column numbers in the IDE. The extra newline at the start is usually fine but does mean you need to start your line counting from the opening string delimiter. We have over 3000 places in tests using assertScript without the parens and just a small handful with the parens and a single multiline string param. I can see arguments for both styles - I am inclined to leave things consistent with the status quo unless we have a compelling reason.

danielsun1106:
@paulk-asert What do you think about the PR? It has been tweaked for many times.

danielsun1106:
Merged. Thanks!

PR #940, opened 2019-05-28
GROOVY-4438: enumConstants: reduce ambiguity within loop

eric-milles:
Allows Eclipse IDE to work with the project.  Without this change, the indexer thread(s) and others get into the infinite loop described in GROOVY-4438.


eric-milles:
**NOTE**: Please hold on merging this.  During additional testing I have found that parsing fails for an enum like this:

```groovy
enum Orientation
{
    LANDSCAPE, PORTRAIT

    @Override
    String toString()
    {
        name().toLowerCase().capitalize()
    }
}
```

PR #943, opened 2019-05-30
Refactor and allow default CompilerConfiguration as parent allows

eric-milles:
Additional cleanup for sub-types of ProcessingUnit


danielsun1106:
Merged. Thanks!

PR #945, opened 2019-06-01
GROOVY-9074: Allow chooseBestMethod to match "? super T"

eric-milles:
See comment in GROOVY-9074 for next step(s)


danielsun1106:
Eric, I find some test names starting with `_FIXME_`, so I am not sure whether the PR is completed or not.

eric-milles:
This PR gets the code to a stable state where at least "? super T" is better supported.  There is discussion on options where to go from here in 9074.  But I wanted to at least get this in to be sure it does not have negative impact.

danielsun1106:
Gotcha. 
Merged. Thanks :-)

PR #948, opened 2019-06-03
eclipse: mark antlr source folders as optional

eric-milles:
image

Generate the grammar source folders with:
>gradlew antlr2classes generateGrammarSource eclipse

GrapeMain makes references to inner classes from type annotation.  Eclipse respects the static scoping of type annotations, so qualifiers are required.  GROOVY-8063 has more details.

This fixes 99% of the build errors in Eclipse IDE.


danielsun1106:
Merged. Thanks!

PR #949, opened 2019-06-04
GROOVY-9151: fix variable expressions referring to replaced parameters

eric-milles:
`def meth(int x = 1, def y = { -> x * 2 }) {...}` would generate:
```groovy
    @Generated
    def meth() {
      Reference x = new Reference((int) 1)
      this.meth((int) 1, (Object) { -> x * 2 })
    }
```
and now:
```groovy
    @Generated
    def meth() {
      int x = 1 // with closureShare=true
      this.meth(x, (Object) { -> x * 2 })
    }
```

Generation of the local variable has been expanded to any reference found within the arguments.  This fixes the method case mentioned in GROOVY-9151.  The constructor case is not yet handled since local variable declarations cannot be added before a special constructor call (aka `this(...)`).

Anonymous inner setting of enclosing method was expanded to handle anon. inner appearing anywhere in the default expression.  Previous fix only handled anon. inner ctor call as the whole default value.


danielsun1106:
Merged. Thanks!

PR #950, opened 2019-06-06
Add first batch of tests for @Generated. Relates to GROOVY-9164

aalmiray:
https://issues.apache.org/jira/browse/GROOVY-9164

### DIFF
--- a/src/test/groovy/generated/AbstractGeneratedAstTestCase.groovy
+++ b/src/test/groovy/generated/AbstractGeneratedAstTestCase.groovy
@@ -12,6 +12,14 @@ abstract class AbstractGeneratedAstTestCase {
         new GroovyClassLoader().parseClass(str)
     }
 
+    protected final void assertClassIsAnnotated(Class<?> classUnderTest) {
+        assert classUnderTest.getAnnotation(Generated)
+    }
+
+    protected final void assertClassIsNotAnnotated(Class<?> classUnderTest) {
+        assert !classUnderTest.getAnnotation(Generated)
+    }
+
     protected final void assertConstructorIsAnnotated(Class<?> classUnderTest, Class... paramTypes) {
         assert findConstructor(classUnderTest, paramTypes).getAnnotation(Generated)
     }
--- a/src/test/groovy/generated/AbstractGeneratedAstTestCase.groovy
+++ b/src/test/groovy/generated/AbstractGeneratedAstTestCase.groovy
@@ -9,7 +9,7 @@ import java.lang.reflect.Method
 @CompileStatic
 abstract class AbstractGeneratedAstTestCase {
     protected final Class<?> parseClass(String str) {
-        new GroovyClassLoader().parseClass(str.stripMargin())
+        new GroovyClassLoader().parseClass(str)
     }
 
     protected final void assertConstructorIsAnnotated(Class<?> classUnderTest, Class... paramTypes) {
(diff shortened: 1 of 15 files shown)


src/test/groovy/generated/AbstractGeneratedAstTestCase.groovy
```diff
@@ -0,0 +1,60 @@
...
+
+import groovy.transform.CompileStatic
+import groovy.transform.Generated
+
+import java.lang.reflect.Constructor
+import java.lang.reflect.Method
+
+/**
+ * @author Dmitry Vyazelenko
+ * @author Andres Almiray
```
paulk-asert: Apache generally discourages explicit @author tags. They get out of date quickly, provide less info than git will already tell you and can sometimes discourage participation if certain files are perceived to belong to particular individuals rather than the whole community. I'd suggest removing them but add Dmitry as contributor in pomconfigurer,gradle - you are already there. :-)

src/test/groovy/generated/AutoExternalizeGeneratedTest.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * @author Dmitry Vyazelenko
+ * @author Andres Almiray
+ */
+@CompileStatic
+class AutoExternalizeGeneratedTest extends AbstractGeneratedAstTestCase {
+    final Class<?> implicitAutoExternalize = new GroovyClassLoader().parseClass('''@groovy.transform.AutoExternalize
+       |class ClassUnderTest {
+       |}'''.stripMargin())
```
paulk-asert: Nothing wrong with `stripMargin` but for consistency with other tests (and to reduce unnecessary noise) I'd suggest just getting rid of it and the `|` characters since Groovy ignores whitespace - unless you were wanting to test specific column numbers of error messages or similar where it would make sense to keep.

src/test/groovy/generated/AbstractGeneratedAstTestCase.groovy
```diff
@@ -6,12 +6,12 @@ import groovy.transform.Generated
...
 import java.lang.reflect.Method
 
-/**
- * @author Dmitry Vyazelenko
- * @author Andres Almiray
- */
 @CompileStatic
 abstract class AbstractGeneratedAstTestCase {
+    protected final Class<?> parseClass(String str) {
+        new GroovyClassLoader().parseClass(str.stripMargin())
```
paulk-asert: That's nicer that it is hidden but you can simply remove it and the'|' characters if you want.

aalmiray: Right. We took inspiration on another AST test that also uses '|' and stripMargin, thought it would be OK to keep it consistent.

paulk-asert: The existing `stripMargin` examples are possibly older tests that we haven't tidied up but for me this falls into "beauty is in the eye of the beholder" and I think we can embrace multiple forms of beauty. :-)

aalmiray: Fixed. Also re-instantiated skipped tests as GROOVY-9161/9162/9163 have been fixed in master

paulk-asert: LGTM

paulk-asert:
Are you happy to commit? Or want me to?

aalmiray:
I can commit and port it to 2_5_X. Just wanted to be sure it was reviewed 😄

paulk-asert:
You should have write access via the gitbox url and should be able to cherry pick from the 2_5_X branch from there. But shout if you are having troubles, we'll need to look into why. I have no probs doing it but would be good to have you all set up.

danielsun1106:
Merged. Thanks!

PR #951, opened 2019-06-09
GROOVY-9133 Loop mode for groovy-console

JacobAae:
This pull requests enables:
* Setting to enable loop mode in script menu
* If enabled run the code in a looping fashion with 1000 millis break after it has ended.
* Preference added to update the time between loops

Work started at GR8Conf EU 2019 Hacker Garden - with Credit to the group of https://github.com/karfunkel , https://github.com/sbglasius and https://twitter.com/a_taylor95 - And guidance from Paul King


paulk-asert:
Merged with a rename of prefs variable - hope that's okay

PR #953, opened 2019-06-11
GROOVY-9170: exclude non-interface methods from methods map

eric-milles:
`ClassNodeUtils.addDeclaredMethodsFromInterfaces` and `getDeclaredMethodsFromInterfaces` are returning methods from `java.lang.Object`.  One reason is traits.  You could investigate this further, but the filter conditions added to these methods prevent ever returning non-interface declared methods.


danielsun1106:
Merged. Thanks!

PR #957, opened 2019-06-17
Groovy 3 feature

davidekholm:
Groovy 3.0.0 beta 1 has broken code that earlier worked in Alpha 4:
Say you have a Java class with a setTitle(String) method. Now you override this class with a Groovy class. In it you declare the following member variable:

JLabel title = new JLabel("skin");

This breaks the Groovy compiler in v3.0.0 beta 1 (but not alpha 4) with the following error:
The return type of javax.swing.JLabel getTitle() in Script34$1$12 is incompatible with java.lang.String in se.datadosen.component.ControlPanel
. At [-1:-1]  @ line -1, column -1.

Two things are noteworthy here:
1) The line number references (-1) are broken
2) Instead of letting the declared "title" variable hide the ability to call setTitle on the base class using the syntactic sugar of Groovy (foo = being syntactic sugar for setFoo(...)), it now bails out with this error, breaking Java compatibility. I expected this declaration to be allowed and simply hide the ability to call setTitle on the base class via that syntactic sugar.


paulk-asert:
This looks like an unintended regression as per your description but I am unsure about this PR. Did you mean to file an issue in the issue tracker?

davidekholm:
Yes. I have posted via Jira now instead. Sorry

paulk-asert:
All good! :-)

PR #958, opened 2019-07-01
GROOVY-6809 and GROOVY-9168: AIC before delegate constructor call

eric-milles:
Add support for:
```groovy
class A {
  A() {
    this(new Runnable() {
      @Override
      void run() {
        ...
      }
    })
  }
  private A(Runnable action) {
    ...
  }
}
```
and
```groovy
 class A {
  A(Runnable action = new Runnable() { void run() { ... }}) {
    this.action = action
  }
  Runnable action
}
```


eric-milles:
CI build passed on all JDKs except 10, which was stopped before it could complete.

danielsun1106:
Merged. Thanks!

PR #959, opened 2019-07-01
GROOVY-9176: fix line and column of error for property-generated method

eric-milles:
Errors like covariant or override checks performed by Verifier go to line -1, column -1 for property-generated methods.  This change allows the error to be associated with the property node, which helps remove some of the mysticism of where an accessor method comes from.


paulk-asert:
Merged, thanks.

PR #963, opened 2019-07-08
 GROOVY-9168: emit more specific error for use of uninitialized this

eric-milles:
Improved error message for cases like:
```groovy
class Outer {
  class Inner {}
  Outer(Inner inner) {}
  Outer() { this(new Inner()) } // uninitialized this passed to inner class constructor
}
```
Simplified `StaticVerifier` and handle `this.@foo` and `this.&foo` in addition to `this.foo` and implicit-this `foo()`.  Special constructor call args and parameter defaults have unified handling.

One test case was disabled.  It is not possible to distinguish this case from other error cases.  And you can add `Outer.this.` qualifier to reference the property (next test case in the unit test source).
```groovy
    void _FIXME_testAccessOuterClassMethodFromInnerClassConstructor() {
        assertScript '''
            class Parent {
                String str
                Parent(String s) { str = s }
            }
            class Outer {
                String a

                private class Inner extends Parent {
                   Inner() { super(getA()) } // Is this a reference to instance property or outer property?
                }

                String test() { new Inner().str }
            }
            def o = new Outer(a:'ok')
            assert o.test() == 'ok'
        '''
    }
```

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/InnerClassVisitor.java
+++ b/src/main/java/org/codehaus/groovy/classgen/InnerClassVisitor.java
@@ -48,9 +48,6 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
-import static org.codehaus.groovy.ast.tools.GeneralUtils.classX;
-import static org.codehaus.groovy.ast.tools.GenericsUtils.nonGeneric;
-
 public class InnerClassVisitor extends InnerClassVisitorHelper implements Opcodes {
 
     private ClassNode classNode;
@@ -74,7 +71,7 @@ public class InnerClassVisitor extends InnerClassVisitorHelper implements Opcode
         InnerClassNode innerClass = null;
         if (!node.isEnum() && !node.isInterface() && node instanceof InnerClassNode) {
             innerClass = (InnerClassNode) node;
-            if (!isStatic(innerClass) && innerClass.getVariableScope() == null) {
+            if (innerClass.getVariableScope() == null && (innerClass.getModifiers() & ACC_STATIC) == 0) {
                 innerClass.addField("this$0", ACC_FINAL | ACC_SYNTHETIC, node.getOuterClass().getPlainNodeReference(), null);
             }
         }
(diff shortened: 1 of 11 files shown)


eric-milles:
There is a redundant check in `Verifier.visitConstructor` (included below).  Is this the better place to do checking for uninitiualized this/super references -- instead of in `StaticVerifier`?  I think `Verifier` runs later and could catch references added by the compiler.

```java
    @Override
    public void visitConstructor(ConstructorNode node) {
        GroovyCodeVisitor checkSuper = new CodeVisitorSupport() {
            boolean firstMethodCall = true;
            String type = null;

            @Override
            public void visitMethodCallExpression(MethodCallExpression call) {
                if (!firstMethodCall) return;
                firstMethodCall = false;
                String name = call.getMethodAsString();
                // the name might be null if the method name is a GString for example
                if (name == null) return;
                if (!name.equals("super") && !name.equals("this")) return;
                type = name;
                call.getArguments().visit(this);
                type = null;
            }

            @Override
            public void visitConstructorCallExpression(ConstructorCallExpression call) {
                if (!call.isSpecialCall()) return;
                type = call.getText();
                call.getArguments().visit(this);
                type = null;
            }

            @Override
            public void visitVariableExpression(VariableExpression expression) {
                if (type == null) return;
                String name = expression.getName();
                if (!name.equals("this") && !name.equals("super")) return;
                throw new RuntimeParserException("cannot reference " + name + " inside of " + type + "(....) before supertype constructor has been called", expression);
            }
        };
        Statement s = node.getCode();
        if (s == null) {
            return;
        } else {
            s.visit(new VerifierCodeVisitor(this));
        }
        s.visit(checkSuper);
    }
```

eric-milles:
Also fixes GROOVY-9189

eric-milles:
This should close out GROOVY-6777, GROOVY-6809, GROOVY-7609, GROOVY-9168, and GROOVY-9189.

eric-milles:
Builds are timing out at 50 min, which is very close to the amount of time they require now.

danielsun1106:
Yep. Maybe we have to remove `testWithIndy` as before.

danielsun1106:
Merged. Thanks!

PR #964, opened 2019-07-10
GROOVY-8819: set source position of property's new constant expression

eric-milles:
References to static final fields from super type are losing source position on the ConstantExpression that is created under the PropertyExpression.  StaticImportVisitor provides a utility method for setting this.


danielsun1106:
Merged. Thanks!

PR #965, opened 2019-07-14
GROOVY-9192: Implement parser error recovery for Parrot parser

danielsun1106:


danielsun1106:
@eric-milles Pls give it a try when the PR is merged.

eric-milles:
Before applying these changes, Groovy-Eclipse has 161 errors and 12 failures in the compiler/builder/editor test suite under Groovy 3 (Parrot Parser).  I have included a screenshot below so you can see the sorts of things that are tested.

After applying these changes, I ended up with 130+ errors and 53 failures in the same test suite.  Also, it took a much greater time to complete, CPU usage was quite high and I ran into out-of-memory errors about 2/3 of the way through.

image

danielsun1106:
Well... the built-in parser error recovery of antlr4 seems not to meet the requirement...

PR #967, opened 2019-07-15
GROOVY-8547: remove ParserVersion in favor of CC.setPluginFactory(PPF)

eric-milles:
As requested, here is the long overdue PR for GROOVY-8547.

setPluginFactory implements strategy pattern; setParserVersion was redundant and led to incorrect checks for Antlr2 parser via version, when it could have been set via setPluginFactory instead.


danielsun1106:
LGTM

paulk-asert:
Merged. Thanks!

PR #973, opened 2019-07-20
GROOVY-9136: set source position of object expression for error messages

eric-milles:
Set only line and column to prevent overlap with property expression.  If the object expression overlaps with the property expression, our type inferencing in the source editor is compromised.


danielsun1106:
Merged. Thanks!

PR #974, opened 2019-07-20
GROOVY-8002: propagate source position to method in chain assignment

eric-milles:


danielsun1106:
Is it better to set the position of method nodes in the parsers?

eric-milles:
This is a common pattern, the original parsed AST is transformed and the new nodes need to have source ranges transferred.  In this case `PoppingMethodCallExpression trn = new PoppingMethodCallExpression(receiver.transformExpression(transformer), setter, (TemporaryVariableExpression) tmp.transformExpression(transformer));` creates new nodes.  The entire expression does not have source position set.  In addition, the object expression and method expression may be new and lack source position.

danielsun1106:
Agreed.
Merged. Thanks!

PR #978, opened 2019-07-21
Add openjdk13 build to travis CI

danielsun1106:


danielsun1106:
Gradle 5.5.1 does not support JDK13 for now

PR #980, opened 2019-07-22
GROOVY-8279: Adapt the groovysh with the new parser Parrot

paulk-asert:


paulk-asert:
This isn't the complete solution. It enables the test suite to pass and supports numerous forms of Groovy 3 syntax that I tried in an exploratory fashion. It is known to be deficient in terms of completion, e.g. !instanceof/!in infix operators don't have auto-completion whereas in/instanceof do and possibly others. There are possibly additional cases which need to be handled to support the multi-line "approach" (hack?) which groovysh uses. The old parser is obtainable in the usual way, e.g. set JAVA_OPTS=-Dparser.antlr4=false

paulk-asert:
I merged the first commit. I'll include the second commit into a subsequent PR for the cloned issue.

PR #985, opened 2019-08-06
GROOVY-9059: use redirect in generics spec if redirect is placeholder

eric-milles:
When resolving a type parameter like "X extends Y", instead of adding "X: X -> Object" to the generics spec, add "X: Y -> Object" to the spec and continue checking the spec if a returned type is a placeholder.

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
+++ b/src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
@@ -22,7 +22,6 @@ import antlr.RecognitionException;
 import antlr.TokenStreamException;
 import groovy.lang.Tuple2;
 import groovy.transform.stc.IncorrectTypeHintException;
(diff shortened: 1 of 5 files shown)


danielsun1106:
+1

danielsun1106:
I am really happy to know you could set aside time to fix generics bugs, which are usually hard and may cost hours to fix and iron out wrinkles. Thanks a lot for your contribution!

BTW, do you notice GROOVY-8409(https://issues.apache.org/jira/browse/GROOVY-8409)? It is hard to fix but quite important for type system too. If you could set aside time to investigate it, that would be great :-)

eric-milles:
I can take a look at 8409.  I will need to send in my unit test refactors; they allowed me to compile far enough in IDE to run debug on unit tests.

________________________________

danielsun1106:
Merged. Thanks!

PR #942, opened 2019-05-29
Console, GroovyMain and Groovysh (via its Main) use context class loader

eric-milles:
- All: pass system properties to new CompilerConfiguration instances
- Console: use default parameters to reduce constructor variants

As discussed with @paulk-asert; support for starting Colsole, et al. using GroovyStarter.  This would allow separate classpath similar to how compiler has transform classpath (for tooling) and standard classpath (for user code).

See also https://github.com/groovy/groovy-eclipse/issues/890

**NOTE**: I don't have any unit tests created yet.  Essentially I would load Console using GroovyStarter, as I have tested separately.

PR #986, opened 2019-08-08
Console, GroovyMain and Groovysh (via its Main) use context class loader

eric-milles:
- All: pass system properties to new CompilerConfiguration instances
- Console: use default parameters to reduce constructor variants

(Clone of 93bebbe140846117cf16127360217dac1d13f5fc)

See also https://github.com/apache/groovy/pull/942


danielsun1106:
Merged. Thanks for your taking care of syncing branches.
BTW, you should have privilege to commit when you become a committer. You can cherry-pick the commit you want to sync ;-)

PR #989, opened 2019-08-12
Fix issues in tests so Eclipse can compile far enough to debug

eric-milles:
* Add script filters for test support files
* Added missing package statements (or qualifiers)
* Moved some type declarations inside test class to hide them
* Imported public method instead of private field in static import test


danielsun1106:
Merged. Thanks!

PR #990, opened 2019-08-12
GROOVY-9223: Avoid generating common methods for each groovy class

danielsun1106:


danielsun1106:
@blackdrag @paulk-asert 
Please review the PR :-)

blackdrag:
I am actually a bit surprised there has been no problem with the meta class invocation logic here, because I remember this being a little bit fragile... To be specific, if you had a call on a GroovyObject using invokeMethod that does not exist, then the meta class logic for the fallback to invokeMethod was sometimes triggered leading to an endless loop. I think we reliably solved that back then, but I don't remember where this logic was (maybe in the callsite stuff and/or the bytecode interface) exactly. But if all tests pass I guess it is ok

danielsun1106:
Thanks for your definitive reviewing 😃
When I saw you had approved the PR, I merged it.

PR #991, opened 2019-08-12
GROOVY-9031: correct trait property generics before writing stub methods

eric-milles:
Replace use of `Traits.findTraits` with local implementation that does not use `GenericsUtils.getParameterizedType` on returned values.


danielsun1106:
Merged. Thanks!

PR #992, opened 2019-08-14
GROOVY-9215: Incorrect compile time access error is raised when using…

danielsun1106:
… @CompileStatic and/or @TypeChecked


paulk-asert:
+1, the other option would be to give some kind of compiler error but I think I am in favor of the approach you have taken here.

I guess it would be also good to check that when you have `@CompileStatic` turned on via the config script but turned off explicitly using `@CompileDynamic` or `@CompileStatic(TypeCheckingMode.SKIP)` that everything still works as expected but that can be done separately from this PR.

PR #994, opened 2019-08-15
GROOVY-9075: The exception message should be more clear when a Groovy…

danielsun1106:
…Object's metaClass is wrong

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/MetaClassHelper.java
+++ b/src/main/java/org/codehaus/groovy/runtime/MetaClassHelper.java
@@ -97,6 +97,8 @@ public class MetaClassHelper {
         return accessible;
     }
 
+
+
     private static Boolean checkCompatiblePackages(Class at, Constructor constructor) {
         if (at.getPackage() == null && constructor.getDeclaringClass().getPackage() == null) {
             return Boolean.TRUE;
@@ -1015,5 +1017,22 @@ public class MetaClassHelper {
             return prop;
         }
         return BeanUtils.decapitalize(prop);
+    }
+
+    // GROOVY-9075
+    public static void checkMetaClassProperty(MetaClass oldValue, Object newValue) {
+        if (null == newValue) return;
+
+        if (!(newValue instanceof MetaClass)) {
+            throw new IllegalArgumentException("the new metaClass[" + newValue.getClass() + "] is not an instance of " + MetaClass.class);
+        }
+
+        MetaClass newMetaClass = (MetaClass) newValue;
+        if (!newMetaClass.getTheClass().isAssignableFrom(oldValue.getTheClass())) {
+            throw new IllegalArgumentException(
+                    "the new metaClass[" + newMetaClass.getTheClass() + "]"
+                            + " is not compatible with the old metaClass[" + oldValue.getTheClass() + "]"
+            );
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/codehaus/groovy/runtime/MetaClassHelper.java
```diff
@@ -1016,4 +1018,21 @@ public static String convertPropertyName(String prop) {
         }
         return BeanUtils.decapitalize(prop);
     }
+
+    // GROOVY-9075
+    public static void checkMetaClassProperty(MetaClass oldValue, Object newValue) {
+        if (null == newValue) return;
+
+        if (!(newValue instanceof MetaClass)) {
+            throw new IllegalArgumentException("the new metaClass[" + newValue.getClass() + "] is not an instance of " + MetaClass.class);
```
paulk-asert: I think the common convention is upper case for first letter.

danielsun1106: OK. I will refine the message later.

danielsun1106: OK. I will refine the message later.

src/main/java/org/codehaus/groovy/runtime/MetaClassHelper.java
```diff
@@ -1016,4 +1018,21 @@ public static String convertPropertyName(String prop) {
...
+        }
+
+        MetaClass newMetaClass = (MetaClass) newValue;
+        if (!newMetaClass.getTheClass().isAssignableFrom(oldValue.getTheClass())) {
```
paulk-asert: I am unsure that this is in fact the real requirement. The following currently works:
```
class MyMeta extends MetaClassImpl {
    MyMeta(Class klass) { super(klass) }
}
class S2 extends MetaClassImpl {
    S2(Class klass) { super(klass) }
}

def s1 = "foo${1+1}"
def m1 = new MyMeta(GString)
m1.initialize()
s1.metaClass = m1

def s2 = 'bar'
def m2 = new MyMeta(String)
m2.initialize()
s2.metaClass = m2

println m1.theClass.isAssignableFrom(m2.theClass)
println m2.theClass.isAssignableFrom(m1.theClass)
s1.metaClass = s2.metaClass
```
But wouldn't that now fail?

danielsun1106: It's a breaking change... but I am not sure whether it will break real use cases...

blackdrag:
Actually one more comment on the change.. as done it will check only if setting the property via the meta class. Is c1.metaClass = c2.metaClass  really supposed to go through the whole MOP if c1 is a Groovy class? Do we not directly set it (https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/classgen/Verifier.java#L467) directly? And at the very least... if the call is done from Java it will also not work. Maybe this is one of those cases where the DGM added methods shadow the methods defined on the class? I guess so... feels really kind of broken. I really would like to see an annotation on the DGM method that lets us control if the interface method is supposed to shadow a subclass defined method or not.

danielsun1106:
To be frank, I am not sure how the PR will impact existing valid use cases...
Thanks for your reviewing :-)
Let's leave it as it is and not merged.

PR #996, opened 2019-08-17
Fix potential bugs

blindpirate:
This is fixing some potential bugs from https://sonarcloud.io/dashboard?id=danielsun1106_groovy

@danielsun1106

### DIFF
--- a/src/main/java/groovy/lang/ExpandoMetaClass.java
+++ b/src/main/java/groovy/lang/ExpandoMetaClass.java
@@ -481,7 +481,7 @@ public class ExpandoMetaClass extends MetaClassImpl implements GroovyObject {
      * {@inheritDoc}
      */
     @Override
-    public synchronized void initialize() {
+    public void initialize() {
         try {
             writeLock.lock();
             if (!isInitialized()) {
--- a/src/main/java/groovy/lang/ExpandoMetaClass.java
+++ b/src/main/java/groovy/lang/ExpandoMetaClass.java
@@ -477,11 +477,11 @@ public class ExpandoMetaClass extends MetaClassImpl implements GroovyObject {
         ExpandoMetaClassCreationHandle.disable();
     }
 
-    /* (non-Javadoc)
-     * @see groovy.lang.MetaClassImpl#initialize()
+    /**
+     * {@inheritDoc}
      */
-
-    public void initialize() {
+    @Override
+    public synchronized void initialize() {
         try {
             writeLock.lock();
             if (!isInitialized()) {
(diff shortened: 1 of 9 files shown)


src/main/java/groovy/lang/ExpandoMetaClass.java
```diff
@@ -477,11 +477,11 @@ public static void disableGlobally() {
...
-    /* (non-Javadoc)
-     * @see groovy.lang.MetaClassImpl#initialize()
+    /**
+     * {@inheritDoc}
      */
-
-    public void initialize() {
+    @Override
+    public synchronized void initialize() {
         try {
```
danielsun1106: The method implements synchronization via ` lock` API,  the `synchronized` seems redundant.

blindpirate: You're right. Removed `synchronized.`

src/main/java/groovy/lang/ObjectRange.java
```diff
@@ -302,7 +302,7 @@ public int size() {
                 // let's fast calculate the size
                 final BigDecimal fromNum = new BigDecimal(from.toString());
                 final BigDecimal toNum = new BigDecimal(to.toString());
-                final BigInteger sizeNum = toNum.subtract(fromNum).add(new BigDecimal(1.0)).toBigInteger();
+                final BigInteger sizeNum = toNum.subtract(fromNum).add(BigDecimal.valueOf(1.0)).toBigInteger();
```
paulk-asert: We can probably just use: BigDecimal.ONE these days.

src/main/java/groovy/lang/ExpandoMetaClass.java
```diff
@@ -477,10 +477,10 @@ public static void disableGlobally() {
         ExpandoMetaClassCreationHandle.disable();
     }
 
-    /* (non-Javadoc)
-     * @see groovy.lang.MetaClassImpl#initialize()
+    /**
+     * {@inheritDoc}
```
paulk-asert: This is better for our "api" Javadoc but we currently don't support inheritDoc within Groovydoc:
https://issues.apache.org/jira/browse/GROOVY-3782
so it leaves a bit of noise in the "gapi" documentation - still, not a huge issue I guess.

src/main/java/groovy/lang/ObjectRange.java
blindpirate: Nice catch! Will change to `BigDecimal.ONE`.

src/main/java/groovy/lang/ExpandoMetaClass.java
blindpirate: But this is a java file... do you mean javadocs in java files are also processed by `groovydoc`?

paulk-asert: Yes, the gapi (Groovydoc ) is currently the combined documentation produced by converting java -> groovy and then running groovydoc on all. But as I said, not a big deal for this PR. It was more to inform why it was the way it was.

blindpirate: Thanks for the clarification.

danielsun1106:
Merged. Thanks.

PR #1010, opened 2019-09-27
Implement `CurlyLevelCountingGroovyLexer` based on parrot lexer

danielsun1106:
As antlr2 parser will be removed in the future, we should not rely on its API.

### DIFF
--- a/src/antlr/GroovyLexer.g4
+++ b/src/antlr/GroovyLexer.g4
@@ -143,9 +143,14 @@ options {
         }
     });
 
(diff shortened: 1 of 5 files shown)


danielsun1106:
@paulk-asert I find the first commit has been merged into master. How about the second commit in this PR?

paulk-asert:
Yes, I am basically doing second commit just into GROOVY_3_0_X. I'll run tests on that branch first.

danielsun1106:
Gotcha.

PR #1013, opened 2019-09-30
Remove antlr2 related code in `LexerFrame`

danielsun1106:


paulk-asert:
Maybe we should set  up a parallel package, e.g. org.apache.groovy.antlr.LexerFrame
We could leave the codehaus one generic supporting other lexer/tokentype combinations but deprecate it.
Then perhaps for the new apache version hard code the tokenTypesClass class as well as the lexerClass (not just the lexerClass as in this PR)?

danielsun1106:
@paulk-asert The PR has been adjusted. Pls review it again.

paulk-asert:
What about leaving codehaus version as is in GROOVY_3_0_X but just mark as deprecated then have new apache version in 3 and 4? This gives anyone using it the whole of 3 to convert over?

paulk-asert:
Merged! Thanks!

PR #1015, opened 2019-10-02
GROOVY-9266: added closure annotations to DefaultGroovyMethods.metaClass

eric-milles:
and ExpandoMetaClass.define

The actual closure param is always a null, but I didn't know if that could be represented using `@ClosureParams` so I just set "java.lang.Object".


danielsun1106:
Merged. Thanks!

PR #1017, opened 2019-10-03
GROOVY-9237: restore constructors for EmptyExpression and EmptyStatement

eric-milles:
Also, normalize the AST visitation of EmptyStatement and EmptyExpression


danielsun1106:
LGTM

eric-milles:
By overriding `setMetaDataMap`, all the attempts to write metadata will fail.  I don't think we need to enforce singleton in `EmptyExpression` or `EmptyStatement` themselves.  `INSTANCE` is the singleton, much like `ConstantExpression` has `NULL`, `THIS`, etc.

paulk-asert:
I was thinking to only throw UnsupportedOperationException if "locked" was set otherwise call super.setMetaDataMap(...).

danielsun1106:
As we approve most of the PR, merge it now and refine the rest later if necessary

PR #1018, opened 2019-10-04
GROOVY-9244: prevent ambiguous constructor if using @InheritConstructors

eric-milles:
```groovy
class Base {
  Base(Number n) {}
  Base(String s) {}
}
@groovy.transform.InheritConstructors
class Impl {
}
new Impl((String) null) // should trigger Base(String)
```

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/InheritConstructorsASTTransformation.java
+++ b/src/main/java/org/codehaus/groovy/transform/InheritConstructorsASTTransformation.java
@@ -38,6 +38,7 @@ import static org.apache.groovy.ast.tools.ClassNodeUtils.addGeneratedConstructor
 import static org.codehaus.groovy.ast.ClassHelper.make;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.args;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.block;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.castX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.ctorSuperS;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.param;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.varX;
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/transform/InheritConstructorsASTTransformation.java
```diff
@@ -98,31 +96,27 @@ private void addConstructorUnlessAlreadyExisting(ClassNode classNode, Constructo
...
     }
 
     private static boolean isExisting(ClassNode classNode, Parameter[] params) {
-        for (ConstructorNode consNode : classNode.getDeclaredConstructors()) {
-            if (matchingTypes(params, consNode.getParameters())) {
-                return true;
-            }
-        }
-        return false;
+        return classNode.getDeclaredConstructors().stream().anyMatch(ctor -> matchingTypes(params, ctor.getParameters()));
```
paulk-asert: You are probably already aware but just in case ... this is great for master/GROOVY_3_0_X but not for earlier branches. To make applying patches/PRs easier across branches we often even leave older style code in newer branches - we do hope to be focusing on Groovy 3/4 soon, so not a big issue here.

src/test/groovy/bugs/Groovy9244.groovy
```diff
@@ -0,0 +1,117 @@
...
+            @groovy.transform.InheritConstructors
+            class Impl extends Base {
+            }
+
+            def impl = new Impl((String) null)
+            assert impl.which == 'String'
+        '''
+    }
+
+    @Test @NotYetImplemented
```
paulk-asert: Yeah, these are tricky cases for a dynamic language.

eric-milles: Tried to create an easy to hard gradient.  And wanted them in in case a future change manages to fix any of them.

src/main/java/org/codehaus/groovy/transform/InheritConstructorsASTTransformation.java
eric-milles: Did not mean to cause extra work.  Would it be better to submit PR in Java 7 style and then follow with refactor on master if streams or whatever makes for shorter/simpler implementation?

eric-milles: To apply to other branches, you can reduce to a single change:
```java
            // cast argument to parameter type in case the value is null
            theArgs.add(castX(p.getType(), varX(p.getName(), newType)));
```

paulk-asert: yes - we can do that - easiest if that change was it's own commit but we can cope with just one line

danielsun1106:
Merged. Thanks!

PR #1020, opened 2019-10-07
Add JDK13 build to travis-ci

danielsun1106:


danielsun1106:
Gradle 5.6.2 does not support JDK 13 for now

PR #1022, opened 2019-10-08
Minor refactoring: Refine collecting errors when generating bytecode for native method reference

danielsun1106:
…eference


paulk-asert:
+1

PR #1023, opened 2019-10-08
Avoid deprecated `compile` and `testCompile`

danielsun1106:


danielsun1106:
@blindpirate 
After `compile` replaced with `implementation`, `:groovy-groovydoc:compileJava` failed because its dependency was not found. Am I misusing gradle's `implementation`?

```
> Task :groovy-groovydoc:compileJava
/home/travis/build/apache/groovy/subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/SimpleGroovyClassDocAssembler.java:21: error: package antlr.collections does not exist
import antlr.collections.AST;
```

melix:
You can't just replace `compile` with `implementation`. That's a good step, but it's not sufficient. The error you are seeing is a good one because it indicates what transitive compile dependencies are missing.

As the documentation states, `compile` must be replaced _either_ by `api` or `implementation`. The difference is whether the dependency "leaks" into the public API of your library or not. For example, if classes of a dependency are in the public signatures, superclasses, ... then it _must_ be declared as an API dependency (this dependency is effectively part of your API, similarly to `require transitive` in JPMS).

melix:
(In other words, `antlr` is an API dependency of `core`, probably)

danielsun1106:
@melix  Thanks for you help. The build is green now :-)

PR #1034, opened 2019-10-18
Fix "Non-atomic operation on volatile field"

danielsun1106:


danielsun1106:
`count` is guarded by lock

PR #1037, opened 2019-10-18
GROOVY-9281: allow package-private for resolved nested type

eric-milles:
refactor ResolveVisitor to use single method for testing visibility


danielsun1106:
+1

PR #1038, opened 2019-10-19
GROOVY-9282: Allow protected override of package scoped methods

gesellix:
See [GROOVY-9282](https://issues.apache.org/jira/browse/GROOVY-9282) for the Jira issue.

When overriding a method as `protected` with a super method being package-private (no modifier in Java or `@PackageScope` in Groovy), compilation fails with an error similar to this:

```
...
exampleMethod(java.lang.Object -> java.lang.Object) in com.example.Child cannot override exampleMethod in com.example.Base; attempting to assign weaker access privileges; was package-private
...
```

This edge case seems to be missed at [GROOVY-8651](https://issues.apache.org/jira/browse/GROOVY-8651).

Groovy 2.5.7 and 2.5.8 are affected. I didn't check Groovy 3.x.

Signed-off-by: Tobias Gesellchen <tobias@gesellix.de>

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/ClassCompletionVerifier.java
+++ b/src/main/java/org/codehaus/groovy/classgen/ClassCompletionVerifier.java
@@ -466,7 +466,7 @@ public class ClassCompletionVerifier extends ClassCodeVisitorSupport {
             Parameter[] superParams = superMethod.getParameters();
             if (!hasEqualParameterTypes(params, superParams)) continue;
             if ((mn.isPrivate() && !superMethod.isPrivate())
-                    || (mn.isProtected() && !superMethod.isProtected() && !superMethod.isPrivate())
+                    || (mn.isProtected() && !superMethod.isProtected() && !superMethod.isPackageScope() && !superMethod.isPrivate())
                     || (!mn.isPrivate() && !mn.isProtected() && !mn.isPublic() && (superMethod.isPublic() || superMethod.isProtected()))) {
                 addWeakerAccessError(cn, mn, params, superMethod);
                 return;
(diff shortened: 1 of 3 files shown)


danielsun1106:
LGTM

gesellix:
Would that be another check in another verifier? Otherwise I would extend the check in the current method we're looking at. Do you have any preference?

danielsun1106:
It's OK to add more check in the current class. FYI, here is the rule Java obeys, groovy should obey too:

If a method is declared with default access (that is, not private, protected, nor public), it can only be overridden by methods in the same package. If a method of the same signature is defined in a subclass in a different package, it is a completely separate method and no overriding occurs.

danielsun1106:
Could you add more test cases to cover the following rule:

If a method of the same signature is defined in a subclass in a different package, it is a completely separate method and no overriding occurs.

gesellix:
Due to Groovy's handling (some would say "ignorance" :) ) of private visibility at fields and methods I'm unsure how to provide such test cases. See https://issues.apache.org/jira/browse/GROOVY-3010 for a collection of tasks regarding that topic.

Would you mind if we consider that scenario unrelated to the current pull request or maybe can you provide a rough idea how such a test case should look like?

danielsun1106:
Here is the test case in Java to show the Java rule, i.e. "If a method of the same signature is defined in a subclass in a different package, it is a completely separate method and no overriding occurs", but it is hard to test in Groovy as Groovy will invoke methods on the instance of specific type(not the declaring type). 

@paulk-asert  any thoughts on the PR?

```
src
└─groovy9282
    │  Base.java
    │  DerivedSamePkg.java
    │
    └─subpkg
            DerivedDiffPkg.java
```

```java
package groovy9282;

import groovy9282.subpkg.DerivedDiffPkg;

public class Base {
    String x() {
        return "Base";
    }

        Base dsp = new DerivedSamePkg();
        if (!"DerivedSamePkg".equals(dsp.x())) throw new AssertionError();

        Base ddp = new DerivedDiffPkg();
        if (!"Base".equals(ddp.x())) throw new AssertionError(); // the key part
    }
}
```

```java
package groovy9282;

public class DerivedSamePkg extends Base {
    @Override
    protected String x() {
        return "DerivedSamePkg";
    }
}
```

```java
package groovy9282.subpkg;

import groovy9282.Base;

public class DerivedDiffPkg extends Base {
//    @Override
    protected String x() {
        return "DerivedDiffPkg";
    }
}
```

danielsun1106:
@blackdrag Could you review the PR too, thanks in advance :-)

danielsun1106:
Hi @gesellix, The PR was merged into master, but we forgot to `add (closes #1038)` in the commit comment when merging. Could you close it manually? 

https://github.com/apache/groovy/commit/bda2edbeae61b7f7047c40724ec49a434e56e8cb

At last, thanks for your contribution!

gesellix:
Thanks a lot!

PR #1045, opened 2019-10-26
Incorrect documentation layout

albert-github:
The layout for http://docs.groovy-lang.org/latest/html/gapi/org/codehaus/groovy/vmplugin/v7/IndyInterface.html#bootstrap(Lookup,%20java.lang.String,%20java.lang.invoke.MethodType,%20java.lang.String,%20int) is incorrect due to wrong usage of a tag.


danielsun1106:
Merged. Thanks!

PR #1046, opened 2019-10-26
GROOVY-9288: Compilation error when accessing a protected/public/package-private super class field from inside a closure

danielsun1106:


eric-milles:
So is this change designed to emit a compiler error or make it work?

I'm not sure I understand the use of static compilation metadata keys in VariableExpressionTransformer.

Paul and I have been fixing several bugs in this area since 9007.  Could we have a look before you decide on a solution involving passing node metadata around; I don't think any of the previous designs have required stashing metadata.

eric-milles:
Within `VariableExpressionTransformer`, this one addition from your change stops the error.  However, I don't know the implications of setting this metadata.  In theory, a dynamic variable within a "with" closure is supposed to be satisfied from the delegate first.  This change is forcing it to be satisfied from the owner.

```java
    private static Expression tryTransformDelegateToProperty(VariableExpression expr) {
        ...

        ClassNode owner = expr.getNodeMetaData(StaticCompilationMetadataKeys.PROPERTY_OWNER);
        if (owner != null) {
            receiver.putNodeMetaData(StaticTypesMarker.INFERRED_TYPE, owner);
            receiver.putNodeMetaData(StaticTypesMarker.IMPLICIT_RECEIVER, val);
            // add
            receiver.putNodeMetaData(StaticCompilationMetadataKeys.RECEIVER_OF_DYNAMIC_PROPERTY, val);
            // end
        }

**Comment by eric-milles** [comment]
However, if I edit the example only slightly, the error is still present for "owner.protectedField", "thisObect.protectedField" and "this.protectedField".  So I don't think this is a complete solution.

```groovy
package a
abstract class Abstract {
  protected String field
}
```

```groovy
package b
class Concrete extends a.Abstract {
  void method() {
    'whatever'.with {
      println field // add "owner.", "thisObect." or "this." and get an error (on line -1 no less)
    }
  }
}
```

danielsun1106:
Thanks for your reviewing the PR. I will try to tweak it later.

danielsun1106:
@eric-milles  The PR can support your edited version of tests now.
P.S.  the PR does not change the default resolve strategy

eric-milles:
I need some time to run some experiments.  There is an interplay between `StaticTypeCheckingVisitor.checkOrMarkInnerPropertyOwnerAccess`, `VariableExpressionTransformer.tryTransformDelegateToProperty` and `StaticTypesCallSiteWriter.makeGetPropertySite` that is quite complicated.  We have made a series of changes to address some bugs and I need to go back and see what the original problems were so I can understnad the delegation metadata and property transformations.

Also, I posted links to a number of open bugs related to forbidden access errors under SC.  I think there is more going on than just protected superclass field access.

danielsun1106:
> Also, I posted links to a number of open bugs related to forbidden access errors under SC. I think there is more going on than just protected superclass field access.

@eric-milles I added some tests about accessing `public`, `package-private` and `private`(via getter), you can find them in this PR too.

eric-milles:
I've tried a few things and added some test cases.  Can you try just your new tests against the current master branch?  I chose to roll back the fix for GROOVY-7996 since it has caused this issue and a number of others.  I'd rather take a fresh look at that instead of adding one more patch on top of it's delegation data changes.

danielsun1106:
OK. I will try later today. It's 1:00 a.m. in Shanghai, I have to sleep now.

danielsun1106:
Running tests based on the latest master(last commit is: https://github.com/apache/groovy/commit/81ef58e114c83746d105dc12027341c27bc63e9f ), 7 tests failed:

image

danielsun1106:
As master reverted some commit, the PR is replaced with https://github.com/apache/groovy/pull/1048

PR #1048, opened 2019-10-28
GROOVY-9292: Compilation error when accessing a protected(via `thisObject`, `owner` and `delegate`)/package-private super class field from inside a closure

danielsun1106:


eric-milles:
Can you reframe the remaining problems as a new ticket?  The original GROOVY-9288 issue should be fixed in my estimation.

Also, you have test cases disabled with a boolean switch.  Can you use @Ignore or @NotYetImplemented to disable them?  And do you expect those test cases to work?  It was unclear if they indicate a different problem or unsolved part of the current issue.

danielsun1106:
Even if the fix for GROOVY-7996 was rolled back, protected field can not be accessed via `thisObject`, `owner` and `delegate` : https://github.com/apache/groovy/pull/1046#issuecomment-546768985
So the original description of GROOVY-9288 have to be refined too.

eric-milles:
Here is another take on your change that checks protected and package-private access for the closure's enclosing type.
```java
                if (field != null && !field.isPublic()
                        && objectExpression instanceof VariableExpression
                        && typeCheckingContext.getEnclosingClosure() != null
                        && (
                            (field.isProtected() && typeCheckingContext.getEnclosingClassNode().isDerivedFrom(current))
                             ||
                            (!field.isPrivate() && Objects.equals(typeCheckingContext.getEnclosingClassNode().getPackageName(), current.getPackageName()))
                           )
                ) {
                    switch (objectExpression.getText()) {
                    case "it":
                    case "owner":
                    case "delegate":
                    case "thisObject":
                        objectExpression.putNodeMetaData(StaticCompilationMetadataKeys.RECEIVER_OF_DYNAMIC_PROPERTY, Boolean.TRUE);
                    }
                }
```

I'm still not too keen on dynamic properties.  But at least this should stop conversion of forbidden access to missing property.  What does not match up is that explicit-this and implicit-this don't produce compiler errors, but instead result in missing property exceptions.  I'm just not sure if we know enough to proceed on a change...

danielsun1106:
@eric-milles Let's make it work, then work better, at last work perfectly ;-)

Replaced with https://github.com/apache/groovy/pull/1050

danielsun1106:
@eric-milles  BTW, I will add your tweak in PR https://github.com/apache/groovy/pull/1050 with your commit info reserved

danielsun1106:
@eric-milles After your patch applied, the following tests fails, so I will keep the original PR as it is for now:

image

danielsun1106:
@eric-milles parts of your patch is applied: https://github.com/apache/groovy/pull/1050/commits/ae0bc9f5f708d5ea1569217002a91f0236aa9e32

PR #1050, opened 2019-10-29
GROOVY-9292: Compilation error when accessing a protected super class…

danielsun1106:
… field from a closure using owner, delegate or thisObject qualifier (different package)


danielsun1106:
The PR is based on PR https://github.com/apache/groovy/pull/1046 and tweak a bit according to suggestions from @eric-milles . 
As 72 hours has gone, I will merge the PR to avoid code conflicts.

PR #1051, opened 2019-10-29
GROOVY-7996: check for get(String)/set(String,Object) or propertyMissing

eric-milles:
Here is another take on [GROOVY-7996](https://issues.apache.org/jira/browse/GROOVY-7996).  Closure delegate implements `propertyMissing` and owner has field.  Submitter expects `propertyMissing` to be respected since resolve strategy is `DELEGATE_FIRST`.


danielsun1106:
Merged. Thanks.

PR #1052, opened 2019-10-30
GROOVY-9195: SC: uniform treatment for "foo.bar" and "foo.with { bar }"

eric-milles:
refactor [GROOVY-9288](https://issues.apache.org/jira/browse/GROOVY-9288), GROOVY-9292, GROOVY-9293, et al. dynamic property

`AsmClassGenerator` was not using `TypeChooser` to determine type of object expression.

Both cases indicated in [GROOVY-9195](https://issues.apache.org/jira/browse/GROOVY-9195) now compile.  That is, "foo.bar" and "foo.with { bar }" -- where "bar" is private field -- return the field value.  If it is desirable to produce a compiler error for private field access, `StaticTypeCheckingVisitor#checkOrMarkInnerPropertyOwnerAccess` can be altered (see [GROOVY-9136](https://issues.apache.org/jira/browse/GROOVY-9136) for changes that made that case of private access possible).


danielsun1106:
+1

danielsun1106:
Merged. Thanks!

PR #1053, opened 2019-10-30
I found some code that can be refactored and a problem

mattisonzhao:
I refactored some code and i have a problem in
"StaticTypeCheckingVisitor.java" class.
i'm not sure if it's right or wrong.

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
@@ -29,6 +29,7 @@ import groovy.transform.TypeCheckingMode;
 import groovy.transform.stc.ClosureParams;
 import groovy.transform.stc.ClosureSignatureConflictResolver;
 import groovy.transform.stc.ClosureSignatureHint;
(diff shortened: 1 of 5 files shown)


src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -29,6 +29,7 @@
 import groovy.transform.stc.ClosureParams;
 import groovy.transform.stc.ClosureSignatureConflictResolver;
 import groovy.transform.stc.ClosureSignatureHint;
+import jdk.internal.org.objectweb.asm.Opcodes;
```
danielsun1106: We should use `org.objectweb.asm.Opcodes` instead of `jdk.internal.org.objectweb.asm.Opcodes`

danielsun1106:
Merged with some tweaks. Thanks.

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
mattisonzhao: oh, i see, thanks.

PR #1056, opened 2019-10-31
GROOVY-9296: [GEP] Import `TypeChecked`, `CompileStatic` and `Compile…

danielsun1106:
…Dynamic` by default


eric-milles:
It's fine if you want to discuss or experiment with changes like this, but the idea has reached nothing close to consensus.  This would be a breaking change for anyone that has a CompileDynamic, CompileStatic or TypeChecked class already defined.

danielsun1106:
http://groovy.329449.n5.nabble.com/Re-apache-groovy-GROOVY-9296-GEP-Import-TypeChecked-CompileStatic-and-Compile-1056-tp5762000p5762002.html

PR #1057, opened 2019-10-31
GROOVY-9257: ZipException, FNFE -> IOException, InvalidPathException

eric-milles:
Reworked `PackageHelperImpl.getPackageNames(URL)` to handle IOException (parent of ZipException and FileNotFoundException) and InvalidPathException.  This should improve compatibility with Java 9+ for jars within jars.

See [https://issues.apache.org/jira/browse/GROOVY-9257](GROOVY-9257) for maven project that can be used to test this.


danielsun1106:
Merged. Thanks.

PR #1059, opened 2019-11-01
GROOVY-8457: use ClassHelper.make(Class) for resolved exception type

eric-milles:
General note: Any AST transform that uses `ClassHelper.make(String)` is likely to have problems interacting with `@CompileStatic` and possibly other parts of the compiler that expect resolved types after SEMANTIC_ANALYSIS.


danielsun1106:
Merged. Thanks.

PR #1060, opened 2019-11-01
java 8 refacotor

mattisonzhao:
I found several code again that can be changed to a lambda expression.


danielsun1106:
Merged. Thanks!

PR #1063, opened 2019-11-03
GROOVY-8507, GROOVY-9301: support comma on last enum value in more cases

eric-milles:
and reduce ambiguity within (antlr2) enumConstants loop.  **This fixes an infinite loop, so I think it's worthy of merging into 2_4_X and 2_5_X.**

**Note**: This does not address the antlr4 parser. GROOVY-8507 was listed as resolved for Parrot Parser.  GROOVY-9301 is untested -- be sure to include tests for trailing comma followed by nested class, interface, trait and @interface; annotated and unannotated constructor; annotated and unannotated field; annotated and unannotated property; object initializer; static initializer.

There was an edit made after last change for GROOVY-4438 that appears to support "enum E {\n A,\n B,\n ;\n}" with no test case.  Is there a desire to have trailing comma followed by semicolon?


danielsun1106:
Merged. Thanks.

PR #1065, opened 2019-11-04
Refactor code

mattisonzhao:
Refactored some code about new java version and some performance issue.
but i'm not sure,Is the file I am modifying correct?

--Make a contribution to groovy every day.


danielsun1106:
Merged. Thanks!

PR #1068, opened 2019-11-05
Rework check for runtime retention of repeatable annotation collector

eric-milles:
- if repeatee has runtime retention, collector must as well
- if repeatable isn't resolved, evaluateExpression will handle aliasing


danielsun1106:
Merged. Thanks.

PR #1069, opened 2019-11-06
refactor code

mattisonzhao:
I refactor some code about new java version and some unnecessary
objects that I leant from effective java item6.

--Make progress every day!


danielsun1106:
Merged. Thanks!

PR #1071, opened 2019-11-06
GROOVY-8372: GrapeIvy: pass remote (not local) conf to addDependencyArtifact

eric-milles:
add dependency artifacts only if ext, type or classifier is non-default

closes #626


paulk-asert:
+1. This looks like it is doing the right thing and appears to solve the manual test case in the jar attached to this issue.

danielsun1106:
Merged. Thanks.

erwa:
Thanks for fixing this issue!

PR #1075, opened 2019-11-09
Refactor code

mattisonzhao:
I refactor some code about new java version.


mattisonzhao:
oh.i'm sorry.  the checks was failed.

mattisonzhao:
oh.i'm sorry. the inspection failed.

eric-milles:
If you need to use a break label, then the conversion from if-else to switch should probably not be done.

PR #1077, opened 2019-11-09
GROOVY-7722: prevent stack overflow resolving generics placeholders

eric-milles:
The RxJava examples in [GROOVY-7722](https://issues.apache.org/jira/browse/GROOVY-7722) and [GROOVY-7864](https://issues.apache.org/jira/browse/GROOVY-7864) may not execute as expected due to [GROOVY-8409](https://issues.apache.org/jira/browse/GROOVY-8409).  But at least they no longer infinitely recurse in the compiler/editor.


paulk-asert:
Merged. Thanks Eric.

PR #1078, opened 2019-11-10
GROOVY-8310: STC: check closure return type when method type has generic

eric-milles:
Signature for DGM.collectMany needed to be fixed.  Once STC checks closure return type compatibility (against method with return type generics), a list returned from the body of a `collectMany` will not match `Closure<Collection<? extedns T>>` but does satisfy `Closure<? extends Collection<? extedns T>>`.


danielsun1106:
LGTM

danielsun1106:
Merged. Thanks.

PR #1085, opened 2019-11-15
GROOVY-9175 migrate private methods to protected

martin-grofcik:
I would like to also extend SecureASTCustomizer in version 2.6.0

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
+++ b/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
@@ -183,53 +183,53 @@ import java.util.Map;
  */
 public class SecureASTCustomizer extends CompilationCustomizer {
 
-    private boolean isPackageAllowed = true;
-    private boolean isMethodDefinitionAllowed = true;
-    private boolean isClosuresAllowed = true;
+    protected boolean isPackageAllowed = true;
+    protected boolean isMethodDefinitionAllowed = true;
+    protected boolean isClosuresAllowed = true;
 
     // imports
-    private List<String> importsWhitelist;
-    private List<String> importsBlacklist;
+    protected List<String> importsWhitelist;
+    protected List<String> importsBlacklist;
 
     // static imports
-    private List<String> staticImportsWhitelist;
-    private List<String> staticImportsBlacklist;
+    protected List<String> staticImportsWhitelist;
+    protected List<String> staticImportsBlacklist;
 
     // star imports
-    private List<String> starImportsWhitelist;
-    private List<String> starImportsBlacklist;
+    protected List<String> starImportsWhitelist;
+    protected List<String> starImportsBlacklist;
 
     // static star imports
-    private List<String> staticStarImportsWhitelist;
-    private List<String> staticStarImportsBlacklist;
+    protected List<String> staticStarImportsWhitelist;
+    protected List<String> staticStarImportsBlacklist;
 
 
     // indirect import checks
     // if set to true, then security rules on imports will also be applied on classnodes.
     // Direct instantiation of classes without imports will therefore also fail if this option is enabled
-    private boolean isIndirectImportCheckEnabled;
+    protected boolean isIndirectImportCheckEnabled;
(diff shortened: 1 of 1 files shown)


eric-milles:
Are methods being made protected to make them callable or to make them overridable?  That is should any of the formerly private methods be made protected *and* final?

gro-mar:
I would like to change current private methods implementation too. 

If I want to extend customizer I have to access private list of checkers too. I made them protected.

eric-milles:
At this point, wouldn't it be better to export the parts you find valuable to an interface or utility class?  It is hard to understand why you need each and every field and method to be visible.

gro-mar:
"export the parts you find valuable"
Do you mean copy and paste code into my interface? In that case I have to copy all tests related to valuable code.

eric-milles:
What I mean is this: SecureASTCustomizer is a class that extends CompilationCustomizer.  If there are some select parts of SecureASTCustomizer that you wish to share, then move them into an abstract class that sits between SecureASTCustomizer and CompilationCustomizer or into an interface that SecureASTCustomizer implements.

By opening **all** the parts of SecureASTCustomizer to extension, you are completely removing the ability to enhance or refactor SecureASTCustomizer without the possibility of impacting extensions.

martin-grofcik:
"without the possibility of impacting extensions."
That's true. 

Methods are already made protected in master. Checkers are still private (without getters) so I can overwrite method without access to checker. 
I am sorry for misunderstanding, but is there any class between SecureASTCustomizer <-> CompilationCustomizer? By interface you mean CompilationUnit.PrimaryClassNodeOperation.
Thank you for your time and answers.

eric-milles:
No, there is not a class that sits between SecureASTCustomizer and CompilationCustomizer currently.  What I am suggesting is to add a class that does (instead of moving shared code up into CompilationCustomizer).  The interface idea also is suggesting creating something new.

Would you be willing to share you SecureASTCustomizer extension so that I can get a better idea of what is being shared or overridden?

martin-grofcik:
As an addition to current SecureASTCustomizer I want to add
`doubleStartImportsWhitelist/Blacklist` to allow e.g. `java.io.**` whitelist/blacklist and for static imports too.
So there will be list of doubleStarsImports and I have to overwrite `assertImportIsAllowed`++

eric-milles:
Since you only need to visit import nodes, wouldn't it be easier to create a new, standalone compilation customizer?  Is a "double star" import an enhancement to the language of yours?

gro-mar:
I have to do that in that way now. So what I did is that I took whole code of ```SecureASTCustomizer``` and I added implementation of double stars checks into assertImportIsAllowed + similar methods and all associated setters.
Because it only extends the functionality of SecureASTCustomizer.

My pull request is based on changes already merged to Groovy v3. 

re:  Is a "double star" import an enhancement to the language of yours?
No, I have to restrict imports of classes and sub-packages .... + I have to keep standard imports configurations. That's why I have to use both.

eric-milles:
Maybe I misunderstood.  Are you saying you want to blacklist or whitelist not just 1 package but an entire "tree" of packages?  Is that the significant difference from SecureASTCustomizer?  Could we just fix the existing class to support that?

martin-grofcik:
The pull request is only about allowing overwriting of methods and accessing fields.

My final goal is to support double star convention. I can implement double stars in our custom code, but if you think it make sense to propagate this functionality to the groovy code base I'll be glad to create a pull request.

eric-milles:
I would rather see an enhancement to SecureASTCustomizer that can benefit others than to see all of its members protected, which is of little or no benefit.  In my experience, asking a few questions about the motivation for changes often leads to solutions that are deeper and more widely beneficial.  Sorry if my questions seem strange; I was just looking for a way to move stale pull requests along.

Also this pull request is for the 2.6 branch, which is end of life.

martin-grofcik:
Closed because of 
https://github.com/apache/groovy/pull/1100

danielsun1106:
2_6_X branch is not maintained any more, so it's better to submit your PR for master.

PR #1087, opened 2019-11-15
GROOVY-8579, GROOVY-8930, GROOVY-8944: check for static interface method on all direct calls

eric-milles:
- calling a static interface method requires Java 1.8+ bytecode


danielsun1106:
Merged. Thanks.

PR #1088, opened 2019-11-15
GROOVY-7996: check for get(String)/set(String,Object) or propertyMissing (Groovy 2.5 backport)

eric-milles:
roll back some of the [initial changes](https://github.com/apache/groovy/commit/8ab86fe9c6dd61767d90425d103d33d741b82fe2)


paulk-asert:
Merged. Thanks. Please feel free to close this PR. I have left the Jira issue as is (reopened) since after the rollback, the initial example in the issue is breaking again even after this PR. Is that also your expectation? Perhaps we need to add that testcase in and marked as NotYetImplemented?

PR #1090, opened 2019-11-16
GROOVY-8446: syntax error for void[] or void...

eric-milles:
GROOVY-9203: anonymous inner types start at "$1" for each enclosing class

(Groovy 3+ version coming soon...)


danielsun1106:
Merged. Thanks.

PR #1093, opened 2019-11-18
GROOVY-9245: exclude synthetic constructors from CachedClass

eric-milles:
This has implications for AST transforms (like Builder) that create private methods to be called by other code that they create.  The builder InitializerStrategy makes private constructors and they won't be found by ScriptBytecodeAdapter if marked synthetic.  Maybe `@CompileStatic` can help resolve these constructors at compile time -- also, there is no `setMethodTarget` on `ConstructorCallExpression` to indicate the direct target.

If the risk is not acceptable, then different filter criteria could be used in `CachedClass` to specifically target constructors created for inner class instantiation.


danielsun1106:
Merged. Thanks.

PR #1101, opened 2019-11-22
SecureASTCustomizer double star imports extension

martin-grofcik:
adding ** convention to imports
```
    void testDoubleStarImportWhiteList() {
        def shell = new GroovyShell(configuration)
        customizer.doubleStarImportsWhitelist = ['java.util.**']
        shell.evaluate("""
            import java.util.ArrayList
            new ArrayList()
        """)

        shell.evaluate("""
            import java.util.concurrent.atomic.AtomicInteger
            new AtomicInteger(0)
        """)
        shell.evaluate("""
            import java.util.*
            import java.util.concurrent.atomic.*
            new ArrayList()
            new AtomicInteger(0)
        """)
        assert hasSecurityException {
            shell.evaluate("""
            import java.math.BigDecimal
            new BigDecimal("35")
        """)
        }
    }
```

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
+++ b/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
@@ -176,7 +176,7 @@ import java.util.Map;
  *             config.addCompilationCustomizers(imports, secure)
  *             GroovyClassLoader loader = new GroovyClassLoader(this.class.classLoader, config)
  *  </pre>
- *  
+ *
  * @since 1.8.0
  */
 public class SecureASTCustomizer extends CompilationCustomizer {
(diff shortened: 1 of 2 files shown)


eric-milles:
@martin-grofcik Is it possible to implement this feature without adding two lists?  That is, could the starImportsWhitelist support "*" and "**" syntax?

```groovy
def configuration = new CompilerConfiguration()
def customizer = new SecureASTCustomizer()
customizer.starImportsBlacklist = ['javax.**', 'jdk.internal.*']
configuration.addCompilationCustomizers(customizer)
def shell = new GroovyShell(configuration)
shell.evaluate '''
    import java.util.List // okay
    import javax.swing.Action // error
'''
```

eric-milles:
Because you are removing the trailing "." and "*"s from the double-star form and then checking with `startsWith`, you will get false matches.  Example "java.**" will be truncated to "java" and `"javax.swing.Action".startsWith("java")` will be true.

PR #1102, opened 2019-11-23
GROOVY-9316: Improved toString for AnnotationNode

paulk-asert:


danielsun1106:
+1

eric-milles:
Is there a before and after string for comparison?

paulk-asert:
Before is normal Object toString, e.g.:
`org.codehaus.groovy.ast.AnnotationNode@e8ff0`
After is:
`org.codehaus.groovy.ast.AnnotationNode@e8ff0[@Deprecated(since: 3.0.0, forRemoval: true)]`

eric-milles:
Thanks.  Could that be indicated in the issue since most people are not going to dig into the PR ?

paulk-asert:
Done. Nice suggestion.

PR #1103, opened 2019-11-23
GROOVY-9315: Bump bytecode version to 1.8

danielsun1106:


eric-milles:
How about using `CompilerConfiguration.DEFAULT.getTargetBytecode()` if there is no active configuration instead of adding another global constant?

danielsun1106:
Good idea.
I'm trying to fix the bytecode issues...

danielsun1106:
Weired...  Java9+ builds are green, but Java8 build failed...
```
groovy.ant.Groovy9105Test > testGenerateStubs FAILED
    java.lang.IllegalAccessError: tried to access method groovy.ant.AntTestCase.doInTmpDir(Lgroovy/lang/Closure;)V from class groovy.ant.AntTestCase$doInTmpDir
        at groovy.ant.AntTestCase$doInTmpDir.callCurrent(Unknown Source)
        at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCallCurrent(CallSiteArray.java:51)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:156)
        at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callCurrent(AbstractCallSite.java:168)
        at groovy.ant.Groovy9105Test.testGenerateStubs(Groovy9105Test.groovy:23)
```

danielsun1106:
@paulk-asert Please review the PR too :-)

eric-milles:
So, I tried to keep from having extra constants added to CompilerConfiguration and now there are 3.  What is the purpose of `ASM_API_VERSION`, `ASM_COMPUTE_MODE` and `ASM_PARSE_MODE`?  Who would use them (that is whay are they in CompilerConfiguration and public)?

eric-milles:
Could you please explain the logic behind taking a string literal and raising it up to a static final member even though that constant is only used within a particular low-level compiler class?  I understand that there may be old code style checks that recommend this, but for a non-shared literal, this is completely unnecessary and just makes it harder to read the compiler code.

danielsun1106:
As you can see, it is not an easy task. We should handle stackmap frames for higher versions of bytecode, or the generated bytecode can not pass the verification of JVM:
```
Expected stackmap frame at this location.
```

eric-milles:
Could you describe that in the issue?  All it says is that we require Java 8 and so we "should" use Java 8 bytecode version.

Is there a minimal set of changes just to handle the stackmap frame issue?  What I am concerned about is that the original goal of upping the bytecode version at a couple of sites has exploded into a bunch of extra files and a bunch of new global constants that, in my opinion, should be kept hidden not made public.

danielsun1106:
> What is the new CompilerConfiguration#getAsmTargetBytecode for and why is it separate from getTargetBytecode?

ASM accepts its own target bytecode version, e.g. `Opcodes.V1_8`, but `getTargetBytecode` returns something like `JDK8`. I do not want to write mapping logic everywhere...

danielsun1106:
In the old codebase, we can not manage the bytecode version well, for example, we use many bytecode versions to generate bytecode, some very old bytecode versions like `1.3` are still used.

What I propose to do is manage the default bytecode code version in a place, same for its relevant compute mode and parse mode. If we want to change them, just modify only one place. 

As for the global static variables, could you give me suggestions to hide and hold them better?

PR #1106, opened 2019-11-25
double stars for static imports

martin-grofcik:
Does it make sense to support `**` for static imports too?


danielsun1106:
@martin-grofcik, as we can see, the existing conflicts stop us from merging. How about submitting a fresh PR based on the latest codebase of master?

PR #1107, opened 2019-11-27
GROOVY-9207: fixed links in core-metaprogramming

eric-milles:
There a a bunch of errors when building the "doc" task.  But I tried before making any changes and they are there already.  I verified each of the links in the meta-programming section work after the changes.


paulk-asert:
Merged with tweaks. Thanks!

PR #1108, opened 2019-11-27
GROOVY-9174: try junit-platform-launcher as compile-time-only dependency

eric-milles:
I'm not sure of the impact of this change.  But it does manage to build successfully.  Here is a bit of reference material: https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_dependency_management_overview

Additionally or alternatively, the `groovy-all-X.Y.Z.pom` could include `groovy-test-junit5` as a runtime dependency so it is not part of the compile classpath of a project.  The "all" pom is created in the `upload.gradle` script.

### DIFF
--- a/subprojects/groovy-test-junit5/build.gradle
+++ b/subprojects/groovy-test-junit5/build.gradle
@@ -24,12 +24,13 @@ ext {
 
 dependencies {
     implementation rootProject
-    compile "org.junit.platform:junit-platform-launcher:$junit5PlatformVersion"
-    runtime "org.junit.jupiter:junit-jupiter-engine:$junit5Version"
-    testImplementation "org.junit.jupiter:junit-jupiter-params:$junit5Version"
-    testRuntime "org.junit.platform:junit-platform-engine:$junit5PlatformVersion"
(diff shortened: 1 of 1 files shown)


paulk-asert:
I think we need to think a little more about this change.

Currently, there is an expectation that downloading the standard Groovy distribution provides JUnit5 functionality "out-of-the-box" whereas after the change the required jars won't be there. Also, the current expectation is that by depending on the "groovy-test-junit5" artifact, all the required dependencies would be available on the classpath but could be overridden with a different version with appropriate excludes and/or additional dependencies.

It is possibly time to review some of these expectations especially in light of modules but it is probably best not do it in a piecewise fashion since we already made various changes for 2.5 and 3 which caused confusion for some of our users, so we should aim to minimise further disruption.

paulk-asert:
I think we can close this PR. I think we want the groovy-test-junit5 pom as is being created for 3.0.2-SNAPSHOT:
https://oss.jfrog.org/artifactory/oss-snapshot-local/org/codehaus/groovy/groovy-test-junit5/3.0.2-SNAPSHOT/
This has only core groovy as a compile dependency and all other artifacts as runtime. There is a follow-up question but I will place it into the Jira.

PR #1109, opened 2019-11-29
GROOVY-9183: use property initial value as default for map.getOrDefault

eric-milles:
if no initial value expression was supplied:
- primitive number types: zero
- primitive boolean type: false
- reference types: null

```groovy
class Pogo {
  int prop
}
@NamedVariant
def m(@NamedDelegate Pogo pogo) {
   ...
}
@Generated
def m(@NamedParam(value="prop", type=int.class) Map __namedArgs) {
  ...
  //previously: return this.m((Pogo) [prop: __namedArgs.prop])
  return this.m((Pogo) [prop: __namedArgs.getOrDefault("prop", (Integer) 0)])
}
```


paulk-asert:
+1, though I wonder if it is worth making special logic for primitive boolean - currently set to null

paulk-asert:
merged with a minor tweak - let me know if tweak looks okay to you

PR #1114, opened 2019-12-01
Update README.adoc

lidiameftah98:


paulk-asert:
Merged.  Thanks!

I think _developers' productivity_ is the correct form for a possessive plural but I tweaked the wording after the merge anyway to make it clearer. Thanks again!

PR #1117, opened 2019-12-06
GROOVY-9329: OOME raised while `parseClass(scriptText)`

danielsun1106:


eric-milles:
If the goal is to create a unique script name, what about using `java.io.File#createTempFile(String, String)` instead of processing the script text at all?

danielsun1106:
The goal is to generate unique ID for same script text. The following code will cause OOME even if the script text has not been changed:

```groovy
def gcl = new GroovyClassLoader()

while (true) {
    gcl.parseClass('def x = 1')
}
```

Error message:
```
java.lang.OutOfMemoryError: Compressed class space

	at ConsoleScript6.run(ConsoleScript6:4)

	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
```

danielsun1106:
@paulk-asert Please review the PR

eric-milles:
So if you are repeatedly parsing the same script text, is the root of the problem really here?  Is the problem that you are holding a strong reference to something that has a limited lifetime?

danielsun1106:
Actions speak louder than words.  Why not try the PR? Check whether is works or not.
Or if you have better solution, please tell us.

eric-milles:
So I looked at the original and thought the String.hashCode was the source of the OOM.  I questioned switching from one hash algorithm (hashCode) to another (MD5).  If creating a unique filename was the goal, UUID is much, much cheaper.

Now I understand that you are creating scripts in a loop.  So I provide the advice that a change in this particular bit of code may only be addressing the symptom.  I could create the same OOM by slightly altering the script text with each parse.

If this is a real world scenario, then what about checking for duplicate parsing in the calling code or providing an API where the script name is passed in so the caller can control the "uniqueness" of the name.

If you are not open to feedback, why submit for review?

danielsun1106:
In the real world, many users store their script text and parse it with GroovyClassLoader again and again.
The PR can solve the usual issues. As for the strong reference, class loaders will store classes loaded to avoid loading for better performance.
If we want to solve the issue parsing changed constantly script text, we have to remove the class cache, which will impact the performance.

P.S. I am open to feedback and appreciate suggestions. But I wish reviewers could figure out what I am doing first, then give feedback. (my 2 cents)

PR #1119, opened 2019-12-10
GROOVY-9328: add outer class accessors to all outer classes

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9328


danielsun1106:
LGTM

PR #1122, opened 2019-12-12
Mark enum next() and previous() as generated, not modifier synthetic

eric-milles:
Since the methods are useful, why not mark them available/accessible.  "values()" and "valueOf(String)" are not modifier synthetic.


danielsun1106:
LGTM

danielsun1106:
Merged. Thanks.

PR #1126, opened 2019-12-19
GROOVY-9340: return type, not type redirect

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9340


danielsun1106:
Merged. Thanks.

PR #1127, opened 2019-12-20
GROOVY-9347: extract super-interface generics information for SAM coerce

eric-milles:
`['x'].forEach(e -> e._) // forEach(Consumer<? super T>) from Iterable<T>`

`List<X>` redirects to `List<E>` extends `Collection<E>` extends `Iterable<T>`

https://issues.apache.org/jira/browse/GROOVY-9347


danielsun1106:
Merged. Thanks.

PR #1128, opened 2019-12-21
GROOVY-9351: Make `NumberRange` serializable to align with `IntRange`

danielsun1106:


eric-milles:
Does the type parameter need to indicate serializable as well?  Why is it `Comparable` and not `Number`?
```java
class NumberRange extends AbstractList<Comparable> implements Range<Comparable>, Serializable
```

danielsun1106:
> Does the type parameter need to indicate serializable as well? 

As we can see, the constructors have indicated type parameters serializable( `Number` implements `Serializable` ), e.g.
```java
    public <T extends Number & Comparable, U extends Number & Comparable>
    NumberRange(T from, U to) {
        this(from, to, null, true);
    }
```

> Why is it Comparable and not Number?

As for your 2nd question, it is out of the scope of the PR, so it's better to ask its original author...

paulk-asert:
+1, `Number` doesn't have to be `Comparable` but the `Range` abstraction assumes `Comparable` elements and there were limitations on types in certain places if you try to push through alternative designs. We can revamp for 4 if we wish.

PR #1129, opened 2019-12-21
GROOVY-9352: Static compilation fails with NoClassDefFoundError

danielsun1106:


danielsun1106:
@melix Thanks for your reviewing. I'll port the fix to 2_5_X later. It's very late for me to sleep... (5:09 a.m.)

eric-milles:
This is a lot of change with very little review.  Already merged into 3.0...  The issue is hardly 7 hours old.

Why isn't a change of this scope on a branch that has others able to pull down and try out to judge benefit vs cost.

PascalSchumacher:
`LazyConstructorNode` etc. override `hashCode`. Shouldn't `equals` also be overriden? Or am I missing something?

Thanks,
Pascal

danielsun1106:
@PascalSchumacher I will refine it later. Thanks for your reviewing :-)

PR #1130, opened 2019-12-23
GROOVY-9327: handle STC for AIC in non-STC class but STC method

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9327

`visitClass` usually processes inner classes.  However, when specific methods are annotated for STC/SC, `visitClass` is not executed (see org.codehaus.groovy.transform.StaticTypesTransformation#visit).  In this case, `visitClass(anonType)` must be driven and the metadata that `visitClass` would normally set must be provided.


paulk-asert:
+1

danielsun1106:
Merged. Thanks.

PR #1133, opened 2019-12-29
AsmClassGenerator: Remove private field genericParameterNames which i…

PascalSchumacher:
…s never accessed.


danielsun1106:
Merged. Thanks!

PR #1134, opened 2020-01-01
GROOVY-9120: check for local variable references within AIC

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9120


danielsun1106:
Merged. Thanks.

PR #1135, opened 2020-01-03
GROOVY-8298: Slow Performance Caused by Invoke Dynamic

danielsun1106:


danielsun1106:
@blackdrag 
Jochen, it would be great if you could review the PR :-)

blackdrag:
Just to be sure I understand the code correctly... We install a CachedCallableCallsite, which will have similar logic to before, meaning we do the normal method selection, then install a method as primary target and if all the guards fail (or all callsites get reset by meta class logic) we fallback to now not a new select, but to another guard that first checks with cacheHits if the cache contains the method and sets the new target in a ThreadLocal. If it exists we use fromCache, which basically gets the handle from the ThreadLocal and in case of a cache miss we do select again, but select will always add the new selections to the cache. To determine if it is a cache hit we use the class name of the receiver. Oh, and select will also (as before) set a new target for the callsite.

Assuming this is correct I do have a few problems with this:
- you should always have a micro benchmark for this sort of thing. Yes, micro benchmarks are evil, but in this case you are trying to optimize something very small, so it is going to be micro, if not nano. 
- unless changed ThreadLocal is a performance killer. It introduces a barrier which is preventing method inlining.
- a call x.foo(y) may add a handle X#foo(Y1) in the cache, but if y changes to Y2 your cache will still return X#foo(Y1), even though y may not be an instance of Y1. You will have to check the argument types. In java and static Groovy the receiver check would be enough, in Groovy (because of double dispatch) not. And of course if X changes to a class of same name but different loader, then the handle for X#foo(Y1) would be invalid as well. 

Let me suggest an alternative approach... if the end up to select the method again, then use old callsite target as fallback for the new handle. That way you create a giant method handle that is the cache, forcing the JVM to optimize the whole thing. Of course there are x more steps to improve that, but it would be a simple start. (we may want to limit the length, reorder according to the number of calls, maybe have some weakly referenced to avoid class unloading issues, and maybe other things, but all that needs a benchmark to go by and prove it is an improvement really)

danielsun1106:
Jochen, the main idea can be described as the following code:

```
if (cacheHits(receiverCassName)) { // Try to find the cached MethodHandle
      // the main logic of `fromCache`
      MethodHandle mh = lruCache.get(receiverCassName); // find the cached MethodHandle again...
      return mh.invokeExact(...);
} else { // fallback, and put the fallback methodhandle to the inline cache(LRU) for reuse
      // the main logic of `selectMethod`
      Selector selector = Selector.getSelector(...);
      selector.setCallSiteTarget();
      MethodHandle mh = selector.handle.asSpreader(...).asType(...);
      lruCache.put(receiverCassName, mh); // cache the methodhandle 
      return mh.invokeExact(...);
}
```

As the inline cache is implemented as LRU cache, and `cacheHits` and `fromCache` are executed in two steps(not atomic), so if `cacheHits` returns `true`(means cached methodhandle found), `fromCache` find the same cached methodhandle from LRU cache, the result probably is `null` as cached methodhandle may be cleared according to the rule of LRU... I use `ThreadLocal` to avoid the above concurrent issue.

Also, we have inline cache for callsite, so `setTarget` can be avoided, which causes the poor performance as the GROOVY-8298 shown.

blackdrag:
frankly I think I am missing something completely right now. You say setTarget is causing the bad performance... there is a hit involved with that yes, but in my experience creating all those MethodTypes is just as bad. Maybe that part got improved. In my opinion your implementation should be much slower now for the regular case... Especially you hitting the cache every time should cause some bad effects... I guess there is only testing.. then let us call cs the normal groovy callsite caching as we have, indy1 the current indy on master, indy2 master with your changes including the removal of the ThreadLocal. Also better would be to run multiple times and only get the best ones (using java11u28)

```Java
def m(i){1}
for (int i = 0; i < 100_000_000; i++) {
 if (m(i)!=1) throw new Error("??")
}
```
I run this up to 20 times to get the best stable time
CS: ~510ms
indy1: ~256ms
indy2: ~12680ms

which more or less shows what I expected. The effect of indy1 being faster than CS, is because the JVM actually has more options to optimize in indy and the guards here are very simple. With more parameters in the call indy would loose performance compared to CS. For indy2 I did expect it to be worse than indy1, but actually not that much. On the other hand I was already expecting form the code, that going to the cache every time is going to hurt.

```Java
def same(String obj) { return obj }
def same(int obj) { return obj }
def same(float obj) { return obj }
for (int i = 0; i < 100_000; i++) {
    [1, 1.0f, '1.0'].each { same(it) }
}
```
CS: ~64ms
indy1: ~2250ms
indy2: ~2000ms

This was the case I was really curious about. There is an improvement, but too small yet. So I did an experiment... 
```diff
@@ -883,7 +884,7 @@ public abstract class Selector {
             if (handle == null) return;
             if (!cache) return;
 
-            MethodHandle fallback = makeFallBack(callSite, sender, name, callType.ordinal(), targetType, safeNavigationOrig, thisCall, spread);
+            MethodHandle fallback = callSite.getTarget();
 
             // special guards for receiver
             if (receiver instanceof GroovyObject) {
```
which means instead of always falling back to the select method I instead reuse the existing handle in the callsite as fallback. That means, that each select call will cause a bigger handle, the newest on top, having the older as fallback. If I then have a callsite [1, 1.0f, '1.0'].each { same(it) } I will get a one handle for the string, which falls back to the float, which falls back to the int, which falls back to the select method. I did not test this for being correct or anything, and even if correct it is still having overhead it does not need, but well... let's compare...

In the first test the performance of my version and the old indy version are about the same, because the handle, that is produced is the same. Anything else would have been strange. In the second test I get 65ms, so roughly the same as CS. Sure, there may be a deopt because of setTarget, which happens in the inner loop 3 times. But after that the callsite is stable and has still 99999 iterations to run and optimize.

I was wondering why your versions is so extremely slow and I think it is because of the receiver based cache you use. Sure you cache the handle, but the key is the receiver.
```Java
for (int i = 0; i < 1_000_000; i++) {
    [1, 1.0f, '1.0'].each { it.toString() }
}
```
In this variant the receiver will change, giving your caching a chance to actually do its work and not fail with cache misses all the time

CS: ~770ms
indy1: ~14600ms (I stopped after 4 iterations)
indy2:  ~639ms
indy3:  ~215ms
(static compiled Groovy: ~110ms)

Here I can see your factor 3 you talked about earlier. And the indy3 version shows that there is even more to gain.

One last case I´d like to showcase
```Java
class Cat {
  static m(Script s){println "x"}
}
def m(){}

for (int i = 0; i < 100_000; i++) {
  use(Cat) {
     m()
  }
}
```
In this version we have an category active. This category is not even going to be used, but the effect is really bad:

CS: starts with ~180ms, ends with ~2000ms
indy1: starts with ~2500ms, ends with ~7000ms
indy2: starts with ~2300ms, ends with ~6650ms
indy3: StackOverflowError

I consider indy1 and indy2 being about equal here, but you may ask why there is a StackOverflowError for indy3. That is because at least one of the guards (I think it is the exception guard, which also degrades performance a lot) causes additional stack usage. That is already the case for the handle we produce in the selector. Older versions of Java did still show this, it seems in Java11 this is now hidden... well.. considering https://bugs.java.com/bugdatabase/view_bug.do?bug_id=8016334 this is the case for a much longer time, but with -XX:+UnlockDiagnosticVMOptions -XX:+ShowHiddenFrames you can make them visible again:
```
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1027)
	at java.base/java.lang.invoke.LambdaForm$DMH/0x00000008002a4440.invokeInterface(LambdaForm$DMH:1000029)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bf040.invoke(LambdaForm$MH:1000027)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002be840.invoke(LambdaForm$MH:1000059)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002be040.guardWithCatch(LambdaForm$MH:1000040)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bc840.invoke(LambdaForm$MH:1000113)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002b4840.invokeExact_MT(LambdaForm$MH:1000019)
	at org.codehaus.groovy.vmplugin.v7.IndyInterface.selectMethod(IndyInterface.java:232)
	at java.base/java.lang.invoke.LambdaForm$DMH/0x0000000800294840.invokeStatic(LambdaForm$DMH:1000026)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bfc40.invoke(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000094)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bd840.reinvoke(LambdaForm$MH:1000020)
	at java.base/java.lang.invoke.LambdaForm$MH/0x00000008002bdc40.guard(LambdaForm$MH:1000074)
	at java.base/java.lang.invoke.Invokers$Holder.linkToCallSite(Invokers$Holder:1000014)
	at simple5$_run_closure1.doCall(simple5.groovy:11)
```
This part of the tracer is from indy in master -  the indy3 version is really really big. What is most likely happening though is that the handle is grown for each method call. So I think it is an actual bug in the indy3 version, but I did not claim it is production ready code either ;)

danielsun1106:
Jochen, as you found, the regular case the most inefficient one. So in the latest commit I introduce a threshold to control whether to set target.
```groovy
def m(i){1}
for (int i = 0; i < 100_000_000; i++) {
 if (m(i)!=1) throw new Error("??")
}
```
(indy2: before commit 6092c76: **15s**, after commit 6092c76(threshold: 100_000): **3s**, indy1: **3s**. indy2 costs almost same time to indy1 on my machine)

Please review the PR again, thanks :-)

blackdrag:
@danielsun1106 which version of Java are you using for your tests? In my test CS was slower for the first case than indy1, in your comment they have equal speed. 

As for the category case I think this should be kept outside for now. they require special treatment, as I think even the times in CS are not acceptable.. I mean compare with case one... that is 100 million calls in 500ms and the category is 100k calls in 4000ms, interpolated than 4 million ms - that is a factor of 8000.

Measuring.... Case 1 now like indy1, as you have. Case2, with the different arguments, is now at 1290ms compared to before 2000ms. Case3, with the different receivers, is now at 540ms, so like 50% improvement. But if you check how often setTarget is called on the callsite even with a threshold of 0 you will see that there must be something wrong in the implementation.

danielsun1106:
Jochen, I'm using the JDK 11:
```
C:\Users\Daniel.LAPTOP-H7RKNSIS>java -version
openjdk version "11.0.5" 2019-10-15 LTS
OpenJDK Runtime Environment Corretto-11.0.5.10.1 (build 11.0.5+10-LTS)
OpenJDK 64-Bit Server VM Corretto-11.0.5.10.1 (build 11.0.5+10-LTS, mixed mode)
```
>But if you check how often setTarget is called on the callsite even with a threshold of 0 you will see that there must be something wrong in the implementation.

OK. I'll look into the issue later. 

Thanks for your reviewing again ;-)

danielsun1106:
@blackdrag 
Jochen, `setTarget` is executed twice when `threshold` is `0`: one for `next`(i.e. i++) and another for `m(i)`, so it's correct ;-)

```groovy
def m(i){1}
for (int i = 0; i < 100_000_000; i++) {
    if (m(i)!=1) throw new Error("??")
}
```

danielsun1106:
Here is an edge case: we set `groovy.indy.optimize.threshold` to `0`, and the target of callsite `toString` will be set because the threshold(`0`) reaches. The newly set target is not good for polymorfic calls(the subsequent calls when `i >= 10`), so when fallback count reaches threshold, we will reset to the `fromCache` target, i.e. the default target.
```groovy
for (int i = 0; i < 1_000_000; i++) {
    [1, 2.0f, '3.0'].each {
        if (1 != it && i < 10) return
        
        it.toString() 
    }
}
```
(indy2(after d799d22, `groovy.indy.optimize.threshold`: `0`): **3s**; indy2(before d799d22, `groovy.indy.optimize.threshold`: `0`): **12s**; indy1: **27s**; CS: **4s** on my another machine)

danielsun1106:
@blackdrag 
Jochen, I measured the performance of cs, indy1 and indy2 with JMH ( https://github.com/danielsun1106/benchmark-labs ), the result is shown as follows.
As we can see, the performance of indy2 is improved a lot:

Benchmark |indy2 VS indy1 | indy2 VS cs
---- | ---- | ----
CallsiteBench.dispatch_megamorphic | 3675% | 25.83%
CallsiteBench.dispatch_monomorphic | 5.72% | 188.62%
CallsiteBench.dispatch_polymorphic | 3000% | _-17.11%_

**indy2:**
```
Benchmark                            Mode  Cnt     Score    Error   Units
CallsiteBench.dispatch_megamorphic  thrpt   15   151.552 ±  2.157  ops/ms
CallsiteBench.dispatch_monomorphic  thrpt   15  4768.491 ± 41.386  ops/ms
CallsiteBench.dispatch_polymorphic  thrpt   15   155.040 ±  7.380  ops/ms
```

**indy1:**
```
Benchmark                            Mode  Cnt     Score    Error   Units
CallsiteBench.dispatch_megamorphic  thrpt   15     3.914 ±  0.462  ops/ms
CallsiteBench.dispatch_monomorphic  thrpt   15  4510.525 ± 46.702  ops/ms
CallsiteBench.dispatch_polymorphic  thrpt   15     5.283 ±  0.325  ops/ms
```

**cs:**
```
Benchmark                            Mode  Cnt     Score    Error   Units
CallsiteBench.dispatch_megamorphic  thrpt   15   119.447 ± 45.829  ops/ms
CallsiteBench.dispatch_monomorphic  thrpt   15  1651.584 ± 39.724  ops/ms
CallsiteBench.dispatch_polymorphic  thrpt   15   186.763 ± 14.549  ops/ms
```

blackdrag:
It is simple: your approach is faster than the old one without adding a lot of complicated code. Go for it.

danielsun1106:
@blackdrag
Jochen, the benchmark test code is borrowed from groovy project, I will add more test cases ;-)
As for the complicated code, if you could give me  some suggestion, it would be great.

paulk-asert:
+1

danielsun1106:
I picked the fastest result of the different version of Groovy
indy2: 4693ms; indy1: 5268ms; cs: 6007ms

**indy2 vs indy1:** 10.91% time reduced
**indy2 vs cs:** 21.87% time reduced

```groovy
        final Random random = new Random()
        final int limit = 1_000_000
        
        def factorial
        factorial = { n, acc = 1G -> 1 >= n ? acc : factorial(n - 1, acc * n ) }

        for (int i = 0; i < limit; i++) {
            factorial(30 + random.nextInt(5))
        }
```
(Copied from https://e.printstacktrace.blog/groovy-trampoline-closure-a-step-into-recursive-closures/)

PR #1136, opened 2020-01-03
GROOVY-9337: add extension methods for Optional, OptionalInt, et al.

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9337

Adds:
- `get()` to OptionalInt, OptionalLong, OptionalDouble
- `filter(Class)` to Optional
- `filter([Int|Long|Double]Predicate)` to OptionalInt, OptionalLong, OptionalDouble
- `mapToObj([Int|Long|Double]Function)` to OptionalInt, OptionalLong, OptionalDouble
- `mapToInt(ToIntFunction)` to Optional
- `mapToLong(ToLongFunction)` to Optional
- `mapToDouble(ToDoubleFunction)` to Optional
- `stream()` to Optional -- _exists as normal method since Java 9_

No `stream()` yet for OptionalInt, OptionalLong, OptionalDouble because it is unclear if return should be Stream or IntStream.  Looking into DGSMs IntStream.of(OptionalInt) and Stream.of(OptionalInt) to support both options.  Also, as with `Optional#stream`, OptionalInt, etc. have a stream() method since Java 9.

There are some additional items mentioned in GROOVY-9337; I haven't completed research for them yet.

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/PluginDefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/PluginDefaultGroovyMethods.java
@@ -21,8 +21,11 @@ package org.codehaus.groovy.vmplugin.v8;
 import groovy.lang.Closure;
 import groovy.transform.stc.ClosureParams;
 import groovy.transform.stc.FirstParam;
+import org.codehaus.groovy.runtime.NullObject;
 
 import java.util.Arrays;
+import java.util.Enumeration;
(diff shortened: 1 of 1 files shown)


danielsun1106:
+0
I'm not sure whether we should add new features to 3.0.0 at this moment as 3.0.0 GA will be out very soon.

danielsun1106:
As nobody rejects the PR in the past 72 hours, merged. Thanks.

PR #1137, opened 2020-01-10
GROOVY-8296: Groovydoc Recognition exception with Java 8 class

paulk-asert:


danielsun1106:
+1

subprojects/groovy-templates/src/main/groovy/groovy/text/GStringTemplateEngine.java
```diff
@@ -199,6 +199,7 @@ public Object run() {
             try {
                 groovyClass = loader.parseClass(new GroovyCodeSource(templateExpressions.toString(), "GStringTemplateScript" + counter.incrementAndGet() + ".groovy", "x"));
             } catch (Exception e) {
+                System.err.println("templateExpressions = " + templateExpressions);
```
PascalSchumacher: I guess this line should be reverted?

subprojects/groovy-groovydoc/src/main/java/org/apache/groovy/antlr/GroovydocVisitor.java
```diff
@@ -78,7 +86,15 @@ protected SourceUnit getSourceUnit() {
...
     @Override
     public void visitClass(ClassNode node) {
-        List<String> imports = node.getModule().getImports().stream().map(ImportNode::getClassName).collect(Collectors.toList());
+        final Map<String, String> aliases = new HashMap<>();
+        final List<String> imports = new ArrayList<>();
+        for (ImportNode iNode : node.getModule().getImports()) {
+            String name = iNode.getClassName();
+            imports.add(name);
+            if (iNode.getAlias() != null && !iNode.getAlias().isEmpty()) {
+                aliases.put(iNode.getAlias(), name.replaceAll("\\.", "/"));
```
danielsun1106: `name.replaceAll("\\.", "/"))` can be replaced with `name.replace('.', '/'))` to avoid regex compilation

subprojects/groovy-templates/src/main/groovy/groovy/text/GStringTemplateEngine.java
paulk-asert: Thanks for spotting that, yes I'll revert.

subprojects/groovy-groovydoc/src/main/java/org/apache/groovy/antlr/GroovydocVisitor.java
paulk-asert: Nice suggestion.

PR #1138, opened 2020-01-11
GROOVY-9336: integer target type for shift RHS in constant initializer

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9336


danielsun1106:
Merged. Thanks.

PR #1140, opened 2020-01-15
GROOVY-8966 Exclusive NumberRange is converted to inclusive when used by List.getAt(range)

blindpirate:
Previously, when a exclusive NumberRange constructed from non-int type (e.g. long)
is used in List.getAt(), it will be incorrectly converted to inclusive. This commit
fixes this issue by introducing a subListBorders() method to NumberRange, which is
similar to IntRange.


danielsun1106:
Merged. Thanks.

PR #1141, opened 2020-01-16
Refine some code

blindpirate:
Found some bad code when running analysis.

### DIFF
--- a/src/main/java/groovy/grape/Grape.java
+++ b/src/main/java/groovy/grape/Grape.java
@@ -37,9 +37,9 @@ public class Grape {
     private static final URI[] EMPTY_URI_ARRAY = new URI[0];
     private static final Map[] EMPTY_MAP_ARRAY = new Map[0];
 
-    private static boolean enableGrapes = Boolean.valueOf(System.getProperty("groovy.grape.enable", "true"));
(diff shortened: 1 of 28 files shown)


src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
```diff
@@ -74,7 +74,7 @@ public static ConstantExpression transformBinaryConstantExpression(final BinaryE
                 Expression left = transformInlineConstants(be.getLeftExpression(), targetType);
                 Expression right = transformInlineConstants(be.getRightExpression(), targetType);
                 if (left instanceof ConstantExpression && right instanceof ConstantExpression) {
-                    return configure(be, new ConstantExpression((String) ((ConstantExpression) left).getValue() + ((ConstantExpression) right).getValue()));
+                    return configure(be, new ConstantExpression(((ConstantExpression) left).getValue() + ((ConstantExpression) right).getValue()));
```
danielsun1106: This may cause compilation error ;-)

blindpirate: Haha, you're so fast.

danielsun1106:
Merged. Thanks.

PR #1142, opened 2020-01-16
Update gradle wrapper to 6.1

blindpirate:
Let's see if it breaks anything on CI.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,15 +45,10 @@ buildscript {
 }
 
 plugins {
-    id 'com.gradle.build-scan' version '2.4.2'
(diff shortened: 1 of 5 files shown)


paulk-asert:
Merged, thanks. I had to swap to the experimental spotbugs plugin to work with 6.
We still have an issue with codenarc. Firstly, it uses groovy.util.XmlParser which at least for 4, has now been replaced with groovy.xml.XmlParser. But from memory, and I'll try again on 3_0_X tomorrow, there was some kind of classpath issue on the gradle plugin side of things on top of that.

PR #1144, opened 2020-01-16
Use String#replace instead of String#replaceAll where possible for im…

PascalSchumacher:
…proved performance.


danielsun1106:
Merged. Thanks.

PR #1146, opened 2020-01-16
Use StringBuilder#append(char[] str, int offset, int len) to improve …

PascalSchumacher:
…performance.


danielsun1106:
Merged. Thanks.

PR #1148, opened 2020-01-20
GROOVY-9370: check for non-synthetic propertyMissing

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9370

Not sure if anon. inner class should not add GroovyObject methods, but this check should at least get Groovy 2.5/3.0 into better territory.


danielsun1106:
Merged. Thanks.

PR #1149, opened 2020-01-20
Remove need for compiler config from parser creation

eric-milles:


danielsun1106:
+1

The following changes will have to be reverted because antlr2 parser will be removed since 4.0.0. /cc @paulk-asert 
https://github.com/apache/groovy/pull/1149/files#diff-7b973e8bb30f308ec7d1427173b9abb0L36-R48

paulk-asert:
It's on the roadmap to remove antlr2 for 4 as well as the non-indy bytecode generation. There are a bunch of legacy classes, e.g. Java2Groovy and SourcePrinter which we haven't replaced yet hence why it hasn't been axed just yet.

eric-milles:
Is this roadmap available for public viewing.

paulk-asert:
We need to get better at capturing roadmap. I'll send an email to dev mailing list.

danielsun1106:
Merged. Thanks.
As for removing the old parser in Groovy 4.0.0, let's remove the related code totally.

PR #1150, opened 2020-01-20
GROOVY-9368: handle in-memory stubs system property in compilation unit

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9368

Prevent enabling joint compilation merely by setting "groovy.generate.stub.in.memory" property.  Prevent overwriting "groovy.generate.stub.in.memory" setting when compiler config is created by `FileSystemCompiler`.


danielsun1106:
Merged. Thanks.

PR #1151, opened 2020-01-23
ASM: remove push/pop for if/else blocks

eric-milles:
Calls to pushBooleanExpression() and pop() do not seem necessary.  Simplify AsmClassGenerator and StatementWriter a bit.


danielsun1106:
Merged. Thanks.

PR #1155, opened 2020-02-01
Fix GROOVY-9204

blindpirate:
### Context

https://issues.apache.org/jira/projects/GROOVY/issues/GROOVY-9204?filter=allopenissues

The concrete type information seems to be missing with `makeDeclaringAndActualGenericsTypeMap`. Use `makeDeclaringAndActualGenericsTypeMapOfExactType` instead.


danielsun1106:
Merged. Thanks.

PR #1156, opened 2020-02-01
refact some code into java 8 style

Ten000hours:
Just refactor some code into Java 8 style.

### DIFF
--- a/src/main/java/org/codehaus/groovy/antlr/Main.java
+++ b/src/main/java/org/codehaus/groovy/antlr/Main.java
@@ -47,25 +47,33 @@ class Main {
 
                 // for each directory/file specified on the command line
                 for (String arg : args) {
-                    if (arg.equals("-showtree")) {
-                        showTree = true;
-                    }
-                    //else if ( args[i].equals("-xml") ) {
-                    //    xml = true;
-                    //}
-                    else if (arg.equals("-verbose")) {
-                        verbose = true;
-                    } else if (arg.equals("-trace")) {
-                        GroovyRecognizer.tracing = true;
-                        GroovyLexer.tracing = true;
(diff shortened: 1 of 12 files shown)


danielsun1106:
> I think the errors on CI are caused by this change.

Agreed.

danielsun1106:
Could you squash the commits?

src/main/java/org/codehaus/groovy/transform/ASTTransformationCollectorCodeVisitor.java
```diff
@@ -35,13 +35,7 @@
...
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
paulk-asert: we tend to avoid star imports

src/main/java/org/codehaus/groovy/antlr/Main.java
```diff
@@ -47,25 +47,33 @@ public static void main(String[] args) {
...
+                            GroovyLexer.tracing = true;
+                            break;
+                        case "-whitespaceIncluded":
+                            whitespaceIncluded = true;
+                            break;
+                        default:
+                            doFile(new File(arg)); // parse it
+
+                            break;
                     }
```
paulk-asert: this file will likely be removed in Groovy 4, so this would be a 3 only change

paulk-asert:
Merged with minor changes (as per my comments). Thanks!

Ten000hours:
Cheers! Thanks for y'all efforts.

PR #1159, opened 2020-02-03
GROOVY-9376: add grab/grape resolver to start of chain

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9376


danielsun1106:
Merged. Thanks.

PR #1160, opened 2020-02-04
GROOVY-9384: Groovy should provide a Maven BOM

paulk-asert:


paulk-asert:
When I run this, I get an extra groovy-binary artifact created for groovy-bom due to the trick we use to generate a pom for a missing artifact. The XML manipulation that we do here must be different to the normal manipulation we do. I can probably get rid of the extra artifact with an upload filter but if anyone can work out the magic required to remove it earlier, that would be appreciated.

PR #1161, opened 2020-02-05
GROOVY-9385: add private field accessor for each mutated private field

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9385

This implements the brute-force approach: add private field accessor bridge method whenever a mutator bridge method is required.  It could be made more precise by checking for ++, +=, etc.  Or by visiting "field++" as "field = field ...".

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/sc/StaticCompilationVisitor.java
@@ -58,10 +58,13 @@ import org.codehaus.groovy.transform.stc.StaticTypeCheckingSupport;
 import org.codehaus.groovy.transform.stc.StaticTypeCheckingVisitor;
 
 import java.util.ArrayList;
+import java.util.Collections;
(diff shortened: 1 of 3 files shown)


danielsun1106:
AFAIR, we have forbidden illegal accessing invisible methods in the static Groovy, so invisible fields should not be allowed to access either even if Groovy supports the illegal accessing for now.
For backaward compatibility, we can provide a switch to enable illegal accessing.

eric-milles:
This change is not for access to private fields from outside.  An inner class can read and write private fields of the outer class(es) using bridge methods.  In the special case of GROOVY-9385, "x++" requires both a read and a write of "x".  At the moment, only the write bridge method is available.  Before some recent changes, "x++" was implemented as private field bridge method for read and dynamic property set for write.  Now, it will use private bridge methods for both read and write.

danielsun1106:
Thanks for your detailed explanation. LGTM.

danielsun1106:
Merged. Thanks!

PR #1162, opened 2020-02-08
GROOVY-9387: MissingMethodException: try one more metaClass.invokeMethod

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9387

This extra call could be made conditionally.  Or it could be called instead of "receiver.invokeMethod(name, args)".  Just not sure what to check to decide which way to go.


paulk-asert:
Merged. Thanks!

PR #1163, opened 2020-02-11
GROOVY-9392: prevent recursive invocation in CompilerConfiguration

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9392


paulk-asert:
Merged, Thanks.

PR #1164, opened 2020-02-11
GROOVY-9389: check setter type(s) against field/property/accessor type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9389

Need to check SetterInfo now that "x++" sees "x" as LHS.


paulk-asert:
Merged. Thanks!

PR #1166, opened 2020-02-12
Fix for GROOVY-9396

aalmiray:
Please review the changes to `org.apache.groovy.ast.tools.AnnotatedNodeUtils` as the check prevented the `@Generated` annotation from being added to the target method.

If accepted, this PR should be cherry-picked to 2_X and 3_X branches


danielsun1106:
Merged. Thanks!

PR #1173, opened 2020-02-23
GROOVY-9211: BUG! UNCAUGHT EXCEPTION on OpenJDK14-ea+8

paulk-asert:
Basic fixes required for 2.5.X to work on JDK14. Just one test (currently disabled) that needs further work but even that can possibly wait.

### DIFF
--- a/src/test/groovy/lang/ScriptSourcePositionInAstTest.groovy
+++ b/src/test/groovy/lang/ScriptSourcePositionInAstTest.groovy
@@ -20,6 +20,7 @@ package groovy.lang
 
 import org.codehaus.groovy.control.CompilationUnit
 import org.codehaus.groovy.control.Phases
+import org.codehaus.groovy.runtime.StringGroovyMethods
 
 /**
  * Check that scripts have proper source position in the AST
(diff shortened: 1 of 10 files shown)


paulk-asert:
Note: this doesn't try to bring across the many changes to get rid of illegal warnings etc., nor yet set target bytecode for JDK14/15, but just aims to get the VMPlugin plumbing working again.

src/test/groovy/lang/ScriptSourcePositionInAstTest.groovy
```diff
@@ -45,18 +46,24 @@ class ScriptSourcePositionInAstTest extends GroovyTestCase {
...
 
     void testDoubleStatementScript() {
-        assert positionsForScript("""\
+        def raw = """\
             println 'hello'
             println 'bye'
-        """.stripIndent()) == [[1, 1], [2, 14]]
+        """
+        // avoid stripIndent issues on JDK13+ by calling Groovy's DGM stripIndent explicitly
+        def script = StringGroovyMethods.stripIndent((CharSequence)raw)
```
blackdrag: I wonder if we should even support JDK13 here. It is not a LTS version, it will stop receiving updates after March 2020 and the stripIdent method in JDK13 is deprecated. I would prefer not changing the tests for that

paulk-asert: The @Deprecated on `stripIndent` in JDK13 is "just" how they mark preview features. The deprecation appears gone in 14 as the feature has now been more or less accepted. They still have a bit of a disclaimer in the javadoc text itself but I'd assume it is here bar a few tweaks for the foreseeable future. Unless someone else knows different.

blackdrag: if it is going to stay then we need to think about how the two stripidents differ and what we do about it. We may have to remove our stripident in general for example

paulk-asert: In Groovy 3, we added a `stripIndent(boolean)` method (GROOVY-9423). It doesn't make the problem go away but at least provides a workaround. It isn't part of GROOVY_2_5_X hence the need for the slightly hacky change here.

PR #1178, opened 2020-02-29
Added support for ScriptEngine.FILENAME behaviour

AlbertSSj:
Based on documentation :
https://docs.oracle.com/en/java/javase/11/docs/api/java.scripting/javax/script/ScriptEngine.html#FILENAME
and Nashorn behaviour, setting in the bindings a String value with ScriptEngine.FILENAME as key should affect the filename used for the script.

This patch implements the behaviour.


danielsun1106:
Merged. Thanks!

PR #1180, opened 2020-03-02
GROOVY-9424: Incorrect handling of final variables within switch

paulk-asert:
Needs more tests


paulk-asert:
Proposed PR merged

PR #1183, opened 2020-03-07
GROOVY-9447: Add a parse tree viewer to groovy console

danielsun1106:
See https://issues.apache.org/jira/browse/GROOVY-9447

code
cst


paulk-asert:
It looks and works great! +1

PR #1184, opened 2020-03-08
Update DefaultGroovyMethods.java

qq359130530:
Bug: can't use startIndex


danielsun1106:
LGTM. Thanks for your contribution.
Would you mind adding some unit tests to the PR and create a JIRA ticket to track the issue you found?

paulk-asert:
As an alternative to the unit test file, you can add an inline test in the Javadoc. E.g.
https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java#L10799-L10804

paulk-asert:
Merged. Thanks!

PR #1185, opened 2020-03-09
GROOVY-9454: STC: check for instanceof type before using the field type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9454


danielsun1106:
Merged. Thanks.

PR #1187, opened 2020-03-10
GROOVY-9455: !(x instanceof T) shouldn't propagate T as inferred type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9455

Solution is a bit brute force.  But there was no nice way to determine if BinaryExpression is enclosed by NotExpression without getting into reworking TypeCheckingContext node stacks.

Another solution considered was to transform "!(x instanceof T)" into "x !instanceof T".  That would give uniform handling to both styles of "not instanceof".  But it would not be portable to Groovy 2.5.


danielsun1106:
Merged. Thanks.

PR #1188, opened 2020-03-12
GROOVY-9462: print placeholder using getUnresolvedName()

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9462


danielsun1106:
Merged. Thanks.

PR #1189, opened 2020-03-12
Restore binary compatibility in TypeCheckingContext

eric-milles:
This should restore binary compatibility.  I suspect many of the fields in TypeCheckingContext have been added to support StaticTypeCheckingVisitor and its delegates, not to support user-defined type checking extensions.  I would suggest that TypeCheckingContext be treated more like an interface and transition all state to fully encapsulated.

If for example, I wanted to rework the node stacks so that I can tell what is nested within what or just to make all the parent nodes available, a massive amount of effort/risk is required given the current state of TypeCheckingContext.


danielsun1106:
Merged. Thanks!

PR #1190, opened 2020-03-13
gradle use groovy-all-3.0.2.jar , Unable to find method 'org.codehaus.groovy.runtime.memoize.EvictableCache.getAndPut(Ljava/lang/Object;Lorg/codehaus/groovy/runtime/memoize/MemoizeCache$ValueProvider;)Ljava/lang/Object;'.

cotools:
Unable to find method 'org.codehaus.groovy.runtime.memoize.EvictableCache.getAndPut(Ljava/lang/Object;Lorg/codehaus/groovy/runtime/memoize/MemoizeCache$ValueProvider;)Ljava/lang/Object;'.
Possible causes for this unexpected error include:
Gradle's dependency cache may be corrupt (this sometimes occurs after a network connection timeout.)
Re-download dependencies and sync project (requires network)

The state of a Gradle build process (daemon) may be corrupt. Stopping all Gradle daemons may solve this problem.
Stop Gradle build processes (requires restart)

Your project may be using a third-party plugin which is not compatible with the other plugins in the project or the version of Gradle requested by the project.

In the case of corrupt Gradle processes, you can also try closing the IDE and then killing all Java processes.


cotools:
when I use: 
implementation("org.codehaus.groovy:groovy-all:3.0.2") {force = true; transitive = false;  }
Error vanishing,
use:
implementation("org.codehaus.groovy:groovy-all:3.0.2") {force = true; transitive = true;  }
Error occurrence.

paulk-asert:
I think you should close this and rework your PR. This doesn't seem to be made between your repo and a Groovy project branch but instead between two branches of the Groovy repo. It's a little hard to tell what you are doing at this stage.

paulk-asert:
If I create a simple little hello world build, then this works:
`implementation 'org.codehaus.groovy:groovy-all:3.0.2'`
or this:
`implementation("org.codehaus.groovy:groovy-all:3.0.2") {force = true; transitive = true; }`
but this won't work since it won't have a Groovy jar on the classpath unless you have another Groovy dependency somewhere:
`implementation("org.codehaus.groovy:groovy-all:3.0.2") {force = true; transitive = false; }`

paulk-asert:
I've closed for now since the PR was malformed. Feel free to open a fixed PR if you still believe there is a bug in Groovy.

PR #1192, opened 2020-03-14
GROOVY-9468: Support running as Java code in the Groovy console

danielsun1106:
See https://issues.apache.org/jira/browse/GROOVY-9468

Run the same code twice: 
```java
public class HelloWorld {
        System.out.println("Hello, world!");
    }
}
```
* `Ctrl + Alt + R` to run as Java for the first time
* `Ctrl + R` to run as Groovy for the second time
Run Java Code

### DIFF
--- a/subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
+++ b/subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
@@ -18,7 +18,7 @@
  */
 package groovy.console.ui
 
-
+import com.github.javaparser.ParseProblemException
 import com.github.javaparser.StaticJavaParser
 import com.github.javaparser.ast.CompilationUnit
 import com.github.javaparser.ast.Modifier
@@ -919,15 +919,8 @@ class Console implements CaretListener, HyperlinkListener, ComponentListener, Fo
                     int errorLine = se.line
                     String message = se.originalMessage
 
-                    String scriptFileName = scriptFile?.name ?: DEFAULT_SCRIPT_NAME_START
-
                     def doc = outputArea.styledDocument
-
-                    def style = hyperlinkStyle
-                    def hrefAttr = new SimpleAttributeSet()
-                    // don't pass a GString as it won't be coerced to String as addAttribute takes an Object
-                    hrefAttr.addAttribute(HTML.Attribute.HREF, 'file://' + scriptFileName + ':' + errorLine)
-                    style.addAttribute(HTML.Tag.A, hrefAttr)
+                    Style style = createLinkStyle(errorLine)
 
                     insertString(doc, doc.length, message + ' at ', stacktraceStyle)
                     insertString(doc, doc.length, "line: ${se.line}, column: ${se.startColumn}\n\n", style)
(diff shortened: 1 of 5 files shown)


subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
```diff
@@ -1136,13 +1142,25 @@ class Console implements CaretListener, HyperlinkListener, ComponentListener, Fo
...
+    private enum SourceType {
+        GROOVY, JAVA
+    }
+
+    void runJava(EventObject evt = null) {
+        doRun(evt, SourceType.JAVA)
+    }
+
     // actually run the script
     void runScript(EventObject evt = null) {
```
paulk-asert: Would it be better to use default params here?
`void runScript(EventObject evt = null, SourceType st = SourceType.GROOVY) {`

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
```diff
@@ -1180,11 +1198,20 @@ class Console implements CaretListener, HyperlinkListener, ComponentListener, Fo
         cc.getOptimizationOptions().get(CompilerConfiguration.INVOKEDYNAMIC)
     }
 
+    void runSelectedJava(EventObject evt = null) {
+        doRunSelected(evt, SourceType.JAVA)
+    }
+
     void runSelectedScript(EventObject evt = null) {
```
paulk-asert: ditto here?

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
```diff
@@ -1329,6 +1356,47 @@ class Console implements CaretListener, HyperlinkListener, ComponentListener, Fo
...
+    @CompileStatic
+    private Object doRun(boolean selected, SourceType st, HistoryRecord record) {
+        def result
+        def src = record.getTextToRun(selected)
+        if (SourceType.JAVA === st) {
+            Optional<String> optionalPrimaryClassName = findPrimaryClassName(src)
+            if (optionalPrimaryClassName.isPresent()) {
+                def js = new JavaShell(Thread.currentThread().contextClassLoader)
+                js.run(optionalPrimaryClassName.get(), src)
+            }
```
paulk-asert: I wonder if there should be an else here with a message/error to output area saying "No public class found, nothing to run." or similar.

paulk-asert:
+1, I'd also like to see "Compile to Java" added but that can be a separate issue.

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
paulk-asert: Otherwise, change `public class` to `class` and it looks like it ran but with no output. Whereas, it passes parser but hasn't been compiled or run.

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
danielsun1106: I am not sure whether swingbuilder knows such method signature or not...

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
danielsun1106: It's better to check, but Java compiler/javaparser will complain finally ;-)

danielsun1106:
Thanks for your reviewing ;-)

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
paulk-asert: Groovy will create all the in between signatures, so it should work - we can try later.

subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy
paulk-asert: No, it fails silently. Not the most catastrophic behavior but not as intuitive as possible. We can change later.

PR #1193, opened 2020-03-15
GROOVY-9469: Preserve exception in global AST transformation creation

szpak:
InvocationTargetException is not very meaningful. It worked in Groovy 2.5.

https://issues.apache.org/jira/browse/GROOVY-9469

Related issue in Spock: https://github.com/spockframework/spock/issues/1067


danielsun1106:
LGTM.
@paulk-asert ping ;-)

paulk-asert:
Proposed fix seems okay, but I was first planning to do a `git bisect` to see if I can find a commit between 2.4 and 2.5 that was the cause.

paulk-asert:
I've looked back through the history and the PR looks like a good way to progress.

paulk-asert:
Merged. Thanks!

PR #1194, opened 2020-03-16
GROOVY-9468: Support running as Java code in the Groovy console (refi…

paulk-asert:
…nements for better error messages)


danielsun1106:
+1

PR #1199, opened 2020-03-21
Avoid unnecessary boxing and unbox of primitive values.

PascalSchumacher:
This improves readability because it makes it clear that the value can never be null.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
+++ b/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
@@ -299,12 +299,12 @@ public class DefaultTypeTransformation {
                 return n.floatValue();
             }
             if (type == Double.class) {
-                Double answer = n.doubleValue();
+                double answer = n.doubleValue();
                 //throw a runtime exception if conversion would be out-of-range for the type.
                 if (!(n instanceof Double) && (answer == Double.NEGATIVE_INFINITY
                         || answer == Double.POSITIVE_INFINITY)) {
                     throw new GroovyRuntimeException("Automatic coercion of " + n.getClass().getName()
-                            + " value " + n + " to double failed.  Value is out of range.");
+                            + " value " + n + " to double failed. Value is out of range.");
                 }
                 return answer;
             }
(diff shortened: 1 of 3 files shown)


danielsun1106:
Merged. Thanks.

src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
```diff
@@ -299,12 +299,12 @@ private static Object continueCastOnNumber(Object object, Class type) {
                 return n.floatValue();
             }
             if (type == Double.class) {
-                Double answer = n.doubleValue();
+                double answer = n.doubleValue();
```
blackdrag: Just want to mention, that you do not really save on boxing here since the return will box it then later again

danielsun1106: Yep...

PascalSchumacher: The `INFINITY` comparisons can be done without boxing/unboxing.

blackdrag: @PascalSchumacher that is actually a very good point. 

I was remembering I did break up some of the conversions for Indy, but what I see in TypeTransformers actually makes me wonder if the implementation is "correct" there. Not that it is converting wrong, but the exception following just a few lines lower is missing there. I doubt the callsite caching logic does it. On the other hand the implementation there is not for type casting, but for coercing type parameters in a method call. But then again, why should a simple coerce not have the check for infinity with that GroovyRuntimeException? 

In the end this whole castToType method should be replaces with Indy logic. Even if Indy is slower (which it is not anymore afaik) this method is a blocker for inlining in a big way and we call it often. With indy we have the chance to make a very lean version.

paulk-asert: I'll need to debug further. I tried a simple test script:
```
def bits = 0x7ff0000000000000L - 1 // almost +ive infinity
def pi = Double.longBitsToDouble(bits)
pibd = new BigDecimal(pi) + 1E292 // should now be approx Double.POSITIVE_INFINITY
println pibd // 179...
println pibd as Double // Infinity
println org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation.castToType(pibd, Double) // GRE Out of range
```
It behaved the same regardless of whether `answer` had Wrapper or primitive type.

PR #1200, opened 2020-03-21
 ConversionHandler: Remove unnecessary null check for value of final field which is always initialized.

PascalSchumacher:


src/main/java/org/codehaus/groovy/control/CompilerConfiguration.java
```diff
@@ -441,7 +441,7 @@ public CompilerConfiguration() {
     private void handleOptimizationOption(final Map<String, Boolean> options, final String optionName, final String sysOptionName) {
         String propValue = getSystemPropertySafe(sysOptionName);
         boolean optionEnabled = propValue == null
-                ? (DEFAULT != null && Boolean.TRUE.equals(DEFAULT.getOptimizationOptions().get(optionName)))
+                ? Boolean.TRUE.equals(DEFAULT.getOptimizationOptions().get(optionName))
```
eric-milles: DEFAULT is null while the constructor in its initializer expression is running.

danielsun1106:
Merged. Thanks.

PR #1202, opened 2020-03-22
Refactor of some groovy methods

DarioArena87:
Hi, based on the fact that java.util.List extends java.util.Collection which in turn extends java.lang.Iterable my suggestion is to remove from DefaultgroovyMethods in Groovy 3.0.X all the deprecated methods that takes a list or a collection as 'self' parameter and were deprecated in favor of the same method which have an Iterable as parameter instead. I think this would greatly simplify having support from other tools (IDEs autocompletion, Static code analyzers...) and simplifies java compiler work making it possible to write other methods in a simpler way. 
I also made some of the default groovy methods more type-safe using java generics in method declaration.
Since groovy 3.X targets java8+ jvm i also wrote some methods for the java.util.Stream class and some to obtain a stream based on an Iterable or an Iterator (that i found on groovy GDK documentation but i cannot find them in the source code, maybe they are redundant). Because i was not sure if this is an intended groovy feature i left theese last methods private but you can make them public if they should show up on iterables and iterators.
I refactored also some other methods to leverage stream api in their implementation and make them simpler. I dont't think that theese are breaking changes because i ran all the tests and they all passes, if you like this commit maybe it can be merged on official source base.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,6 @@
+#Sun Mar 22 12:55:51 CET 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-rc-4-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-rc-4-bin.zip
-zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
--- a/src/main/java/groovy/util/GroovyCollections.java
+++ b/src/main/java/groovy/util/GroovyCollections.java
@@ -18,16 +18,19 @@
  */
 package groovy.util;
 
+import org.codehaus.groovy.runtime.ArrayUtil;
 import org.codehaus.groovy.runtime.DefaultGroovyMethods;
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
 import org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singleton;
+import static java.util.stream.Collectors.toCollection;
+import static java.util.stream.Collectors.toList;
 
 /**
  * A Collections utility class
@@ -94,22 +97,20 @@ public class GroovyCollections {
      * @return a List of the combinations found
      * @since 2.2.0
      */
-    public static List combinations(Iterable collections) {
-        List collectedCombos = new ArrayList();
-        for (Object collection : collections) {
-            Iterable items = DefaultTypeTransformation.asCollection(collection);
+    public static <T> List<List<T>> combinations(Iterable<T> collections) {
+        List<List<T>> collectedCombos = emptyList();
+        for (T collection : collections) {
+            Collection<T> items = DefaultTypeTransformation.asCollection(collection);
             if (collectedCombos.isEmpty()) {
-                for (Object item : items) {
-                    List l = new ArrayList();
-                    l.add(item);
-                    collectedCombos.add(l);
-                }
+                collectedCombos = StreamSupport.stream(items.spliterator(), false)
+                                               .map(GroovyCollections::newArrayList)
+                                               .collect(toCollection(ArrayList::new));
             } else {
-                List savedCombos = new ArrayList(collectedCombos);
-                List newCombos = new ArrayList();
-                for (Object value : items) {
-                    for (Object savedCombo : savedCombos) {
-                        List oldList = new ArrayList((List) savedCombo);
+                List<List<T>> savedCombos = new ArrayList<>(collectedCombos);
+                List<List<T>> newCombos = new ArrayList<>(items.size() * savedCombos.size());
+                for (T value : items) {
+                    for (List<T> savedCombo : savedCombos) {
+                        List<T> oldList = new ArrayList<>(savedCombo);
                         oldList.add(value);
                         newCombos.add(oldList);
                     }
@@ -121,6 +122,15 @@ public class GroovyCollections {
                 break;
         }
         return collectedCombos;
+    }
+
+    private static <T> List<T> newArrayList(T... elements) {
+        if (elements == null || elements.length == 0) {
+            return emptyList();
+        }
+        List<T> ret = new ArrayList<>(elements.length);
+        ret.addAll(Arrays.asList(elements));
+        return ret;
     }
 
     public static <T> List<List<T>> inits(Iterable<T> collections) {
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -139,6 +139,8 @@ import java.util.Random;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.SortedSet;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.Stack;
 import java.util.Timer;
 import java.util.TimerTask;
@@ -146,6 +148,9 @@ import java.util.TreeMap;
 import java.util.TreeSet;
 import java.util.concurrent.BlockingQueue;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 import static groovy.lang.groovydoc.Groovydoc.EMPTY_GROOVYDOC;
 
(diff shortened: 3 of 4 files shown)


src/main/java/groovy/util/GroovyCollections.java
```diff
@@ -18,16 +18,19 @@
...
 import org.codehaus.groovy.runtime.ScriptBytecodeAdapter;
 import org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation;
 
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
danielsun1106: We prefer importing qualified classes and avoid using `*`

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
+#Sun Mar 22 12:55:51 CET 2020
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-rc-4-all.zip
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-rc-4-bin.zip
```
danielsun1106: We use `gradle -b wrapper.gradle wrapper` to generate gradle wrapper, which results in `gradle-xxxx-bin.zip`, so it's better to align with the existing rule ;-)

src/main/java/groovy/util/GroovyCollections.java
```diff
@@ -123,6 +124,15 @@ public static List combinations(Iterable collections) {
         return collectedCombos;
     }
 
+    private static <T> List<T> newArrayList(T... elements) {
+        if (elements == null || elements.length == 0) {
+            return emptyList();
+        }
+        List<T> ret = new ArrayList<>(elements.length);
+        ret.addAll(Arrays.asList(elements));
```
danielsun1106: Maybe the following code is a bit more efficient ;-)
```java
Arrays.stream(elements).collect(Collectors.toList())
```

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -2996,15 +2998,6 @@ public static Number count(Iterator self, Object value) {
         return answer;
     }
 
-    /**
-     * @deprecated use count(Iterable, Closure)
-     * @since 1.0
-     */
-    @Deprecated
-    public static Number count(Collection self, Object value) {
```
danielsun1106: We could remove the method in 4.0.0, but it's better to remain it in the released 3.0.x for binary compatibility

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3022,15 +3015,6 @@ public static Number count(Iterable self, Object value) {
         return count(self.iterator(), value);
     }
 
-    /**
-     * @deprecated use count(Iterable, Closure)
-     * @since 1.8.0
-     */
-    @Deprecated
-    public static Number count(Collection self, Closure closure) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3210,18 +3194,6 @@ public static Number count(byte[] self, Object value) {
         return count(InvokerHelper.asIterator(self), value);
     }
 
-    /**
-     * @deprecated Use the Iterable version of toList instead
-     * @see #toList(Iterable)
-     * @since 1.0
-     */
-    @Deprecated
-    public static <T> List<T> toList(Collection<T> self) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3299,16 +3272,6 @@ public static Number count(byte[] self, Object value) {
         return collate((Iterable)Arrays.asList(self), size, true);
     }
 
-    /**
-     * @deprecated use the Iterable variant instead
-     * @see #collate(Iterable, int)
-     * @since 1.8.6
-     */
-    @Deprecated
-    public static <T> List<List<T>> collate( List<T> self, int size ) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3341,16 +3304,6 @@ public static Number count(byte[] self, Object value) {
         return collate((Iterable)Arrays.asList(self), size, step, true);
     }
 
-    /**
-     * @deprecated use the Iterable variant instead
-     * @see #collate(Iterable, int, int)
-     * @since 1.8.6
-     */
-    @Deprecated
-    public static <T> List<List<T>> collate( List<T> self, int size, int step ) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3383,16 +3336,6 @@ public static Number count(byte[] self, Object value) {
         return collate((Iterable)Arrays.asList(self), size, size, keepRemainder);
     }
 
-    /**
-     * @deprecated use the Iterable variant instead
-     * @see #collate(Iterable, int, boolean)
-     * @since 1.8.6
-     */
-    @Deprecated
-    public static <T> List<List<T>> collate( List<T> self, int size, boolean keepRemainder ) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3448,16 +3391,6 @@ public static Number count(byte[] self, Object value) {
         return collate((Iterable)Arrays.asList(self), size, step, keepRemainder);
     }
 
-    /**
-     * @deprecated use the Iterable variant instead
-     * @see #collate(Iterable, int, int, boolean)
-     * @since 1.8.6
-     */
-    @Deprecated
-    public static <T> List<List<T>> collate( List<T> self, int size, int step, boolean keepRemainder ) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -3566,60 +3499,6 @@ public static Collection collect(Object self) {
...
-     * <pre class="groovyTestCase">assert [1,2,3] == [1,2,3].collect()</pre>
-     *
-     * @param self a collection
-     * @return a List of the transformed values
-     * @see Closure#IDENTITY
-     * @since 1.8.5
-     * @deprecated use the Iterable version instead
-     */
-    @Deprecated
-    public static <T> List<T> collect(Collection<T> self) {
```
danielsun1106: Ditto

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -6644,6 +6286,18 @@ public static Object sum(Collection self, Object initialValue, Closure closure)
...
+    /**
+     * Sums all the items in a stream. This is equivalent to invoking the
+     * "plus" method on all items in the stream.
+     * <pre class="groovyTestCase">assert 1+2+3+4 == [1,2,3,4].stream().sum()</pre>
+     *
+     * @param stream The stream containing the values to add together
+     * @param <T> The type of the values in the stream
+     * @return The sum of all of the items
+     */
+    public static <T> T sum(Stream<T> stream) {
```
danielsun1106: It's better to create JIRA ticket to track the improvement

src/main/java/groovy/util/GroovyCollections.java
DarioArena87: I agree, i usually prefer it too but this it was my mistake because i had not adjusted the settings in my ide and it dit the * automatically. Feel free to CTRL + O to adjust the imports

src/main/java/groovy/util/GroovyCollections.java
DarioArena87: I agree, 100X times better!

gradle/wrapper/gradle-wrapper.properties
DarioArena87: Sorry, feel free to remove it!

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
DarioArena87: Sorry this was my first pull request to the groovy repository, can you explain to me how to do it?

paulk-asert:
I merged the Diamond operator changes into master and 3_0_X, and the removal of deprecated methods into master. Hopefully that will make it easier to rework any potential other bits. Thanks.

paulk-asert:
Note: Any deprecated methods in DGM, aren't added to the respective metaClasses, so tools that are smart enough should be avoiding them. They are kept for binary compatibility since some frameworks call those methods directly rather than through the Groovy runtime.

paulk-asert:
To check for binary compatibility see the guidelines (Binary Compatibility section) at:
https://groovy.apache.org/guidelines.html

paulk-asert:
A comment about refactoring to streams. We are in general trying to improve stream support and make use of streams where it makes sense. We don't necessarily want to use them everywhere though. See Brian Goetz's answer to this SO question about when to use streams vs collections where he says "it depends":
https://stackoverflow.com/questions/24676877/should-i-return-a-collection-or-a-stream
(Basically, if you have a chain of DGM methods returning collections and have small data sizes, the overheads of creating streams and back to collections for each step is likely to be high. We'd instead encourage to stay using streams for all steps or not use streams at all in some cases. Our general direction is to add DGM methods that help with that. So we'll end up with iterator-based variants and stream-based variants for most functionality.)

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
danielsun1106: Here is the link to JIRA ;-)

PR #1205, opened 2020-03-26
GROOVY-9480: New v9 ClassFinder throws NoSuchFileException/Unsupporte…

danielsun1106:
…dOperationException and writes it to stderr

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v9/ClassFindFailedException.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v9/ClassFindFailedException.java
@@ -1,0 +1,29 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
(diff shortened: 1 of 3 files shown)


paulk-asert:
+1

eric-milles:
Not sure I agree with this change.  I filed the bug because the exception was coming for 1) trying to close a file system handle that does not support it and 2) tyring to find "groovy" in the root of a project that does not have such a directory.

paulk-asert:
@eric-milles Do you have a reproducer for those cases?

eric-milles:
In my environment, `org.codehaus.groovy.vmplugin.v9.Java9#getDefaultImportClasses` is called and its request `DefaultGroovyMethods.getLocation(gcl.loadClass("groovy.grape.GrapeIvy")).toURI()` returns:
```
file:/C:/Users/.../org.codehaus.groovy30/
```

`getDefaultImportClasses` makes the assumption that the URI returned will then have a "groovy" package under it.  And that's just not how the class loader that provides a GrapeIvy class works in my environment.

This is a different flavor of the existing note "in production environment, groovy-core classes, e.g. `GroovySystem`(java class) and `GrapeIvy`(groovy class) are all packaged in the groovy-core jar file, but in Groovy development environment, groovy-core classes are distributed in different directories".

Also, you cannot assume that in production they are sourced from the same archive since you could use the patch modules feature of JPMS or OSGi to add classes from another archive.

PR #1206, opened 2020-03-26
Added DOCTYPE tag and "lang" attribute

armeris:
Fix for https://sonarcloud.io/project/issues?id=apache_groovy&open=AXD-_jqMDC9Fk2pZiWTt&resolved=false&types=BUG

Added missing tag and attribute in package.html

### DIFF
--- a/src/main/java/groovy/inspect/package.html
+++ b/src/main/java/groovy/inspect/package.html
@@ -18,7 +18,8 @@
      under the License.
 
 -->
-<html>
+<!DOCTYPE html>
+<html lang="en">
   <head>
     <title>package groovy.inspect.*</title>
   </head>


paulk-asert:
The suggested change looks good. I am just wondering whether the expectations of the javadoc tool have changed. The package description doesn't seem to appear in the javadoc at the moment:
https://docs.groovy-lang.org/latest/html/api/groovy/inspect/package-summary.html
(it is appearing in the groovydoc)

armeris:
Um I wasn't aware of that. Maybe if this info is not going to appear anymore this file should be removed or if it is needed it should be added to the javadoc again.
In anycase these changes goes beyond the objective of this bugix and a issue should be reported with this.
Talking about the PR itself, I don't know why travis is failing, it seems that is because this error:
`java.lang.IllegalStateException: You're not authorized to run analysis. Please contact the project administrator.`
Is a special permission or anything needed to run travis?

paulk-asert:
We'd like the information to appear. We'll need to track down why it isn't. It appears here:
https://docs.groovy-lang.org/3.0.0/html/api/groovy/inspect/package-summary.html
but not here (or 3.0.2):
https://docs.groovy-lang.org/3.0.1/html/api/groovy/inspect/package-summary.html
I'll check if we have made any changes (I don't remember any), otherwise we'll need to look for any gradle or java changes in that area.

armeris:
I just have build the doc in the master branch and it is appearing both in groovydoc and in javadoc. Maybe a specific bug of 3.0.x version?

paulk-asert:
I am thinking different JDKs. Which version did you use?

armeris:
I used JDK 1.8 which is a bit old, the issue could be there. I will try with a newer version to see if it is working properly or not.

danielsun1106:
The package description issue is not related to the PR.
Merged.  Thanks for your contribution!

paulk-asert:
Agreed. No harm in merging. I was just holding off in case the preferred way to fix was to remove package.html and replace with package-info.java.

paulk-asert:
FYI, the issue with package.html has been fixed. From JDK9+, javadoc requires sourcepath to be set or the base source directory containing package.html files to be on the classpath.

PR #1207, opened 2020-03-26
GROOVY-9481: Initialize meta class when checkInitalised returns false

danielsun1106:


paulk-asert:
I'd be keen to investigate this a bit more. This seems to me to be patching something like a race condition rather than fixing the race condition.

danielsun1106:
Agreed.

PR #1215, opened 2020-03-30
GROOVY-9452: Multiple Repeatable Annotations Container generated if a…

paulk-asert:
…n explicit container is present and an implicit one required


danielsun1106:
+1

danielsun1106:
Merged. Thanks.

PR #1218, opened 2020-04-10
Switch to faster String#indexOf/String#lastIndexOf methods taking a c…

PascalSchumacher:
…har parameter.


danielsun1106:
Merged. Thanks!

PR #1219, opened 2020-04-11
Switch to faster String#indexOf method taking a char parameter.

PascalSchumacher:
Sorry, my last pull request missed two places. :(


danielsun1106:
Merged. Thanks again ;-)

PR #1223, opened 2020-04-16
remove antlr2

paulk-asert:


danielsun1106:
+1

paulk-asert:
Currently this removes Java2Groovy, MindmapPrinter and SourcePrinter. If we need these I would suggest using an approach similar to what was done for Groovydoc to re-introduce similar tools and probably (at least for Java2groovy) in a separate optional module, perhaps `groovy-tools`.

danielsun1106:
Merged. Thanks!

PR #1224, opened 2020-04-16
GROOVY-9512: Control the fallback threshold of indy with different sw…

danielsun1106:
…itch


paulk-asert:
Looks okay. I do wonder when we go to "indy only" whether we should rename switches to not have indy in the name since that won't make a lot of sense any more. But perhaps we can look at that later.

danielsun1106:
After we enable the indy by default, we can remove `indy` in the property names.

PR #1225, opened 2020-04-18
GROOVY-9513: Enable indy by default

danielsun1106:


danielsun1106:
Replaced with https://github.com/apache/groovy/pull/1226

PR #1226, opened 2020-04-19
GROOVY-9513: Enable indy by default

danielsun1106:


paulk-asert:
merged, nice work - there are a few things I'd like to tweak but we can do that next

danielsun1106:
Thanks a lot for your reviewing and tweaking the PR, Paul :-)

PR #1230, opened 2020-04-21
GROOVY-9518: STC: type check constructor arguments like method call args

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9518


danielsun1106:
+1

danielsun1106:
Merged. Thanks.

PR #1231, opened 2020-04-22
GROOVY-9517: use inferred right expression type for array checks

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9517


danielsun1106:
Merged. Thanks.

PR #1234, opened 2020-04-26
enable Log4J tests bye ensuring indy is really enabled by default. 

blackdrag:
Warning these tests will still fail in a mixed indy/old callsite mode.


paulk-asert:
I merged because I thought it passed locally but I must have been mistaken. It just broke the build quite badly - so I have reverted for now.

blackdrag:
yes, I was wondering why you merged it... but the failures look like they are exposing problems with indy and related to method selection

PR #1235, opened 2020-04-26
GROOVY-9501: enable access to static, non-final, subclassed-outer fields

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9501


danielsun1106:
Merged. Thanks!

PR #1236, opened 2020-04-27
[GROOVY-9308] Added "final var" shortcut keywords "let" and "val"

mojo2012:
I implemented bot `let` and `val` in this repo: https://github.com/mojo2012/groovy
I also added tests (similar to the `var` and `def` keyword. But unfortunately I didn't manage to make `final val` or `final let` invalid. 
Maybe someone of you can point me in the right direction?

Anyway, I don't know if including both keywords is a good idea, but as both kotlin and scala went with the `val` keyword, I guess it's best to go for this as well.

Let me know if you are interested in this feature.

### DIFF
--- a/src/antlr/GroovyLexer.g4
+++ b/src/antlr/GroovyLexer.g4
@@ -366,6 +366,8 @@ fragment DollarSlashyStringCharacter
 // Groovy keywords
 AS              : 'as';
 DEF             : 'def';
+VAL             : 'val';
+LET             : 'let';
 IN              : 'in';
 TRAIT           : 'trait';
 THREADSAFE      : 'threadsafe'; // reserved keyword
--- a/src/antlr/GroovyParser.g4
+++ b/src/antlr/GroovyParser.g4
@@ -131,6 +131,8 @@ modifier
           |   VOLATILE
           |   DEF
           |   VAR
+          |   VAL
+          |   LET
           )
     ;
 
@@ -170,6 +172,8 @@ variableModifier
     |   m=( FINAL
           | DEF
           | VAR
+          | VAL
+          | LET
           // Groovy supports declaring local variables as instance/class fields,
           // e.g. import groovy.transform.*; @Field static List awe = [1, 2, 3]
           // e.g. import groovy.transform.*; def a = { @Field public List awe = [1, 2, 3] }
@@ -1142,6 +1146,8 @@ identifier
     :   Identifier
     |   CapitalizedIdentifier
     |   VAR
+    |   VAL
+    |   LET
     |   IN
 //    |   DEF
     |   TRAIT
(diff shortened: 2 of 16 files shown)


paulk-asert:
Relates to GROOVY-9308

danielsun1106:
Out of curiosity, why do we need two similar new keywords `val` and `let`?

mojo2012:
When initializing a lot of variables it is way more readable to have them properly aligned like this:

```groovy
var testObject = ...
val webClient = ...
var nextObject = ...
```

instead of:
```
groovy
var testObject = ...
final var webClient = ...
var nextObject = ...
```

I remember mentioning Brian Goetz that as one of the main reasons that make `var` so nice to use.

Obviously none uses final variables these days anymore, or why did the java team not think about that :-D

danielsun1106:
Understood. I mean we can choose one of the two new keywords, `val` or `let`.

mojo2012:
I personally prefer val, but some people might argue that let is easier to distinguish  from var. so I‘m fine with let too here.

Too finish the feature though I would need some help of you guys to figure out how to make ‚final let‘ a syntax error.

danielsun1106:
Also, a thread about the new feature should be started at dev mailing list:
http://groovy.329449.n5.nabble.com/Groovy-Dev-f372993.html

When we get three +1 from PMC, the feature can be included in Groovy. This is the Apache rule, we should obey it.

mojo2012:
I created a post on the mailing list. Let's wait for the feedback :-)

paulk-asert:
Seems like you aren't subscribed to the dev mailing list, so your email hit moderation - it helps us avoid lots of spam. I passed it through but if you subscribe, you will see your messages turn up quicker. Also, while email is preferred for official discussions, there is also the "feature-ideas" slack channel if you just want to brain storm ideas. If we do go ahead, we'll make this an official GEP (Groovy enhancement proposal). We can help with getting that done.

mojo2012:
I subscribed through nabble, didn't that work? Can you also point me into the direction of that slack channel? I registered here a while ago: https://groovycommunity.com/ 
But didn't actually get any notification form there.

Yes help is very much appreciated :-) I'd love to help to make groovy even better.

paulk-asert:
Also, you might want to edit the PR title to say GROOVY-9308 instead of GROOVY-9364

paulk-asert:
Just here as far as I know: https://groovy-community.slack.com/

mojo2012:
It says a workspace admin has to invite me (meister dot fuchs AT gmail com)

mojo2012:
I think the feature request has been denied, so we can close this PR

paulk-asert:
I think the use of "let" wasn't widely liked. The use of "val" as an alternative to "final var" had more support but certainly not universal. I think we might look again at "val" at some point in the future. For now, I think there are still some improvements we are wanting to make wrt final anyway which we should make in the meantime.

PR #1238, opened 2020-05-01
GROOVY-9529: bump Object in distance calculation to match interface DGM

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9529


danielsun1106:
LGTM

danielsun1106:
Merged. Thanks!

PR #1239, opened 2020-05-06
GROOVY-9543: Groovydoc generics improvements

vmj:
Fixes the issues listed in https://issues.apache.org/jira/browse/GROOVY-9543

Please, see the commit messages for details.  But I will raise one quote from 724a4da:

> NOTE: This changes how groovydoc generates anchors for Java code.
> groovydoc is using raw types for Groovy code (same as javadoc), but
> types with type parameters for Java code.  With this change, the docs
> for Java code will use raw types, too.

I.e. whereas groovydoc used to generate anchors like `<a name="compare(Class<A>, Class<B>)">` for Java methods, it will now generate `<a name="compare(Class, Class)">`.  Which is what it did for Groovy methods all along, and what javadoc does, too.   If anyone anywhere in the Internet has linked to such Java method docs, the fragment part of those links will be broken.


danielsun1106:
+1

danielsun1106:
Merged. Thanks!

PR #1241, opened 2020-05-08
GROOVY-9545: add extension method for stream to array using class literal

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9545

Looking into STC.  And waiting on comment/feedback in JIRA.

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/PluginDefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/PluginDefaultGroovyMethods.java
@@ -29,6 +29,7 @@ import org.codehaus.groovy.runtime.InvokerHelper;
 import org.codehaus.groovy.runtime.NullObject;
 import org.codehaus.groovy.runtime.RangeInfo;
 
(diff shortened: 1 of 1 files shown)


paulk-asert:
+1 with the following comments:
- I'd suggest `@since` of 3.0.4 if merged soon
- we can probably do a better error message for the case where `type` is null
-  no need to have in `PluginDefaultGroovyMethods` just `DefaultGroovyMethods` since we know we have JDK8+ for 3+
- perhaps a couple of additional tests for arrays of arrays, e.g.:
```
assert Arrays.equals([].stream().toArray(String[]), new String[0][])
assert Arrays.deepEquals([['x'] as String[]].stream().toArray(String[]), [['x'] as String[]] as String[][])
```

paulk-asert:
I merged and made a few tweaks. I ended up leaving it in PluginDefaultGroovyMethods since all the other stream DGM methods were in there. Also, my suggestion for null checking doesn't offer much since it selects the built-in variant in that case anyway. We might add a NullObject variant at some future point when we add proper support for that.

PR #1242, opened 2020-05-10
GROOVY-9465: groovydoc: show class declaration header for all types

vmj:
It seems a mistake that the template required that the type should implement at least one interface before showing the declaration.


paulk-asert:
Good catch. Merged. Thanks!

PR #1243, opened 2020-05-11
GROOVY-9546: groovydoc: fix annotation rendering

vmj:
First stab at fixing the annotation rendering.

Somehow the annotations are not shown on properties at all.

Let me know if this is too hacky.

### DIFF
--- a/subprojects/groovy-groovydoc/src/test/groovy/org/codehaus/groovy/tools/groovydoc/testfiles/anno/Groovy.groovy
+++ b/subprojects/groovy-groovydoc/src/test/groovy/org/codehaus/groovy/tools/groovydoc/testfiles/anno/Groovy.groovy
@@ -1,0 +1,39 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


vmj:
Created a separate issue for the missing property annotations, since I couldn't yet figure it out: https://issues.apache.org/jira/browse/GROOVY-9548

subprojects/groovy-groovydoc/src/test/groovy/org/codehaus/groovy/tools/groovydoc/testfiles/anno/Groovy.groovy
```diff
@@ -0,0 +1,39 @@
...
+ *  under the License.
+ */
+package org.codehaus.groovy.tools.groovydoc.testfiles.anno
+
+import groovy.transform.EqualsAndHashCode
+import groovy.transform.NamedParam
+import groovy.transform.NamedVariant
+
+// The annotations that are used here don't really matter,
+// since groovydoc shows all annotations, not just @Documented ones.
```
paulk-asert, line 26: We do plan to correct this one day. We currently try to minimise resolving classes when not needed but we will likely change that at some point.

paulk-asert:
Merged. Thanks!

subprojects/groovy-groovydoc/src/test/groovy/org/codehaus/groovy/tools/groovydoc/testfiles/anno/Groovy.groovy
vmj: Got it!  [GROOVY-4634](https://issues.apache.org/jira/browse/GROOVY-4634) is probably it, then.

paulk-asert: That's the one. :-)

PR #1244, opened 2020-05-11
GROOVY-9547: groovydoc: resolve tags later for Groovy source code

vmj:
For Java source code, tags are resolved when the template asks for the
comments.  For Groovy, they were resolved as soon as the source file
was processes.  At that point, not all classes in the same package are
processes, so links to those failed to link.


paulk-asert:
Merged, thanks!

PR #1245, opened 2020-05-11
GROOVY-9549: groovydoc: show implemented interfaces in class header

vmj:
Also, add type params if any.


paulk-asert:
Merged, thanks.

PR #1248, opened 2020-05-16
GROOVY-9556: Stub generated without the effect of AST transformation …

danielsun1106:
…makes joint compilation fail

See https://issues.apache.org/jira/browse/GROOVY-9556


paulk-asert:
+1, also fixes the broken test case for GROOVY-9506 on 3/4 branches.

paulk-asert:
Merged, thanks!

PR #1252, opened 2020-05-22
GROOVY-9548: groovydoc: use property modifiers

vmj:
The property seems to contain the correct modifiers (public vs private
that the field has).

Note that this does not fully close the ticket: the property annotations are still not shown.


paulk-asert:
Merged, thanks! I'll look into the annotations aspect sometime soon and comment on the jira issue.

PR #1253, opened 2020-05-23
GROOVY-9569: check outers and uppers for static fields

eric-milles:
- use ArrtibuteExpression for outer field
- use PropertyExpression for upper field

https://issues.apache.org/jira/browse/GROOVY-9569
https://issues.apache.org/jira/browse/GROOVY-9501
https://issues.apache.org/jira/browse/GROOVY-5259


danielsun1106:
Merged. Thanks!

PR #1255, opened 2020-05-24
Sonar: Replace the type specification in this constructor call with the diamond operator ("<>")

mawasak:
Simple Sonar, e.g: https://sonarcloud.io/project/issues?id=apache_groovy&issues=AXIngM_YBdjO4dFcRg1s&open=AXIngM_YBdjO4dFcRg1s

### DIFF
--- a/src/main/java/org/codehaus/groovy/syntax/ASTHelper.java
+++ b/src/main/java/org/codehaus/groovy/syntax/ASTHelper.java
@@ -62,7 +62,7 @@ public class ASTHelper {
     }
 
     public void setPackageName(String packageName) {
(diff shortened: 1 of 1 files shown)


danielsun1106:
Merged. Thanks!

paulk-asert:
Just by way of historical information. We delayed diamond operator usage changes for some time since earlier branches were allowed to be compiled in JDK versions prior to that operator and hence couldn't use that operator. And we wanted to minimise work when merging PRs from later branches to those earlier branches. Since there is almost no activity in those earlier branches now, it is safe to move forward with such changes.

PR #1257, opened 2020-05-25
GROOVY-9571: groovydoc: don't add constructors for types that can't have them

vmj:
just the default ctor removal


vmj:
Hold on, I'll move the `$INIT` thing from this to a separate issue.

PR #1258, opened 2020-05-26
Also output aborted tests in JUnit 5 GroovyJUnitRunnerHelper

Vampire:


Vampire:
I don't think it is my fault that Travis failed.
The sonarqube analysis part failed it seems, with "you are not allowed to analyze" until Travis aborted the build because the log became too big

paulk-asert:
Merged, thanks!

PR #1260, opened 2020-05-27
GROOVY-9575 ASTNode.hashCode violates Object.hashCode contract: does not return consistent value

awallgren:
Remove unsuitable ASTNode.hashCode -- it does not conform to the Object
.hashCode general contract.

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ASTNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ASTNode.java
@@ -18,7 +18,6 @@
  */
 package org.codehaus.groovy.ast;
 
-import java.util.Arrays;
 import java.util.Map;
 
 /**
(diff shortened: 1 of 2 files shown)


danielsun1106:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html#hashCode()

> Whenever it is invoked on the same object more than once during an execution of a Java application, the hashCode method must consistently return the same integer, provided no information used in equals comparisons on the object is modified. **This integer need not remain consistent from one execution of an application to another execution of the same application.**

As the javadoc of `hashCode` shows, "This integer need not remain consistent from one execution of an application to another execution of the same application."

awallgren:
> https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/Object.html#hashCode()
> 
> > Whenever it is invoked on the same object more than once during an execution of a Java application, the hashCode method must consistently return the same integer, provided no information used in equals comparisons on the object is modified. **This integer need not remain consistent from one execution of an application to another execution of the same application.**
> 
> As the javadoc of `hashCode` shows, "This integer need not remain consistent from one execution of an application to another execution of the same application."

I believe you are misreading that.  It doesn't need to be consistent between invocations of the JVM itself (i.e. between different processes). 

It needs to return the same value as long as no properties referenced in equals() changes.

danielsun1106:
Understood :-)
Merged. Thanks for your contributing.

PR #1261, opened 2020-05-27
GROOVY-9576: groovydoc: hide enum $INIT static method

vmj:
This is just a hacky suggestion.  Not sure if this is the way to go because there are similar cases elsewhere:

 * https://issues.apache.org/jira/browse/GROOVY-9572
 * https://issues.apache.org/jira/browse/GROOVY-9576


paulk-asert:
Merged. Thanks!

PR #1262, opened 2020-05-28
GROOVY-9562: same top-level type, not same module for private field read

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9562
https://issues.apache.org/jira/browse/GROOVY-9292

`StaticTypeCheckingVisitor.checkOrMarkPrivateAccess` marked reads and writes in the same module, but later classgen checks did not match field access across top-level types.  This led to a bad type saved for the "this" that is created for variable expressions in `VariableExpressionTransformer` and thus a `ClassCastException` at runtime.


danielsun1106:
Merged. Thanks.

PR #1263, opened 2020-05-28
GROOVY-9499: AIC as argument to this/super constructor call

eric-milles:
Groovy 2.5 backport

https://issues.apache.org/jira/browse/GROOVY-9499
https://issues.apache.org/jira/browse/GROOVY-9168
https://issues.apache.org/jira/browse/GROOVY-6809


paulk-asert:
Merged, thanks!

PR #1264, opened 2020-05-29
GROOVY-9577: maintain insertion order of imports and retain duplicates

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9577


danielsun1106:
Merged. Thanks.

PR #1266, opened 2020-06-02
GroovyShell parse script use given context

caoli5288:
This is useful in embedded scripts somehow.


danielsun1106:
Merged. Thanks!

PR #1267, opened 2020-06-03
GROOVY-9344 (part 1): SC: retain closure shared variable metadata for classgen

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9344

This change prevents the cast exception within the closure due to loss of node metadata:
```groovy
class A {}
class B {}

def test() {
    def var
    var = new A()
    def c = {
        var = new B() // Cannot cast object 'B@4e234c52' with class 'B' to class 'A'
    }
    c()
    //var.toString() // TODO: Cannot cast object 'B@4e234c52' with class 'B' to class 'A' -- coming in part 2
}

test()
```


danielsun1106:
Merged. Thanks!

PR #1268, opened 2020-06-04
GROOVY-7701: do not cast RHS before LHS of assignment is visited

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7701


danielsun1106:
Merged. Thanks

PR #1270, opened 2020-06-08
GROOVY-9558: drop receiver metadata from transformed property expression

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9558


paulk-asert:
Merged, thanks!

PR #1271, opened 2020-06-09
We got a GROOVY-3456 zoombie here.

urs-ikentoo-com:
It seems this got broken again in a code cleanup.
Present in the 2.5 branch, as well.


paulk-asert:
Merged, thanks! See also GROOVY-9584.

urs-ikentoo-com:
@paulk-asert just for my personal curiosity how does that .context property get overwritten through bindings, I wasn't able to find it in the code.

PR #1272, opened 2020-06-09
GROOVY-8840: use inferred type for subscript if operand is placeholder

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-8840

map/list/array[complex expression] may not resolve all generics for "complex expression".  This can prevent matching DGM `getAt(List, int)` for `[]` since "complex expression" needs to resolve to int/Ineteger to make the match.


danielsun1106:
Merged. Thanks

PR #1273, opened 2020-06-10
GROOVY-7549: STC: if "=" RHS type is inaccessible, revert to origin type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7549


danielsun1106:
Merged. Thanks

PR #1274, opened 2020-06-11
GROOVY-7512: add MapStyleConstructorCall override of transformExpression

eric-milles:
- allow PostTypeCheckingExpressionReplacer to replace closure literals

https://issues.apache.org/jira/browse/GROOVY-7512


danielsun1106:
Merged. Thanks.

PR #1276, opened 2020-06-13
GROOVY-9591: no error for dynamic variable in closure in static context

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9591


danielsun1106:
Merged. Thanks.

PR #1277, opened 2020-06-13
GROOVY-9587: don't check for property for non-empty call args

eric-milles:
`transformMethodCallExpression` had become very difficult to understand and modify, so I tried to reorganize it, reduce the repetition, and simplify the conditions as best as I could.

https://issues.apache.org/jira/browse/GROOVY-9587


paulk-asert:
Merged, thanks!

PR #1278, opened 2020-06-14
GROOVY-9592: Replace "groovy.generate.stub.in.memory" with "groovy.me…

danielsun1106:
…m.stub"


eric-milles:
Do you need a system property any more?  For each option that is available to be set on compiler config, why does each also require a system property?

danielsun1106:
As a normal groovy user, we run/compile Groovy code via command line, so the JVM option is still necessary for us.

eric-milles:
Necessary is probably a bit strong.  Why does the system property name need to be changed?  If not removing, why change?

danielsun1106:
As you can find, most of groovy options are short, e.g. `groovy.antlr4`, `groovy.target.indy`, etc.
`groovy.generate.stub.in.memory` is too long to remember and hard to use, it's better to shorten it and make it friendly to users.

PR #1280, opened 2020-06-14
GROOVY-9554: Script: before adding to binding, check for property setter

eric-milles:
I looked for a shared function to check a `Class` for a setter, but came up empty.  Maybe someone knows of a utility call that can replace `hasSetterMethodFor`.

https://issues.apache.org/jira/browse/GROOVY-9554

### DIFF
--- a/src/main/java/groovy/lang/Script.java
+++ b/src/main/java/groovy/lang/Script.java
@@ -18,6 +18,7 @@
  */
 package groovy.lang;
 
+import org.apache.groovy.util.BeanUtils;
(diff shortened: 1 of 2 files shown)


paulk-asert:
For a `ClassNode` we'd do:
```
classNode.getSetterMethod(GeneralUtils.getSetterName(property))
```
Don't think we have the same thing for a `Class`.

paulk-asert:
Merged, thanks. I also tweaked the doco slightly.

PR #1281, opened 2020-06-15
Switch Spock from SNAPSHOT to 2.0-M3-groovy-3.0

szpak:
With '-Dspock.iKnowWhatImDoing.disableGroovyVersionCheck=true' running
with the unsupported Groovy 4 is possible also with non-SNAPSHOT version.

https://github.com/spockframework/spock/pull/1164

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,7 +147,7 @@ ext {
     slf4jVersion = '1.7.25'
     xmlunitVersion = '1.6'
     xstreamVersion = '1.4.12'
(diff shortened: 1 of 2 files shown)


szpak:
Btw, I put `systemProperty "spock.iKnowWhatImDoing.disableGroovyVersionCheck", "true"` to cover embedded Spock tests only in groovy-templates, so it will have to be potentially added also in any other module that will start to use it. Alternatively, you might want to change it to cover all modules automatically (now or in the future).

paulk-asert:
Merged, thanks.
See also: GROOVY-9593

PR #1282, opened 2020-06-16
GROOVY-9594: Rename whitelist/blacklist in SecureASTCustomizer to mor…

paulk-asert:
…e meaningful names

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
+++ b/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
@@ -110,7 +110,7 @@ import java.util.Map;
  * out, but new language features are then implicitly also available and this may not be desirable.
  * The implication is that you might need to update your configuration with each new release.
  * <p>
- * If you set neither an allowed list nor a disallowed list, then everything is permitted.
+ * If neither an allowed list nor a disallowed list is set, then everything is permitted.
  * <p>
  * Combinations of import and star import constraints are authorized as long as you use the same type of list for both.
  * For example, you may use an import allowed list and a star import allowed list together, but you cannot use an import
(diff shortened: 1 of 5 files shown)


src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
```diff
@@ -94,24 +94,28 @@
...
- * Groovy language won't be allowed by defaut. Using blacklists, you can limit the features of the languages by opting
- * out, but new language features would require you to update your configuration.
+ * The recommended way of securing shells is to use allowed lists because it is guaranteed that future features of the
+ * Groovy language won't be accidentally allowed unless explicitly added to the allowed list.
+ * Using disallowed lists, you can limit the features of the language constructs supported by your shell by opting
+ * out, but new language features are then implicitly also available and this may not be desirable.
+ * The implication is that you might need to update your configuration with each new release.
  * <p>
- * If you set neither a whitelist nor a blacklist, then everything is authorized.
+ * If you set neither an allowed list nor a disallowed list, then everything is permitted.
```
jlstrater, line 113:
```suggestion
 * If neither an allowed list nor a disallowed list is set, then everything is permitted.
```

jlstrater: while you're in there... 

I tripped over this when reading it the first time. I think this might be clearer?

src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
```diff
@@ -462,120 +662,200 @@ public void addExpressionCheckers(ExpressionChecker... checkers) {
...
     }
 
     /**
      * Sets the list of classes which deny method calls.
      *
      * Please note that since Groovy is a dynamic language, and
-     * this class performs a static type check, it will be reletively
-     * simple to bypass any blacklist unless the receivers blacklist contains, at
+     * this class performs a static type check, it will be relatively
+     * simple to bypass any disallowed list unless the receivers disallowed list contains, at
```
jlstrater, line 764: both here and below, I think this was supposed to be 'disallowed receivers list' instead of 'receiver[']s disallowed list'?

src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
paulk-asert: both suggestions sound good to me

paulk-asert:
Merged, thanks everyone for the feedback.

PR #1284, opened 2020-06-20
[GROOVY-8100] MarkupBuilder - yield before any markup (state == 0)

eddelplus:
This will let MarkupBuilder yield output at initial state (0), before any markup is created, e.g. allowing the creation of "<!DOCTYPE html>" via MarkupBuilder.mkp.yieldUnescaped.


danielsun1106:
Merged. Thanks.

PR #1285, opened 2020-06-21
GROOVY-9604: STC: add support for accessing additional closure methods and properties

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9604


danielsun1106:
Merged. Thanks.

PR #1286, opened 2020-06-22
GROOVY-8274: allow user-supplied methodMissing in non-static inner class

eric-milles:
InnerClassCompletionVisitor overwrote user's methodMissing or propertyMissing with synthetic dispatcher.

https://issues.apache.org/jira/browse/GROOVY-8274


danielsun1106:
Merged. Thanks!

PR #1287, opened 2020-06-22
GROOVY-9605: leftShift operator does not work on BigInteger (throws U…

paulk-asert:
…nsupportedOperationException)


paulk-asert:
Merged!

PR #1288, opened 2020-06-22
GROOVY-9603: don't apply map literal as constructor shorthand to Object

eric-milles:
`Object o = [:]` is already checked by `isDynamicTyped()`.  `Map<String, Object> m = [:]; m.p = [x:1]` should also not trigger Object's constructor.

https://issues.apache.org/jira/browse/GROOVY-9603


paulk-asert:
Merged, thanks!

PR #1289, opened 2020-06-23
GROOVY-9598: account for static modifier when checking for outer members

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9598


danielsun1106:
Merged. Thanks.

PR #1291, opened 2020-06-26
GROOVY-8999: super.@x should fail for private fields -- no getX() or setX() workaround

eric-milles:
- Error message for `super.x` and `super.@x` now refers to super class; "no such field x for class A" instead of "no such field x for class B", where B extends A and contains `super.@x`
- STC: error for `anything.@x` when field is private (or package-private and "this" is not in same package)

https://issues.apache.org/jira/browse/GROOVY-8999
https://issues.apache.org/jira/browse/GROOVY-8167
https://issues.apache.org/jira/browse/GROOVY-7094


danielsun1106:
Merged. Thanks!

PR #1292, opened 2020-06-28
Joint build Groovy & Grails

erichelgeson:
I believe this will need to be cherry-picked to GROOVY_2_5_X so it runs on that branch as well. 

Master as noted in the comments will not work until Grails has a branch compatible with Groovy 4.

The job currently fails as there is a reported regression in Groovy 3.0.5.

### DIFF
--- a/.github/workflows/grails-joint-validation.yml
+++ b/.github/workflows/grails-joint-validation.yml
@@ -1,0 +1,66 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: "Grails Joint Validation Build"
+#  GROOVY_2_5_X == Grails 4.0.x
+#  GROOVY_3_0_X == grails master
+#  Groovy master branch does not map to any due to changed package names.
+on:
+  push:
+    branches:
+      - GROOVY_2_5_X
+      - GROOVY_3_0_X
+  #      - master
+  pull_request:
+    branches:
+      - GROOVY_2_5_X
+      - GROOVY_3_0_X
+#      - master
+jobs:
+  build:
+    strategy:
+      fail-fast: true
+      matrix:
+        os: [ubuntu-18.04]
+        java: [11.0.6]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up JDK
+        uses: actions/setup-java@v1
+        with:
+          java-version: ${{ matrix.java }}
+      - name: env
+        run: env
+
+      # Select correct Grails branch for this build:
+      - name: Checkout Grails 4.1.x (master)
+        run: cd .. && git clone --depth 1 https://github.com/grails/grails-core.git -b master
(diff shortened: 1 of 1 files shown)


paulk-asert:
Merged, thanks!

danielsun1106:
@paulk-asert 
As @erichelgeson said, the job currently fails as there is a reported regression in Groovy 3.0.5:

https://github.com/apache/groovy/runs/828911951
```

> Task :grails-web-databinding:compileGroovy FAILED

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.

See https://docs.gradle.org/6.5/userguide/command_line_interface.html#sec:command_line_warnings
* What went wrong:
117 actionable tasks: 78 executed, 39 up-to-date
Execution failed for task ':grails-web-databinding:compileGroovy'.
> BUG! exception in phase 'class generation' in source unit '/home/runner/work/groovy/grails-core/grails-web-databinding/src/main/groovy/grails/web/databinding/GrailsWebDataBinder.groovy' At line 541 column 33
  On receiver: obj with message: getAt and arguments: propName
  This method should not have been called. Please try to create a simple example reproducing
  this error and file a bug report at https://issues.apache.org/jira/browse/GROOVY

```

paulk-asert:
Actually, that is a second regression, but yes, we need to fix that one too.

danielsun1106:
@paulk-asert Gotcha. Also, it's better to check out released tags instead of branches because branches will be updated from time to time and may be affected by some buggy commit... thus it's hard for groovy developers to address issues.

erichelgeson:
@danielsun1106 we want to checkout branches and not releases as we want to catch a bug before it's released! Eg: if a commit is made and pushed to 3_0_X that breaks grails, it's better to know right away rather than right before (or after) a release.

paulk-asert:
I think branches on the Groovy side is fine, I am not 100% sure but I think Daniel was suggesting tags on the Grails side. There are pros and cons of each approach. If it is branches on both sides, that is two moving targets and always hard to know on which side caused the error.

erichelgeson:
@paulk-asert Ahh I see now. Both projects now have joint jobs now - so if we want to adjust in the future we can.

paulk-asert:
For 2_5_X, you need to change the return type of the `exit` methods for the classes that extend `groovy.ui.Console`. It now has a bridge method for any pre-compiled usages but re-compiling against that code needs to change the type. This is already done in master, so perhaps just copying that change but without swapping to the new Groovy 3 class (package rename due to split packaging).

erichelgeson:
@paulk-asert I get an error either way:

Change it to groovy 2.5.13 (leaving void) I get:
```
> Task :grails-console:compileGroovy FAILED
startup failed:
/Users/erichelgeson/source/grails-core/grails-console/src/main/groovy/grails/ui/console/support/GroovyConsoleApplicationContext.groovy: 47: The return type of boolean exit(java.util.EventObject) in grails.ui.console.support.GroovyConsoleApplicationContext$1 is incompatible with void in groovy.ui.Console
. At [47:13]  @ line 47, column 13.
               @Override
```

Change it to boolean, I get:
```
/Users/erichelgeson/source/grails-core/grails-console/src/main/groovy/grails/ui/console/support/GroovyConsoleApplicationContext.groovy: 47: The return type of void exit(java.util.EventObject) in grails.ui.console.support.GroovyConsoleApplicationContext$1 is incompatible with boolean in groovy.ui.Console
. At [47:13]  @ line 47, column 13.
               @Override
```

paulk-asert:
Are your error messages around the wrong way? The void variant is meant to be a synthetic bridge method and shouldn't be checked against. Does removing `@Override` help?

erichelgeson:
Yes, copy/pasted in reverse order, but issue stays the same :) Removing override does not change the issue.

paulk-asert:
We are checking for synthetic/bridge only on the overriding method (`overridingMethod`) not parent method (`oldMethod`):
https://github.com/apache/groovy/blob/GROOVY_2_5_X/src/main/java/org/codehaus/groovy/classgen/Verifier.java#L1347
I'll have to trace through whether the bridge method is getting passed through as `oldMethod`.

paulk-asert:
Can you try the latest SNAPSHOT after changing to boolean?

PR #1293, opened 2020-06-29
GROOVY-7971: @CS flow typing incorrectly casting to map at runtime

paulk-asert:


eric-milles:
Does this change accommodate checking `instanceof` under a ternary expression or as part of a larger boolean expression?
```groovy
int m(o) {
  (o instanceof List || o instanceof Map ? o.size() : 1)
}
// or
def x = (o instanceof List || o instanceof Map) && o.isEmpty()
```

#1269 is not specific to if/else, but as you noted is more restrictive.

Does this change fix the runtime case exception of your Deque vs. Stack example under static compilation?

paulk-asert:
@eric-milles No, those cases aren't covered but the if case is. The Stack/Deque case is one of the test cases. The full solution I think will require us to push through Union and Intersection types (though I don't like those names) throughout. I started doing that but never finished. There are lots of hidden assumptions about the current approach throughout the codebase and will take some time. I see this as fixing up a common case while we work longer on the more complete solution. With the if in place, I believe we could have workarounds for the other cases. We could target them as special cases too but it makes the codebase messier the more of these hacks we have.

danielsun1106:
PR #1269  or PR #1293 ?

danielsun1106:
The PR seems to break the build grails:
https://github.com/apache/groovy/runs/836663285?check_suite_focus=true

PR #1294, opened 2020-07-01
GROOVY-4945, GROOVY-9615: call invokeMissingMethod on MetaClass of super

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-4945
https://issues.apache.org/jira/browse/GROOVY-9615


danielsun1106:
Merged. Thanks.

PR #1295, opened 2020-07-02
GROOVY-9597: STC: update receiver after method selection

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9597


danielsun1106:
Merged. Thanks.

PR #1297, opened 2020-07-04
GROOVY-9607: chase down param, local, field, ... for variable expression

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9607
https://issues.apache.org/jira/browse/GROOVY-9344


danielsun1106:
+1
P.S. Grails joint build is green:
https://github.com/apache/groovy/runs/837081914?check_suite_focus=true

danielsun1106:
Merged. Thanks.

PR #1299, opened 2020-07-05
GROOVY-7304: handle private field access from closure for ++x and x++

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7304

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java
@@ -41,6 +41,7 @@ import org.codehaus.groovy.classgen.asm.CompileStack;
 import org.codehaus.groovy.classgen.asm.MethodCallerMultiAdapter;
 import org.codehaus.groovy.classgen.asm.OperandStack;
 import org.codehaus.groovy.classgen.asm.TypeChooser;
+import org.codehaus.groovy.classgen.asm.VariableSlotLoader;
 import org.codehaus.groovy.runtime.InvokerHelper;
 import org.codehaus.groovy.syntax.SyntaxException;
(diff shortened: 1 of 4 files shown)


eric-milles:
To merge into Groovy 2.5, you would also need: https://github.com/apache/groovy/commit/8c0c5b018805fcab7217f5cdc784fc88354cfb15, https://github.com/apache/groovy/commit/543c97b6e6bbb67cd6c9c71cac1ca37ac8bd39f4, and https://github.com/apache/groovy/commit/77a11e89b70b4e785fab36b9c3e93c11913288a2

paulk-asert:
This looks close to me. When I have @CS just on the method like in Jochen's example it still fails, but @CS on the class and it works.

eric-milles:
I'll have a look

paulk-asert:
Yes, it is not uncommon. I'll just have to refresh my memory. I'll take a look in parallel if I get time and if I spot something obvious I'll report back here. If we find an obvious fix we can add it here (since that was the original case in the issue) but otherwise we can just clone the issue and fix separately.

eric-milles:
I think I found the culprit.  I will amend the pull request if testing bears out.

eric-milles:
There still seems to be a problem when static compilation is applied using compiler configuration, like is done in the Groovy7276 test.  When I added a test case there with `@CompileDynamic` on the class to cancel out the config and `@CompileStatic` on the method, the generated methods did not get static compilation.  But I think that is a more general issue with mixing compiler config with annotations and generated methods.

eric-milles:
Also, there is some room for improvement in `org.codehaus.groovy.transform.sc.StaticCompilationVisitor#isStaticallyCompiled` and `org.codehaus.groovy.classgen.asm.sc.StaticTypesWriterController#updateStaticCompileFlag`.  An inner class may have an enclosing method that should take precedence over the outer class.  And the update method appears to repeat some logic that is or could be encompassed in the isser or the SC visitor.

PR #1300, opened 2020-07-05
GROOVY-9620: do not loop over findPropertyInClassHierarchy

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9620


paulk-asert:
Merged, thanks!

PR #1301, opened 2020-07-07
GROOVY-9422: non-static inner class constructor call sets type of "this"

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9422
https://issues.apache.org/jira/browse/GROOVY-6904


paulk-asert:
Merged, thanks!

PR #1303, opened 2020-07-08
GROOVY-7996: STC: error for mismatched closure resolve strategies

eric-milles:
Provide the user some clue that runtime behavior will not match declared behavior (in explicit or implicit `@DelegatesTo` metadata).

image

https://issues.apache.org/jira/browse/GROOVY-7996
https://issues.apache.org/jira/browse/GROOVY-7597

### DIFF
--- a/src/test/org/codehaus/groovy/classgen/asm/sc/DelegatesToStaticCompileTest.groovy
+++ b/src/test/org/codehaus/groovy/classgen/asm/sc/DelegatesToStaticCompileTest.groovy
@@ -92,7 +92,7 @@ class DelegatesToStaticCompileTest extends DelegatesToSTCTest implements StaticC
                 }
 
                 class Handler {
-                    def doWithEntity(@DelegatesTo(Entity) Closure c) {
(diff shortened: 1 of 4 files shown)


src/test/org/codehaus/groovy/classgen/asm/sc/DelegatesToStaticCompileTest.groovy
```diff
@@ -111,4 +111,29 @@ class DelegatesToStaticCompileTest extends DelegatesToSTCTest implements StaticC
...
+                protected List bars = []
+                def baz() {
+                    new Foo().build {
+                        bars.isEmpty() // fails if executed; delegate's propertyMissing takes precedence
+                    }
+                }
+            }
+
+            new Bar().baz()
+        ''', 'Closure parameter with resolve strategy 0 passed to method with resolve strategy 1'
```
paulk-asert, line 137: I'd suggest we need a `getResolveStrategyAsString()` method to make these more meaningful.

eric-milles: Added method to ClosureUtils and moved the test so it is tried for `@CompileStatic` and `@TypeChecked`.

PR #1304, opened 2020-07-08
GROOVY-9601: reduce method calls in findClassMember and getPropertyName

eric-milles:
Complete the conversion of `VariableScopeVisitor#findClassMember` from recursive to iterative.  Rewrite `MethodNodeUtils#getPropertyName` to check characters instead of strings.  You might be able to gain a bit by inlining `BeanUtils#decapitalize`.  You can also test for `void` and `boolean` using == if you get the redirect of the return type.

Can you give this a try with your performance test(s)?  I got about 1 sec improvement (4500ms vs 5500ms) with the Groovy9601_CLASSGEN test.  I'd be interested to see the SVG now.

https://issues.apache.org/jira/browse/GROOVY-9601

### DIFF
--- a/src/main/java/org/apache/groovy/ast/tools/MethodNodeUtils.java
+++ b/src/main/java/org/apache/groovy/ast/tools/MethodNodeUtils.java
@@ -30,6 +30,10 @@ import static org.apache.groovy.util.BeanUtils.decapitalize;
  * Utility class for working with MethodNodes
  */
 public class MethodNodeUtils {
+
+    private MethodNodeUtils() {
+    }
+
     /**
      * Return the method node's descriptor including its
      * name and parameter types without generics.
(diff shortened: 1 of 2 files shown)


danielsun1106:
Here is the output of async profiler for Groovy9601_CLASSGEN executed on the distribution built from master and this PR:
[groovy9601.zip](https://github.com/apache/groovy/files/4893886/groovy9601.zip)

As we can see, `getPropertyName` costs less time, which is reduced `9.97%`(5.16% + 4.77%) to `9.00%`(4.50% + 4.50%)

P.S. master costs 13000ms and this pr costs 12000ms on my machine, something is better than nothing ;-)

danielsun1106:
Merged. Thanks.

PR #1305, opened 2020-07-09
Updated the path to assets/img in groovy-console.adoc

mardukbp:
assets is now in the spec directory, one directory above. 

Now the images are shown in the documentation.

### DIFF
--- a/subprojects/groovy-console/src/spec/doc/groovy-console.adoc
+++ b/subprojects/groovy-console/src/spec/doc/groovy-console.adoc
@@ -30,7 +30,7 @@ This page documents the features of this user interface.
 [[GroovyConsole-Basics]]
 == Basics
 
-image:assets/img/GroovyConsole.gif[image]
+image:../assets/img/GroovyConsole.gif[image]
 
 . Groovy Console is launched via `groovyConsole` or
 `groovyConsole.bat`, both located in `$GROOVY_HOME/bin`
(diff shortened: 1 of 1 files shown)


paulk-asert:
Thanks for looking at this. We do need to fix it. Unfortunately, there would be more to do than just your suggestion. While it fixes the images in the IDE, it then breaks the generated content (which is currently correct). Looking at the history:
https://github.com/apache/groovy/commit/bdaef2a1539d3b3d0a9a5fa549ddd7421673797f
It looks like we made the move to improve consistency. Perhaps we should have moved the other projects instead.

mardukbp:
What matters is that the images are shown in the documentation. 
In particular, I think that visualizing the results using Swing is a 
great feature of the Groovy Console and the image is key to make
people interested in using it.

paulk-asert:
The images do currently appear in the documentation:
http://groovy-lang.org/groovyconsole.html
And also if you have the source checked out from github and then run `gradlew :groovy-console:asciidoc` and look in the generated directory, e.g. `subprojects/groovy-console/target/asciidoc/html5/groovy-console.html` you will see that it is correct.

Where it is currently not showing up correctly is browsing the source online, e.g.
https://github.com/apache/groovy/blob/master/subprojects/groovy-console/src/spec/doc/groovy-console.adoc
or browsing the source from the checked out github repo from within an IDE.

mardukbp:
Thanks for the references. I guess that if the paths in the AsciiDoc files work, then while building the documentation either the files are moved or the paths are modified. So either the AsciiDoc files or the build script(s) have to be modified. Am I missing something?

paulk-asert:
I just moved everything, as per [b388a251aee](https://github.com/apache/groovy/commit/b388a251aee3e03b9bd3d5ed5f2fcbe6158bc85d), to hopefully correct what you observed. Let me know if you spot anything out of place.

PR #1306, opened 2020-07-09
GROOVY-9631: Replace legacy data structure with Java collection

danielsun1106:
See https://issues.apache.org/jira/browse/GROOVY-9631


danielsun1106:
Grails joint build is green:
https://github.com/apache/groovy/runs/858782411

danielsun1106:
@paulk-asert ping ;-)

danielsun1106:
Grails joint build is green:
https://github.com/apache/groovy/runs/860503230

danielsun1106:
@paulk-asert The PR is targeted to 4.0.0 only and is ready for reviewing now.

danielsun1106:
I'm going to replace more legacy data structure, so convert the PR to draft for now.

danielsun1106:
Before the PR applied, the properties in the json output is out of order.
After the PR applied, the properties in the json output keeps the original order in the its class now.

As a result, the PR will break some tests of Grails, but it's easy for Grails developers to fix:
https://github.com/apache/groovy/runs/861801220

@paulk-asert It's ready to review. I'm not sure if it could be merged into 3.0.X branch too...

blackdrag:
@danielsun1106 I am actually wondering why you added the order of the fields in this PR. Be ensured that the only way to read fields in declaration order from a class in Java is by reading the bytecode directly. Reflection may or may not present you the fields in the right order. There is no guarantee. Which means to me:
* keeping the order of fields in Groovy is questionable as long as we use reflection to build such structures
* having tests depending on the order of fields is wrong

danielsun1106:
@blackdrag Actually I have not sorted the fields on purpose. After we use `LinkedHashMap`, the order changes and I find it always keeps the order of fields declared. In order to control the expectation of users, I do not announce keeping field order in the JIRA ticket now ;-)

blackdrag:
@danielsun1106 not announcing the field order is a good move imho ;-) But of course that shows imho two things: LinkedHashMap will not be able to fulfill the prupose of keeping the declaration order and those JSON tests should be order independent - if they should not, then there is a problem.

blackdrag:
@danielsun1106 actual there are two points I am wondering about - well actually only one: performance. How does creating a meta class compare to the old version. And how is the read performance of the new managed map compared to the old. I am asking here especially because those "new ManagedKey" calls sound like they could cost a lot of performance and the map you replace actually came into existence because of us using before maps with a key generation that took too long.

danielsun1106:
@blackdrag As to your worries about performance, I created benchmark project to test the performance of concurrent maps ( https://github.com/danielsun1106/benchmark-concurrentmaps )

Here is the result generated on my local machine:
```
Benchmark                                                 Mode  Cnt    Score    Error   Units
ConcurrentMapBench.testConcurrentIdentityHashMap_get     thrpt   15  176.293 ± 14.399  ops/ms
ConcurrentMapBench.testConcurrentIdentityHashMap_put     thrpt   15   11.263 ±  8.088  ops/ms
ConcurrentMapBench.testConcurrentReferenceHashMap_get    thrpt   15  197.955 ± 12.753  ops/ms
ConcurrentMapBench.testConcurrentReferenceHashMap_put    thrpt   15  194.053 ±  4.108  ops/ms
ConcurrentMapBench.testManagedConcurrentMap_get          thrpt   15  154.559 ±  5.860  ops/ms
ConcurrentMapBench.testManagedConcurrentMap_put          thrpt   15   10.226 ±  9.142  ops/ms
ConcurrentMapBench.testManagedIdentityConcurrentMap_get  thrpt   15  140.041 ±  2.827  ops/ms
ConcurrentMapBench.testManagedIdentityConcurrentMap_put  thrpt   15    6.897 ±  5.559  ops/ms
```

According to the performance test report, `ManagedIdentityConcurrentMap` is almost same efficient to `ManagedConcurrentMap` when executing `get` method, but less efficient when executing `put` method.

Also, `ConcurrentReferenceHashMap` based on `ConcurrentMap` is more efficient than other alternatives, so I am going to polish `ManagedIdentityConcurrentMap` and make it base on `ConcurrentReferenceHashMap` written by Doug Lea.

P.S. `ManagedIdentityConcurrentMap` and `ManagedConcurrentMap` are from Groovy 4.0 (master branch), and `ConcurrentIdentityHashMap` is written by Doug Lea too.

PR #1307, opened 2020-07-09
GROOVY-9601: reduce checks made by findClassMember and getPropertyName

eric-milles:
`findClassMember`
 - java.lang.Object has no fields, no properties, no interfaces, no isX methods, no setX methods and the only getX method is `getClass`, but "class" cannot be a free-standing variable name according to the parser
- it is possible to make bean-style reference to abstract method:
    ```groovy
    abstract class A {
        abstract getX()
        void test() {
          println x
        }
    }
    ```

`getPropertyName`
 - putting the cached class on the right side of `equals` triggers reference check instead of name string check for primitives
    ```java
    public boolean equals(Object that) {
        if (that == this) return true;
        if (!(that instanceof ClassNode)) return false;
        if (redirect != null) return redirect.equals(that); // go here so "that == this" check works for void, boolean, Object, etc.
        return (((ClassNode) that).getText().equals(getText()));
    }
    ```

https://issues.apache.org/jira/browse/GROOVY-9601

### DIFF
--- a/src/main/java/org/apache/groovy/ast/tools/MethodNodeUtils.java
+++ b/src/main/java/org/apache/groovy/ast/tools/MethodNodeUtils.java
@@ -86,17 +86,17 @@ public class MethodNodeUtils {
         if (nameLength > 2) {
             switch (name.charAt(0)) {
                 case 'g':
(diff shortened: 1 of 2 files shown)


danielsun1106:
Merged. Thanks.

PR #1308, opened 2020-07-09
GROOVY-9634: use getDeclaredMethods(String), not getMethods(String)

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9634


danielsun1106:
Merged. Thanks.

PR #1311, opened 2020-07-13
GROOVY-7848: STC: retain generics of list or map elements

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7848

`[[1],[2]]` should infer as `List<List<Integer>>`
`[1:['a'],2:['b']]` should infer as `Map<Integer,List<String>>`


paulk-asert:
Proposed PR merged. Thanks!

PR #1312, opened 2020-07-13
set range and method pointer/reference types like map, list and closure

eric-milles:
- see ClosureExpression, ListExpression and MapExpression constructors


paulk-asert:
merged, thanks

PR #1313, opened 2020-07-13
GROOVY-9642: resolve anon. inner base type in same scope as ctor call

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9642


paulk-asert:
Merged.

PR #1315, opened 2020-07-14
GROOVY-8999: super.@x should fail for private fields; no getX() or setX(_) workarounds

eric-milles:
@paulk-asert: Here are the changes from GROOVY-8999 that straighten out `super.@name` for Groovy 2.5.  The STC error was left out because the field access checks are not as complete as they are in Groovy 3.  The `InvocationWriter` changes are needed in Groovy 2.4 for `super.@name` to actually skip over `this.@name`.


paulk-asert:
Merged. Feel free to close.

PR #1317, opened 2020-07-17
GROOVY-8715: check array in ResolveVisitor.resolveToInner

eric-milles:
`ResolveVisitor#resolveToOuterNested` probably has the same issue with array types.  I just did not want to break it down to understand what kind of test case to create.

I also grouped up some overloads and reduced the checking done by `resolveToInner`.  The path through `resolve` already drives `resolve("p.A$B", false, true, true)` within `resolveFromStaticInnerClasses`.

https://issues.apache.org/jira/browse/GROOVY-8715

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java
+++ b/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java
@@ -322,24 +322,14 @@ public class ResolveVisitor extends ClassCodeExpressionTransformer {
         genericParameterNames = oldPNames;
     }
 
-    protected boolean resolveToInner(final ClassNode type) {
-        // we do not do our name mangling to find an inner class
-        // if the type is a ConstructedClassWithPackage, because in this case we
-        // are resolving the name at a different place already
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/control/ResolveVisitor.java
```diff
@@ -1467,10 +1472,10 @@ public void visitClass(final ClassNode node) {
         }
 
         ClassNode sn = node.getUnresolvedSuperClass();
-        if (sn != null) resolveOrFail(sn, node, true);
+        if (sn != null) resolveOrFailPlus(sn, node);
```
paulk-asert, line 1475: I think `resolveOrFailPlus` isn't a very meaningful name. Agreed the existing parameter is also confusing.

paulk-asert: I'll make a minor tweak which I think improves it. It is private, so we can tweak further if needed.

PR #1318, opened 2020-07-18
GROOVY-5103: allow plain star imports to pick up static inner types

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-5103

- Java allows "import foo.Bar.*" to resolve "Baz" to "foo.Bar$Baz"

- Note: This is also a compatibility/transparency issue for Eclipse Groovy Development Tools, which has supported star imports that resolve inner types for years.

Checking star imports for type reference is limited to "real" star imports (those present in the source text) to prevent trying to resolve "java.lang", "java.util", etc. for every single `ModuleNode`.  Type resolution also enables display of Javadoc information for the imported type:
image

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ImportNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ImportNode.java
@@ -25,7 +25,7 @@ import static java.util.Objects.requireNonNull;
  */
 public class ImportNode extends AnnotatedNode {
 
-    private final ClassNode type;
+    private ClassNode type;
     private final String alias;
     private final String fieldName;
     private final String packageName;
@@ -141,6 +141,10 @@ public class ImportNode extends AnnotatedNode {
 
     public ClassNode getType() {
         return type;
+    }
+
+    public void setType(final ClassNode type) {
+        this.type = type;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


paulk-asert:
This seems okay. It doesn't seem to break the existing import of static classes using "import static", which while possibly unexpected from a Java point-of-view, we have supported for a long time. I'll note that your patch doesn't extend to another "exotic" case we currently support.
```
import java.util.Map as SomeMap
import static SomeMap.Entry as SomeEntry

assert SomeEntry instanceof Class
```
works but not without the static. It is also somewhat unusual but accounts for aliases to some degree and we have supported it for along time too. We can cover that off separately.

paulk-asert:
Merged, thanks.

eric-milles:
> ... your patch doesn't extend to another "exotic" case we currently support.
> 
> ```
> import java.util.Map as SomeMap
> import static SomeMap.Entry as SomeEntry
> 
> assert SomeEntry instanceof Class
> ```
> 
> works but not without the static. It is also somewhat unusual but accounts for aliases to some degree and we have supported it for along time too. We can cover that off separately.

If you wanted to shut down this "exotic" behavior, you could change this bit:
```
    public void visitClass(final ClassNode node) {
        ...
            for (ImportNode importNode : module.getStaticImports().values()) {
                ClassNode type = importNode.getType();
                if (resolve(type, true, true, true)) continue; // set flags to "false, false, true" to not use other imports to help resolve this import
                addError("unable to resolve class " + type.getName(), type);
            }
```

Just an option to consider.  I don't think anyone coming into the language is expecting one import to influence another.

paulk-asert:
Not planning to shut it down at this stage. I have had numerous folks (albeit quite a few years ago now) who've come up to me at conferences and expressed their like for that feature and how it made it easier for DSLs and various porting scenarios.

PR #1319, opened 2020-07-19
Minor tweak: create `Lookup` instance according to vm plugin

danielsun1106:


src/main/java/org/codehaus/groovy/vmplugin/v8/Java8.java
```diff
@@ -636,7 +637,7 @@ private Object getInvokeSpecialHandleFallback(Method method, Object receiver) {
         }
         Class<?> declaringClass = method.getDeclaringClass();
         try {
-            return getLookupConstructor().newInstance(declaringClass, -1).
```
paulk-asert, line 639: Just this line I am unsure about. It will be different params passed through but I am unsure of the impact.

danielsun1106: The Java8 build is green, so I assume it's OK to make this change.

paulk-asert: I was more worried about the impact on other code bases. But there might be no issue. I'll try to find a way to trigger the line in question in an adverse way in the next day or two.

danielsun1106: Any findings?

eric-milles:
What is the overall purpose of this change?  Are you expecting that the way in which `Lookup` instances are created will change?  At the moment, `Java8#of` is the provider.  I don't particularly like the interface, but since it's public, we're somewhat stuck with it.

danielsun1106:
Yep, the static method "of" seems meaningless. We need instance method to apply polymorphism

PR #1321, opened 2020-07-21
GROOVY-9648: Bad error message when attempting to call a missing cons…

paulk-asert:
…tructor with named args

### DIFF
--- a/src/main/java/groovy/lang/MetaClassImpl.java
+++ b/src/main/java/groovy/lang/MetaClassImpl.java
@@ -1762,19 +1762,39 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
         if (res instanceof MetaMethod) return (MetaMethod) res;
         CachedConstructor constructor = (CachedConstructor) res;
         if (constructor != null) return new MetaConstructor(constructor, false);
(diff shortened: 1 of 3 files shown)


danielsun1106:
It would be better to have a test to show the improved  error message.

paulk-asert:
> It would be better to have a test to show the improved error message.

Yes, I plan to add a testcase. I just hadn't decided the best one to add. The example I had in the issue completes "successfully" via the MOP path which is then taken. It supplies a null in the constructor and then coerces the Date to a String for the subsequent setter call. If you swap from a String property to something with no coercion you then get a GroovyCastException which I think is okay. To get the same kind of "failed to invoke constructor" message, we'd need to throw the exception instead of return null in the code which would rule out some MOP tricks like the coercion.

danielsun1106:
Also, we could use lambda to avoid duplicated code:

```java
BiPredicate<Object, Integer> bp = (res, cnt) -> {
   res instanceof CachedConstructor && ((CachedConstructor)res).getParameterTypes().length == cnt
}

if (bp.test(res, 1)) return null;
    ...
if (bp.test(res, 2)) return null;
```

paulk-asert:
Merged.

PR #1322, opened 2020-07-23

paulk-asert:
Co-authored-by: Daniel Sun <sunlan@apache.org>

### DIFF
--- a/src/main/java/groovy/lang/GString.java
+++ b/src/main/java/groovy/lang/GString.java
@@ -100,11 +100,20 @@ public abstract class GString extends GroovyObjectSupport implements Comparable,
         return values;
     }
 
+
     public GString plus(GString that) {
-        Object[] values = getValues();
+        Object[] thisValues = getValuesInternal(this);
         return new GStringImpl(
-                appendValues(values, that.getValues()),
-                appendStrings(getStrings(), that.getStrings(), values.length));
+                appendValues(thisValues, getValuesInternal(that)),
+                appendStrings(getStringsInternal(this), getStringsInternal(that), thisValues.length));
+    }
+
+    private String[] getStringsInternal(GString gs) {
+        return gs instanceof GStringImpl ? ((GStringImpl) gs).getStringList().toArray(EMPTY_STRING_ARRAY) : gs.getStrings();
+    }
+
+    private Object[] getValuesInternal(GString gs) {
+        return gs instanceof GStringImpl ? ((GStringImpl) gs).getValueList().toArray(EMPTY_OBJECT_ARRAY) : gs.getValues();
     }
 
     private static String[] appendStrings(String[] strings1, String[] strings2, int values1Length) {
@@ -164,24 +173,24 @@ public abstract class GString extends GroovyObjectSupport implements Comparable,
     }
 
     private int calcInitialCapacity() {
-        String[] strings = getStrings();
+        final String[] ss = getStringsInternal(this);
 
         int initialCapacity = 0;
-        for (String string : strings) {
+        for (String string : ss) {
             initialCapacity += string.length();
         }
 
-        initialCapacity += values.length * Math.max(initialCapacity / strings.length, 8);
+        initialCapacity += values.length * Math.max(initialCapacity / ss.length, 8);
 
         return Math.max((int) (initialCapacity * 1.2), 16);
     }
 
     @Override
     public Writer writeTo(Writer out) throws IOException {
-        String[] s = getStrings();
+        final String[] ss = getStringsInternal(this);
         int numberOfValues = values.length;
-        for (int i = 0, size = s.length; i < size; i++) {
-            out.write(s[i]);
+        for (int i = 0, size = ss.length; i < size; i++) {
+            out.write(ss[i]);
             if (i < numberOfValues) {
                 final Object value = values[i];
 
@@ -211,12 +220,12 @@ public abstract class GString extends GroovyObjectSupport implements Comparable,
 
     @Override
     public void build(final GroovyObject builder) {
-        final String[] s = getStrings();
+        final String[] ss = getStringsInternal(this);
         final int numberOfValues = values.length;
 
-        for (int i = 0, size = s.length; i < size; i++) {
+        for (int i = 0, size = ss.length; i < size; i++) {
             builder.getProperty(MKP);
-            builder.invokeMethod(YIELD, new Object[]{s[i]});
+            builder.invokeMethod(YIELD, new Object[]{ss[i]});
             if (i < numberOfValues) {
                 builder.getProperty(MKP);
                 builder.invokeMethod(YIELD, new Object[]{values[i]});
--- a/src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
+++ b/src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
@@ -19,6 +19,11 @@
 package org.codehaus.groovy.runtime;
 
 import groovy.lang.GString;
+import org.apache.groovy.ast.tools.ImmutablePropertyUtils;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Default implementation of a GString used by the compiler. A GString consists
@@ -29,6 +34,8 @@ import groovy.lang.GString;
 public class GStringImpl extends GString {
     private static final long serialVersionUID = 3581289038662723858L;
     private final String[] strings;
+    private boolean cacheable;
+    private String cachedStringLiteral;
 
     /**
      * Create a new GString with values and strings.
@@ -48,6 +55,7 @@ public class GStringImpl extends GString {
     public GStringImpl(Object[] values, String[] strings) {
         super(values);
         this.strings = strings;
+        cacheable = checkValuesImmutable();
     }
 
     /**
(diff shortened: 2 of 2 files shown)


src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
```diff
@@ -47,18 +54,67 @@
...
+        cachedStringLiteral = null;
+        return super.getValues();
+    }
+
+    /**
+     * Get the strings of this GString as an unmodifiable list.
+     * This method is preferred over {@code getStrings()} in performance critical scenarios.
+     */
+    public List<String> getStringsList() {
+        return Collections.unmodifiableList(Arrays.asList(strings));
```
danielsun1106, line 86: How about simplifying `Collections.unmodifiableList(Arrays.asList(strings))` to `Arrays.asList(strings)` and renaming the method name to `getStringList()`?

src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
```diff
@@ -47,18 +54,67 @@
...
+    }
+
+    /**
+     * Get the values of this GString as an unmodifiable list.
+     * This method is preferred over {@code getValues()} in performance critical scenarios.
+     */
+    public List<?> getValuesList() {
+        return Collections.unmodifiableList(Arrays.asList(super.getValues()));
```
danielsun1106, line 94: Same to this line

danielsun1106:
I tried the PR and found no performance improvement. I assume `toString()` has not used `cachedStringLiteral`.
```
D:\_TEMP\GROOVY-9637>groovy -v
Groovy Version: 4.0.0-SNAPSHOT JVM: 11.0.7 Vendor: Azul Systems, Inc. OS: Windows 10

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
8758ms

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
8651ms

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
9018ms
```

FYI. https://github.com/apache/groovy/pull/1309#issuecomment-662822845

danielsun1106:
The performance is improved a lot ;-)
The costed time is reduced from `8700ms` to `370ms`

```
D:\_TEMP\GROOVY-9637>groovy -v
Groovy Version: 4.0.0-SNAPSHOT JVM: 11.0.7 Vendor: Azul Systems, Inc. OS: Windows 10

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
399ms

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
365ms

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
361ms

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
290ms

D:\_TEMP\GROOVY-9637>groovy Groovy9637.groovy
337ms

D:\_TEMP\GROOVY-9637>
```

src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
danielsun1106: Forget it... I find the result of `Arrays.asList(strings)` supports changing the elements by `set` method.

paulk-asert:
It is still a breaking change but much fewer cases now:
```
def s = ['The ', ' brown '] as String[]
def v = ['quick', 'fox'] as Object[]
def gstr = new org.codehaus.groovy.runtime.GStringImpl(v, s)
println gstr // The quick brown fox
v[0] = 'sly'
s[1] = ' old '
println gstr // with caching: The quick brown fox, before: The sly old fox
```
Normally the input array isn't accessible, so perhaps not a problem.

paulk-asert:
Using `GString#plus` can return a cacheable GString and doesn't impact the existing GStrings:
```
def g1 = "integer: ${1}, double: ${1.2d}, string: ${'x'}"
def g2 = "class: ${Map.class}, boolean: ${true}"
def g3 = g1 + ' ' + g2
	g1.toString()
	g2.toString()
	g3.toString()
}
println "${e - b}ms"
```

src/main/java/groovy/lang/GString.java
```diff
@@ -101,9 +101,12 @@ public Object invokeMethod(String name, Object args) {
     }
 
     public GString plus(GString that) {
-        Object[] values = getValues();
+        Object[] v1 = this instanceof GStringImpl ? ((GStringImpl)this).getValueList().toArray(new Object[0]) : getValues();
```
danielsun1106, line 104: We could use `EMPTY_STRING_ARRAY` and `EMPTY_OBJECT_ARRAY` to avoid creating arrays repeatly

danielsun1106: `this instanceof GStringImpl ? ((GStringImpl)this).getValueList().toArray(new Object[0]) : getValues()` and `this instanceof GStringImpl ? ((GStringImpl)this).getStringList().toArray(new String[0]) : getStrings()` could be extracted to seperate methods

danielsun1106:
> It is still a breaking change but much fewer cases now:
> 
> ```
> def s = ['The ', ' brown '] as String[]
> def v = ['quick', 'fox'] as Object[]
> def gstr = new org.codehaus.groovy.runtime.GStringImpl(v, s)
> println gstr // The quick brown fox
> v[0] = 'sly'
> s[1] = ' old '
> println gstr // with caching: The quick brown fox, before: The sly old fox
> ```
> 
> Normally the input array isn't accessible, so perhaps not a problem.

Agreed :-)

paulk-asert:
@blackdrag If you get a chance, please cast your eye over this PR instead of the earlier one

danielsun1106:
@paulk-asert Should we take the multi-thread issues into account?

blackdrag:
This again puts a big question mark on the split between GString and GStringImpl. I think code like those instanceof checks should be avoided. Could we shortly summarize the goal? I am not sure if the focus is set right in total.

paulk-asert:
**_Goal_**: improve the performance of GStrings where possible but without foregoing behavior where a value can be dynamically supplied by a closure including lazily evaluated
**_Highly desired_**: don't break binary compatibility unless we really have to to minimise impact on any frameworks that might be extending the existing classes in ways unknown to us
**_Highly desired_**: should be invisible to the user
**_Somewhat desired_**: don't break less commonly used practice of mutating strings/values if possible
**_Current approach_**: cache only where values are simple types and known to be immutable - and skip caching even in that case if the values/strings are accessed externally

The pre-existing design (class split and array inputs/outputs) all work against maximising the potential cases which could be cached, so the current PR is very much "handle quick wins" and nothing more. We could totally re-design those classes and have a major breaking change and deal with more cases but we'd possibly forgo some existing dynamic GString use cases and for a dynamic language you could argue little would be achieved because metaprogramming could alter what should be output each time. We currently have things like GROOVY-2599 and hard-coded InvokerHelper paths actually limiting metaprogramming impact somewhat but they might be things we want to fix down the track.

blackdrag:
> **_Goal_**: improve the performance of GStrings where possible but without foregoing behavior where a value can be dynamically supplied by a closure including lazily evaluated

I see that as the main goal and the other points less as strict constraints or subgoals, more as nice-to-have

> **_Highly desired_**: don't break binary compatibility unless we really have to to minimise impact on any frameworks that might be extending the existing classes in ways unknown to us
> **_Highly desired_**: should be invisible to the user
> **_Somewhat desired_**: don't break less commonly used practice of mutating strings/values if possible
> **_Current approach_**: cache only where all values are simple types and known to be immutable - and skip caching even in that case if the values/strings are accessed externally

is being able to subclass GString still desired. I mean it goes a little bit into binary compatibility, but of course only if you are actually subclassing GString and even then only maybe

> The pre-existing design (class split and array inputs/outputs) all work against maximising the potential cases which could be cached, so the current PR is very much "handle quick wins" and nothing more. We could totally re-design those classes and have a major breaking change and deal with more cases but we'd possibly forgo some existing dynamic GString use cases and for a dynamic language you could argue little would be achieved because metaprogramming could alter what should be output each time. We currently have things like [GROOVY-2599](https://issues.apache.org/jira/browse/GROOVY-2599) and hard-coded InvokerHelper paths actually limiting metaprogramming impact somewhat but they might be things we want to fix down the track.
> 
> @danielsun1106 chime in if you have any differing view

I like that we have the optimization in GStringImpl now. The part with the internal in GString is still bothering me though:

>  return gs instanceof GStringImpl ? ((GStringImpl) gs).getStringList().toArray(EMPTY_STRING_ARRAY) : gs.getStrings();

I have not tried it out, but I would like this to be considered:

* change implementation to work with lists instead of array 
* remove the get*Internal methods and use get*List methods instead

this will allow the list variant to become the fast case in the sense of not having to convert from an array... 

Now I am rethinking the approach... what about if we

* remove the list methods and internal methods in GString
* have a "freeze" method that will return a new GString, which will potentially ignore changes to the array. I would use that in GStringImpl only though
* have GString more or less like it is now, which means no string cache at all
* give GStringImpl control over the optimization
* calling getValues and getStrings on GString and GStringImpl will then still invalidate the cached String, unless the GStringImpl has been frozen
* a frozen GStringImpl will always return the cached string
* have non invalidating methods to return the values and strings arrays in GStringImpl and use them for toString
* move the string building code of GString into a utility class and call it from GString and GStringImpl with their respective method getValues/getStrings methods

API writers can the use "freeze" if they want to look at the structure of a GString, without maybe causing a performance hit - but actually I think the cases where we look at the structure of a GString have much bigger code following that. Actually I think the freeze feature is more a nice-to-have. After all, not having freeze does not worsen the situation compared to today at all. And of course API users that mutate the arrays will still be able to change the GString.

To show what I mean: https://github.com/apache/groovy/pull/1327
Of course that can be improved by a bit of refactoring and moving the static methods as well as the logic for build and writeTo into a utility class, that is then used by GString and GStringImpl to avoid code duplication and have the only the call to the utility method and the getting of values and strings in the class itself.

paulk-asert:
Closing this one but see:

PR #1323, opened 2020-07-25
 JmxMetaMapBuilder.buildOperationMapFrom throw StringIndexOutOfBoundsException

hdfg159:
JmxMetaMapBuilder.buildOperationMapFrom for method "get" or "set" throw StringIndexOutOfBoundsException
example:
```
class TestGet {
	String sb
	
	def get(String str) {
		str
	}
	
	static main(args) {
		def testGet = new TestGet()
		def property = testGet.metaClass.getMethods()
		def ms = testGet.getClass().getDeclaredMethods()*.name
		def jmx = new JmxBuilder()
		def beans = jmx.export {
			bean(
					target: new TestGet(),
					attributes: [],
					operations: "*"
			)
		}
	}
}
```
after run:
```
Exception in thread "main" java.lang.RuntimeException: Failed to create component for 'bean' reason: java.lang.StringIndexOutOfBoundsException: begin 2, end 4, length 3
```


danielsun1106:
It would be better to have a test in this PR

danielsun1106:
Merged with some minor tweaks.
Thanks for your contribution!

danielsun1106:
BTW, I find you are the 300th contributor. Congratulations!

PR #1324, opened 2020-07-25

paulk-asert:
Co-authored-by: Daniel Sun sunlan@apache.org


paulk-asert:
@blackdrag this is the other alternative I alluded to in earlier comments. No change to `GString`, `GStringImpl` enhanced slightly with a flag that enables defensive copy in/out for the string and value arrays. We create GStrings in @CompileStatic code with that flag set. Caching is only enabled when that flag is set and values are simple immutables. This piggy-backs on the idea that @CS turns off "dynamic behaviors" even though mutating those arrays is a bit of a stretch of what constitutes dynamic behavior.

paulk-asert:
Daniel's simple benchmark:
```
def gstr = makeGString()
	gstr.toString()
}
println "${e - b}ms"

//@groovy.transform.CompileStatic
def makeGString() {
    def now = java.time.LocalDateTime.now()
    "integer: ${1}, double: ${1.2d}, string: ${'x'}, class: ${Map.class}, boolean: ${true}, now: ${now}"
}

```

blackdrag:
My feeling is that stretch is a bit too far.

paulk-asert:
Closing this one but see:

PR #1325, opened 2020-07-25
GROOVY-9653: SC: improve classgen for multiple setter method candidates

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9653


danielsun1106:
Merged. Thanks!

PR #1326, opened 2020-07-27
GROOVY-9652: after "x" to "delegate.x", don't make "delegate.delegate.x"

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9652


danielsun1106:
Merged. Thanks.

PR #1327, opened 2020-07-28
GROOVY-9637: GString variant with frozen

blackdrag:


paulk-asert:
I'll take a look. If you get a chance to examine the alternative, that would be great:
https://github.com/apache/groovy/pull/1324

paulk-asert:
So, after refactoring, something like this:

PR #1328, opened 2020-07-29
GROOVY-7232: prefer methodMissing over declared method on next candidate

eric-milles:
Here is my discussed alternative that simplifies the implementation using `InvokerHelper#invokeMethod`.  If a delegate or owner implements `methodMissing` it will be called instead of trying `invokeMethod` on the next in line.  This is more consistent with the way property resolution works (see `Closure#getProperty`).

`StaticTypeCheckingVisitor` will need a block similar to the one added for GROOVY-7996 in order for static compilation to line up.

https://issues.apache.org/jira/browse/GROOVY-7232

**Note:** This does not address the change in behavior discussed here: https://github.com/apache/groovy/commit/640e9304fe8950400d94adaad474c0c341209df0#commitcomment-40953759

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/metaclass/ClosureMetaClass.java
+++ b/src/main/java/org/codehaus/groovy/runtime/metaclass/ClosureMetaClass.java
@@ -231,6 +231,7 @@ public final class ClosureMetaClass extends MetaClassImpl {
         }
     }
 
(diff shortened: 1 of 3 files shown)


paulk-asert:
Merged, thanks. Agree we need to look further at static compilation cases.

PR #1329, opened 2020-07-29
GROOVY-9637: Improve the performance of GString (closes #1329)

paulk-asert:
Co-authored-by: Daniel Sun sunlan@apache.org
Co-authored-by: Jochen Theodorou <blackdrag@gmx.org>

### DIFF
--- a/src/main/java/groovy/lang/GString.java
+++ b/src/main/java/groovy/lang/GString.java
@@ -20,6 +20,7 @@ package groovy.lang;
 
 import org.apache.groovy.io.StringBuilderWriter;
 import org.codehaus.groovy.runtime.GStringImpl;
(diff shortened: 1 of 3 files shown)


danielsun1106:
Merged. Thanks!

blackdrag:
I think we got a good solution in the end

danielsun1106:
@blackdrag 
Yep ;-) About only 290ms costs in the simple benchmark on machine when the PR applied.

```groovy
def gstr = makeGString()
	gstr.toString()
}
println "${e - b}ms"

//@groovy.transform.CompileStatic
def makeGString() {
    def now = java.time.LocalDateTime.now()
    "integer: ${1}, double: ${1.2d}, string: ${'x'}, class: ${Map.class}, boolean: ${true}, now: ${now}"
}
```

PR #1330, opened 2020-07-30
The fork-join pool does not provide a good ClassLoader for Groovy types

eric-milles:
- `Thread.currentThread().getContextClassLoader()` was used starting with 46b1fface0b17cf41f9681894bcb8d78fc831258 -- commit does not state that this parentage is required for any specific reason; fallback loader (the one that loaded `ProcessingUnit`) seems like a better option for Groovy.

 - `StaticTypeCheckingSupport.evaluateExpression(...)` relies on loader created by this block; it does not transfer loader(s) from it's context.

This came about for me because IDE is performing content assist asynchronously now and processing `ClosureParams`/`DelegatesTo` metadata relies on `evaluateExpression`.

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/ProcessingUnit.java
+++ b/src/main/java/org/codehaus/groovy/control/ProcessingUnit.java
@@ -101,8 +101,7 @@ public abstract class ProcessingUnit {
         // ClassLoaders should only be created inside a doPrivileged block in case
         // this method is invoked by code that does not have security permissions.
         this.classLoader = loader != null ? loader : AccessController.doPrivileged((PrivilegedAction<GroovyClassLoader>) () -> {
-            ClassLoader parent = Thread.currentThread().getContextClassLoader();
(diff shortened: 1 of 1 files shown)


blackdrag:
@eric-milles what about first getting the context loader and check if it can load ProcessingUnit and if the ProcessingUnit loaded by that is the same as the presented ProcessingLoader class, and only fall back to ProcessingUnit.class.classLoader if that fails?

eric-milles:
I just went with the simplest alternative.  If there is a specific reason that the context class loader is needed, I'm interested to know why.

blackdrag:
@eric-milles that was just to keep backwards compatibility. It is difficult to find a reason for something somebody else wrote 15 years ago and did not create a ticket for, nor is it easy to tell if that "feature" is used by someone

danielsun1106:
Merged with some trivial tweaks for better backward compatibility. Thanks!

PR #1332, opened 2020-07-31
GROOVY-9666: ConcurrentModificationException when processing ModuleNo…

paulk-asert:
…de imports

backwards compatibility trumps potential increased use of immutability

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ModuleNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ModuleNode.java
@@ -91,24 +91,39 @@ public class ModuleNode extends ASTNode implements Opcodes {
         return /*Collections.unmodifiableList(*/classes/*)*/; // modified by MacroClassTransform
     }
 
+    /**
+     * @return the module's methods
+     */
     public List<MethodNode> getMethods() {
-        return Collections.unmodifiableList(methods);
+        return methods;
     }
 
+    /**
+     * @return a copy of the module's imports
+     */
     public List<ImportNode> getImports() {
-        return Collections.unmodifiableList(imports);
+        return new ArrayList<>(imports);
     }
 
+    /**
+     * @return the module's star imports
+     */
     public List<ImportNode> getStarImports() {
-        return Collections.unmodifiableList(starImports);
+        return starImports;
     }
 
+    /**
+     * @return the module's static imports
+     */
     public Map<String, ImportNode> getStaticImports() {
-        return Collections.unmodifiableMap(staticImports);
+        return staticImports;
     }
 
+    /**
+     * @return the module's static star imports
+     */
     public Map<String, ImportNode> getStaticStarImports() {
-        return Collections.unmodifiableMap(staticStarImports);
+        return staticStarImports;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


eric-milles:
Could `ModuleNode#addImport` do copy-on-write instead?

eric-milles:
It is only the plain imports that have this issue because they were previously managed by a map and returned as a list.  The other types of imports were returning the internal representation (list for list, map for map).

paulk-asert:
The following all fail for me in 3.0.5:
```
import org.codehaus.groovy.ast.ModuleNode
import org.codehaus.groovy.control.SourceUnit
import org.codehaus.groovy.ast.ClassHelper

def mn = new ModuleNode((SourceUnit)null)

mn.addStarImport('foo.bar')
assert mn.starImports.size() == 1
for (starImport in mn.starImports) {
    mn.addStarImport(starImport.packageName.toUpperCase())
}
assert mn.starImports.size() == 2

mn.addStaticStarImport('Integer', ClassHelper.Integer_TYPE)
mn.addStaticStarImport('Long', ClassHelper.Integer_TYPE)
assert mn.staticStarImports.size() == 2
for (staticStarImport in mn.staticStarImports) {
  mn.addStaticStarImport('Short', ClassHelper.Long_TYPE)
}
assert mn.staticStarImports.size() == 4

mn.addStaticImport(ClassHelper.Integer_TYPE, 'MIN_VALUE', 'MIN_VALUE')
mn.addStaticImport(ClassHelper.Long_TYPE, 'MIN_VALUE', 'MIN_VALUE')
assert mn.staticImports.size() == 2
for (staticImport in mn.staticImports) {
  mn.addStaticImport(staticImport.value.type, 'MAX_VALUE', 'MAX_VALUE')
}
assert mn.staticImports.size() == 4

mn.addImport('Integer', ClassHelper.Integer_TYPE)
assert mn.imports.size() == 1
for (importNode in mn.imports) {
  mn.addImport('Natural', ClassHelper.Integer_TYPE)
}
assert mn.imports.size() == 2

mn.addMethod(null)
assert mn.methods.size() == 1
for (method in mn.methods) {
  mn.addMethod(null)
}
assert mn.methods.size() == 2
```

eric-milles:
> even returning a copy is a breaking change

That's my take.  Since there is little chance of avoiding a breaking change, why not stick with returning immutables and let Vert.x adjust to making a copy to iterate over or performing changes after iteration is completed.  I don't think the design was to return something for modification; that's exposing the implementation of `ModuleNode`.  The emergent behavior that `getImports()` could be iterated over and `addImport(...)` could be called was not intended IMO.

Worst case, could 3.x get the change just for `getImports()` and master be left as is?

paulk-asert:
I will put it back the way it was and add the test cases from above which work. We have had many long discussions about changing the AST data structures to be more immutable (or even more consistently mutable!) but it isn't something which is easy to half do. We can certainly re-look for 4 but it will be some work.

danielsun1106:
Merged. Thanks.

paulk-asert:
> > even returning a copy is a breaking change
> 
> That's my take. Since there is little chance of avoiding a breaking change, why not stick with returning immutables and let Vert.x adjust to making a copy to iterate over or performing changes after iteration is completed. I don't think the design was to return something for modification; that's exposing the implementation of `ModuleNode`. The emergent behavior that `getImports()` could be iterated over and `addImport(...)` could be called was not intended IMO.
> 
> Worst case, could 3.x get the change just for `getImports()` and master be left as is?

@eric-milles As you will have seen, I reverted everything for now. It would be good to make these all consistent in Groovy 4 but we should provide a consistent approach across all AST nodes and a considered approach for mutating and have any breaking changes (if needed) documented in the release notes.

PR #1339, opened 2020-08-06
Support coerce for @NamedVariant

moonfruit:
Support `useAs` for `@NamedVariant`

For example:

```groovy
@NamedVariant
void run(int number) {
    println(number)
}
```

If we called this method with

```groovy
run(number: "123")
```

Groovy will throw out

```
groovy.lang.MissingMethodException: No signature of method: run() is applicable for argument types: (String) values: [123]
```

After support `useAs`, we can define method like this:

```groovy
@NamedVariant(useAs = true)
void run(int number) {
    println(number)
}
```

The generated method will automatically use the operator `as` convert `String` to `int`, and the will succeed now.

And in order to make the operator `as` work correctly, I also disable `@CompileStatic` for the generated method.

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
+++ b/src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
@@ -55,6 +55,7 @@ import static org.codehaus.groovy.ast.ClassHelper.STRING_TYPE;
 import static org.codehaus.groovy.ast.ClassHelper.make;
 import static org.codehaus.groovy.ast.ClassHelper.makeWithoutCaching;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.args;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.asX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.boolX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.callThisX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.callX;
@@ -95,6 +96,7 @@ public class NamedVariantASTTransformation extends AbstractASTTransformation {
         }
 
         boolean autoDelegate = memberHasValue(anno, "autoDelegate", true);
+        boolean coerce = memberHasValue(anno, "coerce", true);
         Parameter mapParam = param(GenericsUtils.nonGeneric(MAP_TYPE), "__namedArgs");
         List<Parameter> genParams = new ArrayList<>();
         genParams.add(mapParam);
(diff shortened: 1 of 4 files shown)


moonfruit:
Seems fine to me, @eric-milles.

eric-milles:
Why is the generated method set to skip type checking?  Was it producing STC errors?

moonfruit:
When define method in a class like this:

```groovy
@CompileStatic
class App {
    @NamedVariant(coerce = true)
    void run(int number) {
        println(number)
    }
}
```

The generated method will also be compiled statically. So when I call the method:

```groovy
run(number: "123")
```

It will call `DefaultGroovyMethods.asType()` instead of `StringGroovyMethods.asType()`, so the string cannot convert to integer properly. So I must skip `@CompileStatic`.
In order to achieve this goal, I skip type checking explicitly.

moonfruit:
Sorry, I don't know what are the STC errors.

src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
```diff
@@ -238,23 +244,37 @@ private void createMapVariant(final MethodNode mNode, final AnnotationNode anno,
         if (mNode instanceof ConstructorNode) {
             body.addStatement(stmt(ctorX(ClassNode.THIS, args)));
             body.addStatement(inner);
-            addGeneratedConstructor(cNode,
+            skipTypeChecked(addGeneratedConstructor(cNode,
```
danielsun1106, line 247: Why should we `skipTypeChecked` here?

src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
```diff
@@ -238,23 +244,37 @@ private void createMapVariant(final MethodNode mNode, final AnnotationNode anno,
...
                     genParamsArray,
                     mNode.getExceptions(),
                     body
-            );
+            ));
         } else {
             body.addStatement(inner);
             body.addStatement(stmt(callThisX(mNode.getName(), args)));
-            addGeneratedMethod(cNode,
+            skipTypeChecked(addGeneratedMethod(cNode,
```
danielsun1106, line 256: Why should we `skipTypeChecked` here?

danielsun1106:
The relevant test is expected for verifying the changes.

eric-milles:
Instead of calling the `asType` method, you can wrap the expression in a `CastExpression` with coerce set to true.  The compiler will take care of the rest.  `GeneralUtils#castX` will create the `CastExpression` and you can then call `setCoerce(true)` or you could go the extra mile and add an `asX` factory method to GeneralUtils that performs these steps.

moonfruit:
OK, I use `as` expression instead of directly calling `asType()` method.

Add I add a test for this.

src/test/groovy/transform/NamedVariantTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+        def tester = new GroovyClassLoader().parseClass(
+                '''@groovy.transform.CompileStatic
+                  |class MyClass {
+                  |    @groovy.transform.NamedVariant(coerce = true)
+                  |    void run(int number) {
+                  |    }
+                  |}'''.stripMargin()).getConstructor().newInstance()
+
+        tester.run(number: 123)
+        tester.run(number: "123")
```
danielsun1106, line 81: It's better to use `assert` statements here.
Given `run` method return `number` directly:
```
void run(int number) { number }
```
We could tweak the test as follows:
`assert 123 == tester.run(number: 123)`
`assert 123 == tester.run(number: "123")`

src/test/groovy/transform/NamedVariantTest.groovy
```diff
@@ -0,0 +1,83 @@
...
+    void testCoerceMethodCall() {
+        def tester = new GroovyClassLoader().parseClass(
+                '''class MyClass {
+                  |    @groovy.transform.NamedVariant(coerce = true)
+                  |    void run(int number) {
+                  |    }
+                  |}'''.stripMargin()).getConstructor().newInstance()
+
+        tester.run(number: 123)
+        tester.run(number: "123")
```
danielsun1106, line 68: ditto: https://github.com/apache/groovy/pull/1339/files#r467682355

moonfruit:
Already add assert to the test, @danielsun1106.

danielsun1106:
LGTM

danielsun1106:
Merged. Thanks!

moonfruit:
Glad to hear that!

PR #1342, opened 2020-08-07
GROOVY-9675: There should be a mechanism to disable individual macro …

paulk-asert:
…methods & extension methods

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/metaclass/MetaClassRegistryImpl.java
+++ b/src/main/java/org/codehaus/groovy/runtime/metaclass/MetaClassRegistryImpl.java
@@ -47,12 +47,15 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Modifier;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
 /**
  * A registry of MetaClass instances which caches introspection and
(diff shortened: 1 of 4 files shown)


paulk-asert:
Since checking individual methods could have a performance impact, I have used a single system property to turn on the disabling feature. E.g.
```
-Dgroovy.macro.disable=true    // turn on globally
-Dgroovy.macro.disable.foo=true  // disable the 'foo' macro (any macro with that name)

-Dgroovy.extension.disable=true    // turn on globally
-Dgroovy.extension.disable.dump=true  // disable the 'dump' extension method (any with that name)

```

eric-milles:
Is there ever a reason to set one of these properties to true?  Instead how about one property that is set the the name(s) of the disabled macros or extensions?  Like, `-Dgroovy.macro.disable=*` or `-Dgroovy.macro.disable=foo,bar` or a pattern that could be compiled once and matched against the names.

Will it be possible to disable macros or extensions using compiler configuration?  Or is this just a system property mechanism?  IMO there are too many special system property controls/switches.

If I want to disable just one instance of `each` or `getAt` (for example) is that possible or must it be all or none?

paulk-asert:
We could probably do the `-Dgroovy.macro.disable=foo,bar` variant. We want something that is going to be fast, so I'd avoid Patterns.

A compiler configuration option isn't enough - it doesn't cover dynamic Groovy and I think it might be a little messy trying to block off access to extensions/macros once they are in our caches. Good news though for @CompileStatic code is that you only need the switch during compilation.

We could get fancier trying to disable particular variants of extensions/macros but I felt YAGNI calling. I imagine disabling would mostly be for folks already using methods with those names when new macros are introduced. They'd just want to block all variants and perhaps consider refactoring and removing the disable switch later.

paulk-asert:
@eric-milles adjusted as per your suggestion

danielsun1106:
Merged. Thanks.

PR #1343, opened 2020-08-07
GROOVY-9677: provide a library of pre-canned macro methods

paulk-asert:


eric-milles:
Could it be named simply "groovy-macros"?

paulk-asert:
could be but I think that is too close to `groovy-macro`

danielsun1106:
Merged. Thanks.

PR #1345, opened 2020-08-10
GROOVY-9680: Groovy 4 should provide some pre-canned type checker ext…

paulk-asert:
…ensions


settings.gradle
```diff
@@ -32,6 +32,7 @@ gradleEnterprise {
 def subprojects = ['groovy-ant',
         'groovy-astbuilder',
         'groovy-bsf',
+        'groovy-checkers',
```
danielsun1106, line 35: `groovy-typecheckers` may be clearer?

paulk-asert: Yes, we could rename. We need something short for "type checking extensions". Perhaps 'typecheckers' is better. I was trying to avoid people second guessing should it be 'groovy-typeCheckers' or 'groovy-type-checkers' but I guess it is just something they will soon learn.

paulk-asert: @danielsun1106 adjusted as per your suggestion

danielsun1106:
+1

paulk-asert:
I have a few more changes in mind. I'll amend here before committing.

PR #1348, opened 2020-08-18
GROOVY-9692: "call" method call for non-string expression w/ closure arg

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9692


eric-milles:
> Are there any parsing tests that needs to be updated?

There are tests for the parser.  I'm not sure exactly where to add this in.  Probably there is a test for "1 { -> ... }" and there could be an addition of "[] { -> ... }" for your case.

MykolaGolubyev:
@eric-milles do you want me to add the test?

MykolaGolubyev:
oh nice!

PR #1356, opened 2020-08-27
GROOVY-9702: make ASTTest value optional (from a JDK bytecode perspec…

paulk-asert:
…tive)

### DIFF
--- a/src/main/java/groovy/transform/ASTTest.java
+++ b/src/main/java/groovy/transform/ASTTest.java
@@ -18,6 +18,7 @@
  */
 package groovy.transform;
 
+import groovy.lang.Closure;
 import org.codehaus.groovy.control.CompilePhase;
 import org.codehaus.groovy.transform.GroovyASTTransformationClass;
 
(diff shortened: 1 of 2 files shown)


paulk-asert:
@eric-milles To avoid the breaking change on 3_0_X, would something like this keep the IDE happy.

eric-milles:
@paulk-asert What is the breaking change?  I'm doubtful anyone relies on runtime retention since the implementation is to read the closure literal's source text and execute that as a script.  You can't even add the arrow to the closure expression without it failing.  There is nothing about the compiled result that is used by ASTTestTransformation.  Only if someone used the ASTTest annotation afterwards for some other purpose would this be a breaking change.

It is okay not to merge back into earlier Groovy versions if you are at all concerned.  This particular change would stop the IDE from giving a warning if you fail to supply a closure, so I don't think it is desirable.

paulk-asert:
@eric-milles Okay, I'll work on an alternative. Just setting the value to a "nulled out" closure expression as per your PR now merged on the master branch isn't good enough if we leave the retention as RUNTIME. And agreed its a very minor breaking change - so I will document it as such in the release notes.

PR #1358, opened 2020-08-31
GROOVY-9704: load minus one as long integer

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9704


eric-milles:
*Q1:* SC adds support for unary +, - and ~ for the `char` type.  Under STC, these all throw MissingMethodException, which is consistent with `Character`.  Should SC have `char` removed from the 3 unary operator optimizations for consistency?

*Q2:* For the unsupported operation cases (~ on floating-point types, etc.) should STC be producing some error?

paulk-asert:
Merged, thanks. @eric-milles I'll split out your other questions into there own issues if needed. I'll investigate more shortly.

PR #1359, opened 2020-08-31
GROOVY-9635: resolve "V" in T=V; "T" may have been reused as type param

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9635

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -1624,43 +1624,36 @@ public abstract class StaticTypeCheckingSupport {
     static void applyGenericsConnections(final Map<GenericsTypeName, GenericsType> connections, final Map<GenericsTypeName, GenericsType> resolvedPlaceholders) {
         if (connections == null) return;
         int count = 0;
-        while (count < 10000) {
-            count += 1;
-            boolean checkForMorePlaceHolders = false;
+        while (count++ < 10000) {
+            boolean checkForMorePlaceholders = false;
             for (Map.Entry<GenericsTypeName, GenericsType> entry : resolvedPlaceholders.entrySet()) {
-                GenericsTypeName name = entry.getKey();
-                GenericsType replacement = connections.get(name);
-                if (replacement == null) {
-                    GenericsType value = entry.getValue();
-                    GenericsType newValue = applyGenericsContext(connections, value);
-                    entry.setValue(newValue);
-                    checkForMorePlaceHolders = checkForMorePlaceHolders || !equalIncludingGenerics(value, newValue);
-                    continue;
-                }
-                GenericsType original = entry.getValue();
-                if (!original.isWildcard() && !original.isPlaceholder()) {
-                    continue;
-                }
-                boolean placeholderReplacement = replacement.isPlaceholder();
-                if (placeholderReplacement) {
-                    GenericsType connectedType = resolvedPlaceholders.get(name);
-                    if (replacement == connectedType) continue;
-                }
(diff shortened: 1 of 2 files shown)


paulk-asert:
The following script fails with a `VerifyError` with this PR merged:
```
import java.util.function.Function
class C<R extends Number> {
    def <V> V m(Function<C, V> f) { // R from C is confused with R->V from Function
        V result = f.apply(this)
        return result
    }
}

def method() {
    def ret = new C().m(new Function<C, String>() {
        @Override
        String apply(C that) {
            return 'foo'
        }
    })
    assert ret == 'foo'
}

method()
```

eric-milles:
Is it because a Number type is never supplied for C?

eric-milles:
So the issue is somewhere in the anon. inner class created for `Function<C, String>`.  When I replace it with closure coercion, I don't get the error.  I don't know that this problem didn't exist before these changes.
```groovy
Function<C, String> func() {
  /*
  new Function<C, String>() {
    @Override
    String apply(C that) {
      return 'foo'
    }
  }
  */
  { that -> 'foo' } as Function<C, String>
}
```

Here is the error I get.  Any idea what "invariant" refers to?  Is this coming from the contracts module integration?
```
java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    ConsoleScript14$1.invariant_ConsoleScript14$1()V @55: invokevirtual
  Reason:
    Type 'ConsoleScript14' (current frame, stack[0]) is not assignable to 'java/lang/Boolean'
```

Any idea why I get this error when creating a full distribution and running within Groovy Console and not when running via GroovyShell through the unit tests?

eric-milles:
This is the java decompilation that Groovy Console is producing for the anonymous inner class:
```
@groovy.contracts.Contracted
public class script1599241127913$1 extends java.lang.Object implements java.util.function.Function<C, String>, groovy.lang.GroovyObject { 

    final script1599241127913 this$0 
    static final boolean $GCONTRACTS_ENABLED 
    private static org.codehaus.groovy.reflection.ClassInfo $staticClassInfo 
    public static transient boolean __$stMC 
    private transient groovy.lang.MetaClass metaClass 

    script1599241127913$1(script1599241127913 p0) {
        this$0 = p0 
        super()
        metaClass = /*BytecodeExpression*/
    }

    @java.lang.Override
    public java.lang.String apply(C that) {
        return 'foo'
    }

    ...

    protected void invariant_script1599241127913$1() {
        if ( $GCONTRACTS_ENABLED ) {
            try {
                if (org.apache.groovy.contracts.generation.ContractExecutionTracker.track('script1599241127913$1', 'void invariant_script1599241127913$1()', 'invariant', false)) {
                    java.lang.Boolean $_gc_result = false
                    org.apache.groovy.contracts.ViolationTracker.init()
                    $_gc_result = true
                    if (!($_gc_result.booleanValue())) {
                        if (org.apache.groovy.contracts.ViolationTracker.violationsOccurred()) {
                            try {
                                org.apache.groovy.contracts.ViolationTracker.rethrowFirst()
                            } 
                            finally { 
                                org.apache.groovy.contracts.ViolationTracker.deinit()
                            } 
                        }
                    }
                }
            } 
            finally { 
                org.apache.groovy.contracts.generation.ContractExecutionTracker.clear('script1599241127913$1', 'void invariant_script1599241127913$1()', 'invariant', false)
            } 
        }
    }

    ...

    static { 
        $GCONTRACTS_ENABLED = org.apache.groovy.contracts.generation.Configurator.checkAssertionsEnabled('script1599241127913$1')
    }

    @groovy.transform.Generated
    public volatile java.lang.Object apply(java.lang.Object param0) {
    }

}
```

paulk-asert:
Found it. There was a misnamed metadata key in groovy-contracts. I am surprised more things didn't break, but fixed now in any case.

paulk-asert:
Merged, thanks!

PR #1361, opened 2020-09-08
GROOVY-9732: SC: re-implement transformExpression and several setters in CompareToNullExpression

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9732

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/sc/transformers/CompareToNullExpression.java
+++ b/src/main/java/org/codehaus/groovy/transform/sc/transformers/CompareToNullExpression.java
@@ -19,7 +19,6 @@
 package org.codehaus.groovy.transform.sc.transformers;
 
 import org.codehaus.groovy.ast.ClassHelper;
-import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.GroovyCodeVisitor;
 import org.codehaus.groovy.ast.expr.BinaryExpression;
 import org.codehaus.groovy.ast.expr.ConstantExpression;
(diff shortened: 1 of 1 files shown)


src/main/java/org/codehaus/groovy/transform/sc/transformers/CompareToNullExpression.java
```diff
@@ -31,25 +30,47 @@
...
-import org.objectweb.asm.Opcodes;
 
-public class CompareToNullExpression extends BinaryExpression implements Opcodes {
+import static org.objectweb.asm.Opcodes.GOTO;
+import static org.objectweb.asm.Opcodes.ICONST_0;
+import static org.objectweb.asm.Opcodes.ICONST_1;
+import static org.objectweb.asm.Opcodes.IFNONNULL;
+import static org.objectweb.asm.Opcodes.IFNULL;
+
+public class CompareToNullExpression extends BinaryExpression {
```
paulk-asert, line 40: Removing interface `Opcodes` is a binary breaking change. I agree it is ugly to leak that implementation class out through our API and acknowledge that style of usage dates back to before Java had static imports. Never-the-less, I think we should make such a change as a separate issue and target just Groovy 4. I'll create such an issue and merge this one without that change.

paulk-asert:
Merged, thanks!

src/main/java/org/codehaus/groovy/transform/sc/transformers/CompareToNullExpression.java
paulk-asert: See GROOVY-9736 for proposed separate `Opcodes` change.

PR #1362, opened 2020-09-08
GROOVY-9712: type name may start with more than A-Z (CapitalizedIdentifier)

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9712

**Note:** The parser tests that use `TestUtils#doTest` are silently swallowing all parser errors since there is no longer comparison against any antlr2 result.

### DIFF
--- a/src/antlr/GroovyLexer.g4
+++ b/src/antlr/GroovyLexer.g4
@@ -882,7 +882,7 @@ ELVIS_ASSIGN    : '?=';
 
 // §3.8 Identifiers (must appear after all keywords in the grammar)
 CapitalizedIdentifier
-    :   [A-Z] JavaLetterOrDigit*
+    :   JavaLetter {Character.isUpperCase(getInputStream().LA(-1))}? JavaLetterOrDigit*
     ;
 
 Identifier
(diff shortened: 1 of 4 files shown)


eric-milles:
From the issue:
>Another way of looking at this is that field/property uses the generalized-typename rule and a method's return type uses standard-typename rule.  The later requires a capital letter (A-Z) as the first name character.  The former has allowances for lowercase names.  I'm not sure why this distinction exists in the parser.
```
generalClassOrInterfaceType
options { baseContext = classOrInterfaceType; }
    :   qualifiedClassName typeArguments?
    ;

standardClassOrInterfaceType
options { baseContext = classOrInterfaceType; }
    :   qualifiedStandardClassName typeArguments?
    ;
```

src/antlr/GroovyLexer.g4
```diff
@@ -882,7 +882,7 @@ ELVIS_ASSIGN    : '?=';
 
 // §3.8 Identifiers (must appear after all keywords in the grammar)
 CapitalizedIdentifier
-    :   [A-Z] JavaLetterOrDigit*
+    :   JavaLetter {Character.isUpperCase(getInputStream().LA(-1))}? JavaLetterOrDigit*
```
danielsun1106, line 885: `getInputStream().LA(-1)` could be simplified as  `_input.LA(-1)`

paulk-asert: LGTM. I'll merge and include this change.

subprojects/parser-antlr4/src/test/groovy/org/apache/groovy/parser/antlr4/TestUtils.groovy
```diff
@@ -87,17 +81,15 @@ final class TestUtils {
         return [newAST, null]
     }
 
-    static void shouldFail(String path, boolean toCheckNewParserOnly = false) {
```
paulk-asert, line 90: I'll add back in the `(String, boolean)` variant for binary compatibility

PR #1366, opened 2020-09-11
Character filter for MarkupBuilder

tosts:
Could this work?

### DIFF
--- a/subprojects/groovy-xml/build.gradle
+++ b/subprojects/groovy-xml/build.gradle
@@ -21,6 +21,13 @@ dependencies {
     testImplementation rootProject.sourceSets.test.runtimeClasspath
     testImplementation "xmlunit:xmlunit:$xmlunitVersion"
     testImplementation project(':groovy-test')
+    testImplementation ("org.spockframework:spock-core:$spockVersion") {
+        exclude group: 'org.codehaus.groovy'
+    }
+}
+
+test {
+    useJUnitPlatform()
 }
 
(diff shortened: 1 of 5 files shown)


tosts:
Also: Not sure what I should do about a timeout in the sonar checks. Does that have something to do with me?

paulk-asert:
Parts of this look good but we can possibly re-work to increase the likelihood of being able to extend to `StreamingMarkupBuilder`. I'll take a look.

tosts:
Like so?

paulk-asert:
It's been a while since I delved into XML encoding rules. We take a slightly different approach for surrogates (potentially writing them as hex strings) in StreamingMarkupWriter:
https://github.com/apache/groovy/blob/master/subprojects/groovy-xml/src/main/java/groovy/xml/streamingmarkupsupport/StreamingMarkupWriter.java#L141-L170
Do you see anything wrong with that approach?

paulk-asert:
Merged with a few changes. Thanks!

tosts:
Great! \o/

@Surrogates: No, as a default for writing HTML that's probably even desirable.
They just live in the intenional limbo of no MUST in XML, but not discouraged either.

Ha, and you caught me. I copy-pasted a bit too much on my explanatory comments.
Let met fix that real quick...

tosts:
Over here: https://github.com/apache/groovy/pull/1372
Sorry for the extra round-trip.

PR #1368, opened 2020-09-14
GROOVY-9739: prevent typecast expression with null type

eric-milles:
This is a proposed change to help identify the source of the null cast.

https://issues.apache.org/jira/browse/GROOVY-9739


paulk-asert:
Merged - just into master. Thanks!

PR #1373, opened 2020-09-21
GROOVY-9751, GROOVY-9463: STC: add type checking of method pointer expressions

eric-milles:
Previously, the inferred type of `a.&b` was simply `Closure`.  This was a problem for return type inferencing of expressions like `list.collect(service.&transform)` (see GROOVY-9751).  Also, once the reference method(s) can be resolved, it was not much of a leap to add STC error for unknown condition (see GROOVY-9463).

https://issues.apache.org/jira/browse/GROOVY-9751
https://issues.apache.org/jira/browse/GROOVY-9463


paulk-asert:
Proposed PR merged, thanks!

PR #1375, opened 2020-09-22
GROOVY-9754: Provide a record-like equivalent

paulk-asert:


eric-milles:
How close is this to `@Immutable` + `@POJO`?  Is there a future path where a `record` keyword maps to this like `trait` to `@Trait`?

paulk-asert:
Records are indeed somewhat similar to @Immutable + @POJO. There are different getter naming conventions, special serialization rules and some other restrictions on records. The goal here would eventually be to create native records with Groovy syntax, so yes a `record` keyword would be part of that but I don't think that aspect needs to be sorted before alpha-1 and can be done later like `trait` was added later for `@Trait`. The AST transform I think is useful enough in it's own right and would allow record-like classes to be created in the meantime until we consider native record support. We just need to not back ourselves into a corner that would disallow native records as a second step.

PR #1379, opened 2020-09-25
Add a muse configuration

TomMD:
Groovy is going to be part of the ApacheCon bug bash. As part of that, ASF Infra will be installing the github app "muse-dev" in this repository. While there will certainly be a list of issues provided, scores can be kept track of automated via the bot. Beyond ApacheCon, the muse app is much like other CI serves in that there are actions taking on PRs and those actions need some minor amount of configuration.

In short: This PR adds a muse configuration and file, sufficient to allow running Muse CI on groovy.


paulk-asert:
Merged, thanks!

PR #1380, opened 2020-09-25
GROOVY-9759: Regression for accessing private static constants in clo…

paulk-asert:
…sures from Map.
Seems we have subsequent code not expecting this enriched information.


eric-milles:
Wasn't this test case already fixed by GROOVY-9699?  I do not see any STC error for Groovy 3.0.6.SNAPSHOT.

paulk-asert:
You may be right - I'll look into it. Perhaps I was on an old branch at the time.

paulk-asert:
You are correct. I must have been on an older branch when I was running my tests. I was sure I had test on master and was basing my assumptions on that, but it works fine testing just now.

PR #1382, opened 2020-09-29
Make the Gradle build more idiomatic

melix:
This is a PR which refactors the build in order to:

- cleanup bad practices
- introduce configuration avoidance
- remove deprecations
- add support for the `maven-publish` plugin instead of the old one

Initially I only wanted to focus on the last point but I realized there's a LOT of hackery in the build which makes it complicated to migrate without doing the cleanup first.


melix:
@paulk-asert @danielsun1106 I'm setting this for review despite CI failing, I'm not 100% sure why the `dist` command on GH action fails, maybe it's just too late.

As you can see this PR ended up being much larger than I initially wanted it to be. In order to migrate to the `maven-publish` plugin, I ended up refactoring the build to make it cleaner, more idiomatic. Work is not complete but I could spend weeks on that.

The new build is, IMO, much easier to understand. I tried to remove bad practices I found, and I tested that I could publish on a local artifactory server with signing. The new publications are richer and use Gradle Module Metadata.

Also I took the liberty to remove the "backports" modules. Reintroducing wouldn't be too hard, but AFAIK they haven't really been used, and now there's only a 2.3 backport, and we're talking about Groovy 4.

I'm going to continue a bit on this but I'd appreciate that you play a bit with this branch to find out if they are obvious problems: during the refactoring it's possible I've slightly changed the outputs.

melix:
PR looks fine from my POV. Let me know how we can make progress on merging this.

paulk-asert:
I will look at it soon. Might be a few days. Nice work! Thanks so much.

buildSrc/src/main/groovy/org/apache/groovy/gradle/GroovyLibraryExtension.groovy
```diff
@@ -64,16 +66,18 @@ class GroovyLibraryExtension {
...
                            ConfigurationContainer configurations,
-                           TaskContainer tasks,
-                           ProjectLayout layout
+                           TaskContainer tasks
     ) {
         this.objects = factory
+        this.layout = layout
         this.sharedConfiguration = sharedConfiguration
         this.includeInGroovyAll = factory.property(Boolean).convention(true)
         this.grooid = factory.property(Boolean).convention(false)
```
wolfs, line 83: I think you should get away with declaring the `grooid` property abstract and then calling the abstract getter in the constructor setting it to a convention. This would allow you to get rid of using the `ObjectFactory`, if you wouldn't need it to instantiate the attributes.

So something like this should work:

```
class GroovyLibraryExtension {
    abstract Property<Boolean> getGrooid()

    GroovyLibraryExtension(...) {
        grooid.convention(false)
    }
}
```

melix: I could but given that I need to inject other properties in any case, I prefer keeping the things consistent and not have a mix of abstract properties and injected properties

paulk-asert:
Merged, Thanks!! Still a few glitches to unravel but with master continually updating, I think merging and some post amendments will be quickest path to get this in.

PR #1391, opened 2020-10-01
Fix possible null dereference in EnumCompletionVistor.java

nescohen:


src/main/java/org/codehaus/groovy/classgen/EnumCompletionVisitor.java
```diff
@@ -83,7 +83,13 @@ private void completeEnum(ClassNode enumClass) {
     private static void addImplicitConstructors(ClassNode enumClass, boolean aic) {
         if (aic) {
             ClassNode sn = enumClass.getSuperClass();
-            List<ConstructorNode> sctors = new ArrayList<ConstructorNode>(sn.getDeclaredConstructors());
```
eric-milles, line 86: How is it that this could return a null?  I don't see any of the 3 implementation ever returning null.

nescohen: I was looking specifically at this function:
https://github.com/nescohen/groovy/blob/master/src/main/java/org/codehaus/groovy/ast/ClassNode.java#L1014

I made the assumption that it was possible for the return value to be null specifically because of the null check on line 1014. It is possible this was a mistake however. I did not look at/consider the actual call to `redirect().getUnresolvedSuperClass();`. What do you think?

eric-milles: I had a bit of confusion.  `getDeclaredConstructors()` is what I was referring to.  `getSuperClass()` may return null for an interface but should return `java.lang.Enum` for an enum class.  If that isn't the case, there is a problem upstream where `enumClass` is created.

nescohen: So I guess the question would be, would you prefer to have a NPE thrown vs en empty constructor list OR I could even refactor to throw an exception of the semantically correct type. Would you know of one to recommend? I guess I just like to err on the side of never throwing NPEs, but I know that sometimes it is preferable to fail fast. 
Thanks!

paulk-asert: For an enum class, I see no way `null` can arise. If we were trying to make the code ultra robust I'd add:
```
if (sc == null || sc.getDeclaredConstructors() == null) {
    throw new GroovyBugError("enum had null superclass or constructors weren't initialized correctly");
}
```
But I think that is overkill here. I'd prefer over time to start using @NonNull, e.g. @NonNull on `getDeclaredConstructors` would allow the second term in the above to be dropped and so on. Currently the "NonNullness" is deduced by hand but I believe is currently the case. Also, there have no doubt been millions of enums compiled by Groovy and no-one has spotted an NPE in this area of the code to date.
I'd recommend closing. But it's a good exercise to check our assumptions in such places and see if we can't make implicit assumptions more explicit somehow.

paulk-asert:
No further feedback, I'll close this.

PR #1393, opened 2020-10-02
Resource leak fix (#2)

smagill:
Fix a resource leak warning.  While CharBuf.close is currently a no-op, CharBuf does inherit from Writer, which has the contract that it should be closed after use.  This protects against the case where CharBuf is later changed to involve resources that should be freed.

### DIFF
--- a/.muse/codenarc.sh
+++ b/.muse/codenarc.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+commit=$2
+cmd=$3
+
+function version() {
+    echo 1
+}
+
+function applicable() {
+    echo "true"
+}
+
+function gettool() {
+  pushd /tmp >/dev/null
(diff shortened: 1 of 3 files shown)


paulk-asert:
Nice detective work. I have two concerns though. Firstly, I think this is very performance sensitive code in terms of JSON parsing. The wrapping with try might totally bog down parsing. I don't know for sure though and we'd need to run a little benchmark analysis to better guide our decisions. Secondly, returning `null` isn't likely to play nicely with existing code that users this. I think we'd just fail fast and wrap the IOException in a RuntimeException and throw that.
As you already spotted, CharBuf.close is currently a no-op, and the impacted classes aren't likely to be changed much, so all the above seems like a bit of an overkill reaction to the issue.
My recommendation would be just to add a comment saying something like "consider wrapping in a try with resources block if CharBuf is ever refactored to have a non-empty close()". We could also provide an appropriate hush for the resource leak warning.

paulk-asert:
I merged the .muse config settings and added the comment around the currently false positive warning. Thanks!

PR #1394, opened 2020-10-03
GROOVY-9769: avoid unnecessary creation of UnionTypeClassNode

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9769

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
@@ -2265,26 +2265,27 @@ public class StaticTypeCheckingVisitor extends ClassCodeVisitorSupport {
         ).toArray(ClassNode[]::new);
     }
 
-    private ClassNode getInferredTypeFromTempInfo(final Expression exp, ClassNode result) {
-        if (exp instanceof VariableExpression && !typeCheckingContext.temporaryIfBranchTypeInformation.isEmpty()) {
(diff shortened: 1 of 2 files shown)


paulk-asert:
Merged, thanks!

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -2265,26 +2265,27 @@ protected MethodNode typeCheckMapConstructor(final ConstructorCallExpression cal
...
-                if (result != null && !classNodes.contains(result)) types.add(result);
-                types.addAll(classNodes);
-                // GROOVY-7333: filter out Object
-                types.removeIf(OBJECT_TYPE::equals);
+    private ClassNode getInferredTypeFromTempInfo(final Expression expression, final ClassNode expressionType) {
+        if (expression instanceof VariableExpression && !typeCheckingContext.temporaryIfBranchTypeInformation.isEmpty()) {
+            List<ClassNode> tempTypes = getTemporaryTypesForExpression(expression);
+            if (tempTypes != null && !tempTypes.isEmpty()) {
+                List<ClassNode> types = new ArrayList<>(tempTypes.size() + 1);
+                if (expressionType != null && !expressionType.equals(ClassHelper.OBJECT_TYPE) // GROOVY-7333
```
paulk-asert, line 2273: Removing `Object` is currently the correct behavior but I noticed last time that I played around with introducing Intersection types (as an addition to the current Union types) that we had to leave it in but fix in other places for things to work nicely.

PR #1395, opened 2020-10-04
GROOVY-8258: [GEP] Create a LINQ-like DSL

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-8258

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -49,6 +49,7 @@ def subprojects = [
         'groovy-jmx',
         'groovy-json',
         'groovy-jsr223',
+        'groovy-ginq',
         'groovy-macro',
         'groovy-macro-library',
         'groovy-nio',
--- a/src/main/java/org/codehaus/groovy/ast/tools/GeneralUtils.java
+++ b/src/main/java/org/codehaus/groovy/ast/tools/GeneralUtils.java
@@ -43,6 +43,7 @@ import org.codehaus.groovy.ast.expr.ConstructorCallExpression;
 import org.codehaus.groovy.ast.expr.DeclarationExpression;
 import org.codehaus.groovy.ast.expr.Expression;
 import org.codehaus.groovy.ast.expr.FieldExpression;
+import org.codehaus.groovy.ast.expr.LambdaExpression;
 import org.codehaus.groovy.ast.expr.ListExpression;
 import org.codehaus.groovy.ast.expr.MapEntryExpression;
 import org.codehaus.groovy.ast.expr.MapExpression;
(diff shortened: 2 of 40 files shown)


eric-milles:
Is this something that could start life as an external project?  What is the case for beginning as part of the language/runtime offerings from the start?  I'm sorry, I can't reach the JIRA ticket for a look -- what is the significant advantage over something like Hibernate or similar mature technologies?

Is LINQ a trademark of Microsoft's?  In any case, for searchability, I would suggest consideration of another name.

danielsun1106:
I believe most of us are familiar with SQL-like language, so it's good to reuse the knowledge to deal with collection for higher productivity and better readability. As you found, GORM and Hibernate are powerful enough to deal with DB population, so currently groovy-linq is just target to collection. Parsing XML, JSON and YAML could use groovy-linq too because they can be parsed into collections.

> Is this something that could start life as an external project?

No worries, the PR will not be merged util review passed. BTW, groovy-linq may be a new module of Groovy 4.0( http://groovy-lang.org/releasenotes/groovy-4.0.html ) /cc @paulk-asert 


> Is LINQ a trademark of Microsoft's? In any case, for searchability, I would suggest consideration of another name.

Any suggestion for the better name? GINQ?

eric-milles:
Is there any reason this new feature could not be part of the `groovy-sql` module under the `org.apache.groovy.sql.dsl` package?  Then whatever marketing name you choose would not impact the packaging and its association with SQL would be clear.

danielsun1106:
> Is there any reason this new feature could not be part of the groovy-sql module under the org.apache.groovy.sql.dsl package? Then whatever marketing name you choose would not impact the packaging and its association with SQL would be clear.

Good idea. but I am a bit afraid that `groovy-sql` module name will make users misunderstand GINQ supports populating DB too, actually I am going to only support populating collections(XML, JSON, YAML could be parsed into collections) at the beginning because GORM and hibernate are powerful enough as you found and supporting DB is too heavy for a languange.

paulk-asert (approved):
Hi Daniel, this is looking nice. There seems to be a little concern from some in the community about incorporating this into the core language straight away. If you think about the contribution as being made of two parts, front-end and back-end, I think it makes a lot of sense for the back-end to be in groovy core. That would be useful in normal streams style coding. For the front-end, I can see pros and cons for whether it should be in the core language or a separate project. Short term, I think it will be easiest to evolve both parts in parallel if they are together. Since we are targeting an alpha release, I think including the front end will be okay so long as we mark it @Incubating and make the groovy-ginq an optional module. How does that sound to you?

danielsun1106:
@paulk-asert Thanks a lot for your reviewing. I've tweaked the PR as you suggested.

paulk-asert:
Merged! Thanks for a great contribution.

PR #1397, opened 2020-10-06
GROOVY-9762: if not ClassExpression, use TypeChooser to find target type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9762


danielsun1106:
+1

PR #1398, opened 2020-10-06
GROOVY-9763: read helper method generics of trait method not found

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9763


paulk-asert:
Merged, thanks!

PR #1399, opened 2020-10-07
GROOVY-9771: visit receiver for private field access

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9771

partial restore of code removed for GROOVY-9699


paulk-asert:
Merged, thanks!

PR #1402, opened 2020-10-12
Fix missing roles in POM files

melix:
This commit fixes the roles which were missing from the POM files.


paulk-asert:
Merged, thanks!

PR #1403, opened 2020-10-12
GROOVY-9779: before missing method, try "call" method of property value

eric-milles:
- this change expands upon the check for Script binding variable "call"

https://issues.apache.org/jira/browse/GROOVY-9779


paulk-asert:
Merged, thanks!

PR #1404, opened 2020-10-14
GROOVY-7494: ArrayStoreException assigning GStringImpl to String[] wh…

paulk-asert:
…en using Indy


danielsun1106:
Merged. Thanks!

PR #1406, opened 2020-10-19
Minor bugfixes to the build

melix:
The platform definition (the BOM) was wrong


paulk-asert:
Merged, thanks! (Daniel had bumped the gradle version in parallel but I merged the BOM-related change)

PR #1407, opened 2020-10-21
GROOVY-9779: don't chain call properties together

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9779


danielsun1106:
Merged. Thanks!

PR #1408, opened 2020-10-21
GROOVY-9786: STC: propagate variable assignment tracking for if-else-if chaining

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9786


danielsun1106:
Merged. Thanks!

PR #1410, opened 2020-10-23
GROOVY-9791: SC: no dynamic property method calls for reading/writing protected field from subclass in different package

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9791


danielsun1106:
Merged. Thanks!

PR #1411, opened 2020-10-24
GROOVY-9093: SC: add compile-time error for inaccessible field or getter

eric-milles:
This handles the read case; the write case may still need a similar check.

https://issues.apache.org/jira/browse/GROOVY-9093


danielsun1106:
Merged. Thanks!

PR #1414, opened 2020-11-09
GROOVY-9803: translate generics of method reference using SAM parameters

eric-milles:
Since this change resolves additional generics, the test case for GROOVY-9762 fails when comparing `List<? super Integer>` to `List<Integer>`.  Wherever `Collections::singletonList` is matched up with `Function<? super T, ? extends U>` is getting `List<? super Integer>`.  Java just shows `List<Integer>` for `Optional.of(123).map(Collections::singletonList)`.

https://issues.apache.org/jira/browse/GROOVY-9803

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
@@ -131,6 +131,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
 
 import static org.apache.groovy.util.BeanUtils.capitalize;
 import static org.apache.groovy.util.BeanUtils.decapitalize;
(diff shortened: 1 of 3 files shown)


eric-milles:
I don't know if the change in `applyGenericsContextToParameterClass` is the right thing to do.  I have no Java specification section to quote for how return type generics should be handled.  Java gives `Set<Integer>` for `Optional.of(123).map(Collections::singleton).get()`.  I read over [this](https://docs.oracle.com/javase/tutorial/extra/generics/morefun.html) but didn't figure any specific guidance for how `map` should work in this case.

**Note:** `applyGenericsContextToParameterClass` is only used by `StaticTypeCheckingVisitor`.  I would have moved it there if not for binary backwards compatibility.

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -5272,6 +5275,46 @@ private static void extractGenericsConnectionsForSuperClassAndInterfaces(final M
...
+            if (types.length == paramTypes.length) {
+                for (int i = 0, n = types.length; i < n; i += 1) {
+                    if (!types[i].isGenericsPlaceHolder() && !isAssignableTo(types[i], paramTypes[i])) {
+                        return false;
+                    }
+                }
+                return true;
+            }
+            return false;
+        }).findFirst().orElse(null); // TODO: order matches by param distance
```
danielsun1106, line 5296: Are you going to tweak the PR further?

eric-milles: I have no further planned changes.  It should be workable as it is, unless there is another test case that can demonstrate an issue.

danielsun1106: Gotcha

danielsun1106:
Merged. Thanks!

chrylis:
I'm not sure from reading the code change, but does this handle arbitrarily nested generics? The actual real-world type that the compiler is having trouble with is (combining Vavr, Spring MVC, and some of my own record types)

```
<E, A, R> Either<JsonApiErrorResponse<E>, HeadersAndBody<JsonApiDataItemResponse<A, R>>>
```

where invoking `Either#map` expects a `Function<? super HeadersAndBody<JsonApiDataItemResponse<A, R>>` (this part works) but the closure (or Parrot lambda) seems not to correctly preserve the nested type parameters into the closure.

eric-milles:
@chrylis You can try out your specific case using the CI build of groovy or the latest snapshot build of groovy-eclipse.

PR #1421, opened 2020-11-17
GROOVY-9819: bump jsp-api/servlet-api versions to 2.3.3/4.0.1

paulk-asert:


veita:
Would this be the right place to fix https://issues.apache.org/jira/browse/GROOVY-9818?

paulk-asert:
It is indeed related. Both the JSP and servlet api license files are missing for Groovy 3. For Groovy 4, we have a revamped build and it doesn't bundle those jars in the zip (revamped after 4.0.0-alpha-1 was released). You can think of this as better honoring the intended "provided" status of those jars in Groovy 3. But I am still working things through, so things might change. I suspect I will keep this issue separate and fix the license files (for these new versions) in Groovy 3 - probably under the GROOVY-9818 issue.

paulk-asert:
Proposed PR merged.

PR #1424, opened 2020-11-19
Adds workflow to verify Groovy and Micronaut integration

JasonTypesCodes:
This PR may be against the wrong branch.  I wasn't quite sure where it would belong.


JasonTypesCodes:
It appears that this failed due to issues retrieving dependencies.  It had passed on my fork.  I think the GitHub action may have experienced a hiccup retrieving some dependencies

PR #1428, opened 2020-11-20
GROOVY-6882: STC: skip overridden methods when selecting best fit

eric-milles:
First commit fixes `AbstractMap#put(K,V)` to have resolved generics just like `HashMap#put(K,V)`; `chooseBestMethod` rejects `HashMap#put(K,V)` but not `AbstractMap#put(K,V)` in current codebase.

https://issues.apache.org/jira/browse/GROOVY-6882


paulk-asert:
Merged, thanks.

PR #1430, opened 2020-11-23
GROOVY-9821: STC: check "T" from "? extends T" for list/map properties

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9821


paulk-asert:
Merged, thanks!

PR #1431, opened 2020-11-23
GROOVY-9815: object method overrides static interface method in MM index

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9815


paulk-asert:
Merged, thanks!

paulk-asert:
It seems like this broke the Micronaut build:
https://github.com/apache/groovy/runs/1446689075?check_suite_focus=true
After reverting:
https://github.com/apache/groovy/actions/runs/382547533

PR #1439, opened 2020-12-04
GROOVY-7549, GROOVY-9847: getOuterClass, getOuterClasses and getOuterField must check redirect

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9847
https://issues.apache.org/jira/browse/GROOVY-7549

This also applies to `2_4_X` but should already be present in `3_0_X` and newer.


paulk-asert:
Merged into GROOVY_2_5_X. Please feel free to close.

PR #1441, opened 2020-12-06
Fix documentation typo in CoercionTest

jack-berg:
Noticed a small typo reading through the coercion documentation. Unless I'm misunderstanding something, this is the intended syntax.


danielsun1106:
when closure is the last argument, meth (a, {}) is equivalent to meth (a) {}

PR #1443, opened 2020-12-14
GROOVY-9373 (part 1): rework line numbers for statements

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9373
https://issues.apache.org/jira/browse/GROOVY-9199
https://issues.apache.org/jira/browse/GROOVY-9126
https://issues.apache.org/jira/browse/GROOVY-7647

"do" and "try" should not generate line numbers
instructions after last source line are tied to closing brace
instructions before first source line are tied to opening brace

Writing two `LINENUMBER` instructions in a row is an issue.  Given this sequence from `groovy.log.classgen` system property:
```
L1
  LINENUMBER 7 L1
L2
  LINENUMBER 8 L2
```
If user sets breakpoint on line 8 of source, debugger will not stop because it associates line 7 with whatever instructions come next.  One example of this is `try {` on line 7 and some statement on line 8 (see GROOVY-9199).


danielsun1106:
Merged. Thanks!

PR #1444, opened 2020-12-18
GROOVY-9865: Add some DGM methods for primitive arrays

paulk-asert:


danielsun1106:
+1

How about  adding `stream()` too?

eric-milles:
> How about adding `stream()` too?

As in `int[] arr; Arrays.stream(arr)`?  I think `org.codehaus.groovy.vmplugin.v8.PluginDefaultGroovyMethods` covers the relevant cases.

eric-milles:
Does it make sense to implement `max(int[])` using `IntStream`'s `max` method?  Like `IntStream.of(self).max().orElseThrow(() -> new UnsupportedOperationException(...));`?  And similarly for long and double arrays?  Essentially, the `max` DGM is just shorthand for `ints.stream().max().getAsInt()`.

paulk-asert:
Yes, boxed streams are already covered. Possibly an unboxed variant, e.g. intStream(int[]), would also be useful.

As to using the stream-based max, we could certainly consider that. For small arrays the overheads of setting up the stream would be slower than the current implementation. For large streams, there might be a gain. I am inclined to leave as is and make it easy for folks to do that themselves. If we had something like `intStream()` they could choose between `nums.max()`, `nums.intStream().max().asInt`, and `nums.intStream().parallel().max().asInt` as they felt appropriate.

danielsun1106:
> > How about adding `stream()` too?
> 
> As in `int[] arr; Arrays.stream(arr)`? I think `org.codehaus.groovy.vmplugin.v8.PluginDefaultGroovyMethods` covers the relevant cases.

Thanks for your reminding. I forgot we had added `stream()`...

paulk-asert:
I added intStream(), longStream() and doubleStream() for int[], long[] and double[] respectively and refactored quite a few places that were making copies of arrays as lists when just iterating through them was all that was needed.

eric-milles:
Oh, the `stream()` methods are returning `Stream` and not `IntStream` for `int[]`.  Since you are in `PluginDefaultGroovyMethods`, can you add `@since` to `orOptional`?

PR #1445, opened 2020-12-18
GROOVY-9662: StreamingJsonBuilder: set resolve strategy metadata

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9662


paulk-asert:
Merged, thanks!

PR #1446, opened 2020-12-20
GROOVY-9863: save type into synthetic property (backed by getter/setter)

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9863

`VariableScopeVisitor#findClassMember` makes `PropertyNode` for method-only properties but does not set any type information


danielsun1106:
Merged. Thanks!

PR #1447, opened 2020-12-21
GROOVY-9860: makeRawType should reduce generic placeholder to bound type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9860

`Tuple2` constructor parameters are placeholders and should reduce to `Object` for argument matching


danielsun1106:
Merged. Thanks!

PR #1450, opened 2020-12-21
GROOVY-9855: inline string concat sonner

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9855

### DIFF
--- a/src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
+++ b/src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
@@ -309,8 +309,15 @@ public final class ExpressionUtils {
             }
         } else if (exp instanceof BinaryExpression) {
             BinaryExpression be = (BinaryExpression) exp;
-            be.setLeftExpression(transformInlineConstants(be.getLeftExpression()));
-            be.setRightExpression(transformInlineConstants(be.getRightExpression()));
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/groovy/ast/tools/ExpressionUtils.java
```diff
@@ -309,8 +309,15 @@ public static Expression transformInlineConstants(final Expression exp) {
             }
         } else if (exp instanceof BinaryExpression) {
             BinaryExpression be = (BinaryExpression) exp;
-            be.setLeftExpression(transformInlineConstants(be.getLeftExpression()));
-            be.setRightExpression(transformInlineConstants(be.getRightExpression()));
+            Expression lhs = transformInlineConstants(be.getLeftExpression());
+            Expression rhs = transformInlineConstants(be.getRightExpression());
+            if (lhs instanceof ConstantExpression && rhs instanceof ConstantExpression &&
```
danielsun1106, line 314: We should check `be.operation.type == Types.PLUS` too?

eric-milles: Yes, thanks.  I did have that when the code was located elsewhere.  I will get it updated.

danielsun1106:
+1

danielsun1106:
Merged. Thanks!

PR #1453, opened 2021-01-02
GROOVY-9877: Add DGM `asReversed` to `List`

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-9877

Though we have DGM `list.reverse()` already, it will create a new list containing all elements from the original list.
While the proposed `list.asReversed()` will create a view list, which just delegates method calls to the original list.


paulk-asert:
+1

PR #1454, opened 2021-01-04
GroovyClassLoader fileReallyExists super inefficient

davydotcom:
fileReallyExists was doing a case sensitive test by listing all files in the directory. this is heavy on the filesystem API and can be done with simply using getCanonicalPath comparison.


paulk-asert:
Merged, thanks!

PR #1455, opened 2021-01-14
GROOVY-9892: fix operand stack for ++x / x++ in static compilation

Fiouz:
40ee3fd0c19530b4308c7c1b3e30abd4d45ed54a replaced an invocation of
`ScriptBytecodeAdapter.setGroovyObjectProperty()` into an invocation of
the generated `pfaccess$0*` mutator method.
Contrary to `setGroovyObjectProperty()`, the mutator method has a return
value, which ends up pushed to the operand stack.

This commit discards the value being pushed to the stack by pop'ing it.

### DIFF
--- a/src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy9892Bug.groovy
+++ b/src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy9892Bug.groovy
@@ -1,0 +1,43 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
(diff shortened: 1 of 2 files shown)


src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy9892Bug.groovy
```diff
@@ -0,0 +1,21 @@
+package org.codehaus.groovy.classgen.asm.sc.bugs
```
danielsun1106, line 1: license header is expected here:
https://github.com/apache/groovy/blob/494c7ec11799c7b6c91f6d7689470544800b7416/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java#L1-L18

src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy9892Bug.groovy
```diff
@@ -0,0 +1,21 @@
...
+            class OuterClass {
+                int counter
+            
+                def call() {
+                    { ->
+                        "Hello${++counter}World"
+                    }.call()
+                }
+            }
+            assert new OuterClass().call() == 'Hello1World'
```
danielsun1106, line 18: Could you add a test case to cover `counter++` ?

danielsun1106:
LGTM

src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy9892Bug.groovy
Fiouz: License header is now added

src/test/org/codehaus/groovy/classgen/asm/sc/bugs/Groovy9892Bug.groovy
Fiouz: Test modified to verify both prefix and postfix increment

danielsun1106:
Merged. Thanks!

PR #1456, opened 2021-01-14
last() method for LinkedList can be more efficient

davydotcom:
A LinkedList has a method called getLast() which on a DoublyLinkedList is a direct reference. Using size and get() can take more overhead in testing. While this optimization is VERY small, it can have impacts on large list traversals

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -9773,6 +9773,29 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
         return true;
     }
 
+
+     /**
+     * Returns the last item from the LinkedList.
+     * <pre class="groovyTestCase">
+     * def list = [3, 4, 2]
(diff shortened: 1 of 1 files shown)


danielsun1106:
+1

eric-milles:
Isn't this true for any implementer of `java.util.Deque`?

eric-milles:
The actual implementation for `get(int i)` in `LinkedList` starts from head or tail:
```java
    public E get(int index) {
        checkElementIndex(index);
        return node(index).item;
    }
    Node<E> node(int index) {
        // assert isElementIndex(index);

        if (index < (size >> 1)) {
            Node<E> x = first;
            for (int i = 0; i < index; i++)
                x = x.next;
            return x;
        } else {
            Node<E> x = last;
            for (int i = size - 1; i > index; i--)
                x = x.prev;
            return x;
        }
    }
```

So `last(List)` is not O(n) if that was the perception.

paulk-asert:
Merged. I also tweaked to make it for Deque as per Eric's suggestion. As noted by Eric, this may yield almost no benefit for LinkedList but can't be worse and may be significantly better for some Deque implementations.

PR #1457, opened 2021-01-14
GROOVY-9896, GROOVY-4727: add return for last case if no default present

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9896
https://issues.apache.org/jira/browse/GROOVY-9880
https://issues.apache.org/jira/browse/GROOVY-9373
https://issues.apache.org/jira/browse/GROOVY-4727


danielsun1106:
Merged. Thanks!

PR #1458, opened 2021-01-15
GROOVY-9882: add more logic from visitBinaryExpression to visitField

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9882


danielsun1106:
Merged. Thanks!

PR #1460, opened 2021-01-18
GROOVY-9890: always search for default methods

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9890
https://issues.apache.org/jira/browse/GROOVY-7160

Test for GROOVY-7160 is now located here: https://github.com/apache/groovy/blob/292254523545a1ab04e85f31ff4f82de417b0ac7/src/test/groovy/transform/stc/BugsSTCTest.groovy#L681

Selection between `EnumSet.of(Enum,Enum[])` and `Set.of(Object,Object,Object,Object,Object,Object)` is consistent between dynamic and static modes.


danielsun1106:
Merged. Thanks!

PR #1461, opened 2021-01-18
GROOVY-9893: STC: findSetters must look in super types

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9893

There is still a strong preference for field-based properties over setter-only properties.  Mixed cases of "real" and "virtual" properties may not work the same as the test cases added below.


danielsun1106:
Merged. Thanks!

PR #1462, opened 2021-01-19
GROOVY-9881: prefer same parameter count for SAM-type overload selection

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9881


danielsun1106:
Merged. Thanks!

PR #1463, opened 2021-01-20
GROOVY-9901: enable semantic analysis transforms for trait class

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9901
https://issues.apache.org/jira/browse/GROOVY-7293

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
@@ -208,6 +208,7 @@ public final class ASTTransformationVisitor extends ClassCodeVisitorSupport {
             GroovyClassVisitor visitor = new ASTTransformationCollectorCodeVisitor(source, compilationUnit.getTransformLoader());
             visitor.visitClass(classNode);
         }, Phases.SEMANTIC_ANALYSIS);
+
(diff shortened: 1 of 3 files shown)


danielsun1106:
Merged. Thanks!

paulk-asert:
Actually, there is more work to do for this one.

The desired intent of placing an AST transform in a trait could be one of (at least) two things. (1) Apply the AST to the trait class itself (essentially the trait helper class), or (2) apply the transform to the composed class once the trait has been merged (composed). Perhaps for @Memoized, this might amount to the same resulting behavior (give or take memoization between alternative trait implementations) but that can't be guaranteed in general. Hence the general disclaimer that we give - it isn't _just_ about the AST transform being in the same phase as trait compilation (semantic analysis) or trait composition (canonicalization).

Having said that, we know there are numerous cases where we can make a sensible intended behavior, so it doesn't mean we can't find a path forward. It might be as simple as using a meta-annotation on AST transforms to indicate which ones are safe to apply at a given time. E.g. @TraitCompilationSafe could annotate @Memoized. This is just one suggestion however. One design question is whether this should be in the hands of the writer of the AST transform or the trait. For the latter, a different mechanism would be needed, e.g. @TraitConfig(applyDuringCompilation=['groovy.transform.Memoized']). Again, just a suggestion.

PR #1464, opened 2021-01-21
GROOVY-9902: STC: allow method checking with partial generic information

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9902


danielsun1106:
Merged. Thanks!

PR #1465, opened 2021-01-21
STC: normalize inclusion of types in errors

eric-milles:
"Cannot call \<T> java.util.List#leftShift(T) ..."
  was
"Cannot call \<T> java.util.List \<String>#leftShift(T) ..."

Arrays are displayed as "java.lang.String[]" not "[Ljava.lang.String;"


danielsun1106:
Merged. Thanks!

PR #1468, opened 2021-01-23
GROOVY-9905: check for MOP method in inner class but not super types

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9905
https://issues.apache.org/jira/browse/GROOVY-8274


danielsun1106:
Merged. Thanks!

PR #1469, opened 2021-01-24
GROOVY-9681: apply GROOVY-9515 patch to v7.IndyInterface

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9681

This is a lite version of https://github.com/apache/groovy/pull/1227/files#diff-346370ec8511baf4e07e2c5fd7604a9d163d05d107f3baa4ae2f7809e8e6c759 targeted at 3_0_X specifically (v7.IndyInterface does not exist in master).  The test case is a bit difficult to set up since it requires _indy_ compilation that links to v7.IndyInterface (the Groovy 2.5 part of GROOVY-9681) and execution under Groovy 3.

I am trying this test under Groovy 4 now to see if it needs anything...

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v7/IndyInterface.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v7/IndyInterface.java
@@ -32,6 +32,7 @@ import java.lang.invoke.MethodType;
 import java.lang.invoke.MutableCallSite;
 import java.lang.invoke.SwitchPoint;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.logging.Level;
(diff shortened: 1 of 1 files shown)


src/main/java/org/codehaus/groovy/vmplugin/v7/IndyInterface.java
```diff
@@ -258,13 +261,18 @@ public static Object fromCache(MutableCallSite callSite, Class<?> sender, String
         MethodHandleWrapper mhw =
                 doWithCallSite(
                         callSite, arguments,
-                        (cs, receiver) ->
-                                cs.getAndPut(
+                        (CacheableCallSite ccs, Object receiver) ->
+                                ccs.getAndPut(
                                         receiver.getClass().getName(),
-                                        c -> fallback(callSite, sender, methodName, callID, safeNavigation, thisCall, spreadCall, dummyReceiver, arguments)
+                                        c -> Optional.of(fallback(callSite, sender, methodName, callID, safeNavigation, thisCall, spreadCall, dummyReceiver, arguments))
```
danielsun1106, line 267: `Optional.of` will create a new object, which may impact the performance a bit.
Why not just use `... ? ... : ...`?

eric-milles: How much performance are we talking about?  This form allows expression lambda.  Otherwise it would need to be restructured to a statement lambda.

danielsun1106:
Merged. Thanks!

PR #1470, opened 2021-01-25
GROOVY-9707: CompileStatic by configuration and TypeChecked by annotation

eric-milles:
`StaticCompilationTransformer` is not idempotent; do not re-apply `StaticCompileTransformation`

https://issues.apache.org/jira/browse/GROOVY-9707


danielsun1106:
Merged. Thanks!

PR #1471, opened 2021-01-25
GROOVY-9906: check index before comparing parameters

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9906


danielsun1106:
Merged. Thanks!

PR #1472, opened 2021-01-26
GROOVY-9907: apply parameter/variable instanceof to closure return type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9907


danielsun1106:
Merged. Thanks!

PR #1474, opened 2021-01-27
GROOVY-9914: DGM: add return type parameters for collectEntries variant

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9914


danielsun1106:
Merged. Thanks!

PR #1475, opened 2021-01-27
GROOVY-9915: use type(s) from static call site to resolve generics

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9915


danielsun1106:
Merged. Thanks!

PR #1476, opened 2021-01-27
GROOVY-9777: stubgen: add cast for Object parameter in special ctor call

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9777

### DIFF
--- a/src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
+++ b/src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
@@ -627,15 +627,15 @@ public class JavaStubGenerator {
 
     private void printSpecialConstructorArgs(PrintWriter out, ConstructorNode node, ConstructorCallExpression constrCall) {
         // Select a constructor from our class, or super-class which is legal to call,
-        // then write out an invoke w/nulls using casts to avoid ambiguous crapo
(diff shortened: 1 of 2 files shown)


danielsun1106:
Merged. Thanks!

src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
```diff
@@ -627,15 +627,15 @@ private static boolean noExceptionToAvoid(ConstructorNode fromStub, ConstructorN
...
         // Select a constructor from our class, or super-class which is legal to call,
-        // then write out an invoke w/nulls using casts to avoid ambiguous crapo
+        // then write out an invoke w/nulls using casts to avoid ambiguous calls
 
         Parameter[] params = selectAccessibleConstructorFromSuper(node);
         if (params != null) {
             out.print("super (");
 
-            for (int i = 0; i < params.length; i++) {
+            for (int i = 0, n = params.length; i < n; i += 1) {
```
keeganwitt, line 636: I think this is less readable than the original honestly (sorry to say).

keeganwitt:
I thought Jochen (or someone) had said there was a reason for not printing in the case of object, but I forget why now (can't find the message). This was something added in 1a757edbaba508cd3b9c1a74f4f84fc2dd1a0c43 for https://issues.apache.org/jira/browse/GROOVY-2418.

PR #1477, opened 2021-01-28
GROOVY-9916: primitive boolean corner case: null casts to false

eric-milles:
```groovy
class C {
  private boolean b
  void m() {
    b = null; // DefaultTypeTransformation.booleanUnbox(null)
    { ->
      b = null // DefaultTypeTransformation.castToType(null,boolean)
    }.call()
  }
}
```

https://issues.apache.org/jira/browse/GROOVY-9916

I tried moving `if (type.isPrimitive()) return castToPrimitive(object, type);` to the top, but none of the other primitive types allow null and would throw a GroovyCastException.


danielsun1106:
Merged. Thanks!

PR #1478, opened 2021-01-29
Sql.call() throw SQLException instead of Exception

moonfruit:


paulk-asert:
Looking back at the history, I think this was probably an early mistake in Groovy 1.0 that has been carried over for a long time.
It would be a binary breaking change. It probably makes sense to change for Groovy 4. For Groovy users, it makes little impact since the exception will generally be a SqlException anyway from underlying calls. From a Java integration point of view, we should do the change.

moonfruit:
I don’t think this will affect any old code, because `SQLException` is a child of `Exception`.

paulk-asert:
You are correct that this is not the worst of offenders regarding breaking changes. Numerous scenarios won't be affected. Certainly, any Java integration code like below wouldn't compile with older versions of Groovy. For Grails, where applications and plugins might be compiled with different Groovy versions, this can be an issue.

```
try {
    sql.call();
} catch(java.sql.SQLException ex) { }

public void run() throws java.sql.SQLException {
    sql.call(...);
}
```

moonfruit:
OK, I get it. Looking forward to Groovy 4.0.
Do I need to do something to merge this to 4.0? @paulk-asert

paulk-asert:
No, I'll merge if no further feedback in the next day or two.

danielsun1106:
+1

paulk-asert:
Merged. Thanks!

PR #1480, opened 2021-01-29
GROOVY-9918: produce varargs binding when one arg short despite arg type

eric-milles:
- trust upstream method target selection; is default argument path dead?

- add failsafe compiler error instead of throwing a NullPointerException

https://issues.apache.org/jira/browse/GROOVY-9918


danielsun1106:
Merged. Thanks!

PR #1481, opened 2021-01-31
GROOVY-9922: Add DGM `getPid`

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-9922

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -132,7 +132,7 @@ the `gradlew` command instead of `gradle`. (On Unix-like systems, use `./gradlew
 
 To build you will need:
 
-* {jdk}[JDK 9+]
+* {jdk}[JDK 10+]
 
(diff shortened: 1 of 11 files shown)


paulk-asert:
Should we actually add a v10 directory into the Plugin hierarchy and use the direct method[1] when available and fall back only when needed?

[1] https://docs.oracle.com/en/java/javase/11/docs/api/java.management/java/lang/management/RuntimeMXBean.html#getPid()

After all, fall back isn't guaranteed across all JVMs since "a Java virtual machine implementation _can choose_ to embed platform-specific useful information in the returned name string". Then the question becomes whether return type should be long or String. I have no strong preference.

danielsun1106:
OK. Let me refine it later according to your suggestions. BTW, I prefer returning `String`(See the second example of the JIRA ticket)

PR #1482, opened 2021-01-31
GROOVY-9801: Print default on interface methods with default implementation

keeganwitt:


src/test/org/codehaus/groovy/tools/stubgenerator/GROOVY9801.groovy
```diff
@@ -0,0 +1,42 @@
...
+class GROOVY9801 extends StringSourcesStubTestCase {
+
+    Map<String, String> provideSources() {
+        [
+                'Interface9801.groovy': '''
+                    interface Interface9801 {
+                        default String method() { return 'test' }
+                    }
+                ''',
+                'Class9801.groovy': '''
```
danielsun1106, line 30: The extension of the file should be 'java'?

src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
```diff
@@ -709,6 +709,8 @@ private void printMethod(PrintWriter out, ClassNode clazz, MethodNode methodNode
         }
 
         printGenericsBounds(out, methodNode.getGenericsTypes());
+        if (methodNode.isDefault())
+            out.print(" default ");
```
danielsun1106, line 713: The tailing space could be removed, otherwise `assert stubSource.contains('default void method')` will fail, because `out.print(" ");` in the next line will add an additional space.

src/test/org/codehaus/groovy/tools/stubgenerator/GROOVY9801.groovy
keeganwitt: Yea, that's probably a better test. Maybe I don't even need to be testing this, it's not part of the stub generation.

src/test/org/codehaus/groovy/tools/stubgenerator/GROOVY9801.groovy
```diff
@@ -0,0 +1,44 @@
...
+        ]
+    }
+
+    void verifyStubs() {
+        def stubSource = stubJavaSourceFor('Interface9801')
+        assert stubSource.contains('default java.lang.String method')
+        def classLoader = new URLClassLoader([targetDir.toURI().toURL()] as URL[], loader)
+        classLoader.loadClass('Interface9801')
+        def test = classLoader.loadClass('Class9801').getDeclaredConstructor().newInstance()
+        assert test.method() == 'test'
```
keeganwitt, line 42: Currently fails with `java.lang.AbstractMethodError: Receiver class Class9801 does not define or inherit an implementation of the resolved method 'abstract java.lang.String method()' of interface Interface9801.` because the interface method is abstract, as traits are currently used for interfaces.

PR #1484, opened 2021-02-10
GROOVY-9935: java.lang.Number can accommodate any number type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9935


danielsun1106:
Merged. Thanks!

PR #1485, opened 2021-02-11
GROOVY-9936: Difference between old/new parser for variable declarati…

paulk-asert:
…on optimisation

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/SemanticPredicates.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/SemanticPredicates.java
@@ -179,10 +179,12 @@ public class SemanticPredicates {
         token = ts.LT(index);
         tokenType = token.getType();
         tokenType3 = ts.LT(index + 2).getType();
+        int nextCodePoint = token.getText().codePointAt(0);
 
         return // VOID == tokenType ||
                 !(BuiltInPrimitiveType == tokenType || Arrays.binarySearch(MODIFIER_ARRAY, tokenType) >= 0)
-                        && Character.isLowerCase(token.getText().codePointAt(0))
+                        && !Character.isUpperCase(nextCodePoint)
+                        && nextCodePoint != '@'
                         && !(ASSIGN == tokenType3 || (LT == tokenType2 || LBRACK == tokenType2));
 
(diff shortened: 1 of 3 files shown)


danielsun1106:
+1

src/main/java/org/apache/groovy/parser/antlr4/SemanticPredicates.java
```diff
@@ -179,10 +179,12 @@ public static boolean isInvalidLocalVariableDeclaration(TokenStream ts) {
         token = ts.LT(index);
         tokenType = token.getType();
         tokenType3 = ts.LT(index + 2).getType();
+        int nextCodePoint = token.getText().codePointAt(0);
 
         return // VOID == tokenType ||
                 !(BuiltInPrimitiveType == tokenType || Arrays.binarySearch(MODIFIER_ARRAY, tokenType) >= 0)
-                        && Character.isLowerCase(token.getText().codePointAt(0))
+                        && !Character.isUpperCase(nextCodePoint)
+                        && nextCodePoint != '@'
```
danielsun1106, line 187: checking character @ seems unnecessary because @ is not allowed to appear in identifier.
Even if we use quotes, e.g. ”@foo”, the first character of the identifier is ".

paulk-asert: I would have thought so too but without it the following annotation on a local variable declaration failed:

> > Task :compileTestGroovy FAILED
> startup failed:
> D:\projects\groovyCopy\src\test\gls\annotations\closures\AnnotationClosureOwnerCallTest.groovy: 45: Unexpected input: '@' @ line 45, column 9.
>            @AnnWithClassElement(elem = { owner.answer() })
>            ^
> 
> 1 error
> 

```
        @AnnWithClassElement(elem = { owner.answer() })
        def aLocal
```
Perhaps there is a better tweak to cover that case?

danielsun1106: Understood. Let's go ahead with the your solution, which looks good to me.

danielsun1106:
Merged.Thanks!

PR #1486, opened 2021-02-12
GROOVY-6360: GString performance is slow with String's trim method

paulk-asert:


paulk-asert:
This avoids using reflection for two common String methods and, for Groovy 4, will piggy back on the already implemented improved caching.

danielsun1106:
How about adding all methods of String to GStringImpl, e.g. substring, indexOf, lastIndexOf, etc.

danielsun1106:
Merged. Thanks!

vy:
Hello, I have two questions regarding this PR:

1. Does this help with `StringBuilder#append(GString)`?
2. Will this be backported to 3.x?

PR #1488, opened 2021-02-12
Make Gradle build fully cacheable

serandel:
As a new hire in Gradle, part of my onboarding has been updating the build of an OSS project to be fully cacheable and we've chosen Groovy for this purpose. Also, the Gradle wrapper and the Gradle Enterprise plug-in have been upgraded to the latest released version. 

Please, have a look at my proposed changes, and merge them if you agree with them. :)


danielsun1106:
LGTM.
According to the Apache rules, binary files are not allowed to include in codebase, e.g. gradle/wrapper/gradle-wrapper.jar

paulk-asert:
> According to the Apache rules, binary files are not allowed to include in codebase, e.g. gradle/wrapper/gradle-wrapper.jar

Actually, the jar can be in the codebase but must not be in a release (in particular the source release). So, that part of the PR looks okay to me. We already have bootstrap steps described in the README for getting that jar when using the source release.

Also, we track dependency and tool version changes via a new issue in Jira. Would you be able to create that @serandel?

danielsun1106:
Merged. Thanks!

serandel:
> > According to the Apache rules, binary files are not allowed to include in codebase, e.g. gradle/wrapper/gradle-wrapper.jar
> 
> Actually, the jar can be in the codebase but must not be in a release (in particular the source release). So, that part of the PR looks okay to me. We already have bootstrap steps described in the README for getting that jar when using the source release.
> 
> Also, we track dependency and tool version changes via a new issue in Jira. Would you be able to create that @serandel?

It would be a pleasure, of course. 

https://issues.apache.org/jira/browse/GROOVY-9942

Can you please have a look at the Fix Version and edit it if it's not the right one, please? Thanks!

gradle/verification-metadata.xml
```diff
@@ -1,42 +1,4 @@
...
-   troubleshooting guide at https://docs.gradle.org/current/userguide/dependency_verification.html
-
-   This build uses PGP verification and SHA512 checksum verification for unsigned artifacts:
-
-   ./gradlew -write-verification-metadata sha512,pgp -export-keys <tasks>
-
-   Then check the diff, remove the unnecessary entries (for old dependencies) and make sure
-   the new ones are correct.
-
-   Do not remove this header.
```
melix, line 37: :x: This says "Do not remove this header" and the PR removes the header. It shows that the file hasn't been reviewed, when it should ;)

danielsun1106: Good catch ;-)

PR #1489, opened 2021-02-13
GROOVY-9938: SC: inner class needs implicitThis false for method calls

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9938


danielsun1106:
Merged. Thanks!

PR #1492, opened 2021-02-16
GROOVY-9946: Add `@Pure` to mark constant result of method

danielsun1106:
See https://issues.apache.org/jira/browse/GROOVY-9946

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
+++ b/src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
@@ -20,10 +20,12 @@ package org.codehaus.groovy.runtime;
 
 import groovy.lang.GString;
 import groovy.lang.GroovyObject;
+import groovy.transform.Pure;
 import org.apache.groovy.ast.tools.ImmutablePropertyUtils;
 
 import java.io.IOException;
 import java.io.Writer;
+import java.lang.reflect.Method;
 import java.nio.charset.Charset;
 import java.util.Locale;
 
@@ -56,7 +58,7 @@ public class GStringImpl extends GString {
      * @param strings the string parts
      */
     public GStringImpl(Object[] values, String[] strings) {
-        this(values, strings, checkValuesImmutable(values), null, false);
+        this(values, strings, checkValuesStringConstant(values), null, false);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


eric-milles:
If this annotation is a marker and not a transform, is there a better package for it?  I am not understanding the value proposition here.  The ticket does not describe why this would be useful.  Is the recreation of GString value a performance problem in some context?

danielsun1106:
> If this annotation is a marker and not a transform, is there a better package for it?

Any suggestion is welcome. Currently I put it with `@KnownImmutable` in the same package.  /cc @paulk-asert 

> The ticket does not describe why this would be useful.

Currently we just cache literal of `GString` when its all values are immutable, but for more common cases, if the result of value's `toString` is constant, we could cache too.

paulk-asert:
I am thinking this could be generalised. If instead of having a very specific class annotation, we had a `@Pure` or `@Constant` annotation and place it on the `toString` method itself, then the annotation could be used in more places. `@Memoized` could annotate the methods it generates (giving the caching optimization for free on memoized `toString` methods) and future type checkers could ensure that the return value of "pure" methods only comes from expressions involving constants or other pure methods - kind of what Frege does.

danielsun1106:
OK. Let me try to implement `@Pure` instead of `@ConstantString`.

paulk-asert:
LGTM. The one thing which I think we should do is move the `@Pure` marker interface into a groovy.annotations package in its own groovy-annotations mandatory module. However, I can do that move later if you like. That would be a good place for other such marker interfaces to also go. Then folks writing Java libraries could make them more Groovy friendly in some places and only need worry about having that one module available as a compile-time only dependency for their builds.

danielsun1106:
@paulk-asert 

> LGTM. The one thing which I think we should do is move the @Pure marker interface into a groovy.annotations package in its own groovy-annotations mandatory module. However, I can do that move later if you like. 

Thanks for your reviewing and help ;-)

src/main/java/org/codehaus/groovy/runtime/GStringImpl.java
```diff
@@ -351,15 +353,26 @@ public String toString() {
         return str;
     }
 
-    private static boolean checkValuesImmutable(Object[] values) {
+    private static boolean checkValuesStringConstant(Object[] values) {
         for (Object value : values) {
             if (null == value) continue;
-            if (!(ImmutablePropertyUtils.builtinOrMarkedImmutableClass(value.getClass())
+            if (!(toStringMarkedPure(value.getClass())
```
paulk-asert, line 359: We can probably swap the order of this clause and the next since this one is probably less common than next line.

paulk-asert:
Merged, thanks!

PR #1493, opened 2021-02-16
GROOVY-9945: findParameterizedType: fill type parameters of super types

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9945

There is quite a history to `findParameterizedType` and related methods.  I did not fully understand the reason for some of the looser checks.  I think they were to account for incomplete generics, which are now more complete.  Rather then building the whole type hierarchy at once, I took an approach of build as needed.  This should reduce some of the cost of searching.  I would have reduced the method count further, but it's all public api now.


danielsun1106:
Merged. Thanks!

PR #1494, opened 2021-02-17
GROOVY-6360: GString performance is slow with String's method (add sp…

paulk-asert:
…ecial GString DGM method for StringBuilder)


paulk-asert:
I see this as complementary to Daniel's other suggestion around `@Pure`.

danielsun1106:
+1

PR #1495, opened 2021-02-17
GROOVY-9948: infer type arguments for CCE diamond using argument type(s)

eric-milles:
`inferReturnTypeGenerics` uses first type seen for each placeholder.  If enhanced, `new C<>(1,'b')` could infer `C<Object>` instead of `C<Integer>`.

https://issues.apache.org/jira/browse/GROOVY-9948


danielsun1106:
Merged. Thanks!

PR #1496, opened 2021-02-22
GROOVY-6360: move @Pure to separate subproject.

paulk-asert:


paulk-asert:
Draft, still needs some work.

src/main/java/org/codehaus/groovy/reflection/ReflectionUtils.java
```diff
@@ -190,6 +191,15 @@ public static boolean trySetAccessible(final AccessibleObject ao) {
         return false;
     }
 
+    public static boolean hasAnnotationWithSimpleName(AccessibleObject obj, String simpleName) {
```
danielsun1106, line 194: Both `@groovy.annotations.Pure` and `@foo.bar.Pure` have same simple name `Pure`. If user uses `@foo.bar.Pure` to mark methods, which will be treated as "pure" methods too.

So I recommend using full class name instead of simple name.

paulk-asert: We can use full names. There are other annotations, e.g.:
checker framework's `org.checkerframework.dataflow.qual.Pure`
pure4j's `org.pure4j.annotations.pure.Pure`
JPure's `@Pure`
JML's `@Pure`
XText has `org.eclipse.xtext.xbase.lib.Pure`
Effect tracked Java uses `@Pure` (and `@Impure`)
C4J uses `de.vksi.c4j.Pure`
and `net.digitalid.utility.annotations.method.Pure`
which do essentially the same thing as ours.
Counter examples: Frege has a `pure` keyword, JetBrains has `@Contract` with a `pure` boolean annotation attribute.

paulk-asert: I haven't worked out a way to make it extensible without making it ugly.

And my next goal will be to write a purity type checker extension which would hopefully be able to be made to work with any of the above. Although I can think of a few options for this second use case.

paulk-asert:
I'll close this. The intention of moving this to a separate jar was so other folks could start using the annotation without depending on (all of) Groovy. I am not sure the world is a better place with yet another choice of a general purpose annotation library. We'll support Groovy's annotation if you take the Groovy jar and we'll work out which other libraries to support as a separate issue.

PR #1498, opened 2021-02-24
GROOVY-9951: set implicit-this to false for method call expressions

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9951

### DIFF
--- a/src/main/java/org/apache/groovy/ast/tools/ConstructorNodeUtils.java
+++ b/src/main/java/org/apache/groovy/ast/tools/ConstructorNodeUtils.java
@@ -24,7 +24,7 @@ import org.codehaus.groovy.ast.Parameter;
 import org.codehaus.groovy.ast.PropertyNode;
 import org.codehaus.groovy.ast.expr.ConstructorCallExpression;
 import org.codehaus.groovy.ast.expr.Expression;
-import org.codehaus.groovy.ast.expr.ListExpression;
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
 import org.codehaus.groovy.ast.expr.VariableExpression;
 import org.codehaus.groovy.ast.stmt.BlockStatement;
 import org.codehaus.groovy.ast.stmt.ExpressionStatement;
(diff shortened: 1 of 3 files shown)


src/test/groovy/transform/stc/BugsSTCTest.groovy
```diff
@@ -938,4 +938,22 @@ Printer
...
     }
+
+    // GROOVY-9951
+    void testInnerImmutablePOJO() {
+        assertScript '''
+            import groovy.transform.Immutable
+            import groovy.transform.stc.POJO
+
+            class Outer {
+                @Immutable @POJO
```
danielsun1106, line 949: AFAIR, `@POJO` should be used with `@CompileStatic`.

eric-milles: This test runs under STC and SC.

danielsun1106: OK

src/main/java/org/apache/groovy/ast/tools/ConstructorNodeUtils.java
```diff
@@ -88,17 +91,20 @@ public static Statement checkPropNamesS(final VariableExpression namedArgs, fina
...
+        MethodCallExpression names = callX(namedArgs, "keySet");
+        names.setImplicitThis(false);
+
+        MethodCallExpression isNameValid = callX(validNames, "contains", varX(name));
+        isNameValid.setImplicitThis(false);
+
+        return block(
+            declS(validNames, listX(props.stream().map(p -> constX(p.getName())).collect(toList()))),
+            forS(name, names, ifS(notX(isNameValid),
+                    throwS(ctorX(EXCEPTION, plusX(constX("Unknown named argument: "), varX(name))))))
```
paulk-asert, line 107: This all LGTM but we can probably go one step further for the plusX expression if `pojo` is true. At the moment it ends up with a call to `StringGroovyMethods#plus` but we could do such a thing without needing the Groovy runtime (for the `pojo` case). There might be multiple ways to do that and we can do that as a next step.

paulk-asert:
Merged, thanks!

PR #1500, opened 2021-02-24
Groovy 7968

PatrikTorn:


danielsun1106:
I am not sure if we have to introduce the breaking change into Groovy 4 right now.
/cc @paulk-asert

paulk-asert:
We'd certainly need more testing around inherited fields/properties for such a change. I'd suggest an amendment.

The current assumption is that if the field is found it is used instead of creating a backing field. It is a little strange/cumbersome but gives wiggle room in rare circumstances, e.g. it would allow for a normal public getter/setter but with for instance a protected backing field, or a synchronized backing field but not synchronized getters/setters (albeit I don't know why you'd want that).

The main thing I'd like to detect is where two initial values are supplied (one on each of the field and property). I believe that should be an error since it would be ambiguous and which value to use is currently not specified. Currently, any initial value given for the property is ignored if a field is also found - though this is really just the current implementation - it is certainly unintuitive where no initial value is given for the field.

@PatrikTorn Are you interested in amending in this way?

paulk-asert:
Merged with some slight tweaks. Hope that's okay.

PR #1501, opened 2021-02-24
GROOVY-9953: lookup instanceof type for variable on right of binary expression

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9953


danielsun1106:
Merged. Thanks!

PR #1502, opened 2021-02-25
GROOVY-9954: Stop incrementing index after comments to avoid skipping…

kawatea:
https://issues.apache.org/jira/browse/GROOVY-9954

Stopped incrementing index in handleComment() and handleBashComment().
These are called in three places.
In decodeJsonObjectLax(), the index is incremented by the for-loop, but startIndexOfKey needs to be set to the next character after the comment.
In decodeValueInternal(), the index is incremented by the for-loop.
In decodeJsonArrayLax(), the index is not incremented by the do-while-loop, it needs to be manually incremented.


danielsun1106:
Merged. Thanks!

PR #1503, opened 2021-02-26
GROOVY-9955: access static fields and invoke static methods using object expression type not owner

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9955


danielsun1106:
Merged. Thanks!

PR #1504, opened 2021-02-27
Fix GROOVY-9797: println(-0.0f) does not comply with IEEE754

Exploder98:
[Jira](https://issues.apache.org/jira/browse/GROOVY-9797)

This PR fixes the mentioned issue and implements tests to make sure the issue does not surface again.

## Some details

This bug was actually a compiler bug. When inserting constants to the Java Bytecode, Groovy compiler checks if a float is 0.0f and inserts the instruction for this special constant in case it is. However, the `==` operator works according to the IEEE-754 standard, and compares negative and positive zero as equal. So, if there's a negative zero (-0.0f) in the original Groovy code, the instruction for a positive float would be inserted in the bytecode.

This PR fixes this issue by using `Float.equals()` instead, which, according to the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#equals-java.lang.Object-), does not compare negative and positive zero as equal.

This PR was a joint effort by four people:

@Exploder98
@EricssonWilli
@iiroki
@OttoVayrynen

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
@@ -551,7 +551,8 @@ public class OperandStack {
                 mv.visitLdcInsn(value);
             }
         } else if (ClassHelper.double_TYPE.equals(type)) {
-            if ((Double)value==0d) {
+            // GROOVY-9797: Use Double.equals to differentiate between positive and negative zero
+            if (value.equals(0d)) {
                 mv.visitInsn(DCONST_0);
             } else if ((Double)value==1d) {
                 mv.visitInsn(DCONST_1);
--- a/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
@@ -540,7 +540,8 @@ public class OperandStack {
                 mv.visitLdcInsn(value);
             }
         } else if (ClassHelper.float_TYPE.equals(type)) {
-            if ((Float)value==0f) {
+            // GROOVY-9797: Use Float.equals to differentiate between positive and negative zero
+            if (value.equals(0f)) {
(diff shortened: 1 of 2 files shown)


danielsun1106:
+1

tomikoskinen:
This fix leads to  `-0.0f == 0.0f` being `false`, which might be unwanted.

Exploder98:
@tomikoskinen 

This could be expected as the [Groovy documentation](https://docs.groovy-lang.org/docs/latest/html/documentation/#_identity_operator) states that *"In Groovy, it [the `==` operator] is calling equals"*, and for the `Float.equals` method the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#equals-java.lang.Object-) quite explicitly states that `-0.0f` and `0.0f` aren't compared equal.

jonspe:
Hmm, true. Float equality comparison should usually be done using a small threshold epsilon value, so it should not be a big deal.

danielsun1106:
@paulk-asert `0.0f == -0.0f` yields `true`, is it the expected result in Groovy? As we could see, Java yields `false`:
```java
jshell> ((Float)0.0f).equals(-0.0f)
$1 ==> false
```

paulk-asert:
The `==` operator is deemed to have _business friendly_ behavior. Equating `==` with `equals` is an over simplification. For comparable objects, comparator logic is used and `NumberAwareComparator` would become relevant for the cases mentioned here. This falls into the same category as when we compare different number types:
```
assert [false, true, 0] == [0.0f.equals(0.0d), 0.0f == 0.0d, new NumberAwareComparator().compare(0.0f, 0.0d)]
```
In summary, this PR looks good but I think Groovy 4 only because of the breaking change.

paulk-asert:
Actually, let me investigate further. We have some short-circuit code for some number types.

paulk-asert:
Thanks for the PR which is merged into master. There is some follow up work we need to do. I'll clone the issue and do that work in the new issue.

PR #1505, opened 2021-02-27
GROOVY-6232, GROOVY-9956: check LHS type witness for diamond constructor

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-6232
https://issues.apache.org/jira/browse/GROOVY-9956


danielsun1106:
Merged. Thanks!

PR #1506, opened 2021-02-28
GROOVY-9797 bug fix (DRAFT)

tomikoskinen:
https://issues.apache.org/jira/browse/GROOVY-9797

This fixes the above bug by using `Float.equals()` which also changes `-0.0f == 0.0f` to `false` instead of `true`. This corresponds with the https://docs.groovy-lang.org/docs/latest/html/documentation/#_identity_operator and https://docs.oracle.com/javase/8/docs/api/java/lang/Float.html#equals-java.lang.Object- documentations.

This PR was made by:
@tomikoskinen
@jonspe
@Justsofun
@rhanlol


paulk-asert:
superseded by merging of #1504

PR #1507, opened 2021-03-04
GROOVY-9963: resolve ctor call diamond for attribute/property expression

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9963


danielsun1106:
Merged. Thanks!

PR #1510, opened 2021-03-05
GROOVY-9968: STC ClosureParams: for "T -> Type" use Type not Object

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9968


danielsun1106:
Merged. Thanks!

mgroovy:
Thank you for fixing this so quickly, that will save me a lot of trouble :-)

PR #1511, opened 2021-03-06
GROOVY-9964: Groovy 3 compilation is much slower than Groovy 2 compil…

danielsun1106:
…ation for Gradle build scripts


eric-milles:
Are there any performance measurements for this change?  Nothing is mentioned in the linked issue about what is taking the time or how much improvement to expect.

danielsun1106:
I tested with "project0/build.gradle" file  locally, about 10% time saved.
As for the goal, tune the performance as good as we could...

paulk-asert:
So, we have to eliminate the lambda chaining to gain the performance? I.e. GROOVY-8991 would be broken?

danielsun1106:
It is not expected... but parser suffers from ambiguities while choosing alternatives... Maybe we have to find better way to improve the parsing performance...

PR #1512, opened 2021-03-08
GROOVY-9970: resolve return type generics from arguments then enclosing

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9970


danielsun1106:
Merged. Thanks!

PR #1513, opened 2021-03-08
GROOVY-9972: STC: infer ctor call diamond type for ternary branches

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9972


danielsun1106:
Merged. Thanks!

PR #1514, opened 2021-03-09
GROOVY-9974: extract generics connections: Closure<T> vs SAM return type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9974


danielsun1106:
Merged. Thanks!

PR #1515, opened 2021-03-10
GROOVY-9807: limit receiver replacement for implicit-this in inner class

eric-milles:
When user creates a method call expression, the default is for implicit
this to be true and this has caused a number of problems for transforms
when applied to inner class scenarios.  Only when the object expression
is truly "this" should it be rewritten to "Outer.this" when referencing
an outer class method.

https://issues.apache.org/jira/browse/GROOVY-9807

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/GroovyCodeVisitorAdapter.java
+++ b/src/main/java/org/codehaus/groovy/ast/GroovyCodeVisitorAdapter.java
@@ -1,0 +1,309 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


danielsun1106:
GroovyCodeVisitorAdapter is for specific purpose, so how about making it AIC instead of standalone new class?

eric-milles:
> `GroovyCodeVisitorAdapter` is for specific purpose, so how about making it AIC instead of standalone new class?

What is missing from `GroovyCodeVisitor` and `CodeVisitorSupport` is the ability to override select methods of interest and restrict traversal.  If I were to create a straight AIC of `GroovyCodeVisitor` in `InvocationWriter#isThis`, the method would be hundreds of lines long and the number of imports for the class would double.  This adapter is useful outside of this particular application -- in fact I imported the implementation from groovy-eclipse.

eric-milles:
Any code that uses a sequence of `instanceof` checks to distinguish `ASTNode`/`Expression`/`Statement` types could instead leverage `GroovyCodeVisitorAdapter` and the visitor pattern for type dispatch.

danielsun1106:
Merged. Thanks!

paulk-asert:
Yep, I am happy with the `GroovyCodeVisitorAdapter` class but if we can think of a better name, I'd also be +1 for that.

PR #1516, opened 2021-03-11
GROOVY-9397: Make respondsTo thread safe

broecirr:
Because ClosureMetaClass.loadsMetaInfo() changes the initialized state 
of the Closure's Metaclass without any locking, if the Closure is visible to
another thread, then any call to invokeMethod can see the closure as
uninitialized and throw an exception.

This fix removes the need to change the initialized state of the
closure in loadMetaInfo().


broecirr:
I don't see any Exceptions when running the example in the linked JIRA ticket with this change.

danielsun1106:
LGTM.
Could you add the test you mentioned?

danielsun1106:
Merged. Thanks!

PR #1517, opened 2021-03-11
GROOVY-9977: share logic for functional interface assignment

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9977

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
@@ -744,25 +744,15 @@ public class StaticTypeCheckingVisitor extends ClassCodeVisitorSupport {
                 if (ensureValidSetter(expression, leftExpression, rightExpression, setterInfo)) {
                     return;
                 }
+                lType = getType(leftExpression);
             } else {
                 lType = getType(leftExpression);
-                boolean isFunctionalInterface = isFunctionalInterface(lType);
-                if (isFunctionalInterface && rightExpression instanceof MethodReferenceExpression) {
-                    LambdaExpression lambdaExpression = constructLambdaExpressionForMethodReference(lType);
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -1834,19 +1838,22 @@ public void visitField(final FieldNode node) {
         try {
             typeCheckingContext.isInStaticContext = node.isInStaticContext();
             currentField = node;
-            super.visitField(node);
+            visitAnnotations(node);
```
paulk-asert, line 1841: Normal visitor pattern would call super here. Perhaps a comment here to indicate that were replicating some of super.visitField to avoid processing init expression twice?

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -1821,7 +1823,9 @@ public void visitProperty(final PropertyNode node) {
         try {
             typeCheckingContext.isInStaticContext = node.isInStaticContext();
             currentProperty = node;
-            super.visitProperty(node);
+            visitAnnotations(node);
```
paulk-asert, line 1826: ditto with my other comment?

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
eric-milles: If changed back, there are test cases that will fail.  It is technically possible to structure this so that the functional interface stuff comes before "super.visitField(node)" and the assignment type checking comes after (where it currently sits).

The property visit could have its initialization expression visit dropped, since it is redundant with the field.  I thought that would be a riskier change to make.

PR #1518, opened 2021-03-12
Groovy9981

paulk-asert:


workingjubilee:
It looks like this means that a "naïve" use of `0.0f == -0.0f` is still `true`, and that otherwise it is advisable to deliberately select the method that is intended for use. Good!

PR #1519, opened 2021-03-15
GROOVY-9984: null arguments provide no value for return type generics inference

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9984


danielsun1106:
Merged. Thanks!

PR #1520, opened 2021-03-16
GROOVY-9985: STC does not report type mismatches in array initializers

paulk-asert:
This covers the array initializer but not the 'as' coercion


danielsun1106:
+1

PR #1521, opened 2021-03-16
GROOVY-7945: resolve type of "super" constructor call including type arguments

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7945


danielsun1106:
Merged. Thanks!

PR #1522, opened 2021-03-17
GROOVY-7686: AIC: place local variable references above super ctor call

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7686


danielsun1106:
Merged. Thanks!

PR #1523, opened 2021-03-17
GROOVY-9989: remove JCenter Bintray from default Ivy resolvers

apupier:
JCenter Bintray will be decommisioned
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

it migth worth a note in the changelog. I have not foudn a changelog file. is it genrated from JIRA? Or in a specific file?


paulk-asert:
The release note/changelog info comes from the groovy-website repo:
https://github.com/apache/groovy-website

There is an automated changelog generated during the release process, e.g.:
http://groovy-lang.org/changelogs/changelog-3.0.7.html
You just need the appropriate fix version set in Jira which we set when we resolve an issue.

There is also a manual release notes, e.g.:
https://groovy-lang.org/releasenotes/groovy-3.0.html
Source:
https://github.com/apache/groovy-website/blob/asf-site/site/src/site/releasenotes/groovy-3.0.adoc

paulk-asert:
Merged, thanks!

PR #1524, opened 2021-03-17
GROOVY-9989: remove JCenter Bintray from default Ivy resolvers

apupier:
JCenter Bintray will be decommisioned
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

see also https://github.com/apache/groovy/pull/1523

### DIFF
--- a/src/resources/groovy/grape/defaultGrapeConfig.xml
+++ b/src/resources/groovy/grape/defaultGrapeConfig.xml
@@ -28,7 +28,6 @@
       </filesystem>
       <ibiblio name="localm2" root="${user.home.url}/.m2/repository/" checkmodified="true" changingPattern=".*" changingMatcher="regexp" m2compatible="true"/>
       <!-- TODO: add 'endorsed groovy extensions' resolver here -->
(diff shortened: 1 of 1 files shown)


apupier:
the [failing job](https://github.com/apache/groovy/pull/1524/checks?check_run_id=2131410713#step:5:1211) has error:

`junit.framework.AssertionFailedError: Exception in constructor: testScriptMethodAnnotation (java.lang.RuntimeException: Error grabbing Grapes -- [unresolved dependency: com.jidesoft#jide-oss;[2.2.1,2.3): not found]`

it is failing only on Windows.
This dependency is available on maven Central https://mvnrepository.com/artifact/com.jidesoft/jide-oss/2.2.2.01
I suppose that it is a flaky test.

paulk-asert:
It is a flaky test. We know about it but haven't fixed it yet.

paulk-asert:
Instead of removing jcenter for Groovy 2/3, I went with moving jcenter to the last entry in the resolver chain list. As per comment in GROOVY-9989, this should minimise the impact for users who have already moved away from jcenter/bintray but not break things for users still in the progress of moving. We should remove in a later Groovy release.

PR #1526, opened 2021-03-17
GROOVY-9524: enum const class extends enum class, so no this$0 for calls

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9524


danielsun1106:
Merged. Thanks!

PR #1527, opened 2021-03-18
GROOVY-7306: stubgen: apply type arguments of super to its constructor parameters

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7306

### DIFF
--- a/src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
+++ b/src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
@@ -576,25 +576,31 @@ public class JavaStubGenerator {
         }
     }
 
-    private static Parameter[] selectAccessibleConstructorFromSuper(ConstructorNode node) {
-        ClassNode type = node.getDeclaringClass();
-        ClassNode superType = type.getUnresolvedSuperClass();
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/tools/javac/JavaStubGenerator.java
```diff
@@ -576,25 +576,31 @@ private void printConstructor(PrintWriter out, ClassNode clazz, ConstructorNode
...
-            Parameter[] copy = new Parameter[parameters.length];
-            for (int i = 0; i < copy.length; i++) {
-                Parameter orig = parameters[i];
-                copy[i] = new Parameter(orig.getOriginType().getPlainNodeReference(), orig.getName());
-            }
-            if (noExceptionToAvoid(node,c)) return copy;
-            if (bestMatch==null) bestMatch = copy;
+        for (ConstructorNode target : superType.getDeclaredConstructors()) {
+            // only look at things we can actually call
+            // TODO: package-private and types are peers
```
danielsun1106, line 586: more  tweak planned?

eric-milles: No, just a note that there is a latent issue here.

PR #1528, opened 2021-03-19
GROOVY-6747: ensure enum ctor private and add bridge for const with body

eric-milles:
```groovy
  enum E {
    X {
    },
    Y
  }
```
E provides a private, implicit constructor (String,int) for Y and a package-private, synthetic constructor (String,int,E) for X (aka E$1).  This is what the Java compiler does as well.

https://issues.apache.org/jira/browse/GROOVY-6747


danielsun1106:
Merged. Thanks!

PR #1529, opened 2021-03-20
GROOVY-9991: closure without arrow works for 0 and 1 parameter methods

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9991


danielsun1106:
Merged. Thanks!

PR #1530, opened 2021-03-20
GROOVY-5728: anonymous inner class can use private constructor when super is outer

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-5728


danielsun1106:
Merged. Thanks!

PR #1531, opened 2021-03-21
GROOVY-8104: fix this$0 for closure with anon. inner class of inner type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-8104

```groovy
class Outer {
  class Inner {
    // requires Outer reference
  }
  def x = { ->
    new Inner() { // this.@this$0 returns closure; call getThisObject()
    }
  }
}
```


danielsun1106:
Merged. Thanks!

PR #1532, opened 2021-03-22
GROOVY-9995: infer ctor call diamond type from closure target type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9995
https://issues.apache.org/jira/browse/GROOVY-9997


danielsun1106:
Merged. Thanks!

PR #1533, opened 2021-03-23
GROOVY-9996: infer return type generics using declared type of variables

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9996


danielsun1106:
Merged. Thanks!

PR #1534, opened 2021-03-27
GROOVY-10002: STC: fix checks for list literal assignment

eric-milles:
- empty and non-empty list literals can initialize Set variables
- list literal generics checking skipped when constructor may be used
- list literal as constructor arguments does not apply to abstract types

https://issues.apache.org/jira/browse/GROOVY-6912
https://issues.apache.org/jira/browse/GROOVY-7128
https://issues.apache.org/jira/browse/GROOVY-10002


danielsun1106:
Merged. Thanks!

PR #1536, opened 2021-03-28
GROOVY-7106, GROOVY-7274, GROOVY-8909, GROOVY-9844: [] and [:] call args

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7106
https://issues.apache.org/jira/browse/GROOVY-7274
https://issues.apache.org/jira/browse/GROOVY-8909
https://issues.apache.org/jira/browse/GROOVY-9844


danielsun1106:
Merged. Thanks!

PR #1538, opened 2021-03-30
GROOVY-9993: clarify boundary cases for split property definition (field and property of same name)

paulk-asert:
When defining a property, getters and setters are automatically supplied unless an existing one is found. This same rule also applies to the backing field but to date many aspects of this rule have been under specified, e.g. what happens if the type given for the field and property aren't the same? This PR clarifies these edge cases and provides a more formal definition of the behavior (the different types mentioned earlier for example becomes a compiler error).


paulk-asert:
See also GROOVY-8659 (might require further work)

PR #1541, opened 2021-04-03
GROOVY-8983: STC: support "Type[] array = collectionOfTypeOrSubtype"

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-8983

**NOTE:** `GenericsUtils.parameterizeType("List<? super Type>","Collection")` returns `Collection<Type>` and not `Collection<? super Type>`.  I attempted to address this, but was not successful.  This should probably be fixed at some point because it breaks the semantics of "? super Type".


danielsun1106:
Merged. Thanks!

PR #1542, opened 2021-04-04
GROOVY-9971: STC: support "Closure<String> c = { -> return "$v" }"

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9971


danielsun1106:
Merged. Thanks!

PR #1544, opened 2021-04-05
GROOVY-10010: check GString element vs String collection for method call

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10010


danielsun1106:
Merged. Thanks!

PR #1545, opened 2021-04-06
GROOVY-10011: STC: instanceof refines declared type for diamond operator

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10011


paulk-asert:
Merged, thanks!

PR #1549, opened 2021-04-07
GROOVY-10006: STC: unify types of multiple type witnesses in method call

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10006


danielsun1106:
Merged. Thanks!

PR #1551, opened 2021-04-09
GROOVY-10027: NamedParam: check against declared/inferred argument type

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10027


danielsun1106:
Merged. Thanks!

PR #1552, opened 2021-04-10
GROOVY-10024: Dollar slashy /$ string ending in backslash "breaks" th…

paulk-asert:
…e parser

### DIFF
--- a/src/test/groovy/StringTest.groovy
+++ b/src/test/groovy/StringTest.groovy
@@ -241,6 +241,12 @@ foo
 
         text = $/$$$$$//$
         assert text == '$$/'
+
+        //GROOVY-10024
+        def s1 = $/abc\\/$
+        def s2 = $/def\/$
+        def s3 = $/ghi/$
+        assert s1 == 'abc\\\\' && s2 == 'def\\' && s3 == 'ghi'
     }
 
     void testSplit() {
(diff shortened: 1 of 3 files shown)


danielsun1106:
LGTM. Let's wait for the test result:
https://github.com/danielsun1106/groovy-parser/tree/GROOVY-10024

danielsun1106:
The test result is green ;-)

src/test/groovy/StringTest.groovy
```diff
@@ -241,6 +241,11 @@ foo
 
         text = $/$$$$$//$
         assert text == '$$/'
+
+        //GROOVY-10024
+        def s1 = $/abc\/$
+        def s2 = $/def/$
+        assert s1.size() == 4 && s2.size() == 3
```
pbielicki, line 248: I’d assert the expected values instead of the size.

timyates: Does

```
def s1 = $/Failing string\\/$
```

With double backslash at the end work (as per the Jira example)?

paulk-asert: The double backslash case works - I'll adjust the test accordingly and also with expected values.

timyates: Awesome as always Paul 😎👍

paulk-asert:
Proposed PR merged.

PR #1553, opened 2021-04-10
Implement GROOVY-9649: left-open and full-open ranges

Exploder98:
[JIRA Issue](https://issues.apache.org/jira/browse/GROOVY-9649)

This Pull request implements the above feature and adds test cases related to left- and full-open ranges.
We have also added documentation.
This change *should* be backwards-compatible, as old right-exclusive ranges work as before.
We have tried to preserve all old constructors and methods so that this change would be as backwards-compatible as possible.

A few things to note:

- ranges like `0<..<1` where to and from differ by one, the range is full exclusive and to and from share the same sign (positive numbers and zero are considered to have the same sign) are created as `EmptyRange`s.
This is because ranges like `0<..<-1` can still be used to index into lists.
- while testing, we noticed that the following equality test produces different results when the operands are switched:
```groovy
IntRange r1 = new IntRange(false, 1, 10) // "Inclusive-aware" range
IntRange r2 = new IntRange(1, 10) // Non-inclusive-aware range
r1.equals(r2) // Returns false
r2.equals(r1) // Returns true
```
we have not changed this behaviour.

While implementing the feature, we noticed some inconsistencies between different ranges, which we have not altered due to backwards compatibility:
- `NumberRange` and `IntRange` contain information about their inclusivity, but `ObjectRange` inclusivity is taken into account when the range is created
- in `IntRange`, the `getFrom` and `getTo` methods return the actual bounds of that range, whereas `NumberRange`'s corresponging methods return the bounds that the object was created with, even if the bounds are not actually included in the range

This PR was a joint effort by 4 people:

@Exploder98 
@iiroki 
@OttoVayrynen 
@EricssonWilli

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
+++ b/src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
@@ -27,17 +27,32 @@ import org.codehaus.groovy.ast.GroovyCodeVisitor;
  * <pre>for i in 0..10 {...}</pre>
  */
 public class RangeExpression extends Expression {
-
     private final Expression from;
     private final Expression to;
-    private final boolean inclusive;
+    private final boolean inclusive; // Kept to keep old code depending on this working
+    // GROOVY-9649
+    private final boolean exclusiveLeft;
+    private final boolean exclusiveRight;
 
+    // Kept until sure this can be removed
     public RangeExpression(Expression from, Expression to, boolean inclusive) {
         this.from = from;
         this.to = to;
         this.inclusive = inclusive;
+        this.exclusiveLeft = false;
+        this.exclusiveRight = !inclusive;
         setType(ClassHelper.RANGE_TYPE);
     }
+
+    public RangeExpression(Expression from, Expression to, boolean exclusiveLeft, boolean exclusiveRight) {
+        this.from = from;
+        this.to = to;
+        this.inclusive = !exclusiveRight;
+        this.exclusiveLeft = exclusiveLeft;
+        this.exclusiveRight = exclusiveRight;
+        setType(ClassHelper.RANGE_TYPE);
+    }
+
 
     @Override
     public void visit(GroovyCodeVisitor visitor) {
@@ -46,7 +61,7 @@ public class RangeExpression extends Expression {
 
     @Override
     public Expression transformExpression(ExpressionTransformer transformer) {
-        Expression ret = new RangeExpression(transformer.transform(from), transformer.transform(to), inclusive);
+        Expression ret = new RangeExpression(transformer.transform(from), transformer.transform(to), exclusiveLeft, exclusiveRight);
         ret.setSourcePosition(this);
         ret.copyNodeMetaData(this);
         return ret;
--- a/src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
+++ b/src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
@@ -44,10 +44,11 @@ public class RangeExpression extends Expression {
         setType(ClassHelper.RANGE_TYPE);
     }
 
+    // GROOVY-9649
     public RangeExpression(Expression from, Expression to, boolean exclusiveLeft, boolean exclusiveRight) {
         this.from = from;
         this.to = to;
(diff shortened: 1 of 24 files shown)


src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
```diff
@@ -27,26 +27,42 @@
  * <pre>for i in 0..10 {...}</pre>
  */
 public class RangeExpression extends Expression {
-
     private final Expression from;
     private final Expression to;
-    private final boolean inclusive;
+    private final boolean inclusive; // Kept to keep old code depending on this working
```
eric-milles, line 32: Can this field be dropped and `isInclusive()` is implemented in terms of "right inclusivity"?

src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
```diff
@@ -27,26 +27,42 @@
...
+    private final boolean exclusiveRight;
 
+    // Kept until sure this can be removed
     public RangeExpression(Expression from, Expression to, boolean inclusive) {
         this.from = from;
         this.to = to;
         this.inclusive = inclusive;
+        this.exclusiveLeft = false;
+        this.exclusiveRight = !inclusive;
+        setType(ClassHelper.RANGE_TYPE);
```
eric-milles, line 44: Can this constructor delegate to the new constructor?

src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
```diff
@@ -64,10 +80,20 @@ public boolean isInclusive() {
...
 
+    public boolean isExclusiveLeft() {
+        return exclusiveLeft;
+    }
+
+    public boolean isExclusiveRight() {
+        return exclusiveRight;
+    }
+
     @Override
```
eric-milles, line 91: Would "isLeftExclusive" or "isLeftOpen" be better than "isExclusiveLeft"?  And same for right?

src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
```diff
@@ -27,26 +27,42 @@
  * <pre>for i in 0..10 {...}</pre>
  */
 public class RangeExpression extends Expression {
```
eric-milles, line 29: Is there any hope that code like AST transforms that depend on "from" and "to" and "inclusive" will be able to handle a left-open or full-open range literal?  Is it considered a breaking change to use "<" on the left?

src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
iiroki: Agreed, I'll implement this.

src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
iiroki: Agreed, I'll make this change.

src/main/java/org/codehaus/groovy/ast/expr/RangeExpression.java
iiroki: I agree that it would be clearer to use "open" keyword instead of "exclusive". `isOpenLeft` and `isOpenRight` are more similiar to each other than `isLeftOpen` and `isRightOpen` so it think it would be better to go with that way. IDE autocompletion will also benefit from this, e.g. writing `isOpen...` will pop up both.

iiroki: On second thought, "exclusive" keyword is widely used elsewhere, so keeping the names as they are now would keep things more consistent.

EricssonWilli:
There seems to be a conflict, should we resolve that?

paulk-asert:
Merged - with a minor tweak for the conflict. Thanks! I suspect we will need to tweak a few other things, e.g. containsWithinBounds and perhaps time to relook at equals symmetry.

paulk-asert:
FYI, we will likely only put this in Groovy 4.

PR #1554, opened 2021-04-11
GROOVY-4990 Make File.write() create any missing directories. Add test.

xaaria:
This PR will:
- Fix issue where `java.io.FileNotFoundException` is thrown when parent directories don't exists
- Add test

https://issues.apache.org/jira/browse/GROOVY-4990

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
@@ -852,6 +852,8 @@ public class ResourceGroovyMethods extends DefaultGroovyMethodsSupport {
     public static void write(File file, String text, String charset, boolean writeBom) throws IOException {
         Writer writer = null;
         try {
+            file.getCanonicalFile().getParentFile().mkdirs();
+
             FileOutputStream out = new FileOutputStream(file);
             if (writeBom) {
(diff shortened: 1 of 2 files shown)


paulk-asert:
Hi, thanks for your PR. While automatically creating the parent directories is in some ways super nice, it might also be surprising for folks familiar with current Java behavior. I think I prefer the `mkParentDirs` suggestion mentioned in the Jira issue, though I'd probably call it `createParentDirectories` these days to match the newer nio styling. Then it could also be used for `setBytes` and so on, not just `write`. What do you think about that suggestion and if you like the idea, would you feel up for altering the PR?
Suggested methods to add:
```
File ResourceGroovyMethods#createParentDirectories(File self)
Path NioExtensions#createParentDirectories(Path self, FileAttribute<?>... attrs)
```
Usage:
```
file.createParentDirectories().write(...)
path.createParentDirectories().bytes = ...
```
What do you think? Or if you want me to progress, just let me know.

xaaria:
@paulk-asert We implented this (easy) feature because participating an open source project is a mandatory part of our university's course _Software Engineering Methodologies_. So to be honest I [we] have no further opinion about this.

paulk-asert:
I merged and tweaked your test and modified the DGM methods as per the earlier suggestion. Thanks.

PR #1555, opened 2021-04-12
GROOVY-10031: write NEWARRAY/ANEWARRAY for all single-dimensional arrays

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10031


danielsun1106:
Merged. Thanks!

PR #1556, opened 2021-04-12
GROOVY-10034: do not cast non-primitive array when target is Object[]

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10034

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
@@ -338,6 +338,12 @@ public class OperandStack {
             return;
         }
 
+        // GROOVY-10034: no need to cast non-primitive array when target is "java.lang.Object[]"
+        if (targetType.isArray() && targetType.getComponentType().equals(ClassHelper.OBJECT_TYPE)
+                && top.isArray() && !ClassHelper.isPrimitiveType(top.getComponentType())) {
+            return;
+        }
+
         boolean primTarget = ClassHelper.isPrimitiveType(targetType);
         boolean primTop = ClassHelper.isPrimitiveType(top);
 
--- a/src/test/groovy/bugs/Groovy10034.groovy
+++ b/src/test/groovy/bugs/Groovy10034.groovy
@@ -1,0 +1,17 @@
+package groovy.bugs
+
+import org.codehaus.groovy.classgen.asm.AbstractBytecodeTestCase
+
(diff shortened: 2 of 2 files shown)


src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
```diff
@@ -338,6 +338,12 @@ private void doConvertAndCast(ClassNode targetType, final boolean coerce) {
             return;
         }
 
+        // GROOVY-10034: no need to cast non-primitive array when target is "java.lang.Object[]"
```
danielsun1106, line 341: How about making the check more general? If the `top` component type is inherited from `target` component type, no need to cast type. The following code could be tweaked as:
```java
        if (targetType.isArray() && !ClassHelper.isPrimitiveType(targetType.getComponentType()) 
                  && top.isArray() && !ClassHelper.isPrimitiveType(top.getComponentType())
                  && targetType.getComponentType().isAssignableFrom(top.getComponentType())) {
            return;
        }
```

danielsun1106: It seems that we could make the check simpler, we do not add extra checks but tweak the existing check:

https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java#L334
```java
if (targetType == top) return;
```
tweaked as:
```java
if (targetType.getTypeClass().isAssignableFrom(top.getTypeClass())) return;
```

eric-milles: I wanted to keep the scope limited.  This is so widely used.  You could certainly add a new issue that references this one and I could take a look at it after I get through with array stuff.

eric-milles: `getTypeClass()` is a known unsafe method to use.  I would caution against that without thorough testing.

danielsun1106: OK. 
Here is the JIRA ticket to track the discusstion: https://issues.apache.org/jira/browse/GROOVY-10035

danielsun1106:
Merged. Thanks!

danielsun1106:
I created a PR https://github.com/apache/groovy/pull/1561 based on your current PR to tweak the check a bit.

PR #1557, opened 2021-04-12
GROOVY-10029: SC: add toArray call for "Type[] array = collectionOfType"

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10029

I have verified that the end result no longer passes through `DefaultTypeTransformation#castToType`.  However there are more total bytecode instructions in the after state due to safe method call handling and cast of `toArray` result to the target type.  I suspect the new sequence is faster but I have no direct evidence of that.  Is that a concern?

**Before**
```
  public static void main(java.lang.String... args);
     0  ldc <Class C> [2]
     2  invokedynamic 0 invoke(java.lang.Class) : java.lang.Object [53]
     7  invokedynamic 1 cast(java.lang.Object) : java.lang.String[] [59]
    12  astore_1 [strings]
    13  aload_1 [strings]
    14  pop
    15  return
```

**After** _I have edited because GROOVY-10031 and GROOVY-10034 have reduced the sequence at offset 10_
```
  public static void main(java.lang.String... args);
     0  invokestatic C.m() : java.util.List [42]
     3  dup
     4  astore_1
     5  ifnull 27
     8  aload_1
     9  iconst_0
    10  anewarray java.lang.String [44]
    xx  invokeinterface java.util.List.toArray(java.lang.Object[]) : java.lang.Object[] [64] [nargs: 2]
    xx  goto xx
    xx  aconst_null
    xx  invokedynamic 0 cast(java.lang.Object[]) : java.lang.String[] [67]
    xx  astore_2 [strings]
    xx  aload_2 [strings]
    xx  pop
    xx  return
```

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ClassHelper.java
+++ b/src/main/java/org/codehaus/groovy/ast/ClassHelper.java
@@ -64,6 +64,7 @@ import java.lang.ref.SoftReference;
 import java.lang.reflect.Modifier;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


eric-milles:
For reference, the ecj/javac equivalent for `String[] strings = m().toArray(new String[0]);` is:
```
  public static void main(java.lang.String[] args);
     0  invokestatic C.m() : java.util.List [30]
     3  iconst_0
     4  anewarray java.lang.String [18]
     7  invokeinterface java.util.List.toArray(java.lang.Object[]) : java.lang.Object[] [32] [nargs: 2]
    12  checkcast java.lang.String[] [38]
    15  astore_1 [strings]
    16  return
```

danielsun1106:
Merged. Thanks!

PR #1558, opened 2021-04-12
GROOVY-9017 DGM methods

tomikoskinen:
https://issues.apache.org/jira/browse/GROOVY-9017

Instead of making keywords for and, or and not, we added a Boolean not() method in DefaultGroovyMethods to get most of what was wanted. DGM already had a or() and and() methods.

This PR was made by:
@tomikoskinen
@jonspe
@Justsofun
@rhanlol


eric-milles:
`and` and `or` work because they are binary operations so you can write "if (a.and(b))".  `not` is a unary operation semantically.  So the way you have it you could have "if (a.not())".  Are you looking for "if (not(a))"?  For that, you could supply a DGM like this:
```java
public static Boolean not(Object self, Boolean target)
```
With this, you could write "if (not(a))" and "self" would the the object that is the implicit "this", which you can ignore.

eric-milles:
Also, when you submit for Pull Request, could you squash your commits to just one unless there is value in the separation?  The ranges change got merged without a squash and now we have over a dozen commit lines for a single feature change.

eric-milles:
Also, you have GROOVY-9797 changes as part of this PR.

eric-milles:
There is a commented-out `negate` method in there.  Can you delete it as part of this change?  Also, why return null?  Do `and`, `or`, `xor` and `implies` return null?  Would "!Boolean.TRUE.equals(left)" satisfy your truth table?

tomikoskinen:
@eric-milles We made a new Pull Request.
Also is it intended that when you call `null.or(true)` it throws an exception, but when you call `true.or(null)` it returns true?

eric-milles:
If you wan't "null.or(true)" to work, you would need to provide DGM "or(NullObject,Boolean)".  This may be worth creating a separate JIRA issue for consideration.  Or you can use "or" in this fashion to be null-safe "null?.or(true)".

PR #1559, opened 2021-04-12
GROOVY-10028: STC: support "Type[] array = streamOfTypeOrSubtype"

eric-milles:
and "Collection<Type> collection = streamOfTypeOrSubtype"
and "List<Type> list = streamOfTypeOrSubtype"
and "Set<Type> set = streamOfTypeOrSubtype"

https://issues.apache.org/jira/browse/GROOVY-10028


danielsun1106:
Merged. Thanks!

PR #1561, opened 2021-04-14
GROOVY-10035: Eliminate redundant type cast

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-10035

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
@@ -332,7 +332,8 @@ public class OperandStack {
 
         ClassNode top = stack.get(size - 1);
         targetType = targetType.redirect();
-        if (ClassNodeUtils.isCompatibleWith(top, targetType)) return;
+        if (top == targetType /* for better performance */
+                || ClassNodeUtils.isCompatibleWith(top, targetType)) return;
 
         if (coerce) {
             controller.getInvocationWriter().coerce(top, targetType);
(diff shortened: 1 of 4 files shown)


src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
```diff
@@ -331,19 +332,13 @@ private void doConvertAndCast(ClassNode targetType, final boolean coerce) {
...
         if (coerce) {
             controller.getInvocationWriter().coerce(top, targetType);
             return;
         }
 
-        // GROOVY-10034: no need to cast non-primitive array when target is "java.lang.Object[]"
-        if (targetType.isArray() && targetType.getComponentType().equals(ClassHelper.OBJECT_TYPE)
-                && top.isArray() && !ClassHelper.isPrimitiveType(top.getComponentType())) {
-            return;
-        }
```
eric-milles, line 345: I thought the goal was to simplify from GROOVY-10034.  Now we have multiple new public APIs to support?  `Object[]` can accept any kind of array regardless of dimension, except one-dimensional primitive array.  This was accounted for in the original condition.

`targetType == top` is a much faster comparison then always making a method call.  `==` or `.equals` should probably be kept before making a method call.

danielsun1106: > I thought the goal was to simplify from GROOVY-10034. Now we have multiple new public APIs to support? Object[] can accept any kind of array regardless of dimension, except one-dimensional primitive array. This was accounted for in the original condition.

Actually the goal is to tweak the fix for GROOVY-10034, not only simplify the check but also make it general to eliminate more redundant type cast.
( See https://github.com/apache/groovy/pull/1556#discussion_r611815705 )

> targetType == top is a much faster comparison then always making a method call. == or .equals should probably be kept before making a method call.

OK. Let's add the check `targetType == top` before `isCompatibleWith` method call for better performance.

PR #1563, opened 2021-04-14
GROOVY-10036: STC: pass explicit type arguments of extension method call

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10036


danielsun1106:
Merged. Thanks!

PR #1565, opened 2021-04-17
GROOVY-9800: ClassNode#toString includes generics of type arguments

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9800


danielsun1106:
Merged. Thanks!

PR #1569, opened 2021-04-23
Fix typoes

zhang13690:
1. `hash` should be `dash`, right?
2. Should be `java.math.BigDecimal`


paulk-asert:
Correct with "dash", merged, thanks!

PR #1570, opened 2021-04-24
GROOVY-10056: Inferred parameter type of lambda expression for multi-…

danielsun1106:
…dimensions array is not correct

https://issues.apache.org/jira/browse/GROOVY-10056

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -428,7 +428,7 @@ public abstract class StaticTypeCheckingSupport {
      * @return -1 if no match, 0 if the last argument is exactly the vararg type and 1 if of an assignable type
      */
     static int lastArgMatchesVarg(final Parameter[] parameters, final ClassNode... argumentTypes) {
-        if (!isVargs(parameters)) return -1;
+        if (!isVargs(parameters, argumentTypes)) return -1;
         // case length ==0 handled already
         // we have now two cases,
         // the argument is wrapped in the vargs array or
@@ -480,8 +480,14 @@ public abstract class StaticTypeCheckingSupport {
         return false;
     }
 
-    static boolean isVargs(final Parameter[] parameters) {
+    static boolean isVargs(final Parameter[] parameters, ClassNode[] argumentTypes) {
         if (parameters == null || parameters.length == 0) return false;
+
+        // GROOVY-10056: Inferred parameter type of lambda expression for multi-dimensions array is not correct
+        if (null != argumentTypes && argumentTypes.length == parameters.length && argumentTypes[argumentTypes.length - 1].isArray()) {
+            return false;
+        }
+
         return (parameters[parameters.length - 1].getType().isArray());
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -480,8 +480,14 @@ public static boolean isAssignableTo(ClassNode type, ClassNode toBeAssignedTo) {
...
 
-    static boolean isVargs(final Parameter[] parameters) {
+    static boolean isVargs(final Parameter[] parameters, ClassNode[] argumentTypes) {
         if (parameters == null || parameters.length == 0) return false;
+
+        // GROOVY-10056: Inferred parameter type of lambda expression for multi-dimensions array is not correct
+        if (null != argumentTypes && argumentTypes.length == parameters.length && argumentTypes[argumentTypes.length - 1].isArray()) {
+            return false;
+        }
+
```
eric-milles, line 490: Instead of changing the definition of what is a variadic method, can you change the lastArgMatchesVarg result only?  A method is variadic regardless of the arguments passed.

eric-milles:
Is there any overlap with https://issues.apache.org/jira/browse/GROOVY-9813 and this?  Or are they satisfied out of very separate paths?

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
danielsun1106: I tried to revert all changes and just to refine the check as you suggested:

**Original**
https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java#L5286

**Refined**
```
boolean isVargs = -1 != lastArgMatchesVarg(parameters, expressions.stream().map(e -> getDeclaredOrInferredType(e)).toArray(ClassNode[]::new));
```

but the build will be broken, then we can not debug... 
`lastArgMatchesVarg` is much more complicated than `isVargs`, it is hard to replace `isVargs`.

danielsun1106: In order to keep the method `isVargs` as it is, I even tried the following approach, but the build will still be broken...

**Original**
https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java#L5286

**Refined**
```java
        boolean isVargs = isVargs(parameters);
        if (isVargs) {
            ClassNode[] argumentTypes = expressions.stream().map(e -> getDeclaredOrInferredType(e)).toArray(ClassNode[]::new);
            if (null != argumentTypes && argumentTypes.length == parameters.length && argumentTypes[argumentTypes.length - 1].isArray()) {
                isVargs = false;
            }
        }
```

eric-milles: I could take a look at this case when I get through GROOVY-10053 and GROOVY-10054.  I'm trying to look over all the generics fallout from 4.0a3 as a whole.

danielsun1106: OK. Thanks.
I'll close the PR without merging as some further tweak could be done.

PR #1571, opened 2021-04-25
GROOVY-10056: Inferred parameter type of lambda expression for multi-…

danielsun1106:
…dimensions array is not correct

https://issues.apache.org/jira/browse/GROOVY-10056

This may be the minimum change for the issue.

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
@@ -5305,10 +5305,7 @@ public class StaticTypeCheckingVisitor extends ClassCodeVisitorSupport {
                                             applyGenericsContextToParameterClass(resolvedPlaceholders, paramType));
                         }
                     }
-                    if (isVargs && lastArg && argumentType.isArray()) {
-                        argumentType = argumentType.getComponentType();
-                    }
-                    if (isVargs && lastArg && paramType.isArray()) {
+                    if (isVargs && lastArg && paramType.isArray() && !argumentType.isArray()) {
                         paramType = paramType.getComponentType();
                     }
                     argumentType = wrapTypeIfNecessary(argumentType);
--- a/src/test/groovy/bugs/Groovy10056.groovy
+++ b/src/test/groovy/bugs/Groovy10056.groovy
@@ -1,0 +1,44 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -5305,10 +5305,7 @@ protected ClassNode inferReturnTypeGenerics(final ClassNode receiver, final Meth
                                             applyGenericsContextToParameterClass(resolvedPlaceholders, paramType));
                         }
                     }
-                    if (isVargs && lastArg && argumentType.isArray()) {
-                        argumentType = argumentType.getComponentType();
-                    }
-                    if (isVargs && lastArg && paramType.isArray()) {
+                    if (isVargs && lastArg && paramType.isArray() && !argumentType.isArray()) {
                         paramType = paramType.getComponentType();
                     }
```
eric-milles, line 5310: What about tying the checks together?  I have tried this for the test case, but have run out of time today to run the full test suite.
```java
                    if (isVargs && lastArg && argumentType.isArray() && paramType.isArray()) {
                        argumentType = argumentType.getComponentType();
                        paramType = paramType.getComponentType();
                    }
```

danielsun1106: As the following code will make `argumentType` and `paramType` non-array type, so `isVargs && lastArg && argumentType.isArray() && paramType.isArray()` will be always `false` IMO.

https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java#L5297-L5300

Anyway, let me try according to you suggestion.

eric-milles: So probably not correct to reduce the arrays together.  But I don't think it is correct to reduce the parameter type only in the argument is not still an array.  There are more complete variadic checks in other blocks; I think the ones in `inferReturnTypeGenerics` are quite simplistic.  Like if the param type is `Object[]`, it accepts any array type.

I did not get the error for "new ListExpression(Arrays.asList(..." and I'm not sure why that is.

danielsun1106: > There are more complete variadic checks in other blocks; I think the ones in inferReturnTypeGenerics are quite simplistic.

Yep. The variadic checks here is not complete.

> So probably not correct to reduce the arrays together.

The cause of GROOVY-10056 and the above error `new ListExpression(Arrays.asList(...` is reducing type array not correctly.
```java
                    if (isVargs && lastArg && argumentType.isArray()) {
                        argumentType = argumentType.getComponentType();
                    }
                    if (isVargs && lastArg && paramType.isArray()) {
                        paramType = paramType.getComponentType();
                    }
```
(https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java#L5310-L5315)

> I did not get the error for "new ListExpression(Arrays.asList(..." and I'm not sure why that is.

You could try run tests with the gradle option `--no-build-cache`.

To be honest, I have no better idea to fix GROOVY-10056 for now.

PR #1572, opened 2021-04-26
"del" ASCII character also should be escaped

eustas:
Otherwise, after printing out the result, it does not represent input!


danielsun1106:
Merged. Thanks.
https://github.com/apache/groovy/commit/4476af5e67bedcdb2c3b80528afcb991610fbcbc

PR #1573, opened 2021-04-28
Move stream-related extension methods to StreamGroovyMethods

eric-milles:
Now that Java 8 is the minimum, stream methods do not need to be located in the vmplugin hierarchy.  Separate the growing number of stream-based extension methods into their own class.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -231,10 +231,10 @@ tasks.named('test') {
 
 if (!JavaVersion.current().java10Compatible) {
     logger.lifecycle '''
-    **************************************** WARNING ********************************************
-    ******   You are running the build with an older JDK. NEVER try to release with 1.8.   ******
-    ******   You must use a JDK 10+ in order to compile all features of the language.     ******
-    *********************************************************************************************
+**************************************** WARNING ********************************************
+******   You are running the build with an older JDK. NEVER try to release with 1.8.   ******
+******   You must use a JDK 10+ in order to compile all features of the language.      ******
+*********************************************************************************************
 '''
 }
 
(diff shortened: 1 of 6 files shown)


paulk-asert:
One thing which we have found in the past when moving DGM methods is that some frameworks like Grails, Micronaut or Spock might use the DGM methods directly. They are public static methods after all even though we regard them as somewhat internal and expose them as methods "only" available through the runtime or when using @CS. So, we adopted the convention of marking old variants with @Deprecated which are ignored when loading DGM methods. Plugin DGM methods might be considered even slightly more internal since in theory they might not be activated on older JVM versions but they are still there as static methods, so it might pay to give them the same treatment. We tend to then delete those deprecated methods after a major version or two. For the same reason, I'd target only Groovy 4 for this change.

paulk-asert:
Actually, Groovy 4 might be a good time to refine how we deactivate extension methods. Re-purposing @Deprecated has always been a compromise. Perhaps we should add an @Inactive annotation and use that instead. I'll start a discussion on the mailing list, but either way, we'd put back the plugin DGM methods and indicate they are inactive. We often truncate the Javadoc when we make such DGM methods inactive.

eric-milles:
I'm fine with deprecated as a way to disable as extension but leave for direct use.  I would be more on board with a way to add extension methods without having to make them public so that they cannot be used outside of the extension mechanism.

So when the v5-v7 vmplugins were combined, was binary compatibility not affected similarly? 
 https://github.com/apache/groovy/commit/6e10c35b7fdc7c6b213372d7fa771f723ca5906a

I could add the deprecated methods back to PluginDefaultGroovyMethods, if there was a string reason to do so.

eric-milles:
You can merge one or both commits

paulk-asert:
All of the v5-v7 classes exist (but aren't used and are deprecated) on the GROOVY_3_0_X branch but they are removed in master. I'd suggest we do the same thing once Groovy 4 is released and we prepare for Groovy 5.

paulk-asert:
Merged, thanks!

PR #1574, opened 2021-05-03
GROOVY-9804: Support TOML

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-9804

### DIFF
--- a/subprojects/groovy-toml/src/test/groovy/groovy/toml/TomlJavadocAssertionTest.groovy
+++ b/subprojects/groovy-toml/src/test/groovy/groovy/toml/TomlJavadocAssertionTest.groovy
@@ -1,0 +1,28 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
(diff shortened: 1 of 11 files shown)


subprojects/groovy-toml/src/test/groovy/groovy/toml/TomlJavadocAssertionTest.groovy
```diff
@@ -0,0 +1,28 @@
...
+ */
+package groovy.toml
+
+import groovy.test.JavadocAssertionTestSuite
+import org.junit.runner.RunWith
+import org.junit.runners.Suite
+
+@RunWith(Suite)
+@Suite.SuiteClasses(JavadocAssertionTestSuite)
+class TomlJavadocAssertionTest {
```
paulk-asert, line 27: Only needed if we have `class="groovyTestCase"` style tests within the Javadoc/Groovydoc. There doesn't currently seem to be any such tests but it might be nice if we did in TomlSlurper or TomlBuilder.

danielsun1106: Understood, so you mean we should remove it?

paulk-asert: I'd remove if we don't enhance the documentation with such tests. I'd prefer to enhance the doco, e.g. one or two examples like this one from `YamlBuilder`:
```
/**
     * Named arguments can be passed to the YAML builder instance to create a root YAML object
     * <p>
     * Example:
     * <pre><code class="groovyTestCase">
     * def yaml = new groovy.yaml.YamlBuilder()
     * yaml name: "Guillaume", age: 33
     *
     * assert yaml.toString() == '''---
     * name: "Guillaume"
     * age: 33
     * '''
     * </code></pre>
     *
     * @param m a map of key / value pairs
     * @return a map of key / value pairs
     */
    public Object call(Map m) {
        return jsonBuilder.call(m);
    }
```

danielsun1106: OK. Let me add some sample codes in the javadoc later.

danielsun1106: @paulk-asert The sample codes have been added: https://github.com/apache/groovy/pull/1574/commits/560ed65abe82c1260778dd03e6bf114caa433cc6

paulk-asert (approved):

PR #1577, opened 2021-05-10
GROOVY-3015: add GroovyInterceptable check to closure resolve strategy

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-3015


eric-milles:
see also GROOVY-4610

paulk-asert (approved):
GroovyInterceptable isn't frequently used these days, so probably why this issue hasn't received any attention. If we had ever completed our MOP 2 redesign we were thinking of absorbing it into other functionality but things have moved in a different direction since then. So we might as well fix and your change LGTM.

PR #1578, opened 2021-05-12
GROOVY-9632: Java 8 Type Param Annotation Not Generated in Byte Code

paulk-asert:
Still needs tests. To illustrate what is enabled, consider the following Bean Validation 2 example:
```
@Grab('org.hibernate.validator:hibernate-validator:7.0.1.Final')
@Grab('org.hibernate.validator:hibernate-validator-cdi:7.0.1.Final')
@Grab('org.glassfish:jakarta.el:4.0.0')
import jakarta.validation.constraints.*
import jakarta.validation.*
import groovy.transform.*

@Canonical
class Car {
    @NotNull @Size(min = 2, max = 14) String make
    @Min(1L) int seats
    List<@NotBlank String> owners
}

def validator = Validation.buildDefaultValidatorFactory().validator

def violations = validator.validate(new Car(make: 'T', seats: 1))
assert violations*.message == ['size must be between 2 and 14']

violations = validator.validate(new Car(make: 'Tesla', owners: ['']))
assert violations*.message.toSet() == ['must be greater than or equal to 1', 'must not be blank'] as Set

violations = validator.validate(new Car(make: 'Tesla', owners: ['Elon'], seats: 2))
assert !violations
```
And the following Jqwik example:
```
@Grab('net.jqwik:jqwik:1.5.1')
import net.jqwik.api.*
import net.jqwik.api.constraints.*

class PropertyBasedTests {
    @Property
    def uniqueInList(@ForAll @Size(5) @UniqueElements List<@IntRange(min = 0, max = 10) Integer> aList) {
        //println aList
        assert aList.size() == aList.toSet().size()
        assert aList.every{ anInt -> anInt >= 0 && anInt <= 10 }
    }
}
```

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -343,6 +343,7 @@ import static org.codehaus.groovy.ast.tools.GeneralUtils.assignX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.stmt;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.varX;
 import static org.codehaus.groovy.classgen.asm.util.TypeUtil.isPrimitiveType;
+import static org.codehaus.groovy.classgen.asm.util.TypeUtil.isPrimitiveVoid;
 import static org.codehaus.groovy.runtime.DefaultGroovyMethods.asBoolean;
 import static org.codehaus.groovy.runtime.DefaultGroovyMethods.last;
 
@@ -873,7 +874,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     @Override
     public List<ClassNode> visitCatchType(final CatchTypeContext ctx) {
         if (!asBoolean(ctx)) {
-            return Collections.singletonList(ClassHelper.OBJECT_TYPE);
+            return Collections.singletonList(ClassHelper.DYNAMIC_TYPE);
         }
 
         return ctx.qualifiedClassName().stream()
@@ -1152,13 +1153,13 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
                     outerClass,
                     outerClass.getName() + "$" + className,
                     modifiers,
-                    ClassHelper.OBJECT_TYPE
+                    ClassHelper.OBJECT_TYPE.getPlainNodeReference()
             );
         } else {
             classNode = new ClassNode(
                     packageName + className,
                     modifiers,
-                    ClassHelper.OBJECT_TYPE
+                    ClassHelper.OBJECT_TYPE.getPlainNodeReference()
             );
         }
 
@@ -1186,7 +1187,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
                 }
                 superClass = scs[0];
             } else {
-                superClass = ClassHelper.OBJECT_TYPE;
+                superClass = ClassHelper.OBJECT_TYPE.getPlainNodeReference();
             }
             classNode.setSuperClass(superClass);
             classNode.setInterfaces(this.visitTypeList(ctx.is));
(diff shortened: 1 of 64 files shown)


src/main/java/org/codehaus/groovy/ast/ClassNode.java
```diff
@@ -1420,6 +1421,16 @@ public void setUsingGenerics(boolean usesGenerics) {
...
+        n.setRedirect(redirect());
+        if (isArray()) {
+            n.componentType = redirect().getComponentType();
+        }
+        return n;
+    }
+
     public ClassNode getPlainNodeReference() {
         if (ClassHelper.isPrimitiveType(this)) return this;
         ClassNode n = new ClassNode(name, modifiers, superClass, null, null);
```
danielsun1106, line 1436: We could remove the duplicated code by invoking the method:
`return getPlainPrimitiveReference()`

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -873,7 +873,7 @@ public ExpressionStatement visitResource(final ResourceContext ctx) {
     @Override
     public List<ClassNode> visitCatchType(final CatchTypeContext ctx) {
         if (!asBoolean(ctx)) {
-            return Collections.singletonList(ClassHelper.OBJECT_TYPE);
+            return Collections.singletonList(ClassHelper.OBJECT_TYPE.getPlainNodeReference());
```
danielsun1106, line 876: Out of curiosity, why should we invoke `getPlainNodeReference()` on `OBJECT_TYPE` instead of using  `OBJECT_TYPE` directly? Also, I find many similar changes in the PR, so I assume lots of changes from `==` to `equals(...)` is caused by the change too.

paulk-asert: Yes, it is a change in numerous previous assumptions. It would be good for others to think the change through too. It is somewhat of a compromise to get the needed functionality and minimise my perceived potential breakages in existing frameworks and existing code. Basically, for this:
```
@Anno1
def x = 'foo'
@Anno2
def y = 42
```
The type of the field is `Object`. If `Anno1` and `Anno2` are field annotations, they are stored on the `FieldNode` itself and all is well. It is trickier if `Anno1` and `Anno2` are type annotations. We can either try to stored them along with other annotations in the `FieldNode` and extract them out when writing bytecode. Indeed that is what we did for our initial implementation for classes and methods/method parameters.

I in fact started heading further down this path when adding support for generic types. I made `GenericsType` extend `AnnotatedNode` but then you need to also account for annotations on the generic type bounds and later for casts, new expressions, catch target expressions and the list goes on. So I headed down the current path and store the type annotations on the types. Again there are implementation details which impacted so choices I made, e.g. we currently get normal annotations from the redirect classnode if it exists. Adding a separate list of type annotations seemed the safest path forward. So going back to `getPlainNodeReference`, we obviously don't want `Anno1` and `Anno2` to be present in all places where `OBJECT` occurs. For memory efficiency we currently share OBJECT, primitives and void throughout our data structures as pseudo immutable types which we can't do if type annotations are present.

I have tried to keep the changes so that, as a second pass, I think we could look at retaining the existing approach in places where no annotations occur. This might save memory but could cause strange behavior if AST transforms try to add type annotations to nodes. We'd probably need to enforce our pseudo immutable types if we do that second pass.

src/main/java/org/codehaus/groovy/ast/ClassNode.java
paulk-asert: That would be possible. Let's see first if others feel the method makes sense. We could remove the new method and just always call `getPlainNodeReference()` if we removed the primitive check line. I just felt that might be more likely to break existing external code.

danielsun1106:
+1

danielsun1106:
Merged. Thanks!

PR #1580, opened 2021-05-22
GROOVY-9272: Support switch expression

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-9272


eric-milles:
Looks good so far.  When using `callX` to create a method call to the closure, you should `setImplicitThis(false)` on the returned `MethodCallExpression`.

Any idea of level of compatibility with the Java switch expression variants?  Does this cover them all or are there known limitations?  Is there any measure of performance impact versus current parser?

danielsun1106:
> Looks good so far. When using callX to create a method call to the closure, you should setImplicitThis(false) on the returned MethodCallExpression.

Tweaked. Thanks!

> Any idea of level of compatibility with the Java switch expression variants?

Groovy should be able to support all variants of switch expression of the latest JEP.

> Is there any measure of performance impact versus current parser?

No, it should be a bit impact because the build time does change obviously.

eric-milles:
Is there any more coming for this change?  Or is the plan to wait until after 4.0b1?  I'd like to see it as part of the beta myself.

danielsun1106:
> Is there any more coming for this change? Or is the plan to wait until after 4.0b1? I'd like to see it as part of the beta myself.

No more changes for this PR, but I am waiting @paulk-asert to review it ;-)

paulk-asert:
Yes, I've been busy on other things the last few weeks. I'm looking at it this week. I won't proceed with beta-1 until finished.

PR #1581, opened 2021-05-23
better distinction between Object and dynamic types

nineninesevenfour:
Signed-off-by: Harald Fassler <harald.fassler+9974@gmail.com>

A proposal (draft) for a step towards better distinction between types of java.lang.Object and dynmic types.
More tests need to be added.

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -874,7 +874,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     @Override
     public List<ClassNode> visitCatchType(final CatchTypeContext ctx) {
         if (!asBoolean(ctx)) {
-            return Collections.singletonList(ClassHelper.DYNAMIC_TYPE);
+            return Collections.singletonList(ClassHelper.makeDynamicType());
         }
 
         return ctx.qualifiedClassName().stream()
@@ -1706,7 +1706,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     @Override
     public ClassNode visitReturnType(final ReturnTypeContext ctx) {
         if (!asBoolean(ctx)) {
-            return ClassHelper.OBJECT_TYPE.getPlainNodeReference();
+            return ClassHelper.makeDynamicType();
         }
 
         if (asBoolean(ctx.type())) {
@@ -3633,7 +3633,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     public Parameter[] visitStandardLambdaParameters(final StandardLambdaParametersContext ctx) {
         if (asBoolean(ctx.variableDeclaratorId())) {
             VariableExpression variable = this.visitVariableDeclaratorId(ctx.variableDeclaratorId());
-            Parameter parameter = new Parameter(ClassHelper.OBJECT_TYPE.getPlainNodeReference(), variable.getName());
+            Parameter parameter = new Parameter(ClassHelper.makeDynamicType(), variable.getName());
             configureAST(parameter, variable);
             return new Parameter[]{parameter};
         }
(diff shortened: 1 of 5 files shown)


nineninesevenfour:
@paulk-asert 
Hello Paul,
while working on tests, I stumbled once more over this method:
```
@Override
    public void visitCatchStatement(final CatchStatement cs) {
        resolveOrFail(cs.getExceptionType(), cs);
        if (ClassHelper.isDynamicTyped(cs.getExceptionType())) {
            cs.getVariable().setType(ClassHelper.make(Exception.class));
        }
        super.visitCatchStatement(cs);
    }
```
see https://github.com/nineninesevenfour/groovy/blob/93ad2a577c6facc3f5620fd4a6c8167daa43ee74/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java#L1561

And I asked myself if the type shouldn't rather be coerced to `Throwable.class` than `Exception.class` to catch unchecked exceptions, too? 🤔
What do you think?

Thank you!
Harald

nineninesevenfour:
> @paulk-asert
> Hello Paul,
> while working on tests, I stumbled once more over this method:
> 
> ```
> @Override
>     public void visitCatchStatement(final CatchStatement cs) {
>         resolveOrFail(cs.getExceptionType(), cs);
>         if (ClassHelper.isDynamicTyped(cs.getExceptionType())) {
>             cs.getVariable().setType(ClassHelper.make(Exception.class));
>         }
>         super.visitCatchStatement(cs);
>     }
> ```
> 
> see https://github.com/nineninesevenfour/groovy/blob/93ad2a577c6facc3f5620fd4a6c8167daa43ee74/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java#L1561
> 
> And I asked myself if the type shouldn't rather be coerced to `Throwable.class` than `Exception.class` to catch ~unchecked exceptions~ `java.langError` and other Throwables, too? 🤔
> What do you think?
> 
> Thank you!
> Harald
> 
> cc: @eric-milles

Discussed on Slack and catching Throwable seems to be a typical anti-pattern. Question is answered. :-)

eric-milles:
On the mailing list I suggested the possibility of deprecating `isDynamicTyped` in `FieldNode`, `Parameter`, etc.  And then removing their use from the compiler/runtime.  IMO there is only a very small bit of code in the compiler/editor that cares about the difference between "Object name" and "def name".

To be able to distinguish between those two, I have considered "ClassNode DYNAMIC_TYPE = ClassHelper.make(Object.class); ClassNode OBJECT_TYPE = DYNAMIC_TYPE.getPlainNodeReference();" so that referential equality is broken between the two constants.  However, I think there is probably a lot of code out there that checks "type == OBJECT_TYPE" or "type.redirect() == OBJECT_TYPE".  So this too would be a breaking change and possibly more impactful than the current condition of `isDynamicTyped`.

WRT your pull request, I don't think adding more API to `ClassNode` is the answer.  I'm still hoping for a reduction in API in areas where a consistent answer cannot be provided.  And backwards compatibility should be maintained if at all possible.

nineninesevenfour:
@eric-milles 
Hello Eric,
thank you very much for your feedback, it is very appreciated!
Independently from the way it is actually solved, I think one benefit of my proposed pull request would have been that through ClassHelper.makeDynamicType() the four relevant places in AstBuilder are very easy to determine.
IMHO found it also more easy to read than just ClassHelper.OBJECT_TYPE.getPlainNodeReference();
Anyway I agree that we have to come up with a better solution. 
I will close this pull request and keep the tests in my fork, they could be helpful later maybe.
Thank you and kind regards, Harald

paulk-asert:
I echo Eric's thanks for your efforts. Some parts of where I think we will end up could be quite similar to some parts of your PR, so nice work in your analysis! I am expecting we won't fully get rid of the dynamic type concept (at least not straight away), perhaps we'll use node metadata though instead of changing the API. I'd actually like to deprecate DYNAMIC_TYPE but we'll have to see how far we can get with that.

PR #1584, opened 2021-05-26
GROOVY-9985, GROOVY-10111: STC: check array sizes and values not as cast

eric-milles:
size checks as described in 9994 comments

https://issues.apache.org/jira/browse/GROOVY-9985
https://issues.apache.org/jira/browse/GROOVY-9994
https://issues.apache.org/jira/browse/GROOVY-10111

### DIFF
--- a/src/test/groovy/transform/stc/ArraysAndCollectionsSTCTest.groovy
+++ b/src/test/groovy/transform/stc/ArraysAndCollectionsSTCTest.groovy
@@ -27,10 +27,11 @@ class ArraysAndCollectionsSTCTest extends StaticTypeCheckingTestCase {
         assertScript '''
             String[] strings = ['a','b','c']
             String str = strings[0]
+            assert str == 'a'
         '''
     }
 
-    void testArrayElementReturnType() {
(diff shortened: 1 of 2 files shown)


src/test/groovy/transform/stc/ArraysAndCollectionsSTCTest.groovy
```diff
@@ -27,26 +27,55 @@ class ArraysAndCollectionsSTCTest extends StaticTypeCheckingTestCase {
...
 
-    void testArrayElementReturnType() {
+    void testArrayElementTypeInference() {
         shouldFailWithMessages '''
             String[] strings = ['a','b','c']
             int str = strings[0]
         ''', 'Cannot assign value of type java.lang.String to variable of type int'
     }
 
+    void testRightComponentTypeInArray() {
```
paulk-asert, line 41: Minor, but I'd suggest "Compatible" or "Correct" instead of "Right". "Right" is the opposite of "Wrong" but also "Left".

paulk-asert (approved):

PR #1585, opened 2021-05-27
GROOVY-10112: IndexedProperty AST transform should only provide the g…

paulk-asert:
…etter for immutable fields


danielsun1106:
Merged. Thanks!

PR #1587, opened 2021-06-05
GROOVY-10113: check class cycle for outer types and type parameters

eric-milles:
Cycle errors are consistent with ecj, with a separate error message for implementing self or own inner.  Errors are now associated with the extended/implemented type instead of the primary type.  This allows distinct errors in the case of cycle issues with multiple super types.

Each scenario added in the new test was lacking a cycle error.  There was one example of a type that implemented its own inner interface (GROOVY-8048), but it was not important for the bug.  I think it is safer to bar this kind of thing so we don't run into cycles when navigating outers, supers and interfaces.

https://issues.apache.org/jira/browse/GROOVY-10113


danielsun1106:
LGTM

PR #1590, opened 2021-06-05
GROOVY-7867: cast to collection: attach exception of constructor as suppressed exception

nineninesevenfour:
Proposed solution to https://issues.apache.org/jira/browse/GROOVY-7867

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
+++ b/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
@@ -382,6 +382,7 @@ public class DefaultTypeTransformation {
         }
 
         Exception nested = null;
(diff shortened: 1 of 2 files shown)


eric-milles:
Instead of stashing `ex` into `e` and then extracting in the `GroovyCastException` constructor, can you save the reference as a peer to the `nested` local variable and then call `addSuppressed` on the new cast exception instance?  I can't really say how many places make use of the `GroovyCastException` constructor.

Also, you don't need a package statement for your test scripts.  They should run just fine from the default package.  Usually we add that only if there is a need to check a scenario involving separate packages.

nineninesevenfour:
@eric-milles 
Thank you so much for your review! I did the requested changes. Honestly I have no idea why I made it unnecessarily complicated like this, it's much better now.

PR #1592, opened 2021-06-08
GROOVY-5245: bean-style property access for boolean "is" category method

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-5245


danielsun1106:
+1

PR #1594, opened 2021-06-11
remove unneccessary package declarations in nested test scripts

nineninesevenfour:
small change, because I stumbled over it, too. :)


paulk-asert:
Haven't checked all of them, but some of these look like they are testing package specific aspects, so probably shouldn't change.

nineninesevenfour:
Have to check those errors anyway :)

PR #1596, opened 2021-06-11
GROOVY-8678: extend test case

nineninesevenfour:
after the code review I had the impression that more tests are needed on that


nineninesevenfour:
Apart from the actual change, I also encountered a problem with the dist build:

> One artifact failed verification: testng-7.4.0.jar

After checking the verification-metadata I thought on one box the Gradle cache might be broken, so I added the --refresh-dependencies option (the equivalent for -U in Maven).
But I am not sure, if we have to consider consumed resources on GH Actions and if we should rather add a "Cache Gradle Repository" step like in the tests. 🤔

danielsun1106:
Merged. Thanks.

PR #1597, opened 2021-06-11
GROOVY-10133: prefer getPropName() over isPropName() when both available

eric-milles:
This is consistent with static compilation method selection for [Bb]oolean properties.

https://issues.apache.org/jira/browse/GROOVY-10133

### DIFF
--- a/src/main/java/groovy/lang/MetaClassImpl.java
+++ b/src/main/java/groovy/lang/MetaClassImpl.java
@@ -2595,7 +2595,7 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
         });
     }
 
-    private static MetaProperty makeReplacementMetaProperty(MetaProperty mp, String propName, boolean isGetter, MetaMethod propertyMethod) {
+    private static MetaProperty makeReplacementMetaProperty(final MetaProperty mp, final String propName, final boolean isGetter, final MetaMethod propertyMethod) {
         if (mp == null) {
             if (isGetter) {
                 return new MetaBeanProperty(propName,
@@ -2625,7 +2625,10 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
         } else if (mp instanceof MetaBeanProperty) {
             MetaBeanProperty mbp = (MetaBeanProperty) mp;
             if (isGetter) {
-                mbp.setGetter(propertyMethod);
+                // GROOVY-10133: do not replace "getPropName()" with "isPropName()" or ...
+                if (mbp.getGetter() == null || !mbp.getGetter().getName().startsWith("get")) {
+                    mbp.setGetter(propertyMethod);
+                }
                 return mbp;
             } else if (mbp.getSetter() == null || mbp.getSetter() == propertyMethod) {
                 mbp.setSetter(propertyMethod);
--- a/src/test/groovy/bugs/Groovy10133.groovy
+++ b/src/test/groovy/bugs/Groovy10133.groovy
@@ -1,0 +1,173 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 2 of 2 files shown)


danielsun1106:
LGTM

paulk-asert:
It would be nice to make this consistent. It is a tricky area though. The JavaBean spec[1] seems to indicate that the "is" method should have precedence:

> This “is\<PropertyName\>” method may be provided instead of a “get\<PropertyName\>” method, or it may be provided in addition to a “get\<PropertyName>” method.
> In either case, if the “is\<PropertyName>” method is present for a boolean property then we will
> use the “is\<PropertyName>” method to read the property value.

That also seems to be how the JavBean Introspector behaves:
```
import java.beans.Introspector

class Bean {
    void setFlag(boolean flag) {}
    boolean isFlag() {}
    boolean getFlag() {}
}

def props = Introspector.getBeanInfo(Bean).propertyDescriptors
assert 'isFlag' == props.find{ it.baseName == 'Flag' }.readMethod.name
```

[1] https://download.oracle.com/otndocs/jcp/7224-javabeans-1.01-fr-spec-oth-JSpec/

eric-milles:
Please see the linked issue ticket.  The getter is being used in most cases today, including under static compilation.  The few cases that use the isser are outliers and cause confusion.  Trying to change it all to prefer the isser is a much bigger change.

danielsun1106:
It's better for Groovy to align with Java's specification to make Java developers as less surprised as we could.

paulk-asert:
> Please see the linked issue ticket. The getter is being used in most cases today, including under static compilation. The few cases that use the isser are outliers and cause confusion. Trying to change it all to prefer the isser is a much bigger change.

Yes, I have seen the issue ticket and understand that it is more work.

I guess there are lots of ways to look at this. We could say that for this edge case, that anyone that has different behavior between the isser and the getter deserves to experience pain since it makes no sense to make them different. And so our existing code is fine and we can take the smallest steps possible to make our existing code consistent. On the other hand, there is an existing (albeit old) spec which we try to adhere to in most other respects. If we don't like that spec, we could justify why we don't comply with the spec or we can define our own spec and justify why it is different and better than the official one but regardless, Groovy will still break when used with tools which adhere to the spec (for this rare edge case - which is why we haven't seen issues reported before).

Given that it's an edge case we hope never to see, I don't see the need to rush to make ourselves less compliant. Instead, I'd prefer to wait and see if any of us has time to explore going down the opposite path. If that does indeed turn out to be a horror path, then we can revisit with a well-informed reasoning behind making ourselves less compliant.

eric-milles:
So my understanding of compliance is that "isPropName()" is preferred over "getPropName()" for `boolean propName` and in all other cases the "is" variant is ignored.  Is that what you would be comfortable with?  There will be changes in MOP, STC, category methods and classgen.

I can check this with:
```groovy
@Grab('commons-beanutils:commons-beanutils:1.9.4')
import org.apache.commons.beanutils.PropertyUtils;

Pojo pojo = new Pojo();
Object value = PropertyUtils.getSimpleProperty(pojo, "dualState");
System.out.println("boolean: " + value);
value = PropertyUtils.getSimpleProperty(pojo, "triState");
System.out.println("Boolean: " + value);
value = PropertyUtils.getSimpleProperty(pojo, "other");
System.out.println("Object : " + value);

  class Pojo {
    public boolean isDualState() {
      return true;
    }
    public boolean getDualState() {
      return false;
    }
    public Boolean isTriState() {
      return Boolean.TRUE;
    }
    public Boolean getTriState() {
      return Boolean.FALSE;
    }
    public Object isOther() {
      return Boolean.TRUE;
    }
    public Object getOther() {
      return Boolean.FALSE;
    }
  }
```

Sorry, not very groovy.  I checked from Java to be sure.  If I remove "getDualState()" it still works fine and switches as expected.  Removing "getTriState()" or "getOther()" does not make use of the is variant.

paulk-asert:
Well, I think compliance is well-defined for primitive boolean and the spec says nothing about "is" ever being supported for the wrapper classes. We know of different frameworks which have made different choices about what the silence in the spec for wrapper classes means. Our past behavior has mostly been about trying not to get in the way of other framework/library choices which is somewhat short of fully clarifying our behavior.

I don't know whether the strict interpretation that you propose is best or perhaps ignore the isser for Boolean unless there is an isser and no getter. That might be more backwards compatible.

eric-milles:
I'd prefer to see Boolean have the same treatment as boolean or have the same treatment as say Object.  My original tickets were along these lines.  Adding special rules for Boolean that do not apply to any other types requires more checks in the compiler and IDE.  To have it work like boolean only requires the method return value checks to include both types.  This would be inconsistent with Apache BeanUtils, but would be more compatible with the historic support for Boolean properties in Groovy.

paulk-asert:
Yes, I think Boolean being the same as all other objects is the way to go. We could document some metaprogramming tricks for dynamic Groovy. I was hoping we could also write a type checking extension that might allow us to enable "Lenient Boolean Properties" through an extension but this scenario triggers the `unresolvedProperty` hook rather than the `methodNotFound` hook and I am unsure whether we can do this tweak from there.

PR #1598, opened 2021-06-13
GROOVY-10145: Support JDK16

danielsun1106:


paulk-asert (changes requested):
+1

I have some unresolved questions about the PR. I would be keen to explore whether we can avoid the extra try/catch by perhaps funnelling certain method invocations directly through the fallback path. This is with a view to not impact performance. However, some kind of initial guard check might impact the normal case anyway. Also, I wonder whether something like the spy would help us solve existing super call bugs. Having said that, we know from the test suite that the number of places affected is relatively small, so I think we can merge this as is and explore further refactoring as a separate activity. We might as well have things working for JDK16 in the meantime.

Apart from that I have some minor suggestions included as single line comments.

paulk-asert:
Current PR LGTM

PR #1599, opened 2021-06-14
GROOVY-9382, GROOVY-10133: prefer isName() over getName() for boolean

eric-milles:
This should be the bulk of the changes to align dynamic and static compilation handling of `boolean` property access as well as aligning `java.lang.Boolean` property access with the other reference types.  No more "name" to "isName()" translation for `Boolean`.

I did not exhaustively search for `"is"` so there may be a few outliers out there.  If you know of them or can produce a test case to help me seek them out, I would appreciate it.

The changes had surprisingly-low impact to the unit tests.

https://issues.apache.org/jira/browse/GROOVY-10133
https://issues.apache.org/jira/browse/GROOVY-9382
https://issues.apache.org/jira/browse/GROOVY-7924
https://issues.apache.org/jira/browse/GROOVY-7300
https://issues.apache.org/jira/browse/GROOVY-6097 (TODO) -- "super.propName" handling
https://issues.apache.org/jira/browse/GROOVY-1736


eric-milles:
One thing to watch out for in XML handling is Node has both "getFirstChild()" and "isFirstChild()".
```
com.sun.org.apache.xerces.internal.dom.NodeImpl.isFirstChild()
com.sun.org.apache.xerces.internal.dom.ParentNode.getFirstChild()
```

eric-milles:
TODO: `org.codehaus.groovy.control.StaticImportVisitor#findStaticFieldOrPropAccessorImportFromModule(String)`

src/test/groovy/CategoryTest.groovy
```diff
@@ -68,8 +68,21 @@ final class CategoryTest extends GroovyTestCase {
         }
     }
 
-    // GROOVY-5245
+    // GROOVY-10133
     void testCategoryDefinedProperties2() {
+        assertScript '''
+            class Cat {
+                static boolean isAbc(self) { true }
+                static boolean getAbc(self) { true }
```
paulk-asert, line 76: don't you want false here?

paulk-asert:
> Node has both "getFirstChild()" and "isFirstChild()"

That's unfortunate. The latter is package private and almost certainly not what folks would want but I can't see a workaround other than your change.

src/test/groovy/CategoryTest.groovy
eric-milles: Fixed, thanks.  I also tested with the methods defined in the reverse order to be sure that wasn't a factor.

PR #1600, opened 2021-06-16
GROOVY-10141: retain parse order of anon. inner types in module classes

eric-milles:
- InnerClassVisitor checks this$0 which isn't set up if C$1$1 before C$1

https://issues.apache.org/jira/browse/GROOVY-10141


paulk-asert:
+1

danielsun1106:
Merged. Thanks!

PR #1602, opened 2021-06-19
GROOVY-9999,GROOVY-8488:STC:make decimal literal parameter can auto c…

YueZhikong:
fix GROOVY-9999,GROOVY-8488, in general, make literal such as "4.0" could be use in double param method such as "Math.sqrt(double a)".

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/expr/ConstantExpression.java
+++ b/src/main/java/org/codehaus/groovy/ast/expr/ConstantExpression.java
@@ -39,7 +39,7 @@ public class ConstantExpression extends Expression {
     public static final ConstantExpression VOID = new ConstantExpression(Void.class);
     public static final ConstantExpression EMPTY_EXPRESSION = new ConstantExpression(null);
 
-    private final Object value;
+    private Object value;
     private String constantName;
 
     public ConstantExpression(Object value) {
@@ -121,5 +121,9 @@ public class ConstantExpression extends Expression {
 
     public boolean isEmptyStringExpression() {
         return "".equals(value);
+    }
+
+    public void setValue(Object value) {
+        this.value = value;
     }
 }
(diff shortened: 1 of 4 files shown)


eric-milles:
For STC, there are a couple methods that check the compatibility of arguments and parameters or source and target types.  I would expect a change in one of these to allow `BigDecimal` as argument to `double` parameter.  I'm not sure of the reason for the change to make `BigDecimal` not convert as before.  It would be good to understand why this stopped working at some point.

In terms of tests, we are slowly migrating away from `GroovyTestCase` and `GroovyShell#evaluate` in favor of JUnit 4 and `GroovyAssert#assertScript`.  It appears you have taken us in the reverse direction.

YueZhikong:
> For STC, there are a couple methods that check the compatibility of arguments and parameters or source and target types. I would expect a change in one of these to allow `BigDecimal` as argument to `double` parameter. I'm not sure of the reason for the change to make `BigDecimal` not convert as before. It would be good to understand why this stopped working at some point.
> 
> In terms of tests, we are slowly migrating away from `GroovyTestCase` and `GroovyShell#evaluate` in favor of JUnit 4 and `GroovyAssert#assertScript`. It appears you have taken us in the reverse direction.

@eric-milles 
Is the method "StaticTypeCheckingVisitor#findMethod" designed to "check the compatibility of arguments and parameters or source and target types"?

That method returns empty mn in "StaticTypeCheckingVistor:3479",and causes "GROOVY-9999","GROOVY-8488". And my idea is to write a new method "StaticTypeCheckingVisitor#bigDecimalCast" to “check the compatibility”。 Maybe I should integrate the logic of "bigDecimalCast" into "findMethod"?

And could you tell me all "methods that check the compatibility of arguments and parameters or source and target types"? That may help to solve GROOVY-8168 which is possible caused by similar method like "StaticTypeCheckingVisitorfindMethod".

PR #1603, opened 2021-07-01
GROOVY-10159: Fixed String casting issue

onlydevelop:


danielsun1106:
It's better to add a test.

paulk-asert:
Or if you are struggling to re-create the test locally, if you can describe the circumstances in more detail, we might be able to help re-create.

PR #1607, opened 2021-08-07
GROOVY-10193: Support sealed type grammar

danielsun1106:


src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -1439,6 +1445,34 @@ public ClassNode visitClassDeclaration(final ClassDeclarationContext ctx) {
...
+
+        if ((isAnnotation || isEnum) && (isSealed || isNonSealed)) {
+            ModifierNode mn = isSealed ? sealedModifierNodeOptional.get() : nonSealedModifierNodeOptional.get();
+            throw createParsingFailedException("modifier `" + mn.getText() + "` is not allowed here", mn);
+        }
+
+        boolean hasPermits = asBoolean(ctx.PERMITS());
+        if (isSealed && !hasPermits) {
+            throw createParsingFailedException("sealed type declaration should have `permits` clause", ctx);
+        }
```
paulk-asert, line 1471: This is not actually a required restriction. The `permits` clause can be elided for the case where the sealed class and all subclasses are in the same source file, in which case it is worked out automatically. Later checks should give us the correct behavior if permits is left off and no implicit subclasses are found in the same source file. So, I think we can just remove this check.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -1439,6 +1445,34 @@ public ClassNode visitClassDeclaration(final ClassDeclarationContext ctx) {
...
+        if (isNonSealed && hasPermits) {
+            throw createParsingFailedException("non-sealed type declaration should not have `permits` clause", ctx);
+        }
```
paulk-asert, line 1474: Actually, we should just check that permits is only ever used with sealed. Currently this is allowed but shouldn't be:
`class Shape permits Circle {}`
Note: no sealed.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -1439,6 +1445,34 @@ public ClassNode visitClassDeclaration(final ClassDeclarationContext ctx) {
...
+
+        if (isFinal && (isSealed || isNonSealed)) {
+            throw createParsingFailedException("type cannot be defined with both " + (isSealed ? "`sealed`" : "`non-sealed`") + " and `final`", finalModifierNodeOptional.get());
+        }
+
+        if ((isAnnotation || isEnum) && (isSealed || isNonSealed)) {
+            ModifierNode mn = isSealed ? sealedModifierNodeOptional.get() : nonSealedModifierNodeOptional.get();
+            throw createParsingFailedException("modifier `" + mn.getText() + "` is not allowed here", mn);
+        }
```
paulk-asert, line 1466: This matches Java (at least for enum). I changed to be slightly more descriptive when using @Sealed. I have a slight preference to make this also more descriptive unless you can see a good reason to keep like Java?

paulk-asert (changes requested):
+1, approved with minor suggestions

paulk-asert:
I merged with my suggested changes. Feel free to refine further if needed.

PR #1608, opened 2021-08-11
GROOVY-10195: Groovy should have a @Final AST transform

paulk-asert:


danielsun1106:
Paul, it looks good to me.
I wonder if it is possible to merge the new `@Final` and the existing `@AutoFinal`:
https://github.com/apache/groovy/blob/master/src/main/java/groovy/transform/AutoFinal.java

paulk-asert:
It would be possible to combine but I think would be more cumbersome. Instead of:
```
class Foo {
  @AutoFinal bar(param1, param2) { ... } // params will be final
  @Final baz(param1, param2) { ... }    // baz will be final
}
```
You would need something like:
```
class Foo {
  @Final(enabled=false, autoParams=true) bar(param1, param2) { ... } // params will be final
  @Final baz(param1, param2) { ... }    // baz will be final
}
```

eric-milles:
I find it confusing to try and sort out the purpose of `final` vs `@Final` vs `@AutoFinal`.  Is the concept of "no extend" or "no override" what is represented here?

src/test/org/codehaus/groovy/transform/FinalTransformTest.groovy
```diff
@@ -0,0 +1,80 @@
...
+package org.codehaus.groovy.transform
+
+import org.junit.Test
+import org.codehaus.groovy.control.MultipleCompilationErrorsException
+
+import static groovy.test.GroovyAssert.assertScript
+import static groovy.test.GroovyAssert.shouldFail
+
+/**
+ * Tests for the {@code @Field} transformation
```
eric-milles, line 28: Need to fix the reference.

src/test/org/codehaus/groovy/transform/FinalTransformTest.groovy
```diff
@@ -0,0 +1,80 @@
...
+            import static groovy.test.GroovyAssert.shouldFail
+            import static java.lang.reflect.Modifier.isFinal
+
+            @AnnotationCollector
+            @NullCheck
+            @Final
+            @interface MyNullCheck {}
+
+            class Foo {
+                @MyNullCheck static upper(String s) { s.toUpperCase() }
```
eric-milles, line 62: Does it make sense to add final to a static method?

src/test/org/codehaus/groovy/transform/FinalTransformTest.groovy
paulk-asert: Well spotted. Done.

paulk-asert:
> I find it confusing to try and sort out the purpose of `final` vs `@Final` vs `@AutoFinal`. Is the concept of "no extend" or "no override" what is represented here?

You can think of `@Final` is just an alias for `final` but one that you would almost never use directly (since you can just use `final` after all). However, in meta-annotation scenarios, it can be useful to be able to "mix in" that functionality.

PR #1610, opened 2021-08-12
GROOVY-10199: ASTTest: evaluate with compliation unit's transform loader

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10199


danielsun1106:
Merged. Thanks!

PR #1611, opened 2021-08-12
GROOVY-10191: catch LinkageError during static inline

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10191


danielsun1106:
Merged. Thanks!

PR #1614, opened 2021-08-22
GROOVY-8947: convert "new A().new B()" into "new A.B(new A())"

eric-milles:
New take on GROOVY-8947.  Unburden `ResolveVisitor` by using context to create qualified type name.


eric-milles:
I'm taking a look separately at simplifying or reorganizing the changes for GROOVY-4287/GROOVY-7812/GROOVY-9243 since they apply only to running through `GroovyMain` (aka runtime compilation).  Can the one regression for 9243 be accepted until then?

danielsun1106:
LGTM except for the regression.

> Can the one regression for 9243 be accepted until then?

Correctness is more important than elegance IMO, so it's better not to introduce regression for refactoring.

paulk-asert:
There are examples (like below) which work after the change - so that's good. I know you are working on a cleaner medium-term change but I don't know if we can find a short-term fix for the regression in the meantime.
```
public class Main {
    public class Foo {
        public class Bar {
            private String prop;
            public Bar(String prop) { this.prop = prop; }
            public String getProp() { return prop; }
        }
    }
        System.out.println(new Main().new Foo().new Bar("bar").getProp());
    }
}
```

eric-milles:
@paulk-asert Is the new example a scenario that did not work previously?

My concern with all the GROOVY-4287/GROOVY-7812/GROOVY-9243 code is that it has increased the complexity of `ResolveVisitor` significantly only to benefit runtime compilation scenarios.  Eclipse tooling does not make use of runtime compilation so it is extra complexity for no benefit.  I carved off the 8947 rework because it applies to traditional compilation.  And my initial testing did not catch the 9243 regression.

What got me going on this is the double execution of `VariableScopeVisitor` and `ResolveVisitor`[1][2].  I have a solution that runs these each once during normal compilation, but the change affects the inner class resolution for runtime compilation (aka 4287/7812/9243).

[1] https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/control/CompilationUnit.java#L623
[2] https://github.com/apache/groovy/blob/master/src/main/java/org/codehaus/groovy/control/CompilationUnit.java#L203

paulk-asert:
@eric-milles Correct, that example didn't work previously.

paulk-asert:
I am inclined to hold off the change until 3.0.9 is out. There is already a lot in that release.

eric-milles:
Regarding this change request, my single-pass resolver solution, and refactored support for runtime resolution of inner types:

I'd like to be able to put this in as a first step, work out the 9243 regression as part of the planned refactor and then supply the resolver changes.  Do you see that as a possibility for Groovy 4.0b2 or is it better to save this for Groovy 5 (or whatever comes after 4)?

PR #1618, opened 2021-09-08
GROOVY-10216: Avoid unnecessary security checks for each invocation

danielsun1106:


eric-milles:
So if `isAccessible()` returns true, you still should be calling `setAccessible(true)`?  What are the security checks that are tried for an accessible object?  Do you have any documentation on this behavior?

Also, what do the rows and values in the performance chart represent?  I'm not clear on what the data is intended to communicate.

PR #1620, opened 2021-09-17
GROOVY-10240: Support record grammar

danielsun1106:


src/main/java/org/codehaus/groovy/ast/MethodNodeHelper.java
```diff
@@ -0,0 +1,40 @@
...
+ *  under the License.
+ */
+package org.codehaus.groovy.ast;
+
+/**
+ * Helper for {@link MethodNode}
+ *
+ * @since 4.0.0
+ */
+public class MethodNodeHelper {
```
paulk-asert, line 26: Would it make sense for this to move to `MethodNodeUtils`?

danielsun1106: Agreed ;-)

paulk-asert:
Merged. Thanks! I think there are still a few things to tweak but we can do that on master.

PR #1624, opened 2021-09-24
Make consistent with #1141

T45K:
In #1141, `Node#depthFirst` was modified to use `Boolean#parseBoolean` instead of `Boolean#valueOf` ([here](https://github.com/apache/groovy/pull/1141/files#diff-6ce3c6667b435203886c05db2ba7cd666bd9918d2af8881530a56f554ef1c30fR610)), but its similar method, `Node#breadthFirst` was not.
I modified the method to make consistent with the change.


paulk-asert:
Merged, thanks!

PR #1626, opened 2021-09-28
GROOVY-10234: handle raw types in class signature without looping

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10234

`interface ConversionService<Impl extends ConversionService>`

https://github.com/micronaut-projects/micronaut-core/blob/3.0.x/core/src/main/java/io/micronaut/core/convert/ConversionService.java


danielsun1106:
LGTM

PR #1628, opened 2021-09-30
GROOVY-10275: Align intersect which takes a closure/comparator orderi…

paulk-asert:
…ng to be the same as other languages.


paulk-asert:
In addition to altering the order to align with other languages, there is also a new Closure variant added for consistency with other operators which accept a closure or comparator.

danielsun1106:
+1

PR #1635, opened 2021-10-04
GROOVY-10282: GenericsUtils: do not mix type parameter contexts

eric-milles:
collect for `BiFunction<U, ? super T, U>` yields `[T:U, U:? super T, R:U]` not `[T:? super T, U:? super T, R:? super T]`

https://issues.apache.org/jira/browse/GROOVY-10282


danielsun1106:
LGTM

PR #1636, opened 2021-10-06
GROOVY-10284: AnnotationCollector should not collect @Incubating whic…

paulk-asert:
…h is a meta-annotation in Groovy


danielsun1106:
+1

PR #1643, opened 2021-10-18
fix ArrayIndexOutOfBoundsException

suoyi123wang:
in this method  org.codehaus.groovy.tools.GrapeUtil.getIvyParts, when running the testcase in below,it will  fail for ArrayIndexOutOfBoundsException
[image
](url)

I think you should either declare the legal format of the parameters, or judge the length of the array parts in the code in advance


paulk-asert:
Merged, thanks!

PR #1651, opened 2021-11-09
GROOVY-10353: `evaluateExpression` can handle simple expression directly

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10353

When inspecting the `@ClosureParams` or `@DelegatesTo` metadata of methods, it is typical to require:
```
eval((java.lang.Integer) 1)
eval((java.lang.Integer) 1)
eval((java.lang.String) "")
eval((java.lang.Integer) 1)
eval((java.lang.Integer) -1)
eval((java.lang.String) "self")
eval((java.lang.String) "self")
eval((java.lang.Class<T>) groovy.transform.stc.FirstParam)
eval((java.lang.String[]) [])
eval((java.lang.Integer) 1)
eval((java.lang.Integer) 1)
eval((java.lang.String) "")
eval((java.lang.Integer) 1)
eval((java.lang.Integer) -1)
eval((java.lang.String) "self")
eval((java.lang.String) "self")
eval((java.lang.Class<T>) groovy.transform.stc.FirstParam)
eval((java.lang.String[]) [])
eval((java.lang.Class<T>) groovy.transform.stc.SimpleType)
eval((java.lang.String[]) ["java.io.File"])
eval("java.io.File")
```

I tried to return the class directly from `ClassExpression` where `getType()` returns a `ClassNode` with a class loaded -- `groovy.transform.stc.FirstParam` above -- but it did not like the typecast `Class<? extends ClosureSignatureHint) evaluateExpression(...)` since there were two different class loaders involved.  So these still go through the slow path.

### DIFF
--- a/src/main/java/groovy/lang/GroovyClassLoader.java
+++ b/src/main/java/groovy/lang/GroovyClassLoader.java
@@ -703,10 +703,10 @@ public class GroovyClassLoader extends URLClassLoader {
     }
 
     /**
-     * open up the super class define that takes raw bytes
+     * Converts an array of bytes into an instance of {@code Class}.
      */
-    public Class defineClass(String name, byte[] b) {
-        return super.defineClass(name, b, 0, b.length);
+    public Class defineClass(final String name, final byte[] bytes) throws ClassFormatError {
+        return super.defineClass(name, bytes, 0, bytes.length);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


danielsun1106:
Merged. Thanks!

src/main/java/groovy/lang/GroovyClassLoader.java
```diff
@@ -703,10 +703,10 @@ public Collection getLoadedClasses() {
     }
 
     /**
-     * open up the super class define that takes raw bytes
+     * Converts an array of bytes into an instance of {@code Class}.
      */
-    public Class defineClass(String name, byte[] b) {
```
paulk-asert, line 708: Is it normal to add an unchecked exception to a throws clause?

danielsun1106: It seems that the PR derives the exceptiion definition from the following method:

**java.lang.ClassLoader#defineClass(java.lang.String, byte[], int, int)**
```java
    protected final Class<?> defineClass(String name, byte[] b, int off, int len)
        throws ClassFormatError
    {
        return defineClass(name, b, off, len, null);
    }
```

PR #1653, opened 2021-11-17
GROOVY-10374: Provide more tuple classes

danielsun1106:
https://issues.apache.org/jira/browse/GROOVY-10374


eric-milles:
Why create an issue ticket at all if this was incoming?  Should discussion occur on the issue ticket or the pull request?  Now that you have the sunk cost of implementation, is there really any way to stop this from happening?

-1

danielsun1106:
I do not care the sunk cost very much. In fact, I have closed many PRs without merging in the recent years, though most of them cost me hours.
I agree with you that we had better to discuss in the issue ticket, PR or mailing list. As I am not sure whether it's accepted by others or not, the PR was created for reviewing and not pushing to master directly.

PR #1656, opened 2021-11-24
GROOVY-7001: DGM: return exact type for methods accepting collector

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7001

I stopped short of `collectEntries` methods.  They don't have quite the same problem -- map return is probably okay, whereas the distinction between set, list and collection is often important.

I also changed the return type of some variants to List and added the Collection bridges.

There is some inconsistency between some the variations.  In cases where the input and output types are different, collect (except map variant) is listing the input type first where collectMany and others list input type(s) second.  I did not make any changes (except to use `E` instead of `S`).
```java
    <E, T> List<T> collect(E[] self, @ClosureParams(FirstParam.Component.class) Closure<T> transform)
    <T, K, V> List<T> collect(Map<K, V> self, @ClosureParams(MapEntryOrKeyValue.class) Closure<T> transform)
    <T, E> List<T> collectMany(E[] self, @ClosureParams(FirstParam.Component.class) Closure<...> projection)
    <T, K, V> List<T> collectMany(Map<K, V> self, @ClosureParams(MapEntryOrKeyValue.class) Closure<...> projection)
```

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -180,7 +180,6 @@ import static groovy.lang.groovydoc.Groovydoc.EMPTY_GROOVYDOC;
  */
 public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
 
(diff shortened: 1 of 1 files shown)


paulk-asert (commented):
These changes seem okay to me and japicmp seems happy.

PR #1657, opened 2021-11-28
GROOVY-10383: SC: optimize `a !in b`

eric-milles:
https://github.com/apache/groovy/commit/1cff465ee4eaa159192cbade013033af0b38be04 provides the necessary direct targets for this optimization.  `a !in b` is transformed into `b == null ? a != null : b.isNotCase(a)` instead of `ScriptBytecodeAdapter.isNotCase(a,b)`

https://issues.apache.org/jira/browse/GROOVY-10383


danielsun1106:
Merged. Thanks.

PR #1660, opened 2021-12-04
GROOVY-10259: Groovy4 NV Macros: Rename NV* -> SV*

paulk-asert:


subprojects/groovy-macro-library/src/main/groovy/org/apache/groovy/macrolib/MacroLibGroovyMethods.java
```diff
@@ -29,29 +32,41 @@
...
 import static org.codehaus.groovy.ast.tools.GeneralUtils.constX;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.ctorX;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.listX;
 
 public class MacroLibGroovyMethods {
+    private static final ClassNode NAMED_VALUE = ClassHelper.make(NamedValue.class);
 
-    public static <T> T NV(Object self, Object... args) {
-        throw new IllegalStateException("MacroLibGroovyMethods.NV(Object...) should never be called at runtime. Are you sure you are using it correctly?");
+    public static <T> T SV(Object self, Object... args) {
```
eric-milles, line 44: Why does this return `T`?

src/main/java/groovy/lang/NamedValue.java
```diff
@@ -0,0 +1,24 @@
+package groovy.lang;
+
+public class NamedValue<T> {
```
eric-milles, line 3: Is this just a `Tuple2<String,T>`?  Is this really a valuable addition to `grovy.lang` package?

mgroovy: Yes, semantically it is quite different. The implementation we use also shows that, since toString auto-quotes String/GString/Character (making the output much easier to read), and is exception safe (so NV macros can be used e.g. when supplying the message string of an exception), e.g.:

throw new SqlExecutorException(innerException, "updateValueAtPosition failed for ${NVL(table,schema,pos,val,updateSql)}")

String toString() {
   name + "=" + safeQuote(val)
}

static String safeQuote(final Object o) {
 if(null == o) {
  return null
 }
 else if(o instanceof GString) {
  return '"' + o + '"'
 }
 else if(o instanceof String) {
  return "'" + o + "'"
 }
 else if(o instanceof Character) {
  return "C'" + o + "'"
 }
 else {
  try {
   return o.toString()
  }
  catch(Throwable throwable0) {
   try {
    return "NV(${o.getClass().getSimpleName()}.toString() threw ${throwable0.getClass().getSimpleName()})"
   }
   catch(Throwable throwable1) {
    return "NV(o.getClass() threw ${throwable1.getClass().getSimpleName()})"
   }
  }
 }
}

eric-milles: Yes, but couldn't the NV whatevers just return `Map.Entry<String,T>` or `Tuple2<String,T>` that exhibit that behavior?  Do we need a new core language type just for that?  If this is specific to NV, why doesn't it go there?

mgroovy: Map.Entry<String,T> & Tuple2<String,T>: No, both do not have 
1. The desired toString behavior: The no core application for NV is logging without having to write the typical "person=$pin, user=$user, insertSql='$insertSql'" (the first term is an example of a typical refactoring error here: the code was copy-pasted from another place where "person" was the first var, and only the variable name was refactored, leading to a misleading log output).
2. Speaking member access (name, val).

mgroovy: My argument for NamedValue to be a core language type would be that it has general applicability, but I would not mind if it were located somewhere else.

mgroovy: It is Paul's NamedValue implementation, but, as I said, we have been using our own NV macros for quite some time now, and from this experience we are convinced NV + NamedValue with the  toString functionality I posted above makes the whole concept really convenient & powerful.

eric-milles: `Map.Entry` can have any `toString()` behavior you want.  It is an interface, you choose the implementation within the macro method.  Name and value vs. key and value is not that big of a difference to justify an entirely separate abstraction.

eric-milles: I am not buying that this is beneficial to others.  I still think it should have stayed your own extension that you use.  It is now a core language feature that keeps growing in scope.  Paul tried to take in part of what you offered and I don't understand the need for this entire renaming affair.

eric-milles: I propose that these move from Groovy 4 to Groovy 5 to allow more time to consider the implications of adding macro methods.  The Public API changes in Groovy 4 are already quite large and a source of debate within the PMC.

mgroovy: Basically all software logs, println|s, etc variables, and creates exceptions, which are better the more context information they contain. NV macros make doing this as concise, safe and painless as it can possible be. 
How would that not be beneficial to every Groovy developer... ?-)

mgroovy: The feature is also not growing in scope per se: My initial proposal was to introduce the NV/NVL macros we have used for a long time now for the benefit of all Groovy 2.5 users. These macros literally create a name, together with the corresponding value, hence the macro name "NV". Paul then suggested we postpone this to Groovy 3, and then to Groovy 4, to give people time for feedback, and later proposed a less powerful but more runtime efficient NV macro implementation, that always just returns a "name=$value"  GString, instead of a NamedValue object which has the advantage that it can be stored as such and/or processed further.
Since that gives only part of the functionality we use, I proposed to rename Paul's macros SV for "stringify variable", since a) that is what they do, and b) it would have allowed us to keep our more powerful NV macros, for which it is evidently hard to find another good, concise name. I also proposed that the most powerful solution would of course be to introduce both SV and NV macros in parallel, since they offer different benefits/tradeoffs to the user. That is what Paul evidently decided to do now, some time later, as I learned yesterday. The only thing I then finally suggested, was to add the toString functionality I posted above to the NamedValue class, for the reasons/benefits given.

mgroovy: Using Map.Entry would a) require a user who wants to e.g. return the result of NVL call from a method to write
```List<Map.Entry<String,Object>>``` instead of just ```List<NamedValue>```, and b) also the returned object is, in fact, not a map-entry, and the String value it carries is not a "key" to a map :-)

So it would seem extremely forced, inconvenient & confusing for the user to go with Map.Entry<String,Object> here, just to avoid a small, clearly named class. I do also not understand where this small, completely orthogonal functionality would tie into any discussion (which I am evidently unaware of / not privy to) about the large public API of Groovy... ?

mgroovy: (Even if that is not for me to decide: This feature has been moved from Groovy 2.5, to Groovy 3, then to Groovy 4. It has been in discussion for years now. Keeping it out of users' hands further, seems, quite frankly, ridiculous. 

I am, as you will know from discussions on the Groovy mailing list, very much for discussing changes to Groovy, weighing the pros and cons, trying to find benefits and pitfalls. I even sometimes take the opposite position to an argument which I feel is not strong enough to be for or against a proposal, even if I am myself not convinced what the right decision is, to stimulate further discussion, as I have done for the recent proposal regarding dates and dates-with-time objects.

But in this case, even if you personally never log or println any variables, never need to embed them in exceptions, never need to generate any output where you need the name of variables or fields together with their value, I can, quite frankly not fathom what makes you continually fight this, in my eyes, small, harmless extension to Groovy so hard...)

paulk-asert: `Map.Entry<String,Object>` is not quite the right abstraction being mutable. It isn't really `Tuple2` either. It is more like a _named_ `Tuple1` much like we have the `ginq` specific `NamedTuple`. Adding a more general version of the `ginq` class to core or `Named` variants of our Tuple0..Tuple16 classes seemed like bigger changes to me. `NamedValue` seems like an abstraction that could pull its weight in numerous other situations.

paulk-asert:
I have revised the PR based on earlier feedback.

danielsun1106:
+1
P.S. license header should be added to the new class.

PR #1661, opened 2021-12-06
GROOVY-10278, GROOVY-10281: refactor `CompilerConfiguration` somewhat

eric-milles:
This is not complete, but does give some shape to recent discussions.

- `ASM_API_VERSION` goes to `AsmClassGenerator`
- `DEFAULT_TARGET_BYTECODE` and `defaultTargetBytecode()` were unnecessary additions
- `setTargetBytecode(String)` can use nearest value as described in GROOVY-10278
- `isPostJDKn` can be replaced with use of `isAtLeast(String,String)` -- the recent additions could be dropped, but I just moved them for now
- `LoggableClassVisitor` --> `TraceClassVisitor` (just need to work out the jarjar stuff for unit tests

Thoughts?


danielsun1106:
> LoggableClassVisitor --> TraceClassVisitor

After reading tons of "minor edits" and this PR, I am still not sure why "inline" or "anonymous" code is preferred over code with meaningful names, e.g. method names, class names, etc.

-1 from me.

eric-milles:
Now if the new methods added were `isAtLeastJDK9` and so on and `isPostJDK5`, etc. were removed or scheduled for removal, then my only concern would be the need to add a new method every six months.  That kind of work should be automated IMO.

paulk-asert:
According to a strict interpretation of ASF rules, a -1 on a code submission is a veto:
https://www.apache.org/foundation/voting.html#votes-on-code-modification
I'll close this issue and see if there are any non-controversial parts which can be cherry-picked.

PR #1662, opened 2021-12-07
GROOVY-9391: Cannot cast or coerce `super`

eric-milles:
It may still make sense to cast or coerce `this` (Java allows it), but `super` is a hard error (Java won't even parse).

https://issues.apache.org/jira/browse/GROOVY-9391


paulk-asert:
Looks okay to me.

PR #1666, opened 2021-12-21
GROOVY-10417: MethodNode toString() could add quotes when name contai…

paulk-asert:
…ns spaces (or other non-identifier characters)


danielsun1106:
+1

PR #1670, opened 2022-01-04
GROOVY-5169, GROOVY-7682: JSON output: serialization fixes

eric-milles:
First commit covers the private property portion of 5169.  Third commit covers public fields.  Second commit drops static properties, which are the source of the cycles in 7682 and 8371.  10438 has been split off to determine if DGMs should return properties for private fields or methods.

There were some assumptions made about the return value of MetaClass#getProperties.  These were easy to address and probably represent the kind of breakage other projects might expect.

https://issues.apache.org/jira/browse/GROOVY-5169
https://issues.apache.org/jira/browse/GROOVY-7682
https://issues.apache.org/jira/browse/GROOVY-8371
https://issues.apache.org/jira/browse/GROOVY-10438

### DIFF
--- a/subprojects/groovy-json/build.gradle
+++ b/subprojects/groovy-json/build.gradle
@@ -21,13 +21,19 @@ plugins {
 }
 
 dependencies {
-    api rootProject  // JsonBuilder extends GroovyObjectSupport...
+    api rootProject // JsonBuilder extends GroovyObjectSupport...
+
     testImplementation project(':groovy-test')
     testImplementation project(':groovy-dateutil')
     testImplementation 'net.javacrumbs.json-unit:json-unit:2.28.0'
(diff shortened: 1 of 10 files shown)


subprojects/groovy-json/build.gradle
```diff
@@ -21,13 +21,19 @@ plugins {
 }
 
 dependencies {
-    api rootProject  // JsonBuilder extends GroovyObjectSupport...
```
paulk-asert, line 24: I think it is useful to keep the comment, otherwise it is hard to track down why we use `api` here and not `implementation`. We should have a similar comment for each api dependency.

eric-milles: comment restored

PR #1671, opened 2022-01-04
Add equals() and hashCode() methods to the ImportNode class.  

opeongo:
These methods are useful for developing an alternate JSR223 ScriptEngine class.  In this situation it is necessary to avoid duplicate import entries, and the easiest way to do this is by using methods in the Collections API that rely upon object equals().

This PR contains code that should be benign other than fulfilling the equals contract.

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ImportNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ImportNode.java
@@ -31,6 +31,7 @@ public class ImportNode extends AnnotatedNode {
     private final String packageName;
     private final boolean isStar;
     private final boolean isStatic;
+    private int hashCode;
 
     /**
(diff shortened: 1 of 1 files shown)


eric-milles:
It is possible in a source unit to have multiple imports for the same type.  So it is normal to have multiple AST nodes that represent the same logical import.  It is best IMO you de-duplicate using your own filtering logic.

opeongo:
Yes, precisely.  I would like to use my own filtering logic.  Having this
equals() method in place will make that very easy to do.

paulk-asert:
Given that the `ImportNode` is _slightly_ mutable, via `setType`, you would need to reset the `hashCode` cache in the `setType` method, or not cache, but just resetting should be fine.

paulk-asert:
Merged with the suggested hashCode modification. Thanks!

PR #1673, opened 2022-01-07
Make order of generated MOP methods deterministic to facilitate repea…

idooley:
Switch to a LinkedHashMap to preserve insertion order when creating MOP methods. This will make the MOP methods derived from methods in a superclass be emitted in the same order as the original methods in the superclass. This should make builds more repeatable.


idooley:
Independently fixed in [GROOVY-10704](https://issues.apache.org/jira/browse/GROOVY-10704)

PR #1674, opened 2022-01-07
GROOVY-5001, GROOVY-5491, GROOVY-6144: map property precedence

eric-milles:
There is still some asymmetry between read and write for protected fields.  But I wasn't sure if I should add a protected check to read or remove the protected check from write (see `MetaClassImpl#setProperty` circa GROOVY-5985 comment).  Also, read will use a category-supplied `propertyMissing` method before general `get(String)` method(s) and write does not.

https://issues.apache.org/jira/browse/GROOVY-5001
https://issues.apache.org/jira/browse/GROOVY-5491
https://issues.apache.org/jira/browse/GROOVY-6144

see also:
https://issues.apache.org/jira/browse/GROOVY-662
https://issues.apache.org/jira/browse/GROOVY-8065
https://issues.apache.org/jira/browse/GROOVY-8074


src/main/java/groovy/lang/MetaClassImpl.java
```diff
@@ -2085,73 +2089,65 @@ public void setProperty(Object object, Object newValue) {
...
-        }
-
         Tuple2<MetaMethod, MetaProperty> methodAndProperty = createMetaMethodAndMetaProperty(sender, theClass, name, useSuper, isStatic);
         MetaMethod method = methodAndProperty.getV1();
 
-        //----------------------------------------------------------------------
-        // getter
-        //----------------------------------------------------------------------
-        MetaProperty mp = methodAndProperty.getV2();
+        if (method == null || "class".equals(name) || (isMap && "empty".equals(name))) {
```
danielsun1106, line 2097: It's better to extract the common checking logic, here is same code at line 1981:
`if (method == null || "class".equals(name) || (isMap && "empty".equals(name)))`

eric-milles: Why is it better to extract?  `getProperty` and `getEffectiveMetaProperty` are essentially the same general structure, only differing in a few details.  I don't really have the energy to extract the true repeated structure(s).  You can follow the two methods and see the parallel, so I think it is decent in this state.

paulk-asert:
I agree it would be good for the get and set behaviors to be aligned. The suggested external behavior seems okay and I would add 'metaClass' as an additional case to test for.

What worries me is the change in behavior for closures. The following currently passes:
```
class Foo {
    @Delegate Map m = [:]
    def x = 4
    protected y = 5
    private z = 6
    void method() {
        def w = 3
        assert 3 == { -> w }()
        assert null == { -> x }()
        assert null == { -> y }()
        assert null == { -> z }()
    }
}

class Bar {
    def x = 4
    protected y = 5
    private z = 6
    void method() {
        def w = 3
        assert 3 == { -> w }()
        assert 4 == { -> x }()
        assert 5 == { -> y }()
        assert 6 == { -> z }()
    }
}

new Foo().method()
new Bar().method()
```
This exhibits the "potentially strange" map takes precedence behavior but is at least consistent across `x`, `y`, and `z`.
After the change, just the `x` behavior has altered precedence.

My suspicion is that what you have is close but needs some more work. I am also thinking we might need to define a special Map metaclass that would provide back the old behavior - again it would be good to align get/set even for that case. Users relying on the old behavior could manually apply that metaclass. I would regard this as a legacy compatibility mechanism rather than something we would recommend is used for new development.

eric-milles:
It was intended to make this read the property and not from the map:
```groovy
class C {
    @Delegate Map m = [:]
    def x = 4
    public y = 5
    void method() {
        def v = 3
        assert 3 == { -> v }()
        assert 4 == { -> x }() // previously preferred map lookup
        assert 5 == { -> y }() // previously preferred map lookup
    }
}
```

I thought the first step would be to have read of `public` field or accessor gain precedence over map lookup.  For write it already includes `public` and `protected`.  I took the cautions approach by dealing with `public` first and leaving `protected` as-is.

The idea here is that if you created a type that implements `Map` but then added public API, that API should take precedence over the general map property mechanism.

For legacy support, any code may replace `name` with `get('name')` to stick with the map.

eric-milles:
Added test cases for "metaClass" and "MetaClass" in dynamic groovy.  "class", "Class", "empty", "Empty", "metaClass" and "MetaClass" probably need tests under static compilation.

eric-milles:
GROOVY-6151 is a good illustration of user expectations (comments added):
{code:groovy}
class TestError extends HashMap {
    String id
    boolean equals(Object o) {
        if (this.is(o)) return true
        if (getClass() != o.class) return false // "o.class" may not be "getClass()" for some types
        if (!super.equals(o)) return false
        TestError testError = (TestError) o
        if (id != testError.id) return false // "id" and "this.id" refer to the property; "testError.id" does not refer to "getId()"
        return true
    }
}
{code}

With this PR, all references to "id" will refer to the declared property and not a Map property.  "class" vs "getClass()" and "empty" vs "isEmpty()" remain unchanged.

src/main/java/groovy/lang/MetaClassImpl.java
```diff
@@ -1959,100 +1959,105 @@ public void setProperties(Object bean, Map map) {
...
-        if (method == null && !useSuper && !isStatic && GroovyCategorySupport.hasCategoryInCurrentThread()) {
-            method = getCategoryMethodGetter(sender, PROPERTY_MISSING, true);
-            if (method != null) arguments = new Object[]{name};
+            //------------------------------------------------------------------
+            // non-public field
+            //------------------------------------------------------------------
+            if (mp != null) {
+                try {
+                    return mp.getProperty(object);
+                } catch (IllegalArgumentException | CacheAccessControlException e) {
```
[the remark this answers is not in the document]
eric-milles: @sonatype-lift ignore

PR #1677, opened 2022-01-16
GROOVY-10451: Sealed classes incorrectly allow a self-reference in th…

paulk-asert:
…e permitted subclasses


eric-milles:
You may be able to add the permitted subclasses to `org.codehaus.groovy.control.ResolveVisitor#checkCyclicInheritance`.  Then you'd get errors for direct and indirect cyclic references.

paulk-asert:
> You may be able to add the permitted subclasses to `org.codehaus.groovy.control.ResolveVisitor#checkCyclicInheritance`. Then you'd get errors for direct and indirect cyclic references.

@eric-milles Can you think of an example involving cyclic sealed that wouldn't already be violating cyclic inheritance?

PR #1679, opened 2022-01-23
GROOVY-8433: Category transform implies static methods

eric-milles:
Another way to do this would be to add a second transform class to the annotation that runs during `SEMANTIC_ANALYSIS` (before `InnerClassVisitor`) and adds `static` to methods.

https://issues.apache.org/jira/browse/GROOVY-8433


paulk-asert:
Merged, thanks! Let's keep your alternate possible implementation in mind if anyone has any further issues with this feature.

PR #1681, opened 2022-02-04
GROOVY-9866: process discovered sources before units that depend on them

eric-milles:
> The discovery mechanism is present in `ClassNodeResolver`/`CompileUnit`/`CompilationUnit` to locate a source, add it and start resolve phase over. The map `sources` in `CompilationUnit` is insertion-ordered, which means the sources will be processed in the order they are discovered. A superclass or superinterface needs to be resolved before its subtype(s).

https://issues.apache.org/jira/browse/GROOVY-9866

To enable this order, process `sources` in reverse.


eric-milles:
I could do a counting loop from length-1 to 0 instead of reverse to save that cost.

eric-milles:
I did run into a case of same-package types where the reverse order causes missed inner type resolution.  I'll try to construct a test case.  At least this commit can serve as an experiment for now.

PR #1686, opened 2022-02-15
GROOVY-10484: `@NamedVariant`: bind `@NamedParam` into `@NamedParams`

eric-milles:
Wrapping `NamedParam` annotations in a `NamedParams` container is straightforward.  ~For default values, I wanted to use `map.containsKey(name) ? map.name : defaultValue` instead of `map.name ? map.name : defaultValue` -- in case `null` or `0` or another falsy value is supplied.  This change does not give the map a chance to supply a default value, but I thought that was okay in order to allow user to supply falsy value explicitly.~

~@paulk-asert  One case that was hard to understand comes form the `record` tests.  The example has `null` passed for an `int` and results in `0` whereas this code produces a cast exception.  I wanted to understand if that was desired or just a happy accident.~
```groovy
assert new ColoredPoint(x: 0, y: null).toString() == 'ColoredPoint[x=0, y=0, color=white]'
```

https://issues.apache.org/jira/browse/GROOVY-10484

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
+++ b/src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
@@ -20,6 +20,7 @@ package org.codehaus.groovy.transform;
 
 import groovy.transform.NamedDelegate;
 import groovy.transform.NamedParam;
+import groovy.transform.NamedParams;
 import groovy.transform.NamedVariant;
 import org.apache.groovy.ast.tools.AnnotatedNodeUtils;
 import org.codehaus.groovy.ast.ASTNode;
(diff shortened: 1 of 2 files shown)


paulk-asert:
I am still looking at the changes. Some effort was put into making primitives behave in an "expected" way but whether we are consistent to what happens elsewhere needs further investigation. I am wondering whether it would be better to adapt `StaticTypeCheckingVisitor` to know about the individual `NamedParam` annotations not just the collector. The annotation collector is already currently applied but not until classgen phase.

eric-milles:
This commit has been reduced to 10484 changes only.

eric-milles:
NOTE: This is a solution that I can backport to Groovy 2.5, which lacks the `@Repeatable` annotation: https://github.com/apache/groovy/blob/GROOVY_2_5_X/src/main/groovy/groovy/transform/NamedParam.java

paulk-asert:
I still wonder whether forcing the collation early should be a 2.5 only thing - simply because if other frameworks like Grails or Micronaut did similar NamedParam additions we'd be requiring them to also collate. Let's commit as is for now. If it ever becomes a problem we can always refactor on 3 and above.

paulk-asert:
@eric-milles The alternative would look something like: #1693

paulk-asert:
PR#1693 was merged in Groovy 3 and above. This PR was merged with minor adjustments onto 2_5_X.

PR #1687, opened 2022-02-16
GROOVY-10497: `@NamedVariant`: don't replace explicit value with default

eric-milles:
For default values, I want to use `map.containsKey(name) ? map.name : defaultValue` instead of `map.name ?: defaultValue` -- in case `null` or `0` or another falsy value is supplied.  This change does not give the map a chance to supply a default value, but I thought that was okay in order to allow user to supply falsy value explicitly.

One case that was hard to understand comes form the `record` tests.  The example has `null` passed for an `int` and results in `0` whereas this code produces a cast exception.  I wanted to understand if that was desired or just a happy accident.
```groovy
assert new ColoredPoint(x: 0, y: null).toString() == 'ColoredPoint[x=0, y=0, color=white]'
```

A null test could be inserted for primitives, but it would be more complex and would probably mean one of the operands was evaluated twice.  Probably not a huge concern, but I thought I would meantion it.

https://issues.apache.org/jira/browse/GROOVY-10497

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
+++ b/src/main/java/org/codehaus/groovy/transform/NamedVariantASTTransformation.java
@@ -70,10 +70,12 @@ import static org.codehaus.groovy.ast.tools.GeneralUtils.entryX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.getAllProperties;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.list2args;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.mapX;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.notNullX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.param;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.plusX;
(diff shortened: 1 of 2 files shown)


eric-milles:
`map.getOrDefault(name,defaultValue)` is another option.  The ternary avoids evaluating `defaultValue` if it is not used.

paulk-asert:
I am in favor of removing the truthy and null-handling aspect. I think the intention was just to handle the null case but I think we shouldn't even handle that. We could always add a `nullSafePrimitive` flag to `NamedVariant` and/or `MapConstructor` but I suggest we don't do that for now.

eric-milles:
Not sure I follow fully.  Are you saying go with the replacement of elvisX with ternaryX/containsKey, but leave off the primitive null check (lines 283-285)?

paulk-asert:
Yes. And I was suggesting:

- we can add in some additional throw IllegalArgumentException statements in places where the resulting error message is confusing, e.g. when we call containsKey on a null map
- potentially add in an option to keep the replace null with default (basically keep lines 283-285 or similar as an option) at a later date if there is demand, we are kind of breaking behavior which some might be relying on - having the option would be a workaround for them

PR #1691, opened 2022-02-24
Issue GROOVY-10505 trait composer methods as generated

aulea:


src/main/java/org/codehaus/groovy/transform/trait/TraitASTTransformation.java
```diff
@@ -498,7 +500,7 @@ private void processField(final FieldNode field, final MethodNode initializer, f
                         ClassNode.EMPTY_ARRAY,
                         returnS(initCode.getExpression())
                 );
-                helper.addMethod(fieldInitializer);
+                ClassNodeUtils.addGeneratedMethod(helper, fieldInitializer);
```
paulk-asert, line 503: I am unsure whether this one is needed. The whole helper class is marked as generated. Does the coverage tool not recognize that?

aulea: You are right, whole helper class is marked as generated and this is unnecessary. Reverted it with new commit.

paulk-asert:
Merged, thanks!

PR #1694, opened 2022-02-28
GROOVY-10516: Script @groovy.transform.Field AST transformation gener…

paulk-asert:
…ics handling with fully qualified types


paulk-asert:
Not needed, so closing.

PR #1696, opened 2022-03-03
Groovy 3: Gradle 7.2 for Java 16+

eric-milles:
Here are some changes to support running Groovy 3 builds with Java 16+ for purposes of evaluating issue reports.  `upload.gradle` still needs to be migrated to use the `maven-publish` plugin.


paulk-asert:
I brought across the minor refactoring changes compatible with the existing gradle version

eric-milles:
Now I see that you can use one JDK to build and another to test using `-Ptarget.java.home=...`

PR #1699, opened 2022-03-05
Groovy-10520: Capture correct line numbers for Closure block statement.

johnrengelman:
Closes Groovy-10520

### DIFF
--- a/src/antlr/GroovyParser.g4
+++ b/src/antlr/GroovyParser.g4
@@ -487,7 +487,7 @@ lambdaBody
 
 // CLOSURE
 closure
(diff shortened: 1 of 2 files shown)


paulk-asert:
I suspect L#563 can now also lose the "sep?" but otherwise LGTM.
I haven't tried to measure if it impacts performance but accuracy would trump performance.

johnrengelman:
Thanks @paulk-asert 
I included that change and forced pushed the branch with a commit message that more closely follows existing comments.

daniellansun:
Merged. Thanks!

PR #1704, opened 2022-03-20
GROOVY-8243: support closure for functional interface that extends trait

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-8243


paulk-asert:
Merged, thanks!

PR #1706, opened 2022-03-22
GROOVY-10540: add `GroovyObject` interface before STC and classgen

eric-milles:
Adding `GroovyObject` at the start of `INSTRUCTION_SELECTION` will impact the abstract methods of a class for that phase.  Not many transforms run there besides STC.  I think the risk is low.

I left `Verifier` alone because `JavaStubGenerator` runs it to get `GroovyObject` add-ons, etc.  And in rare cases if a class is added during `INSTRUCTION_SELECTION`.

https://issues.apache.org/jira/browse/GROOVY-10540

### DIFF
--- a/src/main/java/groovy/transform/stc/FromAbstractTypeMethods.java
+++ b/src/main/java/groovy/transform/stc/FromAbstractTypeMethods.java
@@ -19,6 +19,7 @@
 package groovy.transform.stc;
 
 import org.codehaus.groovy.ast.ASTNode;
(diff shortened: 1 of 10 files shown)


eric-milles:
@kriegaex I'm considering the risk of moving the changes back to Groovy 3 and 2.5.  There is a small difference in map property access that could catch people off guard.  It needs to be an object that implements both the `GroovyObject` and `Map` interfaces, so it is not widespread.  But there have been several issues filed over time for such cases, so it does exist in the wild.

PR #1708, opened 2022-03-31
GROOVY-10561: @NamedVariant self referential default values are not c…

paulk-asert:
…orrectly resolved


paulk-asert:
The PR works for the simple case of the initial value being a reference to a previously defined parameter, as in the provided example, but should really be a visitor so that it could handle expressions involving such parameters. I'll do that next.

eric-milles:
Is this the same way `Verifier` handles it for normal methods?

PR #1710, opened 2022-04-06
groovy-bom dependency has no LICENSE

soundasleep:
I've recently upgraded from Groovy 3.x to 4.x. As part of that upgrade, my `:checkLicense` task (using https://github.com/jk1/Gradle-License-Report) has identified that `groovy-bom` has no license:

```json
{
    "dependenciesWithoutAllowedLicenses": [
        {
            "moduleLicense": null,
            "moduleVersion": "4.0.1",
            "moduleName": "org.apache.groovy:groovy-bom"
        }
    ]
}
```

I'm assuming that groovy-bom is under Apache (as the rest of Groovy). I don't know if this is the "correct" way to add a license for `:checkLicense`. 

I'm guessing https://github.com/apache/groovy/blob/master/gradle/licenses.gradle needs to be updated too?

### DIFF
--- a/subprojects/groovy-bom/LICENSE
+++ b/subprojects/groovy-bom/LICENSE
@@ -1,0 +1,216 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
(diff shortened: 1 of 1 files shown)


paulk-asert:
This is a little strange. We don't have separate LICENSE files in general in each subproject since the main LICENSE file covers all subprojects. The LICENSE is embedded into the respective jars during assembly[1]. The BOM artifact has no jar, just a pom which contains among other things the following excerpt:
```
<licenses>
  <license>
    <name>The Apache Software License, Version 2.0</name>
    <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
    <distribution>repo</distribution>
  </license>
</licenses>
```
We'll need to examine what `:checkLicense` is expecting.

[1] https://github.com/apache/groovy/blob/master/build.gradle#L192

paulk-asert:
It looks like it would be safe (and in fact desirable) for you to set the `:checkLicense` `excludeBoms` property to true.
The relevant code is here:
https://github.com/jk1/Gradle-License-Report/blob/master/src/main/groovy/com/github/jk1/license/LicenseReportExtension.groovy#L97
So, if I understand correctly, it will not try to check boms if there is no jar artifact.

paulk-asert:
I'll close this based on my earlier comments. If you don't agree, let's discuss further.

PR #1711, opened 2022-04-08
GROOVY-10543 Publish groovy-all as library

ljacomet:
This is done by updating the attributes of the configurations
created by the Gradle java-platform plugin to match the
attributes of regular libraries.

### DIFF
--- a/buildSrc/src/main/groovy/org.apache.groovy-all.gradle
+++ b/buildSrc/src/main/groovy/org.apache.groovy-all.gradle
@@ -112,4 +112,21 @@ configurations {
             }
         }
     }
+    // GROOVY-10543 - The platform configurations are published as libraries
+    apiElements {
+        attributes {
+            attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category, Category.LIBRARY))
+            attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(LibraryElements, LibraryElements.JAR))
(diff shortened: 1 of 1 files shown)


ljacomet:
Hey folks,

I opened this as a draft PR as it currently contains 0 tests. I could not find tests related to the published Gradle Module Metadata. If I missed them, could you point them out?

@melix Wondering what you think of that simple solution.

buildSrc/src/main/groovy/org.apache.groovy-all.gradle
```diff
@@ -112,4 +112,21 @@ configurations {
             }
         }
     }
+    // GROOVY-10543 - The platform configurations are published as libraries
+    apiElements {
+        attributes {
+            attribute(Category.CATEGORY_ATTRIBUTE, objects.named(Category, Category.LIBRARY))
```
kkocel, line 118: ihmo java version is missing & required

ljacomet: I thought about it and decided to not include it in the first version. Since `groovy-all` does not have a JAR itself, constraining the Java version here is not really needed.
The libraries it depends on do have that attribute and will be the ones failing to resolve if they are consumed in the wrong context.

melix: I agree that the Java version doesn't make sense in this context.

melix:
Sounds like a reasonable fix

paulk-asert:
Merged. Thanks!

We will add tests separately. I have just tried a few manual tests in the meantime and it seems to be behaving as expected.

PR #1713, opened 2022-04-15
GROOVY-10582: Funnel checkPermission through VMPlugin (JEP-411)

paulk-asert:


paulk-asert:
As per comments in the associated issue, we will try a different approach here.

PR #1716, opened 2022-04-23
GROOVY-10593: fixes groovydoc sometimes links to incorrect classes

stalb:
Fixes imports that were not rewritten to replace dots with slashes.

### DIFF
--- a/subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/SimpleGroovyClassDoc.java
+++ b/subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/SimpleGroovyClassDoc.java
@@ -18,6 +18,9 @@
  */
 package org.codehaus.groovy.tools.groovydoc;
 
+import com.github.javaparser.ast.expr.Name;
+import com.github.javaparser.JavaParser;
+
 import org.codehaus.groovy.groovydoc.GroovyAnnotationRef;
 import org.codehaus.groovy.groovydoc.GroovyClassDoc;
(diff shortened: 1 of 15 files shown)


subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/SimpleGroovyClassDoc.java
```diff
@@ -682,11 +686,24 @@ private static boolean isPrimitiveType(String name) {
...
+            // need this for correct resolution of static imports
+            if (targetClassName != null){
+                GroovyClassDoc doc = null;
+                Optional<Name>  name = new JavaParser().parseName(targetClassName.replace('/','.')).getResult();
+                String staticPart = "";
+                while (doc == null && name.isPresent()) {               
+                    doc = ((SimpleGroovyRootDoc)rootDoc).classNamedExact(name.get().asString().replace('.','/')+staticPart);
+                    staticPart = "."+name.get().getIdentifier()+staticPart;
+                    name = name.get().getQualifier();
+                }
```
stalb, line 706: _SimpleGroovyRootDoc.classNamedExact_ is used at several other places in this class. 
Thus I can't exclude that the same problem occurs elsewhere. 

Perhaps it  would be better to change the method itself.

paulk-asert: I had a look at the other places which call that method. I am unsure if any are affected in the same way, so I've merged as is and we can always refine later if we find a case.

paulk-asert:
Merged, thanks!

PR #1724, opened 2022-06-10
Fix missleading error message in GroovyScriptEngine

aalku:
The error message is missleading and needs a fix. Otherwise it can show "Cannot open URL: folderfile" instead of "folder/file" and the user might wonder why is it looking for "folderfile".


paulk-asert:
Merged, thanks.

PR #1726, opened 2022-06-16
Add indention

vanillajonathan:


paulk-asert:
We would not likely include indentation recommendations in this document. Groovy as a language, like Java, is somewhat whitespace agnostic, i.e. it supports multiple indentation practices including 4 spaces, 2 spaces, tabs etc.
The Groovy source code itself mostly follows the 4 space guideline and its not a bad one to follow but this document is mostly about style changes a Java developer might make when moving to Groovy. With that in mind, a developer coming to Groovy can continue whatever whitespace conventions they might already be using in Java.

vanillajonathan:
This leads to the Groovy community applying indention inconsistently.
Developers are free to indent however they see fit, however I think it would serve the Groovy community well to have the recommended indention specified in the style guide. It would make the code style more homogeneous which I think the Groovy developer community would appreciate.

paulk-asert:
Sometimes consistency is nice, sometimes inclusivity is nice.
OO vs functional style, dynamic vs static, tabs vs spaces.
As a general rule Groovy tries to be agnostic.
For those who want additional consistency there are IDE features and tools like CodeNarc which can be utilised:
https://codenarc.org/codenarc-rules-formatting.html#indentation-rule
(The linked rule checks by default for 4 spaces and other rules can also check brace positioning, line length and spacing and numerous other parameters.)

vanillajonathan:
The nice thing with a style guide is that it is just a guide, not rules. so even with indention defined in the style guide developers would still be free to use whatever they want.

PR #256, opened 2016-02-09
Tighten the filtering prefixes

yaccz:
as the current rules are a bit overzealous and may filter relevant
information.

PR #1727, opened 2022-06-16
StackTraceUtils: Add 'jdk.internal' to the list of ignored packages

alewando:
Include 'jdk.internal' in the default list of packages that are filtered out by StackTraceUtils' `sanitze()` methods

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/StackTraceUtils.java
+++ b/src/main/java/org/codehaus/groovy/runtime/StackTraceUtils.java
@@ -64,7 +64,8 @@ public class StackTraceUtils {
                             "java.," +
                             "javax.," +
                             "sun.," +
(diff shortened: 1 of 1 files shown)


paulk-asert:
LGTM. As you will have seen in the code (but for the benefit of others searching here later) you can provide a `groovy.sanitized.stacktraces` system property in the meantime (or for anyone not happy with this change).

paulk-asert:
Actually, I have just recalled a related much earlier request: #256 (GROOVY-7756) which wants to tighten the exclusion list. I'll write an email and see if anyone else has feedback before merging but I will add `jdk.internal` to my suggested path forward in the email.

paulk-asert:
Merged, thanks! I created this issue to track the details: GROOVY-10665.

PR #1729, opened 2022-06-20
Show Path in object explorer.

sandipchitale:
Show Path in object explorer. Show 'classname instance' when starting… path is not specified.


sandipchitale:
For expression ```def p = System.properties```, the Path shown is:
image

sandipchitale:
Need to fix some bad commits.

PR #1730, opened 2022-06-21
Show path to current level in object browser header. Show 'class name … instance' when starting path is not specified.

sandipchitale:


sandipchitale:
For expression def p = System.properties, the Path shown is:

image

paulk-asert:
I quite like the idea. I also played around with putting the path in the frame title instead of the label:
```
frame = swing.frame(title: "Groovy Object Browser: $path", location: [200, 200],
```
I am unsure which I prefer. It is visually cleaner in the title I think but does become truncated more quickly.

sandipchitale:
Something like this:

image

The user can select the path from the non-editable text field.

Also, I am proposing to close the current ObjectBrowser if CTRL key was down when double-clicked and open the next one to prevent the proliferation of frames.

If you like that I can commit.

sandipchitale:
I went ahead and submitted the two commits.  I used Alt-Double-Click instead o Ctrl-Double-Click because the Ctrl key is also used for extending the selection. Please pick and choose those that make sense.

sandipchitale:
@paulk-asert Thanks for merging and making corrections to my possible silly mistakes.

sandipchitale:
Works for me on master. Works great.

paulk-asert:
@sandipchitale No silly mistakes, I was just finessing slightly with an eye on maintenance and consistency.

PR #1731, opened 2022-06-21
Fixed issue in PackageHelperImpl#getPackageNames

justinchuch:
## Issue:

PackageHelperImpl#getPackageNames would not be able to find and return
the package names if the package path or filename of the jar contains
plus signs due to the behavior of URLDecoder#decode.

URLDecoder#decode would decode the plus sign (i.e. `+`) as empty space, causing `java.io.FileNotFoundException`.

## Reproduction

To reproduce the case, include a plus sign in the `groovyVersion` in the `gradle.properties`, e.g. `5.0.0+SNAPSHOT`, run the `PackageHelperImplTest`, and `testLoadAndGetPackagesEmpty` would fail.

```
WARN [org.apache.groovy.groovysh.util.PackageHelperImpl] Error opening jar file : '/home/jus/repos/apache/groovy/build/libs/groovy-raw-5.0.0+SNAPSHOT.jar' : java.io.FileNotFoundException: /home/jus/repos/apache/groovy/build/libs/groovy-raw-5.0.0 SNAPSHOT.jar (No such file or directory)
...
```

## Proposed fix:

Use `Paths.get(url.toURI()).toFile()` instead of `new File(URLDecoder.decode(url.file, 'UTF-8'))`.


paulk-asert:
Looks good - merged!

PR #1732, opened 2022-06-24
GROOVY-10223: support assignment/cast of `Optional` to array/collection

eric-milles:
Adds support for assignment of optional to arrays or collections -- and by extension an "iterator()" method and for-each iteration.  Does not support spread arguments (aka "def list = [*optional]" or "m(*optional)")
```groovy
Object[] array = optional
List<Object> list = optional
Iterator<Object> optional.iterator()
for (value in optional) { /*...*/ } // ... executed only if value is present
```

https://issues.apache.org/jira/browse/GROOVY-10223

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
+++ b/src/main/java/org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.java
@@ -50,6 +50,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Supplier;
 import java.util.stream.BaseStream;
 import java.util.stream.DoubleStream;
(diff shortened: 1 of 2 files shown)


eric-milles:
And this kind of thing is not supported without `getAt(int)` or `getAt(Integer)`:
```groovy
def opt = Optional.of('')
def (String x) = opt
assert x == ''
```

eric-milles:
@paulk-asert I agree, it seems the extension method `Optional<T> collect(Optional<S>,Closure<T>)` is not that valuable.  If deprecated, it would not resolve as an extension, which would then cause `List<T> collect(Object,Closure<T>)` to be selected instead.  I doubt any code that makes use of this is prepared for a list return value.

paulk-asert:
> @paulk-asert I agree, it seems the extension method `Optional<T> collect(Optional<S>,Closure<T>)` is not that valuable. If deprecated, it would not resolve as an extension, which would then cause `List<T> collect(Object,Closure<T>)` to be selected instead. I doubt any code that makes use of this is prepared for a list return value.

Yes, it would be a breaking change. We'd deprecate it now in Groovy 4 and remove in Groovy 5 with explanation in release notes. Folks wanting to prepare could potentially test with -Dgroovy.extension.disable=collect but that isn't very selective.
We can explain the behavior either way, I just think the explanation is much simpler if it behaves like all the other fallback variants.

PR #1733, opened 2022-06-26
GROOVY-10473: add stream `getAt(int)` and `getAt(IntRange)` extension methods for quick access

eric-milles:
`IntStream`, `LongStream`, and `DoubleStream` do not have sensible default for `getAt(int)` except maybe `NaN` for `DoubleRange` unless they return boxed types.  They could have `getAt(IntRange)` added safely.  I don't think `getAt(Range)`, `getAt(EmptyRange)` or `getAt(ObjectRange)` are necessary.

https://issues.apache.org/jira/browse/GROOVY-10473

spread-dot syntax (aka `stream*.property`) is already supported through `Stream#iterator()` as [noted](https://issues.apache.org/jira/browse/GROOVY-10473?focusedCommentId=17489094&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17489094)

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/StreamGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/StreamGroovyMethods.java
@@ -18,6 +18,8 @@
  */
 package org.codehaus.groovy.runtime;
 
+import groovy.lang.IntRange;
+
 import java.lang.reflect.Array;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


eric-milles:
https://issues.apache.org/jira/browse/GROOVY-10666 discusses the possibility of supporting multiple-assignment for streams via `getAt(IntRange)`

daniellansun:
+1

paulk-asert (approved):
LGTM.

My main concern was dedicating another powerful shorthand to finish with a terminal operation. I think we could define a `take` extension method for `IntRange` to provide a reasonable shorthand when wanting to retain the stream nature. Something like:
```
    public static <T> List<T> getAt(final Stream<T> self, final IntRange range) {
        if (range.isReverse()) throw new IllegalArgumentException("reverse range");
        return take(self, range).collect(Collectors.toList());
    }

    public static <T> Stream<T> take(final Stream<T> self, final IntRange range) {
        if (range.isReverse()) throw new IllegalArgumentException("reverse range");
        return self.skip(range.getFromInt()).limit(range.size());
    }
```
Could be used like:
```
def names = ['foo','bar','baz']
List result = names.stream().take(0..1).filter(s -> s.contains('b'))
assert result == ['bar']
```
I'll also note that the error when trying to use multiple assignment once this change is in place goes from a MME for getAt to an IllegalStateException "stream has already been operated upon ..." but if we progress GROOVY-1066, this shouldn't be a problem.

PR #1736, opened 2022-07-03
GROOVY-10676: ErrorReporter (used by FileSystemCompiler) can throw an…

paulk-asert:
… IOOB exception on files with CR only


daniellansun:
+1

PR #1743, opened 2022-07-10
Tooltip and CTRL+BUTTON3 click to copy cell value to clipboard.

sandipchitale:
- Hovering over cells shows a tooltip.
- CTRL+Right Click (BUTTON 3) copies the value to the clipboard.


paulk-asert:
The tooltip stuff looks great. I am wondering whether a simple right-click => "copy cell" popup menu might be more consistent since we have such actions in other places. Thoughts? Let me know if you want to look into that or want me to play at my end.

sandipchitale:
Thanks. I like the use of the popup menu better. I do now know SwingBuilder very well and could not figure out how to set up the popup menu on the table. If you are familiar and have time please go ahead and implement that. If not, I will look at it next weekend. Do not want to push more work to you though.

paulk-asert:
The tooltip part of the PR has been merged. Thanks!
I'll try to have a look at the popup menu idea in the next few days.

PR #1746, opened 2022-07-14
GROOVY-10679: Annotations not being correctly placed in native records

paulk-asert:


paulk-asert:
I haven't yet created a constant for "_SKIPPABLE_ANNOTATIONS" and similarly there is a private but not public constant for "_RECORD_HEADER". We could make public constants or usage could easily be hidden behind some API. I was just avoiding new API in the first instance but I welcome feedback.

daniellansun:
+1

paulk-asert:
Merged!

PR #1749, opened 2022-07-19
GROOVY-10689: Groovydoc for Groovy 3+ documents classes at too early …

paulk-asert:
…a phase of compilation


paulk-asert:
Merged!

PR #1752, opened 2022-07-26
Persist the order of methods

musketyr:
to ensure the generated trait classes are the same for the same source code.

This is a one of the possible solutions to solve [GROOVY-10704](https://issues.apache.org/jira/browse/GROOVY-10704).

see #1753 for an alternative version which only sorts the methods when generating the trait


paulk-asert:
This is the approach I would take to address the issue. I'll check into the errors - one looks explainable at first glance.

musketyr:
Thank you @paulk-asert!

paulk-asert:
Merged, thanks!

PR #1753, opened 2022-07-26
sort the methods from the trait when generating methods

musketyr:
This is a one of the possible solutions to solve [GROOVY-10704](https://issues.apache.org/jira/browse/GROOVY-10704).

see #1752 for an alternative version which always keeps the methods in the original order


paulk-asert:
I think the other offered solution is preferred in terms of styling with the rest of the Groovy codebase. We don't sort except in very particular circumstances.

PR #1756, opened 2022-08-01
GROOVY-5744, GROOVY-10666: multi-assign via `iterator()` or `getAt(int)`

eric-milles:
Support multi-assign for `Stream` and `Iterator`.  Improve efficiency of multi-assign for types that do not support direct indexing like `Set`.

Translate "def (one,two) = rhs" into (pseudo-code):
```groovy
def one, two;
{
  def iter = rhs.iterator(), first;
  if (iter.hasNext() && (first = iter.next()) !== rhs) {
    one = first
    two = iter.hasNext() ? iter.next() : null
  } else {
    one = rhs.getAt(0)
    two = rhs.getAt(1)
  }
}
```

https://issues.apache.org/jira/browse/GROOVY-10666
https://issues.apache.org/jira/browse/GROOVY-5744

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/BinaryExpressionHelper.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/BinaryExpressionHelper.java
@@ -33,7 +33,6 @@ import org.codehaus.groovy.ast.expr.Expression;
 import org.codehaus.groovy.ast.expr.FieldExpression;
 import org.codehaus.groovy.ast.expr.ListExpression;
 import org.codehaus.groovy.ast.expr.MethodCallExpression;
-import org.codehaus.groovy.ast.expr.NotExpression;
 import org.codehaus.groovy.ast.expr.PostfixExpression;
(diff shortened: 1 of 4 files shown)


blackdrag:
I don't see changes from double quotes to single quotes as formatting change, nor do I see a change from {} to {->} as such a change. So where any of these changes required for this code change to pass?

paulk-asert:
On my list to look at tomorrow.

daniellansun:
Merged. Thanks.

paulk-asert:
I'll probably revert this, I have an alternative implementation which I have spoken to Eric about but I have been too busy with ApacheCon to get my PR ready for review.

daniellansun:
Got it.

eric-milles:
@paulk-asert Instead of revert, can you just replace the "multiple declaration or assignment" block?  There are some other fixups, including new test cases that I'd like to keep.

paulk-asert:
@eric-milles I'll see what I can do.

PR #1760, opened 2022-08-06
GROOVY-10278: `CompilerConfiguration`: improve target bytecode selection

eric-milles:
As described in GROOVY-10278, select the nearest target bytecode version.  As described on the mailing list, set minimum Java to 11 for Groovy 5.  This sets the stage for using nestmates ([JEP-181](https://openjdk.org/jeps/181)) without need for a fallback solution or version checking.

https://issues.apache.org/jira/browse/GROOVY-8591
https://issues.apache.org/jira/browse/GROOVY-10278
https://issues.apache.org/jira/browse/GROOVY-10595
https://issues.apache.org/jira/browse/GROOVY-10687
https://github.com/apache/groovy/pull/1661 (original incarnation of this and other proposed changes for `CompilerConfiguration`)

### DIFF
--- a/.github/workflows/groovy-build-test.yml
+++ b/.github/workflows/groovy-build-test.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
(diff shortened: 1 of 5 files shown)


daniellansun:
Merged. Thanks.

PR #1764, opened 2022-08-14
Gradle tweaks

Goooler:


buildSrc/src/main/groovy/org.apache.groovy-common.gradle
```diff
@@ -42,10 +42,6 @@ group = 'org.apache.groovy'
 version = sharedConfiguration.groovyVersion.get()
 
 repositories {
-    if (rootProject.hasProperty('buildInChina')) {
-        // Try to use Aliyun maven repository when building in China
-        maven { url 'https://maven.aliyun.com/nexus/content/groups/public' }
-    }
```
Goooler, line 48: Aliyun maven shouldn't be applied here, if someone needed, can add this to `init.gradle`.

paulk-asert:
Most of the commits seem okay. I'll wait a little longer for any other feedback and I want to try a few more things.

paulk-asert:
Merged. Thanks!

PR #1766, opened 2022-08-17
GROOVY-8788: STC: prefer closer parameter match over self-type match

eric-milles:
Given choice between extension methods `m(String,Object)` and `m(Object,String)`, prefer closer parameter matching.  This aligns with the method selection of the dynamic runtime.  So there are some edge cases that were giving false STC errors or missing them.

This is a breaking change!  `getAt(Map,String)` was added to prevent STC errors for `Type obj = map[str]` expressions.

https://issues.apache.org/jira/browse/GROOVY-8788

https://issues.apache.org/jira/browse/GROOVY-6504
https://issues.apache.org/jira/browse/GROOVY-6849
https://issues.apache.org/jira/browse/GROOVY-6970
https://issues.apache.org/jira/browse/GROOVY-8787
https://issues.apache.org/jira/browse/GROOVY-9069
https://issues.apache.org/jira/browse/GROOVY-9420

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -8290,8 +8290,11 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
 
     /**
      * Support the subscript operator for a Map.
-     * <pre class="groovyTestCase">def map = [a:10]
-     * assert map["a"] == 10</pre>
+     * <pre class="groovyTestCase">
+     * def map = [:]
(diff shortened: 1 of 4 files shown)


eric-milles:
With the addition of `getAt(Map,String)` a small change in `StaticTypeCheckingSupport#getDistance` can resolve the `map["$key"]` issues.  I know Cedric stated in 6668 that this should be an STC error, but if there is just one method and it accepts String, it will be chosen for a GString argument.  This changes just closes the distance between GString and String so that `m(String)` is selected over `m(Object)`.

https://issues.apache.org/jira/browse/GROOVY-6668

daniellansun:
I am keen to vote +1

PR #1767, opened 2022-08-19
GROOVY-8283: field hides getter of super class (not interface)

eric-milles:
For the dynamic side of 8283, field can be selected in `MetaClassImpl`.  If access method is declared by an interface, the interface method is the one indexed so it is considered pervasive in the type hierarchy (cannot be safely hidden).  Not sure if this kind of name hiding should be extended to inner classes; hopefully the commit is enough of a first step to determine if this protocol should be changed or left alone.

@blackdrag @paulk-asert 

TODO:
- setter method
- STC selection of field vs method (verify what `StaticTypeCheckingVisitor` does vs `MetaClassImpl`)
- static compiler classgen -- I think it currently selects the method

https://issues.apache.org/jira/browse/GROOVY-8283


eric-milles:
https://github.com/apache/groovy/pull/1767/commits/c8805170944018f17fcc5f68b760c423d43f23ce might be good to go in on its own.  There is some confusion between `sender` and `theClass` because `MetaClassImpl` uses `theClass` for `sender` in the get and set variants that accept only receiver and name.

eric-milles:
@blackdrag @paulk-asert If I refresh the branch, do you think this is viable?  The test case should illustrate the new hiding behavior.

eric-milles:
For circumstances that go through `ScriptBytecodeAdapter`, the sender class is not propagated to `MetaClassImpl`.  Access checks give different results, since the sender is re-mapped to the object.

* https://github.com/apache/groovy/blob/981928053477e7d7790e0b6000e7a4dd5b45b583/src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java#L472
* https://github.com/apache/groovy/blob/981928053477e7d7790e0b6000e7a4dd5b45b583/src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java#L504

```groovy
            package p
            class A {}
            class B {}
            class C {
                boolean setter
                protected A foo = new A()
                A getFooA() { return this.@foo }
                void setFoo(A a) { setter = true; this.@foo = a }
            }
            class D extends C {
                protected B foo = new B() // hides A#foo; should hide A#setFoo in subclasses
                B getFooB() { return this.@foo }
            }
```

```groovy
import p.*
            class E extends D {}
            def e = new E()
            e.foo = null // not the field from this perspective
            assert e.setter // fails
```

paulk-asert (approved):

PR #1770, opened 2022-08-24
GROOVY-6668, GROOVY-8212, GROOVY-9617: STC: distance of `GString` for `String` target

eric-milles:
Here is the more controversial change of https://github.com/apache/groovy/pull/1766 so it can be debated/evaluated in isolation.

A small change in `StaticTypeCheckingSupport#getDistance` can resolve the `map["$key"]` issues. I know Cedric stated in 6668 that this should be an STC error, but if there is just one method and it accepts String, it will be chosen for a GString argument. This change just closes the distance between GString and String so that `m(String)` is selected over `m(Object)`. Yes, this puts `Map#get` and `Map#getAt` out of alignment for GString arguments, which was a stated concern.  But the gain is in alignment of dynamic and static behaviors.

I checked `put` and `putAt` and I think there is consistency with `get` and `getAt`.  You can read and write via GString or String interchangeably.  I think this is the expectation of most users as described in 6668.

https://github.com/apache/groovy/pull/708
https://github.com/apache/groovy/pull/1766
https://issues.apache.org/jira/browse/GROOVY-6668
https://issues.apache.org/jira/browse/GROOVY-9529
https://issues.apache.org/jira/browse/GROOVY-9617

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -923,6 +923,9 @@ public abstract class StaticTypeCheckingSupport {
         if (receiver.isArray() && compare.isArray()) {
             return getDistance(receiver.getComponentType(), compare.getComponentType());
         }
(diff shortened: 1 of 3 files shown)


daniellansun:
Merged. Thanks.

PR #1777, opened 2022-09-02
GROOVY-10737: Groovydoc deprecation list page is always empty even if…

paulk-asert:
… @deprecated is used in groovydoc comments on some elements


erdi:
LGTM, thanks for fixing so quickly!

PR #1779, opened 2022-09-03
GROOVY-7992: DGM: `Comparator<? super T>` for comparator parameters

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-7992
https://issues.apache.org/jira/browse/GROOVY-7976


daniellansun:
Merged. Thanks.

PR #1780, opened 2022-09-03
Migrate buildSrc to composite build

Goooler:
Ref https://github.com/detekt/detekt/pull/4090

https://docs.gradle.org/7.5.1/userguide/composite_builds.html


build.gradle
```diff
@@ -35,11 +35,6 @@ buildScanRecipes {
 
 archivesBaseName = 'groovy'
 
-repositories {
-    // todo Some repos are needed only for some configs. Declare them just for the configuration once Gradle allows this.
-    maven { url 'https://repository.jboss.org/nexus/content/groups/m2-release-proxy' } // tools
-}
```
Goooler, line 41: Seems we can remove this now.

daniellansun:
@melix @paulk-asert I'm not an expert on Gradle, could you review the PR?

paulk-asert:
It does mostly look good (apart from an import or two which popped above the license header which are certainly fixable later). I just wanted to check it doesn't cause issues for the release process and we should check there are no adverse impacts when using IDEA and Eclipse. It will probably be a couple of weeks before I get to do such checks - the PR is certainly much appreciated and not forgotten.

settings.gradle
```diff
@@ -38,6 +40,14 @@ gradleEnterprise {
     }
 }
 
+dependencyResolutionManagement {
+    repositories {
+        mavenCentral()
+        maven{ url('https://repository.ow2.org/nexus/content/repositories/public/')}
```
melix, line 46: This block looks quite suspicious. Should never use `mavenLocal()` in any case, and the `ow2` repo seems redundant.

Goooler: Should we also remove `mavenLocal()` in `pluginManagement`?

Goooler: 69867ef9509fa8bb74a0ba295db62c739a198f0b

Goooler:
Just revert 969d8b24ffaa9e1ca2fc226979a7385f1057e70a now, maybe we should do it in a follow up.

build.gradle
paulk-asert: I believe so

paulk-asert:
Merged. Thanks. I spotted a few things I was going to try to fix but looks like you found one already - I'll await for the CI checks and merge that too.
If you have some time, it would be super useful to get "rat" and "dependencyUpdates" to work on the build-logic directory!

Goooler:
Thanks, will take a look.

paulk-asert:
Also, sync'ing in IntelliJ currently gives me:
image
I presume we should just point to:
https://repository.apache.org/content/repositories/snapshots/

Goooler:
Yeah, should be picked from `mavenLocal` if you already published to local.

Goooler:
Could you add `mavenLocal` back to `dependencyResolutionManagement` and try again on you device?

paulk-asert:
It is easy to add that back in but in the interests of repeatable setup for all committers, I think pointing to the snapshot site would be better.

Goooler:
Make sense.

paulk-asert:
Let's not worry about the sync issue for now, it works fine for me on Linux, just Windows has the problem and we can track that down as a separate activity.

PR #1781, opened 2022-09-04
GROOVY-10747: Fix illegal access for object clone on JDK16+

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-10747


paulk-asert:
LGTM

daniellansun:
@paulk-asert Thanks for your reviewing. I'll tweak it a bit later

src/main/java/org/codehaus/groovy/runtime/ArrayUtil.java
```diff
@@ -61,31 +55,32 @@
 public class ArrayUtil {
     private static final Object[] EMPTY = new Object[0];
 
-    public static MethodHandle getCloneArrayMethodHandle() {
```
daniellansun, line 64: `getCloneArrayMethodHandle` was added in 4.0.5-SNAPSHOT, so removing it will not break binary compatibility

Melissa1181 (approved):

PR #1783, opened 2022-09-17
GROOVY-10759: groovy-datetime should support collections of fields wh…

paulk-asert:
…en accessing a temporal accessor


paulk-asert:
Merged!

PR #1785, opened 2022-09-21
GROOVY-10682: Provide eachWithIndex for primitive arrays

Shadow-Devil:
Consider the following:
```
@groovy.transform.TypeChecked
void test(int[] ints) {
  ints.eachWithIndex { value, index ->
    println "$index: ${value.doubleValue()}"
  }
}
test(0,1,2,3,4,5)
Compiler reports "[Static type checking] - Cannot find matching method java.lang.Object#doubleValue()"
```
`eachWithIndex` is only provided for reference types, so "value" is seen as Object by the type checker.

Currently, only a draft for int arrays. Should probably also be implemented for other primitive types.
There are three different approaches to do this:
1. Copy-paste the implementation for each overload
2. Delegate to `eachWithIndex(Iterable<T> self, Closure closure)` 
3. Delegate to `eachWithIndex(Iterator<T> self, Closure closure)` 

I found option three most appealing since option 1 adds unnecessary code duplication and option 2 just calls option 3 with the iterator of `self` which would add unnecessary overhead.

https://issues.apache.org/jira/browse/GROOVY-10682

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
@@ -1,0 +1,2040 @@
+package org.codehaus.groovy.runtime;
+
+import groovy.lang.Closure;
+import groovy.lang.IntRange;
+import groovy.transform.stc.ClosureParams;
+import groovy.transform.stc.FirstParam;
+import groovy.transform.stc.FromString;
+import org.codehaus.groovy.runtime.callsite.BooleanClosureWrapper;
+import org.codehaus.groovy.util.*;
+
+import java.math.BigDecimal;
+import java.util.*;
+
+/**
+ * This class defines new groovy methods which appear on primitive arrays inside the Groovy environment.
+ * Static methods are used with the
+ * first parameter being the destination class,
+ * i.e. <code>public static int[] each(int[] self, Closure closure)</code>
+ * provides a <code>each({i -> })</code> method for <code>int[]</code>.
+ * <p>
+ * NOTE: While this class contains many 'public' static methods, it is
+ * primarily regarded as an internal class (its internal package name
+ * suggests this also). We value backwards compatibility of these
+ * methods when used within Groovy but value less backwards compatibility
+ * at the Java method call level. I.e. future versions of Groovy may
+ * remove or move a method call in this file but would normally
+ * aim to keep the method available from within Groovy.
+ */
+public class ArrayGroovyMethods {
+    /* Arrangement of each method (skip any unapplicable types for the methods):
+     * 1. boolean[]
+     * 2. byte[]
+     * 3. char[]
+     * 4. short[]
+     * 5. int[]
+     * 6. long[]
+     * 7. float[]
+     * 8. double[]
+     */
+
+    private ArrayGroovyMethods() {
+    }
+
+    //-------------------------------------------------------------------------
+   	// any
+
+    /**
+     * Iterates over the contents of a boolean Array, and checks whether a
+     * predicate is valid for at least one element.
+     * <pre class="groovyTestCase">
+     * {@code @groovy.transform.TypeChecked}
+     * void test(){
+     *      boolean[] array = [false, true, false]
+     *      assert array.any{ true == it.booleanValue()}
+     * }
+     * test()
+     * </pre>
+     *
+     * @param self      the boolean array over which we iterate
+     * @param predicate the closure predicate used for matching
+     * @return true if any iteration for the booleans matches the closure predicate
+     * @since 5.0.0
+     */
+    public static boolean any(boolean[] self, @ClosureParams(FirstParam.Component.class) Closure predicate) {
+        BooleanClosureWrapper bcw = new BooleanClosureWrapper(predicate);
+        for (boolean item : self) {
+            if (bcw.call(item)) return true;
+        }
+        return false;
+    }
+
+    /**
+     * Iterates over the contents of a byte Array, and checks whether a
+     * predicate is valid for at least one element.
+     * <pre class="groovyTestCase">
+     * {@code @groovy.transform.TypeChecked}
+     * void test(){
+     *      byte[] array = [0, 1, 2]
+     *      assert array.any{ 0 == it.byteValue()}
+     * }
+     * test()
+     * </pre>
+     *
+     * @param self      the byte array over which we iterate
+     * @param predicate the closure predicate used for matching
+     * @return true if any iteration for the bytes matches the closure predicate
+     * @since 5.0.0
+     */
+    public static boolean any(byte[] self, @ClosureParams(FirstParam.Component.class) Closure predicate) {
+        BooleanClosureWrapper bcw = new BooleanClosureWrapper(predicate);
+        for (byte item : self) {
+            if (bcw.call(item)) return true;
+        }
+        return false;
+    }
+
+    /**
+     * Iterates over the contents of a char Array, and checks whether a
+     * predicate is valid for at least one element.
+     * <pre class="groovyTestCase">
+     * {@code @groovy.transform.TypeChecked}
+     * void test(){
+     *      char[] array = ['a' as char, 'b' as char, 'c' as char]
+     *      assert array.any{ 'a' as char == it.charValue()}
+     * }
+     * test()
+     * </pre>
+     *
+     * @param self      the char array over which we iterate
+     * @param predicate the closure predicate used for matching
+     * @return true if any iteration for the chars matches the closure predicate
+     * @since 5.0.0
+     */
+    public static boolean any(char[] self, @ClosureParams(FirstParam.Component.class) Closure predicate) {
+        BooleanClosureWrapper bcw = new BooleanClosureWrapper(predicate);
+        for (char item : self) {
+            if (bcw.call(item)) return true;
+        }
+        return false;
+    }
+
+    /**
+     * Iterates over the contents of a short Array, and checks whether a
+     * predicate is valid for at least one element.
+     * <pre class="groovyTestCase">
+     * {@code @groovy.transform.TypeChecked}
+     * void test(){
+     *      short[] array = [0, 1, 2]
+     *      assert array.any{ 0 == it.shortValue()}
+     * }
+     * test()
+     * </pre>
+     *
+     * @param self      the char array over which we iterate
+     * @param predicate the closure predicate used for matching
+     * @return true if any iteration for the shorts matches the closure predicate
+     * @since 5.0.0
+     */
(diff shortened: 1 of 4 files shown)


src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -2273,6 +2273,26 @@ public static <T> T[] eachWithIndex(T[] self, @ClosureParams(value=FromString.cl
         return self;
     }
 
+    /**
+     * Iterates through an int[],
+     * passing each int and the element's index (a counter starting at
+     * zero) to the given closure.
+     * <pre class="groovyTestCase">
+     * int result = 0
+     * [0, 1, 2].eachWithIndex{ x, index {@code ->} result += x * index }
```
eric-milles, line 2282: Please ensure that the object expression is of type `int[]` to test the annotated method.  This is `List<Integer>` and so it tests another `eachWithIndex` veriant.  You can debug `MainJavadocAssertionTest` to see if your addition is hit by any test cases.

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -2273,6 +2273,26 @@ public static <T> T[] eachWithIndex(T[] self, @ClosureParams(value=FromString.cl
...
+     * assert result == 5
+     * </pre>
+     *
+     * @param self    an int array
+     * @param closure a Closure to operate on each item
+     * @return the self Object
+     * @since 1.0
```
eric-milles, line 2289: Set the since value to 5.0.0

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -2273,6 +2273,26 @@ public static <T> T[] eachWithIndex(T[] self, @ClosureParams(value=FromString.cl
...
+     * zero) to the given closure.
+     * <pre class="groovyTestCase">
+     * int result = 0
+     * [0, 1, 2].eachWithIndex{ x, index {@code ->} result += x * index }
+     * assert result == 5
+     * </pre>
+     *
+     * @param self    an int array
+     * @param closure a Closure to operate on each item
+     * @return the self Object
```
eric-milles, line 2288: Should probably say array instead of Object

src/spec/test/typing/TypeCheckingTest.groovy
```diff
@@ -44,6 +44,24 @@ class TypeCheckingTest extends StaticTypeCheckingTestCase {
         '''
     }
 
+    void testTypeCheckedEachWithIndex(){
+        //TODO: tags?
+        assertScript '''
+            @groovy.transform.TypeChecked
```
eric-milles, line 50: No need to add this annotation within a `StaticTypeCheckingTestCase`.  The test case takes care of this for you and the subclass also applies static compilation.

eric-milles:
This is a good start.  I added some notes on the commit.  You should compare to the javadoc for the `T[]` variant and try to match it, except here you know the array element type so you may be able to make some small improvements.

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -2273,6 +2273,26 @@ public static <T> T[] eachWithIndex(T[] self, @ClosureParams(value=FromString.cl
...
+     * @since 1.0
+     */
+    public static int[] eachWithIndex(int[] self, @ClosureParams(value=FromString.class, options="int,Integer") Closure closure) {
```
eric-milles, line 2291: I think "int" should be "Integer".  The closure params are passed as objects, so there is no optimization like `Consumer` versus `IntConsumer`.

eric-milles:
If we're adding `eachWithIndex` it appears we also need `each` to satisfy STC.  Can you add an `each(int[])` as well?

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
Shadow-Devil: I'm not sure how the debugging works, since I can't place the breakpoint inside the javadocs and I don't think I should step through all of the test cases.
I changed the test case to:
```
int[] nums = [5, 4, 3]
String result = ''
nums.eachWithIndex{ num, index -> result += "$index:${num + 1}" }
assert result == '0:61:52:4'
```

src/spec/test/typing/TypeCheckingTest.groovy
```diff
@@ -44,6 +44,23 @@ class TypeCheckingTest extends StaticTypeCheckingTestCase {
         '''
     }
 
+    void testTypeCheckedEachWithIndex(){
+        //TODO: tags?
```
paulk-asert, line 48: Tests like this either belong here if they appear as part of the documentation in which case we'd also add the documentation and tags or we'd put them somewhere like:
https://github.com/apache/groovy/blob/master/src/test/groovy/transform/stc/DefaultGroovyMethodsSTCTest.groovy#L28-L40

paulk-asert: The current best fit for such doco would currently be core-syntax.adoc or core-semantics.adoc but I'll make a separate "working with arrays" subsection in the gdk doco since it would really belong there.

paulk-asert: I added the doco mentioned in previous comment, so if you want to add to the doco, you would update:
https://github.com/apache/groovy/blob/master/src/spec/test/gdk/WorkingWithArraysTest.groovy (probably `testArrayIteration`)
https://github.com/apache/groovy/blob/master/src/spec/doc/_working-with-arrays.adoc (if needed)
This wouldn't cover static compilation unless you added it explicitly, so moving to the aforementioned `DefaultGroovyMethodsSTCTest` might be just as good an option.

eric-milles:
> If we're adding eachWithIndex it appears we also need each to satisfy STC. Can you add an each(int[]) as well?

The reason I did not pick this ticket up after creating it is there was no reasonable stopping point.  If adding `eachWithIndex`, what about `each`, `reverseEach`, `every` and all the other `T[]` extensions.  I wonder if there is some way to write just one implementation and have all the primitive versions be generated automatically.  @paulk-asert Do you know of such a thing?  Or should we consider the possibility of allowing the `T[]` or `Object[]` apply to primitive arrays as well through Groovy "auto-unboxing"?

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
eric-milles: To debug a junit test class, you can start gradle like this:
> gradlew :test --tests MainJavadocAssertionTest --no-daemon --debug-jvm

Gradle will start up and build enough to start the requested test(s) and wait for a connection on port 5005.  You can connect and debug with your favorite debugger / development environment.

paulk-asert:
We have used code generators (hand-hacked throwaway Groovy scripts) from time to time to generate repeated patterns.

We could enhance both the dynamic and static behaviours to do some kind of array-level "auto-unboxing" for T[] and the like but we have opted not to in the past for better Java integration and (at the time) efficiency with the hand-coded primitive variants likely to be more efficient.

We also have `IntArrayIterable` (and Long/Double variants). So we could write:

```
public static int[] eachWithIndex(int[] self, @ClosureParams(value=FromString.class, options="Integer,Integer") Closure closure) {
    eachWithIndex(new IntArrayIterable(self), closure);
    return self;
}

```
This saves work when writing the new variants but is probably slightly less efficient that the current code in this PR.

src/spec/test/typing/TypeCheckingTest.groovy
Shadow-Devil: Thank you very much for your help. I think I will add a simple test case here https://github.com/apache/groovy/blob/master/src/spec/test/gdk/WorkingWithArraysTest.groovy
that also works without my extension 
and one with stc into here https://github.com/apache/groovy/blob/master/src/test/groovy/transform/stc/DefaultGroovyMethodsSTCTest.groovy#L28-L40 that only works if we add the extension. Otherwise, I think it would be quite weird for the reader of the doco to see a method annotated with `@groovy.transform.TypeChecked`. 
One small note: I thought the doco file should be about arrays, but in the second paragraph you are talking about "Iterating on a list". Is this a maybe typo?

src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
```diff
@@ -0,0 +1,85 @@
+package org.codehaus.groovy.runtime;
+
+import groovy.lang.Closure;
+import groovy.transform.stc.ClosureParams;
+import groovy.transform.stc.FirstParam;
+import groovy.transform.stc.FromString;
+
+/**
+ * This class defines new groovy methods which appear on prrimitive arrays inside the Groovy environment.
```
paulk-asert, line 9: type in 'prrimitive'

src/test/groovy/transform/stc/DefaultGroovyMethodsSTCTest.groovy
```diff
@@ -37,6 +37,20 @@ class DefaultGroovyMethodsSTCTest extends StaticTypeCheckingTestCase {
                 println it // DGM#println(Object,Object)
             }
         '''
+
+        assertScript '''
+            int[] nums = [0, 1]
+            nums.each { it -> // ArrayGroovyMethods#each(int[], Closure)
+                println it.toDouble() // DGM#println(Object,Object)
```
paulk-asert, line 44: Though we have many legacy tests which use println, we tend to avoid adding additional such tests when we can since they slow down the build and don't actually get automatically tested (even though the main objective here is for compilation to pass).

Shadow-Devil: I think I would stop writing tests in this test class and only use the tests inside the Javadoc since otherwise the number of tests would explode for ~60 methods with each 8 primitive types (roundabout 480 tests).

paulk-asert:
@sonatype-lift ignore

eric-milles:
I'm trying to fight through the sonatype comments.  But it is looking ready for review to me.

Shadow-Devil:
Could someone please approve the workflow?

paulk-asert:
LGTM

paulk-asert:
Merged, thanks! I might make some minor formatting changes post merge. I am also thinking have the TypeChecked annotation in the groovyTestCase tests in javadoc probably adds noise that doesn't need to be there. I'll ponder whether we should tweak that somehow.

PR #1786, opened 2022-09-22
Improve generic-ClassNode compiler error message

chrylis:
Explain how (rather than what) should be done to refer to classes with generic types in AST transforms.

GROOVY-10758


chrylis:
My run of `./gradlew test` had one failing test, but it was in `groovy-ant` and related to a missing XML file, so I don't think that's related to this code change (which is only a string, the contents of which don't appear to be asserted specifically).

paulk-asert:
The other scenario which I think applies sometimes is if you were using ClassHelper.make, you can swap to using ClassHelper.makeWithoutCaching.

chrylis:
@paulk-asert That was what I'd originally tried, but (perhaps due to some other interactions; see GROOVY-10763) it does not resolve the error. Per @eric-milles, the two approaches I outlined are the "official" ones.

paulk-asert:
Merged, thanks!

PR #1787, opened 2022-09-28
Migrate JavaPluginConvention to JavaPluginExtension

Goooler:
`JavaPluginConvention` has been deprecated, and will be removed in Gradle 8, we should migrate to `JavaPluginExtension`. 

https://docs.gradle.org/current/userguide/upgrading_version_7.html#java_convention_deprecation


daniellansun:
Merged. Thanks!

PR #1788, opened 2022-10-02
GROOVY-10783: propertyMissing of category does not work when indy ena…

daniellansun:
…bled

https://issues.apache.org/jira/browse/GROOVY-10783


daniellansun:
Not sure. As the PR title described, it just tries to fix `propertyMissing` when indy is enabled.

PR #1799, opened 2022-10-18
Fix minor typos in documentations in spec/doc

marcwrobel:
A few trailing whitespaces were removed in the process.


src/spec/doc/_working-with-io.adoc
```diff
@@ -281,7 +281,7 @@ There are also variations of `consumeProcessOutput` that make use of `StringBuff
 `OutputStream` etc... For a complete list, please read the
 http://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/Process.html[GDK API for java.lang.Process]
 
-In addition, these is a `pipeTo` command (mapped to `|`
+In addition, these are `pipeTo` commands (mapped to `|`
```
paulk-asert, line 284: I think this should be "there is a" or "there's a"

marcwrobel: You are right, fixed.

paulk-asert:
Merged, thanks!

PR #1802, opened 2022-10-18
Fix minor typos in tests documentation

marcwrobel:

- meta data -> metadata (incorrect, see https://www.merriam-webster.com/dictionary/metadata),

Some trailing whitespaces were removed and a few typos in failure messages were also fixed.


paulk-asert:
Merged, thanks!

PR #1805, opened 2022-10-18
Polish groovy.lang package documentation

marcwrobel:
Fix typos and improve documentation (javadoc, comments) in the groovy.lang package.

- return back -> returns (when possible because it is easier to read),
- multi-threaded -> multithreaded (more common, see https://en.wikipedia.org/wiki/Multithreading)
- whether or not -> whether (when possible because it is easier to read),

Trailing whitespaces were also removed in the process.


paulk-asert:
Merged, thanks!

PR #1807, opened 2022-10-18
Polish org.codehaus.groovy package documentation

marcwrobel:
Fix typos and improve documentation (javadoc, comments) in the org.codehaus.groovy package. A typo in the Grape resolve command were also fixed.

- all of the -> all the (when possible because it is easier to read),
- whether or not -> whether (when possible because it is easier to read),
- sub-type -> subtype (more common, see https://docs.oracle.com/javase/tutorial/java/generics/subtyping.html),
- sub-class -> subclass (more common, see https://docs.oracle.com/javase/tutorial/java/IandI/subclasses.html),

Trailing whitespaces were also removed in the process.


paulk-asert:
Merged, thanks!

PR #1809, opened 2022-10-18
Polish groovy package documentation

marcwrobel:
Fix typos and improve documentation (javadoc, comments) in the groovy package.

- sub-class -> subclass (more common, see https://docs.oracle.com/javase/tutorial/java/IandI/subclasses.html),


paulk-asert:
Merged, thanks!

PR #1813, opened 2022-10-24
Cleanup build-logic

Goooler:
Some changes addressed from #1780.

- Replace `with` with `using` 
https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/DependencySubstitutions.Substitution.html#with-org.gradle.api.artifacts.component.ComponentSelector-
- Use HTTPS links
- Optimize imports
- Remove deprecated `forUseAtConfigurationTime` call https://docs.gradle.org/current/userguide/upgrading_version_7.html#for_use_at_configuration_time_deprecation


build-logic/src/main/groovy/org/apache/groovy/gradle/JarJarTask.groovy
```diff
@@ -112,7 +114,7 @@ class JarJarTask extends DefaultTask {
         def manifestFile = new File(temporaryDir, 'MANIFEST.MF')
         // First step is to create a repackaged jar
         outputFile.parentFile.mkdirs()
-        def tstamp = Date.parse('yyyy-MM-dd HH:mm', '1980-02-01 00:00').time.toString()
+        def tstamp = (new SimpleDateFormat('yyyy-MM-dd HH:mm')).parse('1980-02-01 00:00').time.toString()
```
paulk-asert, line 117: The original is more idiomatic Groovy

Goooler: Cause `Date.parse` is deprecated.

build-logic/src/main/groovy/org/apache/groovy/gradle/ReleaseInfoGenerator.groovy
```diff
@@ -69,7 +69,7 @@ class ReleaseInfoGenerator extends DefaultTask {
 #  "License"); you may not use this file except in compliance
 #  with the License.  You may obtain a copy of the License at
 #
-#    http://www.apache.org/licenses/LICENSE-2.0
+#    https://www.apache.org/licenses/LICENSE-2.0
```
paulk-asert, line 72: In most places we have left "http" inside the license text. That was conservatively because that is what is actually in the license file we refer to and we didn't want to open up an even insignificantly small debatable legal point, but also waiting for tooling to catch up. I believe lots of other projects now use https in their headers and the tooling has for the most part caught up.

build-logic/src/main/groovy/org.apache.groovy-published-library.gradle
```diff
@@ -765,7 +767,7 @@ publishing {
                 licenses {
                     license {
                         name = 'The Apache Software License, Version 2.0'
-                        url = 'http://www.apache.org/licenses/LICENSE-2.0.txt'
+                        url = 'https://www.apache.org/licenses/LICENSE-2.0.txt'
```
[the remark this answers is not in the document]
paulk-asert: This should be fine. The actual license file is at (logically at least) both http and https.

build-logic/src/main/groovy/org/apache/groovy/gradle/ReleaseInfoGenerator.groovy
Goooler: So if I should revert this line?

build-logic/src/main/groovy/org/apache/groovy/gradle/JarJarTask.groovy
Goooler: My bad, it might be a bug of IDEA, actually no deprecation on `parse`, will revert.

build-logic/src/main/groovy/org/apache/groovy/gradle/ReleaseInfoGenerator.groovy
Goooler: Reverted.

paulk-asert: Yeah, we'll no doubt probably eventually change this but we'll probably do all 5k places in one commit.

paulk-asert:
Merged, thanks!

PR #1821, opened 2022-11-08
Replace gradlew command with relative path

tisonkun:
Current working directory is not always in PATH


paulk-asert:
We already have the relative path for unix systems as a comment the first time the command appears.

tisonkun:
@paulk-asert Thanks! I read it.

Closed as invalid.

PR #1822, opened 2022-11-08
GROOVY-8668: add property groovysh.disableDocCommand

tisonkun:
The AWT window is launched by DocCommand. I want to exclude this functionality to avoid opening a new window so I suggest to add an option to disable it.


paulk-asert:
Merged, thanks!

PR #1824, opened 2022-11-10
GROOVY-10813: Fix NullPointerException for lambdas with raw types

Shadow-Devil:
I've added a test case and a fix for the bug reported in GROOVY-10813. 
I'm not sure if there are any other edge cases that will now break because of this change, but one [test](https://github.com/apache/groovy/blob/9210cfab6aac681a865c57dd97e96aac73f38fd4/src/test/groovy/transform/stc/GenericsSTCTest.groovy#L740), that failed with another solution that I tried, passes with this solution.

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
+++ b/src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
@@ -855,10 +855,21 @@ public class GenericsUtils {
             GenericsType[] targetGenericsTypes = parameterizedType.redirect().getGenericsTypes();
             if (targetGenericsTypes != null) {
                 GenericsType[] sourceGenericsTypes = parameterizedType.getGenericsTypes();
-                if (sourceGenericsTypes == null) sourceGenericsTypes = EMPTY_GENERICS_ARRAY;
                 map = new LinkedHashMap<>();
-                for (int i = 0, m = sourceGenericsTypes.length, n = targetGenericsTypes.length; i < n; i += 1) {
-                    map.put(targetGenericsTypes[i], i < m ? sourceGenericsTypes[i] : targetGenericsTypes[i]);
+                if (sourceGenericsTypes == null) {
+                    //We have no sourceGenerics, so we need to return the erasures for the raw types.
+                    // If the generic has upper bounds use the first one, else use Object (Groovy-10813)
+                    for (GenericsType targetGenericsType : targetGenericsTypes) {
+                        map.put(targetGenericsType,
+                                targetGenericsType.getUpperBounds() != null ?
+                                    targetGenericsType.getUpperBounds()[0].asGenericsType() :
+                                    ClassHelper.OBJECT_TYPE.asGenericsType()
+                        );
+                    }
+                } else {
+                    for (int i = 0, m = sourceGenericsTypes.length, n = targetGenericsTypes.length; i < n; i += 1) {
+                        map.put(targetGenericsTypes[i], i < m ? sourceGenericsTypes[i] : targetGenericsTypes[i]);
+                    }
                 }
             }
         }
--- a/src/test/groovy/transform/stc/ClosureParamTypeInferenceSTCTest.groovy
+++ b/src/test/groovy/transform/stc/ClosureParamTypeInferenceSTCTest.groovy
@@ -1709,4 +1709,24 @@ class ClosureParamTypeInferenceSTCTest extends StaticTypeCheckingTestCase {
             assert i == 1
         '''
     }
+
+    void testGroovy10813() {
+        assertScript '''
+            @groovy.transform.CompileStatic
+            void testClosure(){
+                java.util.function.Consumer c = x -> print(x)
+                c.accept('hello world!')
+            }
+            testClosure()
+        '''
+        assertScript '''
+            interface A<T extends String>{void accept(T input);}
+            @groovy.transform.CompileStatic
+            void testClosure(){
+                A c = x -> print(x)
+                c.accept('hello world!')
+            }
+            testClosure()
+        '''
+    }
 }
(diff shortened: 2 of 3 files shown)


src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
```diff
@@ -855,10 +855,16 @@ private static Map<GenericsType, GenericsType> doMakeDeclaringAndActualGenericsT
...
             if (targetGenericsTypes != null) {
                 GenericsType[] sourceGenericsTypes = parameterizedType.getGenericsTypes();
-                if (sourceGenericsTypes == null) sourceGenericsTypes = EMPTY_GENERICS_ARRAY;
                 map = new LinkedHashMap<>();
-                for (int i = 0, m = sourceGenericsTypes.length, n = targetGenericsTypes.length; i < n; i += 1) {
-                    map.put(targetGenericsTypes[i], i < m ? sourceGenericsTypes[i] : targetGenericsTypes[i]);
+                if (sourceGenericsTypes == null) {
+                    //We have no sourceGenerics, so we need to return the erasures for the raw types (Groovy-10813)
+                    for (GenericsType targetGenericsType : targetGenericsTypes) {
+                        map.put(targetGenericsType, ClassHelper.OBJECT_TYPE.asGenericsType());
```
eric-milles, line 862: The erasure type is not always Object.  Can you create an example where it is another type?

src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
```diff
@@ -950,7 +956,7 @@ public static boolean hasUnresolvedGenerics(final ClassNode type) {
     public static Tuple2<ClassNode[], ClassNode> parameterizeSAM(final ClassNode samType) {
         MethodNode abstractMethod = ClassHelper.findSAM(samType);
 
-        Map<GenericsType, GenericsType> generics = makeDeclaringAndActualGenericsTypeMapOfExactType(abstractMethod.getDeclaringClass(), samType);
+        Map<GenericsType, GenericsType> generics = makeDeclaringAndActualGenericsTypeMap(abstractMethod.getDeclaringClass(), samType);
```
eric-milles, line 959: We have been moving away from `makeDeclaringAndActualGenericsTypeMap` towards `makeDeclaringAndActualGenericsTypeMapOfExactType`.  Can you explain this change?

eric-milles: Is this the stuff that sits under parametwrizeSAM?  That was the key method for fixing lambda param types.

eric-milles (changes requested):

src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
FelixSchnabel: I'm sorry, this was a mistake. 
I will revert this change.

Shadow-Devil: I'm sorry, this change was not intended to be commited, it was just for testing. 
I will revert it.

Shadow-Devil: > Is this the stuff that sits under parametwrizeSAM? That was the key method for fixing lambda param types.

parameterizeSAM uses this makeDeclaringAndActualGenericsTypeMapOfExactType method to map declaring generics to actual generics. I thought that I needed to adjust this mapping so for unresolved generics it will output Object, but I'm still unsure if that's the right solution (as you mentioned in the other comment).

src/main/java/org/codehaus/groovy/ast/tools/GenericsUtils.java
Shadow-Devil: I will try to find an example.

Shadow-Devil: I've played around with generics in java, but couldn't find an example where the erasure isn't of type Object.
I've also tried something like this:
```java
import java.util.function.*
interface T<E> extends Function<String, E>{}
class TT implements T{
  @Override
  Object apply(Object o) {
    return null
  }
}
```
and it still loses all type information. Maybe I'm just not advanced enough into generics and how they work. Could you perhaps help me here or give me a resource where to learn more about this? I've found some information about the topic [here](https://docs.oracle.com/javase/tutorial/java/generics/rawTypes.html) and [here](https://docs.oracle.com/javase/tutorial/java/generics/erasure.html)

Shadow-Devil: Another example where the type parameter has an upper bound: 
´´´java
interface A<T extends String> extends Function<T, T> {}
class B implements A{
    @Override
    Object apply(Object input) {
        return null
    }
}
´´´

Shadow-Devil: Uh, now I got an example, that has not Object as its type:
```java
interface A<T extends String> {
    T apply(T input);
}
class B implements A{
    @Override
    String apply(String input) {
        return null
    }
}
```

Shadow-Devil: I've added some tests that have an upper bound. I was a bit suprised that `T extends String & Object` just selects the first one when you use raw types, so the result is different from `T extends Object & String`

eric-milles:
I relocated the lambda tests and reworked the generics tests to ensure the placeholders are not returned.

https://github.com/apache/groovy/commit/ef9b58463e83f842b56f24dff0ac4df0e02f112f

PR #1828, opened 2022-11-21
refactored TimeDuration, added AgoDate, added TimeAgoFromData;

h1alexbel:


paulk-asert:
Hi Aliaksei Bialiauski. While there is nothing wrong with your refactorings, we'd typically want some discussion in a forum or mailing list so we could discuss merits of this versus other possible changes we could make. Please reopen and start a discussion if you feel there is a real improvement in functoinality that Groovy needs.

PR #1836, opened 2022-12-21
GROOVY-10881: Make FastStringUtils work in an OSGi environment

royteeuwen:


eric-milles:
Will this cause many passes of class scanning and `DefaultFastStringService` instances?  How are `loadService()` and `getService()` used?

royteeuwen:
> Will this cause many passes of class scanning and `DefaultFastStringService` instances? How are `loadService()` and `getService()` used?

INSTANCE s a static field, meaning that at the class loading of FastStringUtils it will be initialized by calling loadService(). This will only be done once

getService() will just return that same static instance all the time

paulk-asert:
@royteeuwen Should we close this now given the alternative fix?

royteeuwen:
> @royteeuwen Should we close this now given the alternative fix?

Done!

PR #1839, opened 2022-12-31
GROOVY-10891: We should have variants of findResult/s with no Closure…

paulk-asert:
… which use Closure.IDENTITY in that case

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -4715,6 +4715,28 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     /**
+     * Treats the object as iterable, iterating through the values it represents and returns the first non-null value, otherwise returns null.
+     * <p>
+     * <pre class="groovyTestCase">
+     * class Foo {
+     *     List items
+     *     Iterator iterator() {
+     *         items.iterator()
+     *     }
+     * }
+     * assert new Foo(items: [null, 2, 4]).findResult() == 2
+     * assert new Foo(items: [null, null]).findResult() == null
+     * </pre>
+     *
+     * @param self      an Object with an iterator returning its values
+     * @return the first non-null result of the closure
+     * @since 4.0.8
+     */
+    public static Object findResult(Object self) {
+        return findResult(self, Closure.IDENTITY);
+    }
+
+    /**
      * Treats the object as iterable, iterating through the values it represents and returns the first non-null result obtained from calling the closure, otherwise returns the defaultResult.
      * <p>
      * <pre class="groovyTestCase">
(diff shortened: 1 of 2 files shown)


eric-milles:
-1 for me unless some description is provided.  What is the specific purpose of this and what does to replace that is currently available?  More discussion from Feb 2019: https://www.mail-archive.com/dev@groovy.apache.org/msg04292.html

paulk-asert:
It is filling a hole in existing functionality. Take `find` and `findAll` for instance, they have the no Closure variants:
```
var list = [null, '', 'A', null, 'B']
assert list.findAll() == ['A', 'B']
assert list.find() == 'A'
```
These return all (or the first) element that satisfies Groovy truth. It is common in Groovy to provide these shortcuts even though `findAll { it }` and `find { it }` would do in the above examples. The same applies for `groupBy`, `collect`, `collectEntries`, and others. We have never gotten around to providing the findResult/s variants which is what this PR does. It improves Groovy's consistency in the API.

eric-milles:
I don't think it is hole in existing functionality though.  I'm missing the clear and obvious part of all this.  Leaving the default value stuff aside,
```groovy
def list = [null, '', 'A', null, 'B', false, []]
assert list.find() == 'A'
assert list.findAll() == ['A', 'B']
// Is this what you are suggesting?
assert list.findResult() == 'A'
assert list.findResults() == ['A', 'B']
```
Why do I need the no-arg variants of `findResult` and `findResults` then?  If not transforming, I justuse `find` or `findAll`.

eric-milles:
And since `find` and `findAll` are based on groovy truth, to get a default value, you can just pepper in elvis:
```groovy
def list = [null,false,'',[]]
def result = list.find() ?: 'defaultValue'
assert result == 'defaultValue'
```

eric-milles:
Okay, just picked up on the difference.  `find` and `findAll` apply groovy truth and `findResult` and `findResults` check `null` only.  Is that too subtle a difference?

If this is what I am looking for:
```groovy
def list = [null, '', 'A', null, 'B', false]
assert list.find() == 'A'
assert list.findAll() == ['A', 'B']
assert list.xxx == ''
assert list.yyy == ['','A', 'B',false]
```

For myself, I'd look to sub "xxx" with `find{it != null}` or `find(Objects::nonNull)` and "yyy" with `findAll{it != null}` or `findAll(Objects::nonNull)`.  Nice and clear IMO.

I'm just really stuck on "result" in `findResult` meaning something other than the input value.

paulk-asert:
We picked the name `findResult` many years ago. Naming is hard. I wasn't going to open the naming can of worms myself. Just plugging that hole in the existing named functions.

But on naming, I think the idea was that if you got a result (being "non-null") that was a match whereas what was regarded as more common case of wanting Groovy truth was find/findAll. If we hadn't already locked in the find/findAll naming many years earlier again, we might not have chosen findResult/findResults for the non-null case.

paulk-asert:
Is the "further clarification request" on your -1 veto satisfied? Perhaps you can formally remove it?

eric-milles:
Not sure why the new methods are listed as "since 4.0.8".  Are these such a must have that they should be available immediately.  If kept for 5+, the array-based overloads can start off in `ArrayGroovyMethods` with no deprecated artifact.

paulk-asert:
My thinking was since this is missing functionality it makes sense for 4.0.8. We have the find/findAll variants (Groovy truth) in 4.0.7 why not the missing findResult/s variants (null-based) also? We just never got around to it before.
The `ArrayGroovyMethods` are just for primitive arrays. There is no need for primitive variations for these since they couldn't have null and moreover the Object variant picks up those anyway.

PR #1843, opened 2023-01-13
Add `asBoolean` extension methods to `File` and `Path`

Matyrobbrt:
This PR adds `asBoolean` extension methods to `Path` and `File` in order to allow Groovy Truth usage.  
I have also added an `exists(LinkOption...)` extension method in `NioExtensions` but I can remove it if it is too out-of-scope for this PR.


paulk-asert:
This PR doesn't look bad, but just by way of feedback, the idea has been considered in the past and, while many thought such functionality might be useful, there was also debate about possible other implementations, e.g. as one possibility, whether Groovy truth should return true or false for an existing but empty file. So, we'll probably let it sit for a little bit longer to see if anyone comments. I think this was thought of as best among multiple possibilities but at the time since there was not unanimous consensus, we decided to not go forward at the time.

Goooler (approved):

paulk-asert:
Merged. Thanks!

PR #1844, opened 2023-01-13
Replace usages of `path.toFile().exists()` in `NioExtensions` with `Files#exists`

Matyrobbrt:
This PR replaces the different usages of `Path#toFile` in `NioExtensions` with the equivalent methods in the NIO API (`Files`). This change is done because there is no guarantee a `Path` can always be converted to a `File`, even if you can open a stream to it:
> Throws: UnsupportedOperationException – if this Path is not associated with the default provider


daniellansun:
Merged. Thanks!

paulk-asert:
I don't know whether the original code was chosen to avoid the exception being thrown when using the NIO method (for the not exists case) but it shouldn't be too much of a drama.

PR #1849, opened 2023-01-26
GROOVY-7319: Add primitive array min() and max()

paulk-asert:
Inspiration from Yu Kobayashi (https://github.com/groovy/groovy-core/pull/616 from 2015) Currently IntBinaryOperator and IntUnaryOperator are used instead of introducing the proposed IntComparator class (it's not a perfect abstraction but saves us creating additional classes.). Feedback welcome.


paulk-asert:
I'll tweak and reissue.

PR #1859, opened 2023-02-18
GROOVY-10942 Add `extension` & `baseName` extension methods

Goooler:
Ref:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/java.nio.file.-path/extension.html
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/java.nio.file.-path/name-without-extension.html

### DIFF
--- a/subprojects/groovy-nio/src/main/java/org/apache/groovy/nio/extensions/NioExtensions.java
+++ b/subprojects/groovy-nio/src/main/java/org/apache/groovy/nio/extensions/NioExtensions.java
@@ -102,6 +102,36 @@ public class NioExtensions extends DefaultGroovyMethodsSupport {
     }
 
     /**
+     * Provide the extension of this path.
+     *
+     * @param self a path object
+     * @return the extension of this path (not including the dot), or an empty string if it doesn't have one.
+     * @since 5.0.0
+     */
+    public static String getExtension(Path self) {
+        final Path fileName = self.getFileName();
+        final String name = fileName == null ? "" : fileName.toString();
+        final int index = name.lastIndexOf('.');
+        if (index == -1) return "";
+        return name.substring(index + 1);
+    }
+
+    /**
+     * Provide the path's name without an extension.
+     *
+     * @param self a file object
+     * @return the name of this file or directory without an extension, or an empty string if this path has zero path elements.
+     * @since 5.0.0
+     */
+    public static String getBaseName(Path self) {
+        final Path fileName = self.getFileName();
+        final String name = fileName == null ? "" : fileName.toString();
+        final int index = name.lastIndexOf('.');
+        if (index == -1) return name;
+        return name.substring(0, index);
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


paulk-asert:
I like the idea behind these extensions. Can you create a corresponding Jira issue?

Also, some minor points:

- I think using `getExtension` rather than `extension` would be closer to existing methods. WDYT?
- Also, `extension` doesn't currently return empty string when there is no extension. This is what the javadocs state but not what the implementation does. The javadoc matches how Kotlin and commons-io FilenameUtils behave.
- An alternative to `nameWithoutExtension` is `baseName` (what commons-io uses) or just `basename` (again, I'd go with the getter variant). I guess there is some merit in following Kotlin but `basename` would certainly be shorter. WDYT?

Naming is always tricky, we can always ask on the mailing list if we want to confirm preferences.

Goooler:
GROOVY-10942

1. Agreed.
2. I misjudged `Path.getFileName()` to return a non-null String, will fix it later.
3. Will add a `getBaseName()`.

subprojects/groovy-nio/src/main/java/org/apache/groovy/nio/extensions/NioExtensions.java
```diff
@@ -101,6 +101,36 @@ public static long size(Path self) throws IOException {
...
+    /**
+     * Provide the extension of this path.
+     *
+     * @param self a path object
+     * @return the extension of this path (not including the dot), or an empty string if it doesn't have one.
+     * @since 5.0.0
+     */
+    public static String getExtension(Path self) {
+        final Path fileName = self.getFileName();
+        final String name = fileName == null ? "" : fileName.toString();
```
Goooler, line 113: If we can resue this part, add a `Path.name()` like [`Path.name` in Kotlin](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/java.nio.file.-path/name.html)?

paulk-asert:
Can you elaborate what you mean by "too implicit"? (Just checking I understood) You mean "too obscure" whereas the longer one is "more self-describing"? I guess the shorter one relies on some familiarity with *nix commands which might not always be the case.

subprojects/groovy-nio/src/main/java/org/apache/groovy/nio/extensions/NioExtensions.java
paulk-asert: You want `Path.getName()`, with shorthand `Path.name` as a shorthand for `Path.getFileName()`/`Path.fileName`?
That could work but there is the existing `getName(int index)` method which might confuse some folks.

Goooler: Yeah, that's what I'm afraid.

Goooler:
Renamed to `getBaseName`.

PR #1867, opened 2023-02-28
GROOVY-10943: Consider additional use of _ as a placeholder

paulk-asert:


paulk-asert:
This covers just multi-assignment so far. Also, it leaves the underscore as a dynamic variable for dynamic Groovy. While not recommended, if you do happen to use the underscore (in a multi-assignment for more than one underscore) later, it will be the last set value. The variable is undeclared from a static type checking point of view.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -3906,13 +3906,17 @@ private void validateParameterList(final List<Parameter> parameterList) {
         for (int n = parameterList.size(), i = n - 1; i >= 0; i -= 1) {
             Parameter parameter = parameterList.get(i);
 
+            String name = parameter.getName();
+            if (name.equals("_")) {
```
Goooler, line 3910:
```suggestion
            if ("_".equals(name)) {
```

src/main/java/org/codehaus/groovy/control/PlaceholderVisitor.java
```diff
@@ -0,0 +1,68 @@
...
+    @Override
+    public void visitDeclarationExpression(DeclarationExpression expression) {
+        if (expression.isMultipleAssignmentDeclaration() && expression.getTupleExpression().getExpressions().size() > 1) {
+            long underscoreCount = StreamSupport.stream(expression.getTupleExpression().spliterator(), false)
+                .map(e -> ((VariableExpression) e).getName())
+                .filter(s -> s.equals("_"))
+                .count();
+            if (underscoreCount > 1) {
+                expression.getTupleExpression().getExpressions().forEach(e -> {
+                    if (((VariableExpression) e).getName().equals("_")) {
```
Goooler, line 46:
```suggestion
                    String name = ((VariableExpression) e).getName();
                    if ("_".equals(name)) {
```

paulk-asert:
Support for lambda and closure parameters also added.

orubel:
The underscore (_) is used in groovy.cli.commons.CommandLineInterface when there is no shortDesc . I don't think this would be an issue

ex.

```
			cliBuilder.with {
				// HELP OPT
				h(longOpt: 'help', 'Print this help text and exit (usage: -h, --help)')
				_(longOpt:'compress','flag used to signify compression of classes (usage: --compress)')
				_(longOpt:'decompress','flag used to signify decompression of classes (usage: --decompress)')
			}
```

paulk-asert:
> The underscore (_) is used in groovy.cli.commons.CommandLineInterface when there is no shortDesc . I THINK this is the same usage(???)
> 
> ex.
> 
> ```
> 			cliBuilder.with {
> 				// HELP OPT
> 				h(longOpt: 'help', 'Print this help text and exit (usage: -h, --help)')
> 				_(longOpt:'compress','flag used to signify compression of classes (usage: --compress)')
> 				_(longOpt:'decompress','flag used to signify decompression of classes (usage: --decompress)')
> 			}
> ```
> 
> ... but this may be worth checking.

No change to that scenario. There are two separate method call expressions with name `_` in your example.

orubel:
@paulk-asert 

>  There are two separate method call expressions with name `_` in your example.

AFAIK this is following the example (works great anyway) to set it so that only 'longOpt' is available; basically sets a 'null' for a shortOpt I think(???)

PR #1885, opened 2023-04-27
fix for NPE

weih-kahoot:
java.lang.NullPointerException
	at org.codehaus.groovy.util.ListHashMap.get(ListHashMap.java:120)
	at org.codehaus.groovy.ast.NodeMetaDataHandler.getNodeMetaData(NodeMetaDataHandler.java:44)

### DIFF
--- a/src/main/java/org/codehaus/groovy/util/ListHashMap.java
+++ b/src/main/java/org/codehaus/groovy/util/ListHashMap.java
@@ -114,15 +114,15 @@ public class ListHashMap<K,V> implements Map<K,V> {
 
     @Override
     public V get(Object key) {
-        if(size==0) return null;
-        if (innerMap==null) {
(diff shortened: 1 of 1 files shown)


weih-kahoot:
I've seen this NPE occurred for quite a few times, and it's weird the fix it's backported to 4.0.x release branch.
Kindly let me know if the PR can be accepted, :)

eric-milles:
Is the request just to cherry-pick a commint from `master`?  If so, can you create a JIRA ticket with a quick description so there is tracking for the next Groovy 4 release?  Then I can handle the git ops.

paulk-asert:
After the proposed change, the code will still be prone to a NPE if called with a null 'key'.
What about using the following as the replacement for the condition in question:
{code}
Objects.equals(key, listKeys[i])
{code}

PR #1890, opened 2023-05-17
GROOVY-11062: Annotation collection isn't properly handling enum anno…

paulk-asert:
…tation attribute values


paulk-asert:
This doesn't include a test yet. It is just a little tricky to produce a test since it is only the pre-compiled case where there is a problem.

PR #1901, opened 2023-06-13
GROOVY-8499: `combinations` metadata

eric-milles:
@paulk-asert Can you have a look at this?  GROOVY-8499 examples can work under STC if `combinations` indicates it returns a list of lists.  It is up to the caller to determine the number of elements in the sub-lists, but this is not unique to use of `combinations`.

https://issues.apache.org/jira/browse/GROOVY-8499
https://issues.apache.org/jira/browse/GROOVY-11092

### DIFF
--- a/src/main/java/groovy/util/GroovyCollections.java
+++ b/src/main/java/groovy/util/GroovyCollections.java
@@ -34,19 +34,83 @@ import java.util.List;
 import java.util.Set;
 import java.util.TreeSet;
 
-/**
- * A Collections utility class
- */
 public class GroovyCollections {
+
     /**
      * Finds all combinations of items from the given collections.
      *
      * @param collections the given collections
-     * @return a List of the combinations found
+     * @return A list of the combinations found.
      * @see #combinations(Iterable)
      */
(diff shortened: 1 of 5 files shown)


src/main/java/groovy/util/GroovyCollections.java
```diff
@@ -34,50 +34,21 @@
...
     /**
      * Finds all combinations of items from the given collections.
      *
      * @param collections the given collections
-     * @return a List of the combinations found
+     * @return A list of the combinations found.
      * @see #combinations(Iterable)
      */
-    public static List combinations(Object[] collections) {
+    public static List<List> combinations(Object[]    collections) {
```
paulk-asert, line 46: A few spurious spaces after Object[]?

paulk-asert (approved):
This looks good for Groovy 5. I think the erasure will still be the same.

src/main/java/groovy/util/GroovyCollections.java
eric-milles: When collapsed, they all lign up:
image

PR #1904, opened 2023-06-19
GROOVY-9848: `in` operator: key set membership for `isCase(Map,Object)`

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-9848


eric-milles:
I added tests for GROOVY-2456 as well.  This is probably a good time to revisit that item and decide if `isCase` should be left alone for maps and strings and some other implementation provided for the `in` operator.

paulk-asert:
I think your idea about leaving isCase alone and changing "in" to use "contains" has merit and is worth exploring as an alternative to this change.

PR #1913, opened 2023-07-11
`SecureASTCustomizer` can't see types set prior by a type-checking extension

inad9300:
This is an incorrect, incomplete proposal to suggest that methods such as `SecureASTCustomizer.SecuringCodeVisitor.visitMethodCallExpression()` should probably be using `TypeCheckingExtension.getType()` as opposed to `Expression.getType()`.

This is because, as suggested in the [docs](https://docs.groovy-lang.org/next/html/documentation/#Typecheckingextensions-TheAPI), type-checking extensions (via `ASTTransformationCustomizer` with `CompileStatic`) may be calling `TypeCheckingExtension.storeType()` to dynamically assign a type to a variable, but `SecureASTCustomizer` won't see that type even if run after such type-checking extension.


eric-milles:
The `SecureASTCustomizer` is run as part of the _canonicalization_ compilation phase (5).  Static type checking is run as part of the _instruction selection_ phase (6).

But to your point, the types you can get from the AST directly are not what STC works out.  The class generator uses `TypeChooser` interface to read static types.

PR #1875, opened 2023-03-16
GROOVY-10964: List.minus() slow for Numbers

paulk-asert:

PR #1915, opened 2023-07-13
GROOVY-11127: Set operator extension methods

merscwog:
Many languages conventionally allow sets to use '|' as union, '&' as intersection, and '^' as symmetric difference operations on sets.  This PR adds those operations to Groovy Sets and SortedSets.

This is a draft PR as there are a few design considerations that have not been discussed:

- Maybe there should be an explicit symmetricDiff() method included
- The second argument for the methods possibly should be Collection or Iterable when possible instead of Set.
- All of these operations are well-defined for Sets, but maybe they could all apply to generic Collections -- so not sure if making them explicit to Set/SortedSet is desired.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -10735,6 +10735,258 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     /**
+     * Create a Collection composed of the union of both collections.  Any
+     * elements that exist in either collections are added to the resultant collection, such
+     * that no elements are duplicated in the resultant collection.
+     * For collections of custom objects; the objects should implement java.lang.Comparable
+     * <pre class="groovyTestCase">assert [1,2,3,4,5,6.7,8] == [1,2,3,4,5].union([4,5,6,7,8])</pre>
+     * By default, Groovy uses a {@link NumberAwareComparator} when determining if an
+     * element exists in the resultant collection.
+     *
+     * @param left  a Collection
+     * @param right a Collection
+     * @return a Collection as a union of both collections
+     * @see #union(Collection, Collection, Comparator)
+     * @since 5.0.0
+     */
+    public static <T> Collection<T> union(Collection<T> left, Collection<T> right) {
+        return union(left, right, new NumberAwareComparator<>());
+    }
+
+    /**
+     * Create a Collection composed of the union of both collections.  Any
+     * elements that exist in either collections are added to the resultant collection, such
+     * that no elements are duplicated in the resultant collection.
+     * For collections of custom objects; the objects should implement java.lang.Comparable
+     * <pre class="groovyTestCase">
+     * assert [1,2,3,4,5,6] == [1,2,3,4].union([3,4,5,6], Comparator.naturalOrder())
+     * assert [4,8,12,16,20,1,3] == [4,8,12,16,20].union([1,2,3,4], (x, y) {@code -> x * x <=> y})
+     * </pre>
+     * <pre class="groovyTestCase">
+     * def one = ['a', 'B', 'c', 'd']
+     * def two = ['b', 'C', 'd', 'e']
+     * def compareIgnoreCase = { a, b {@code ->} a.toLowerCase() {@code <=>} b.toLowerCase() }
+     * assert one.union(two) == ['a', 'B', 'c', 'd', 'b', 'C', 'e']
+     * assert two.union(one) == ['b', 'C', 'd', 'e', 'a', 'B', 'c']
+     * assert one.union(two, compareIgnoreCase) == ['a', 'B', 'c', 'd', 'e']
+     * assert two.union(one, compareIgnoreCase) == ['b', 'C', 'd', 'e', 'a']
+     * </pre>
+     *
+     * @param left  a Collection
+     * @param right a Collection
+     * @param comparator a Comparator
+     * @return a Collection as a union of both collections
+     * @since 5.0.0
+     */
+    public static <T> Collection<T> union(Collection<T> left, Collection<T> right, Comparator<? super T> comparator) {
+        if (left.isEmpty() && right.isEmpty()) {
+            return createSimilarCollection(left, 0);
+        }
+
+        Collection<T> result = createSimilarCollection(left, left.size() + right.size());
+        //creates the collection to look for values.
+        Collection<T> compareWith = new TreeSet<>(comparator);
+
+        for (final T leftElement : left) {
+            if (compareWith.add(leftElement)) {
+                result.add(leftElement);
+            }
+        }
+
+        for (final T rightElement : right) {
+            if (compareWith.add(rightElement)) {
+                result.add(rightElement);
+            }
+        }
+        return result;
+    }
+
+    /**
+     * Create a Collection composed of the union of both iterables.  Any
+     * elements that exist in either iterables are added to the resultant collection, such
+     * that no elements are duplicated in the resultant collection.
+     * For iterables of custom objects; the objects should implement java.lang.Comparable
+     * <pre class="groovyTestCase">
+     * assert [1,2,3,4,5,6,7,8] == [1,2,3,4,5].union([4,5,6,7,8])
+     * </pre>
+     * By default, Groovy uses a {@link NumberAwareComparator} when determining if an
+     * element already exists in the resultant collection.
+     *
+     * @param left  an Iterable
+     * @param right an Iterable
+     * @return a Collection as a union of both iterables
+     * @see #union(Iterable, Iterable, Comparator)
+     * @since 5.0.0
+     */
+    public static <T> Collection<T> union(Iterable<T> left, Iterable<T> right) {
+        return union(asCollection(left), asCollection(right));
+    }
+
+    /**
+     * Create a Collection composed of the union of both iterables.  Any
+     * elements that exist in either iterables are added to the resultant collection, such
+     * that no elements are duplicated in the resultant collection.
+     * For iterables of custom objects; the objects should implement java.lang.Comparable
+     * <pre class="groovyTestCase">
+     * assert [1,2,3,4,5,6] == [1,2,3,4].union([3,4,5,6], Comparator.naturalOrder())
+     * </pre>
+     *
+     * @param left  an Iterable
+     * @param right an Iterable
+     * @param comparator a Comparator
+     * @return a Collection as a union of both iterables
+     * @since 5.0.0
+     */
+    public static <T> Collection<T> union(Iterable<T> left, Iterable<T> right, Comparator<? super T> comparator) {
+        return union(asCollection(left), asCollection(right), comparator);
+    }
+
+    /**
+     * Create a Collection composed of the union of both iterables.
+     * Elements from the first iterable and second iterable are added to the result if the elements are not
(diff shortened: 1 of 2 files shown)


paulk-asert:
I like this PR. Could you create a Jira ticket for it also to make it easier for tracking. My suggestions for the PR:

- Let's stick with [Sorted]Set for the self/left parameter (since we already have `plus`/`intersect` for `Iterable`)
- We already use `and`, `or`, and `xor` for `BitSet`, so let's stick with `xor` rather than `symmetricDiff`
- I'd suggest `Iterable` as the second parameter like we use for `plus`, `minus` and `intersect`
- I'd suggest additional variants of all methods taking an extra `Comparator` like we do for `intersect`, this would allow the default `NumberAwareComparator` to be skipped if desired although only using the longhand syntax, e.g. `assert [1,2,5,6] as Set == ([1,2,3,4] as Set).xor([3,4,5,6], Comparator.naturalOrder())`

merscwog:
I'm still waiting on my JIRA access, but I've updated the code and tests based upon your suggestions.

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-11127

Since the stated goal is to support the `&`, `|` and `^` operators for sets, the minimal addition is 3 methods:
```java
public static <T> Set<T> and(Set<T> left, Iterable<T> right) { }
public static <T> Set<T>  or(Set<T> left, Iterable<T> right) { }
public static <T> Set<T> xor(Set<T> left, Iterable<T> right) { }
```

I'd suggest leaving everything else on the cutting room floor.  `SortedSet` implements `Set` and your extension implementation does not require the distinction, correct?

My other concern is that the operators are not commutative.  "set & iter" goes here and "iter & set" will be method missing.

DGM is a bit unorganized at the moment, but it is best to keep overloads grouped, so new and methods should be added where `and` is declared or all `and`s should be moved as a group to somewhere appropriate.  Same for `or` and `xor`.  `ArrayGroovyMethods` demonstrates alphabetical ordering.

merscwog:
Moving the methods to follow a convention would not be difficult.

Without constraining the two types to Set, there would definitely be much more work to make the operations available for all possible Iterable types as both arguments and keep the return type semantics.

paulk-asert:
@eric-milles Just some notes about historical feature requests/discussions (which we can re-evaluate if we think they no longer hold but shouldn't totally ignore):

- Folks asked us to retain the sorted nature of sets for as many operators as possible which is why there are 48 existing DGM methods which follow that pattern.
- Folks using Groovy operators on heterogeneous data very much like `NumberAwareComparator` but in some data science scenarios with homogeneous data, folks didn't want the performance hit. That's why I asked for those methods to be added. But I note that while we previously added that capability for `intersect` and `minus` we don't currently have it for `plus` (though we probably should).
- There have also been numerous discussions on commutativity in the past. We don't expect commutativity for normal method calls, e.g. `List#add(String)` isn't the same as `String#add(List)` but for operators there is a higher expectation that commutativity will hold. But even for mathematical operators, minus isn't commutative for instance. So, we while it has been a desirable feature for operators, we haven't had a strict requirement to always support commutativity. Do a search on "commutativity" on markmail (or elsewhere) and you will see numerous discussions.

merscwog:
Before I make any additional modifications, it seem there are essentially 4 options to choose from at the moment:

1. Do not add these operators at all.
2. Revert the secondary argument to Set.
3. Keep the secondary argument as Iterable.
4. Update to include Iterable, List, and Collection types as a first argument for the '|', '&', and '^' operators.

For options 2-4, adjusting the location of the new operators within DGM to satisfy desired conventions should be assumed -- I had originally placed all the new methods grouped together, as that's how I interpreted the existing grouping pattern.

Option [1] is basically what will happen if there is no consensus between 2-4 :-)

Option [2] has the advantage that it is self contained for the type that the operators are mathematically defined for, Sets.  The disadvantage, is that it requires a Set type for each argument -- kind of like BitSet and Boolean (and even Number) have for their '|', '&', and '^' operators.

Option [3] makes it easier to utilize other Iterable types without having to coerce them into a Set first to utilize the operators, but as noted, that could surprise a developer since they would receive a compilation error attempting to invert the call for a non-Set type as the first argument.

Option [4] would require adding Iterable, List and Collection first arguments to bring into close parity with the coverage of the existing intersect() method.  This would define the operators over all of the collections, assuming that actually makes sense.  Union and plus are generally treated equivalent for Sets (assuming no custom comparator -- meaning equality is same as internal Set equals() logic -- maybe this is why plus() doesn't take a custom comparator?) But for non-Sets the behavior of dealing with duplicates would need to be defined.

Now I'm also starting to believe that a dedicated union() operator for Sets is necessary, at least to get the xor() logic correct, since there is a custom NumberAwareComparator involved with intersection and minus -- and no matter which way xor() is composed from the other primitives, it needs to utilize a union() operation.

I do not believe that plus() and union() should be considered the same thing for non-Set types, as plus should just blindly add things to the collection, and union() should "ensure" that the result doesn't contain unnecessary duplicates -- which is hard to define for non-sets.

I'm starting to lean towards option [2] with an explicit union() method being added with a default NumberAwareComparator() utilized to determine whether something should be added to the resulting Set (where it will choose the value from the first set when there are duplicates)

NOTE:  This will _NOT_ guarantee true commutativity because consider ["A", "b", "C"]  as Set unioned with ["a", "B", "d"] as Set with a case-insensitive comparator:
The result will be ["A", "B", "C", "d"] or ["a", "B", "C", "d"] depending upon which was used first.  Based upon the custom comparator, they technically are equivalent, but for future uses with a different custom comparator, they would not be.

Thoughts?

paulk-asert:
I think option 3 aligns best with existing functionality. That let's us add option 4 down the track if there is interest, but we already have plus, intersect etc. on collections and that is probably sufficient whereas and, or, xor are more commonly associated with sets.

I don't think comparator variants make sense for plus for non-sets. It makes a lot of sense for sets but would be a breaking change. A union method would be a nice side-step to avoid a breaking change.

I agree with a lot of your thinking around union and non-sets. It may be that if we did look at adding option 4 later, we might just cater for "and" and "or" and not look at xor/union.

merscwog:
@paulk-asert I'm interpreting your response that your vision is for me to retain the Iterable secondary arguments, and incorporate union() methods with Set and SortedSet first arguments and Iterable as second arguments, along with a comparators version, all to be utilized by '|' and '^'.

eric-milles:
I don't think any of these additions represent something that a user cannot do with existing extension methods, except for symmetric difference, which you could do with a simple `((a - b) + (b - a))` expression.  So since these are just names to support operator overloads, I don't understand the push for a wider interface.

I was initially going to say we should not bother with this change since we have `+` and `-` operators already.  `&`, `|` and `^` make some sense for sets.  I think it is more confusion than it is worth if you add non-commutative operators.

merscwog:
@eric-milles @paulk-asert So, just for grins, I tried to see exactly what Python allows for those operators, and they are only supported on Sets.  Obviously, a different language implementation doesn't require Groovy to behave identically, but this makes me lean towards Eric's recommendation to only support the operators on Sets.

paulk-asert:
I am still keen to keep Iterable as the second argument (option 3). It makes most sense with how the DGM methods and the Java APIs are designed. I agree that non-symmetry can potentially lead to confusion but I don't see that applies here.

If we have explained that "union" is like a Set-specific "plus", then I don't see why folks might expect it to work for Collections.
If I have `Set<String> wordsSeen` and `List<String> nextSentence`, then doing `wordsSeen |= nextSentence` makes perfect sense just like we currently would allow `wordsSeen += nextSentence`. Folks dealing with large datasets aren't going to want the extra time or memory hit of creating the superfluous set. I don't see folks suddenly expecting `nextSentence |= wordsSeen` to be available when we have explained this as a set operation.

Or, to say it a different way, Set already has `containsAll(Collection)`, `addAll(Collection)`,` retainAll(Collection)`, `removeAll(Collection)`, as well as the Collection constructor, e.g. in HashSet. I see no problem with having `union(Collection)` or in our case we'd have `union(Iterable)` and `or(Iterable)` as an alias. We already have `minus(Iterable)`, `plus(Iterable)`, and `intersect(Iterable)`.

Much like we know that subtraction, power, division aren't symmetric and the order of the parameters is intrinsic to the problem domain, I see it that for the set-specific operators, the order is important and the iterable can be the second argument and have a different type to the first. Operations with a custom comparator aren't going to be symmetric behavior-wise in general, I don't see it as a given that set operations must be symmetric type-wise.

paulk-asert:
@merscwog Sounds good to me.

src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
```diff
@@ -10734,6 +10734,258 @@ public static boolean disjoint(Iterable left, Iterable right) {
         return true;
     }
 
+    /**
+     * Create a Collection composed of the union of both collections.  Any
+     * elements that exist in either collections are added to the resultant collection, such
+     * that no elements are duplicated in the resultant collection.
+     * For collections of custom objects; the objects should implement java.lang.Comparable
+     * <pre class="groovyTestCase">assert [1,2,3,4,5,6.7,8] == [1,2,3,4,5].union([4,5,6,7,8])</pre>
```
paulk-asert, line 10742: typo? "6.7" instead of "6,7"

paulk-asert: I added a commit to fix this

merscwog:
In trying to implement and run various symmetric difference tests, I discovered what might be considered a bug in the minus() method.  It contains this very innocent looking code:
```groovy
answer = new TreeSet<>(comparator);
answer.addAll(self1);
answer.removeAll(removeMe1);
```

```groovy
Set baseValues = ['a', 'B', 'c'].toSet()

Set tree1 = new TreeSet(String.CASE_INSENSITIVE_ORDER)
tree1.addAll(baseValues)
tree1.removeAll(['A', 'b'])
assert tree1 == ['c'] as Set

Set tree2 = new TreeSet(String.CASE_INSENSITIVE_ORDER)
tree2.addAll(baseValues)
tree2.removeAll(['A', 'b', 'D'])
assert tree2 == ['c'] as Set  // this fails...
```

The removeAll() method on TreeSet() is actually the one in AbstractSet, which has some very unexpected behavior, in that if the removal collection is the same length or longer than that initial collection, it flips the comparisons (which means that the TreeSet comparator is effectively ignored -- not a good thing)

AbstractSet documents the behavior, but it really seems like TreeSet (because it supports a custom comparator) should override removeAll() to be always correct.

I guess I could update minus() to do a simple iteration and call the remove() method for each element in the removal collection

paulk-asert:
@merscwog There is an open issue for `minus` that overlaps with that scenario (see GROOVY-10964). Our users have been particularly desiring of an efficient minus which also caters for NumberAwareComparator. There is a PR (see https://github.com/apache/groovy/pull/1875) also for that but I haven't progressed that since the initial PR also is affected by the AbstractSet "bug". It looks like the JDK will fix that soon (see https://bugs.openjdk.org/browse/JDK-6394757). I was planning on updating the PR to try to align our behavior with how it will end up if possible.

So, feel free to give that a crack if you want (i.e. incorporate GROOVY-10964), or ignore it altogether and skip test cases that are affected or add in a comment that the test case might change if GROOVY-10964 is resolved.

paulk-asert:
But, I am happy to merge your PR as it stands and we can add further test cases/fixes as a separate PR.

PR #1922, opened 2023-07-25
introduce interface `MetaMember`

eric-milles:
I'd like to be able to refer to elements of the meta-method index's `FastArray` uniformly.  What I want to do is check access modifiers.  However `CachedConstructor` and `MetaMethod` only share the non-abstract class `ParameterTypes`.  That does not seem the appropriate place to add member name, modifiers and anything else that is common to a class member.

I looked at `java.lang.reflect.Member` and it specifies `getName()`, `getModifiers()`, `isSynthetic()` and `getDeclaringClass()`.  `CachedConstructor` could implement `Member` but `MetaMethod` (the other subclass of `ParameterTypes`) cannot due to a conflict with the `getDeclaringClass()` return type.  So I'd like to introduce the interface `groovy.lang.MetaMember` that is inspired by `Member` but does not declare `getDeclaringClass()`.  With this, the name and modifiers (and common modifier conveniences) are available without a bunch of `instanceof` checks.

Here is a sample from `MetaClassImpl` where `ParameterTypes` and `MetaMethod` can be used in tandem:
```java
    private Object chooseMethodInternal(String methodName, Object methodOrList, Class[] arguments) throws MethodSelectionException {
        ...
        List<MetaMember> matchingMethods = new ArrayList<>(methodCount);
        Object[] methodArray = methods.getArray();
        for (int i = 0; i < methodCount; i += 1) {
            var method = (ParameterTypes & MetaMember) methodArray[i]; // extends PT and implements MM
            if (method.isValidMethod(arguments)) {
                matchingMethods.add(method);
            }
        }
        methodCount = matchingMethods.size();
        if (methodCount == 0) {
            return null;
        } else if (methodCount == 1) {
            return matchingMethods.get(0);
        }
        return doChooseMostSpecificParams(theClass.getName(), methodName, matchingMethods, arguments, false);
    }
```

NOTE: the diff shows `MethodHelper` renamed to `MetaMember` but it is really delete the former and add the latter.


daniellansun:
LGTM

PR #1926, opened 2023-08-03
GROOVY-11146: Support flattening of Optionals in flatten

paulk-asert:


paulk-asert:
I'll see if I can rework that bit.

PR #1933, opened 2023-08-17
GROOVY-11157: Provide a GINQ aggregate function shorthand for returni…

paulk-asert:
…ng a list of the selected items


daniellansun:
+1

PR #1934, opened 2023-08-18
GROOVY-11158: deprecate callsite code

blackdrag:
to give an impression on the impact of deprecating the callsite caching code


paulk-asert:
@blackdrag With recent changes, and in particular, #2730, should we close this?

PR #1935, opened 2023-08-22
Move JDK 18 into groovy-build-test-aux matrix

Goooler:
Seems JDK 18 should be seated at `Build and test additional JDK versions`.

### DIFF
--- a/.github/workflows/groovy-build-test-aux.yml
+++ b/.github/workflows/groovy-build-test-aux.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04]
-        java: [12, 13, 14, 15, 16]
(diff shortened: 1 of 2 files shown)


paulk-asert:
+1, I think you have all the necessary permissions to merge yourself.

Goooler:
Can we use squash commits for some PRs, is there any reason to use rebase only?

paulk-asert:
Yes, we can certainly squash commits when it makes sense. In fact, it is often preferred - though if there are pieces of a PR which might be cherry-picked individually for backporting to branches, we often keep those separate.

We often don't need the commit history, so merge isn't usually needed and rebase suffices giving us a cleaner history (when it makes sense). Also, in the past, some tooling like `git bisect` seemed to get more confused from merges than rebases but I haven't tracked whether that has improved over the years.

PR #1936, opened 2023-08-23
Refactor assertFalse(equals()) with assertNotEquals

Taher-Ghaleb:
I am working on research investigating test smell refactoring in which we identify alternative implementations of test cases, study how commonly used these refactorings are, and assess how acceptable they are in practice.

This smell occurs when inappropriate assertions are used, while there exist better alternatives. For example, in [GStringTest.java](https://github.com/apache/groovy/commit/f0bfe65ed5ae2b223b189ee600afc948af125fe4#diff-a8a3008dfada3fafce795f2a075bdbec7b14e24856857129ae21d24869ed083b), I refactored `assertFalse("a != c", a.equals(c));` with `assertNotEquals("a != c", a, c);`.

I would like to get your feedback on this particular test smell and its refactoring. Thanks in advance for your input.

### DIFF
--- a/src/test/groovy/GStringTest.groovy
+++ b/src/test/groovy/GStringTest.groovy
@@ -21,6 +21,8 @@ package groovy
 import groovy.test.GroovyTestCase
 import groovy.transform.Pure
 
+import static org.junit.Assert.assertNotEquals;
+
 class GStringTest extends GroovyTestCase {
 
     void check(template, teststr) {
@@ -265,8 +267,8 @@ class GStringTest extends GroovyTestCase {
         assertTrue("literal == template false", literal == template)
 
         // these fail
-        assertFalse("literal not equal to template", literal.equals(template))
-        assertFalse("template not equal to literal", template.equals(literal))
+        assertNotEquals("literal not equal to template", literal, template)
+        assertNotEquals("template not equal to literal", template, literal)
         assertTrue("hash codes not equal", literal.hashCode() != template.hashCode())
     }
 
--- a/src/test/groovy/lang/GStringTest.java
+++ b/src/test/groovy/lang/GStringTest.java
@@ -21,6 +21,8 @@ package groovy.lang;
 import groovy.test.GroovyTestCase;
 import org.codehaus.groovy.runtime.FormatHelper;
 
+import static org.junit.Assert.assertNotEquals;
+
 /**
  * Tests the use of the structured Attribute type
  */
@@ -70,7 +72,7 @@ public class GStringTest extends GroovyTestCase {
 
         assertTrue("a == b", a.equals(b));
         assertEquals("hashcode a == b", a.hashCode(), b.hashCode());
-        assertFalse("a != c", a.equals(c));
+        assertNotEquals("a != c", a, c);
         assertTrue("hashcode a != c", a.hashCode() != c.hashCode());
         assertEquals("a <=> b", 0, a.compareTo(b));
         assertEquals("a <=> b", -1, a.compareTo(c));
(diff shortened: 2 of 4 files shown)


blackdrag:
First of all thanks for the contribution. I was wondering if the change in the GString.groovy is ok, but since it does really test the equals method on GString, that is as that also accessible in Java it should be fine. I do have some remarks:
-  "This smell occurs when inappropriate assertions are used". Actually in this case you should probably say that they became inappropriate. When this test infrastructure was build assertNotEquals did not even exist. Just look at the base class and what package that comes from. So why is it no test smell to use an long outdated Junit?
- Obviously the reason of why assertNotEquals is better than assertFalse when dealing with two operands is because then the error message from assertNotEquals can use the operand values. In this case the error message is given. Is there still an advantage to the change then?

Taher-Ghaleb:
Thanks, @blackdrag, for your input.

> Actually in this case you should probably say that they became inappropriate. When this test infrastructure was build assertNotEquals did not even exist. Just look at the base class and what package that comes from.

I see your point. Perhaps we could look at such refactoring as a way of keeping unit tests up to date with the latest advancements.

> Obviously the reason of why assertNotEquals is better than assertFalse when dealing with two operands is because then the error message from assertNotEquals can use the operand values. In this case the error message is given. Is there still an advantage to the change then?

Right. `assertNotEquals` improves the readability of a test case and makes equality checks more explicit and intuitive, unlike `assertFalse` combined with `equals`, which indirectly checks if two objects are not equal. Also, as you have indicated, `assertNotEquals` typically provides more informative failure messages, as it can include details about the expected and actual values if the assertion fails, making it easier to diagnose and fix the issue. The implementation of `assertNotEquals` is also designed to handle a variety of data types, as its implementation typically includes type checks and appropriate comparison logic to determine if the values or objects are unequal, regardless of their types, unlike `assertFalse`, which typically relies on the `equals()` method of the objects being compared.

paulk-asert:
Merged, thanks!

blackdrag:
@Taher-Ghaleb 
assertNotEquals may improve readability a bit. I think that does not apply to me so much, but I can imagine others think differently. So I agree in that point. 
As for the failure message... the assertFalse may have a failure message only because it is otherwise quite a bad message in the reporting that something was true and should have been false. But then it would be consequent to check if the failure message can be remove when transforming from assertFalse to assertNotEquals. Just a suggestion. I don't know if you guys use a tool for this, or do that by hand. If it would be a tool only, like a simple refactoring, it can of course not tell if the error message in the assertFalse is still required. I did not look in detail for the PR, But I think in all cases it could have been removed. Just as a point to think about.
Then there is one point I do not agree with you and that is what you wrote about the implementation of assertNotEquals. It has a signature that takes Object for the object to check. That means there is no static type information and compiler failure that prevents me using a String and an Integer. And if you look in the PR, there are equal checks on String and GString, which are not closely related in the java world. There is a equal check on ArrayList and Range. Which might be closer, but still Rang is not a ArrayList. assertNotEquals as used here does only 2 things, checking for null and maybe  null==null, otherwise it will only call equals. So yes, there is the null check that could help, but I am usually very picky when it comes to null. I prefer testing that state explicitly if expected or not expected. That kind of negates the usefulness of that one additional step in assertNotEquals for me. Others may see it different.

Anyway... I wrote this only to give you thoughts for your research... actually one more thought. In case of Groovy power asserts are almost always a good choice: "assert string == gstring". It will trace string and gstring values as well as the expression string. In the specific case here we wanted to make very sure the equals method is checked only, which is why it was done different. See http://groovy-lang.org/semantics.html#_power_assertion

Taher-Ghaleb:
Thanks for your detailed response, @blackdrag. I really appreciated it.

PR #1939, opened 2023-08-28
fix an AsciiDoc inline code block

alex-sherwin:
needed to use "unconstrained formatting pair" since there was no space after the closing backtick (see https://docs.asciidoctor.org/asciidoc/latest/text/#unconstrained)


paulk-asert:
Merged, nice catch.

PR #1941, opened 2023-08-29
Improve method selection for script body in SecureASTCustomizer

blackdrag:


paulk-asert:
This ended up being all that was required for 4_0_X and earlier, so I merged this PR on those branches. Only for master the improved solution catering for JEP445 was needed.

PR #1948, opened 2023-09-07
GROOVY-11167: JsonOutput should handle Records like POGOs

paulk-asert:


glaforge (commented):
Looking great!
I was surprised by the java.awt.Image and had forgotten JavaBeans had icons!

src/main/java/org/codehaus/groovy/vmplugin/VMPlugin.java
```diff
@@ -179,4 +181,17 @@ static String getJavaVersion() {
...
         return Collections.emptyMap();
     }
+
+    /**
+     * Returns the list of record component names or the empty list if
+     * the class is not a record or running on a pre16 JDK.
+     *
+     * @param maybeRecord the class in question
+     * @return the default list of names
+     * @since 4.0.14
```
daniellansun, line 191: 4.0.14 was released, so the version should be 4.0.15?

paulk-asert: Indeed, fixed.

PR #1953, opened 2023-09-11
GROOVY-8299: default methods

blackdrag:
First incomplete implementation of default methods.
  

- [x] interface with default methods is not seen as trait anymore
- [x] stub generation adds default modifier to stub
- [x] allow I.super in ResolveVisitor
- [x] implement checking I.super.m(2) in static compiler
- [x] implement correct bytecode for I.super.m(2) in static compiled code
- [x] implement correct method call logic for I.super.m(2) in dynamic compiler


paulk-asert:
Jochen, looks great to me. I'll rebase and merge tomorrow if no further feedback.

paulk-asert:
Merged! Thanks!

PR #1954, opened 2023-09-13
Use the tap() method instead of with(), as tap() returns the object instead of returning the last expression of the closure like with with().

glaforge:
Without this change, variable p3 actually contained the int 1951


paulk-asert:
Because `born = 1951` is inside a closure, it is routed through `ScriptBytecodeAdapter.setGroovyObjectProperty`, if that line had been `setBorn(1951)`, it would have been okay to stick with `with`.

PR #1956, opened 2023-09-20
Add Bundle-License to the generated jars

coheigea:
SBOM tools like Syft rely on the Bundle-License in the Manifest to detect licenses when generating SBOMs. Please backport to all active branches if/when applied.


aalmiray:
Must the value be a link to the text or could it be a matching SPDX identifier?

coheigea:
I don't think it has to be a link, but it's what is in most ASF jars (e.g. CXF)

aalmiray:
More context on SPDX license id as value https://github.com/gradle/native-platform/issues/40

coheigea:
OK updated @aalmiray

aalmiray:
LGTM

PR #1973, opened 2023-10-22
fix(XmlUtil): escapeControlCharacters off-by-one bug

pinicarus:


daniellansun:
It would be better if you could provide the related test as well.

pinicarus:
> It would be better if you could provide the related test as well.

I added the range bounds to the existing test

paulk-asert:
Merged, thanks!

PR #1974, opened 2023-10-22
GROOVY-11200: (Second Approach) JsonSlurper should throw exception when parsing invalid String ended with a right curly brace

yuhengfdada:
This approach avoids array copy as in https://github.com/apache/groovy/pull/1972 by using a simple counter.


paulk-asert:
This looks okay to me. I plan to merge tomorrow if no further comments.

PR #1990, opened 2023-11-21
Fix typo dor -> for in MetaClassImpl docstrings

AndreyBozhko:
Looks like there is a typo in the docstrings for MetaClassImpl constructors


paulk-asert:
Merged, thanks!

PR #1994, opened 2023-11-25
GROOVY-11229: Support pattern matching for `instanceof`

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-11229


eric-milles:
Do you have some description how this will be integrated?  Adding something to `ClassExpression` without changing its `transformExpression` or `visit` does not make it clear where this is going.  Perhaps a description in the issue ticket before code changes would be helpful.

Was it considered to use something besides a `ClassExpression` on the RHS of an `instanceof` binary expression to represent this?  I'm pleased to see "p" represented by a variable expression, but the semantic of "Person" is represented in the `ClassExpression` and then again in the `PatternVariableExpression`.  Where will the source ranges be associated?

There is also the possibility of raising "instanceof" to a new kind of `BinaryExpression` subclass that includes the pattern-matching stuff instead of adding it to `ClassExpression`.

daniellansun:
I am looking for better AST for pattern matching for both `instanceof` and `switch`.
The PR is just for `instanceof`, but I try to make flexible to change easily for other pattern matching and destructure.

The reason of adding pattern variable to `ClassExpression` is making the change smoothly. As you known, all visitors process `ClassExpression` well, but do not know new expressions, we could refactor later, let's make it work at the beginning.

eric-milles:
It would have been better to state in the original issue ticket that the goal is to support `instanceof` and `switch` pattern matching.  It is still not clear from this where your plan takes us.  You just have something stuck onto every `ClassExpression` that is not referenced by anything and does not tie into variable scopes or resolve visitor or anything else.

daniellansun:
Closed. I misunderstood the usage of draft pull request.

PR #1997, opened 2023-11-26
GROOVY-10060: Support static & private interface methods

paulk-asert:
Currently for dynamic dispatch, default interface methods which access private methods must use the syntax `InterfaceName.this.privateMethod(args)`. We might be able to remove that limitation with further work. The qualification is not needed for static compilation.

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -1288,7 +1288,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
         configureAST(classNode, ctx);
         classNode.setSyntheticPublic(syntheticPublic);
         classNode.setGenericsTypes(this.visitTypeParameters(ctx.typeParameters()));
-        boolean isInterfaceWithDefaultMethods = (isInterface && this.containsDefaultMethods(ctx));
+        boolean isInterfaceWithDefaultMethods = (isInterface && this.containsDefaultOrPrivateMethods(ctx));
 
         if (isSealed) {
             AnnotationNode sealedAnnotationNode = makeAnnotationNode(Sealed.class);
@@ -1395,7 +1395,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
         }
     }
 
-    private boolean containsDefaultMethods(final ClassDeclarationContext ctx) {
+    private boolean containsDefaultOrPrivateMethods(final ClassDeclarationContext ctx) {
         List<MethodDeclarationContext> methodDeclarationContextList =
                 (List<MethodDeclarationContext>) ctx.classBody().classBodyDeclaration().stream()
                         .map(ClassBodyDeclarationContext::memberDeclaration)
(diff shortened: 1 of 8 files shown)


daniellansun:
+1

eric-milles:
Now `private` methods are supported in interfaces.  Are `static` also supported?  It was less clear from the changes if this is the case.  Also, are there tests for `protected` and `@PackageScope` method errors?  I can't quite find them among the tests.  Is there a statement about property support -- I assume now "def foo" is a `static final` constant -- or any other differences from the Groovy 4 trait solution?

Regarding `InterfaceName.this.method()`, is that an issue of the method indexing or the access restriction or the `invokeinterface` instruction?  Just curious what that's about and if there is some implication to an implementer calling the interface's private method from outside the interface's own methods.

eric-milles:
Something is missing in the stub generator, because when I try to add a static method to a groovy interface, it states that the java implementer needs to implement the abstract method.

Here is my `org.codehaus.groovy.classgen.InterfaceTest` that is updated to try out the Java interop if these new changes:
```groovy
package org.codehaus.groovy.classgen

import org.codehaus.groovy.control.CompilerConfiguration
import org.codehaus.groovy.tools.javac.JavaAwareCompilationUnit
import org.junit.Test

final class InterfaceTest {

    @Test
    void testJavaImplementsGroovyInterface() {
        def config = new CompilerConfiguration(
            targetDirectory: File.createTempDir(),
            jointCompilationOptions: [memStub: true]
        )
        def parentDir = File.createTempDir()
        try {
            new File(parentDir, 'test').mkdir()

            def a = new File(parentDir, 'test/GClass.groovy')
            a.write '''package test
                class GClass {
                }
            '''
            def b = new File(parentDir, 'test/GInterface.groovy')
            b.write '''package test
                interface GInterface {
                    GClass[] getGC()
                    default String foo() { 'foo' + GInterface.this.bar() }
                    private String bar() { 'bar' }
//                    static  String baz() { 'baz' }
                }
            '''
            def c = new File(parentDir, 'test/JClass.java')
            c.write '''package test;
                public class JClass implements GInterface {
                    public GClass[] getGC() {
                        return new GClass[0];
                    }
                    public String toString() {
                        return this.foo();
                    }
                }
            '''
            def d = new File(parentDir, 'Main.groovy')
            d.write '''
                def jc = new test.JClass()
                assert jc.getGC().length == 0
                assert jc.toString() == 'foobar'
//                assert test.JClass.baz() == 'baz'
            '''

            def loader = new GroovyClassLoader(this.class.classLoader)
            def cu = new JavaAwareCompilationUnit(config, loader)
            cu.addSources(a, b, c, d)
            cu.compile()

            loader.loadClass('Main').main()
        } finally {
            config.targetDirectory.deleteDir()
            parentDir.deleteDir()
        }
    }
}
```

eric-milles:
@blackdrag @paulk-asert  I was able to implement a groovy `interface` with both `default` and `private` methods in it using a java `class` (test case above).  Very encouraging progress!

paulk-asert:
I'll split the issue up since I don't think I will get static finished for alpha-3. There is still some more testing and doco to come too.

paulk-asert:
Merged the private methods commits.

PR #1999, opened 2023-11-27
GROOVY-11238: Logical implication operator revisited

paulk-asert:


src/antlr/GroovyParser.g4
```diff
@@ -848,6 +848,9 @@ expression
     // logical or (||)  (level 13)
     |   left=expression nls op=OR nls right=expression                                      #logicalOrExprAlt
 
+    // implication (==>)  (level 13.5)
+    |   <assoc=right> left=expression nls op=IMPLIES nls right=expression                                 #implicationExprAlt
```
daniellansun, line 852: Logical Operators are left-associative, I think we do not need the assoc option. 

See also https://stackoverflow.com/questions/39317938/associativity-and-precedence-of-java-operators-in-antlrv4-grammar

paulk-asert: Looking into the reasoning behind Dafny's choice it is because implication is often read like the ternary operator (which is right associative).

So, elaborating on the example I gave in the issue, you might see (if Groovy had a framework like Dafny's):
```
@Ensures(@ForAll int i, @ForAll int j, { i, j ->
    i in 0..<items.size() && j in 0..<items.size() ==>
        i <= j ==>
            result[i] <= result[j]
})
List<String> sort(List<String> items) {
    // sort method here
}
```
All of Dafny's other operators match what we have, i.e. right associative for assignment/ternary but left elsewhere.

But if you have a strong counter example/counter argument we can switch it back.

daniellansun: LGTM

PR #2001, opened 2023-11-28
GROOVY-11237: Support static interface methods

paulk-asert:


src/main/java/org/codehaus/groovy/classgen/ClassCompletionVerifier.java
```diff
@@ -418,10 +418,6 @@ private void checkMethodsForIncorrectModifiers(final ClassNode cn) {
                 addError("The " + getDescription(method) + " from " + getDescription(cn) +
```
eric-milles, line 418: In an interface, is final allowed on private or static methods?  If allowed, this check needs a second look.  The message may need to check for method with and without body before stating "abstract by definition".

src/main/java/org/codehaus/groovy/classgen/ClassCompletionVerifier.java
```diff
@@ -418,10 +418,6 @@ private void checkMethodsForIncorrectModifiers(final ClassNode cn) {
...
                         " must not be final. It is by definition abstract.", method);
             }
-            if (method.isStatic() && !method.isStaticConstructor()) {
-                addError("The " + getDescription(method) + " from " + getDescription(cn) +
-                        " must not be static. Only fields may be static in an interface.", method);
-            }
         }
     }
```
eric-milles, line 423: `checkInterfaceMethodVisibility` may need some changes.  I rewrote it like this:
```java
    private void checkInterfaceMethodVisibility(final ClassNode node) {
        if (!node.isInterface()) return;
        for (MethodNode method : node.getMethods()) {
            if (!method.isPublic()) {
                if (method.isAbstract()) {
                    addError("Method '" + method.getName() + "' from " + getDescription(node) +
                            " must be public as it is declared as an abstract method.", method);
                } else if (!method.isPrivate() && !method.isStaticConstructor()) {
                    addError("Method '" + method.getName() + "' is " +
                            (method.isProtected() ? "protected" : "package-private") +
                            " but should be private in " + getDescription(node) + ".", method);
                }
            }
        }
    }
```

src/test/groovy/InterfaceTest.groovy
```diff
@@ -91,4 +91,19 @@ final class InterfaceTest extends CompilableTestSupport {
...
+            interface Foo {
+                static hello(where) { "hello $where" }
+                static String BAR = 'bar'
+                String BAA = 'baa' // implicit static
+            }
+
+            assert Foo.hello('world') == 'hello world'
+            assert Foo.BAR == 'bar'
+            assert Foo.BAA == 'baa'
+        '''
```
eric-milles, line 107: Can you also check that property methods are not generated:
```groovy
            groovy.test.GroovyAssert.shouldFail(MissingMethodException) {
                Foo.getBAR()
            }
            groovy.test.GroovyAssert.shouldFail(MissingMethodException) {
                Foo.getBAA()
            }
```

eric-milles (commented):
Also JavaStubGenerator needs to write "static".  And if you support non-public static methods, the visibility modifier.

```java
    private void printMethod(final PrintWriter out, final ClassNode classNode, final MethodNode methodNode) {
        if (methodNode.isStaticConstructor()) return;
        if (methodNode.isPrivate() || !Utilities.isJavaIdentifier(methodNode.getName())) return;
        if (methodNode.isSynthetic() && (methodNode.getName().equals("$getStaticMetaClass"))) return;

        printAnnotations(out, methodNode);
        if (!isInterfaceOrTrait(classNode)) {
            int modifiers = methodNode.getModifiers();
            if (classNode.isEnum()) modifiers &= ~Opcodes.ACC_ABSTRACT;
            else if (isDefaultTraitImpl(methodNode)) modifiers ^= Opcodes.ACC_ABSTRACT;
            if (methodNode.isSyntheticPublic() && hasPackageScopeXform(methodNode, PackageScopeTarget.METHODS)) modifiers &= ~Opcodes.ACC_PUBLIC;

            printModifiers(out, modifiers);
        } else if (methodNode.isDefault()) {
            out.print("default ");
        } else if (methodNode.isStatic()) { // ADDED
            out.print("static ");
        }
```

eric-milles:
I have all these changes together locally if you'd like me to push a branch or commit.

paulk-asert:
> With static method support so close at hand, did you think of re-rolling 5.0.0-alpha-3?

Let's go with the current release and shoot for another before the end of the year.

src/main/java/org/codehaus/groovy/classgen/ClassCompletionVerifier.java
paulk-asert: I don't think there is any case where final is allowed.

src/test/groovy/InterfaceTest.groovy
paulk-asert: Added

paulk-asert:
> I have all these changes together locally if you'd like me to push a branch or commit.

I've included some of them with tweaks. I'll merge when ready but feel free to amend further if needed.

PR #2004, opened 2023-11-30
Renovate updates

Goooler:
I'm enabling @renovate-bot in my fork, it would send dep updates more efficiently than @dependabot. I'll create updates from it if we can accept.

https://docs.renovatebot.com


Goooler:
I'm planning to send Renovate updates manually, I'll view their docs later for looking if there is any automatically way.

PR #2017, opened 2023-12-20
GROOVY-11255: DGM#flatten flattens nested structures including arrays…

paulk-asert:
… but not primitive arrays


orubel:
THANK YOU!!!

PR #2018, opened 2023-12-20
Partial cleanup of ReflectionCache

blackdrag:
A bit refactoring of ReflectionCache
I think we should actually completely get rid of this class. The leftovers could go to CachedClass. And actually this searchMethods method in CachedClass https://github.com/apache/groovy/blob/94c7b6c2bbc8c7ddfbb7c83cac71a0a4348fca47/src/main/java/org/codehaus/groovy/reflection/CachedClass.java#L271 should also go away.

### DIFF
--- a/src/main/java/org/codehaus/groovy/reflection/CachedClass.java
+++ b/src/main/java/org/codehaus/groovy/reflection/CachedClass.java
@@ -242,14 +242,6 @@ public class CachedClass {
         modifiers = klazz.getModifiers();
         isInterface = klazz.isInterface();
         isNumber = Number.class.isAssignableFrom(klazz);
-
-        for (CachedClass inf : getInterfaces()) {
-            ReflectionCache.isAssignableFrom(klazz, inf.cachedClass);
-        }
-
-        for (CachedClass cur = this; cur != null; cur = cur.getCachedSuperClass()) {
-            ReflectionCache.setAssignableFrom(cur.cachedClass, klazz);
-        }
     }
 
     public CachedClass getCachedSuperClass() {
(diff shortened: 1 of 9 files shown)


daniellansun:
+1

src/main/java/org/codehaus/groovy/reflection/CachedClass.java
```diff
@@ -282,14 +274,36 @@ public CachedMethod searchMethods(String name, CachedClass[] parameterTypes) {
...
+        }
+
+        for (int i = 0; i < a1.length; i++) {
+            if (a1[i] != a2[i]) {
+                return false;
+            }
+        }
+
+        return true;
+    }
```
eric-milles, line 305: Is there a DGM or JVM method for this now?

src/main/java/org/codehaus/groovy/reflection/CachedClass.java
```diff
@@ -339,7 +353,7 @@ public String getTypeDescription() {
         return BytecodeHelper.getTypeDescription(getTheClass());
     }
 
-    public final Class getTheClass() {
+    public final Class<?> getTheClass() {
```
eric-milles, line 356: I think this is a binary compat issue.

blackdrag: > I think this is a binary compat issue.

Could you explain how?

src/main/java/org/codehaus/groovy/reflection/CachedClass.java
blackdrag: > Is there a DGM or JVM method for this now?

for line 298+ maybe, though I did not see a DGM for this. There is of course DefaultTypeTransformation#compareArrayEqual, but I would not want to use Array.get if I don't have to. Arrays.equals(a1,a2) could be used for line 294+... It does call equals instead of doing a reference check, but since the Objects in question here are Class instances and Class#equals does also only a reference check it is possible.

paulk-asert: Arrays.equals would be less code to maintain but is likely less efficient. I'm inclined to leave as is for now.

src/main/java/org/codehaus/groovy/reflection/CachedClass.java
paulk-asert: I think it is okay in this instance

paulk-asert:
Merged - Thanks! I took everything except the import re-ordering for CachedSAMClass. We have until now always just used the default ordering which Intellij uses (original Sun coding style I believe). I backported to Groovy 4, deprecating the relevant methods and carrying over the trivial refactorings.

PR #2019, opened 2023-12-21
make use of Runtime.version().feature()

blackdrag:
Maybe special attention should be given to the change of ClassNodeResolver. So far whenever Groovy tried to compile on a newer JDK it did fail because it could not read the class file version. This is mostly due to the decompiler (using ASM to read class files).  https://github.com/apache/groovy/blob/60bde29d2d6546d0cf2c4f83a5f2ad9c641710e6/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java#L260
This would normally throw a IllegalArgumentException, which is by this change now swallowed and Groovy falls back to ClassLoader based reading of the file instead. This is of course slower then, but it will still work.

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
+++ b/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
@@ -257,7 +257,7 @@ public class ClassNodeResolver {
                     // this may happen under Windows because getResource is case-insensitive under that OS!
                     asmClass = null;
                 }
-            } catch (IOException e) {
+            } catch (Exception e) {
                 // fall through and attempt other search strategies
             }
         }
--- a/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
+++ b/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
@@ -188,15 +188,16 @@ public class ClassNodeResolver {
     private LookupResult tryAsLoaderClassOrScript(final String name, final CompilationUnit compilationUnit) {
         GroovyClassLoader loader = compilationUnit.getClassLoader();
         Map<String, Boolean> options = compilationUnit.configuration.getOptimizationOptions();
+        boolean noClassLoaderResolving = Boolean.FALSE.equals(options.get("classLoaderResolving"));
 
         if (!Boolean.FALSE.equals(options.get("asmResolving"))) {
-            LookupResult result = findDecompiled(name, compilationUnit, loader);
+            LookupResult result = findDecompiled(name, compilationUnit, loader, noClassLoaderResolving);
             if (result != null) {
                 return result;
             }
         }
 
-        if (!Boolean.FALSE.equals(options.get("classLoaderResolving"))) {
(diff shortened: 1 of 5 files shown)


blackdrag:
> Can you just quickly explain the different return values between `VMPlugin.getJavaVersion()` and `Runtime.version().feature()`?

`Runtime.version().feature()` is available since Java10 and returns the major version of the used JVM. For Oracle 17.36 it would thus return 17. I have no idea why they called that feature.

> Do we want to make capture of non-`IOException` dependent on "classLoaderResolving" optimization option in `CompilerConfiguration`? See `ClassNodeResolver#tryAsLoaderClassOrScript`.

You mean if class loader resolving is turned off? Well... in that case Groovy won't be able to load the class. My current version would mean Groovy will in that case fail to resolve the class and eventually fail resolving completely. I guess you are right. It would e better to give a proper error message instead for this case

paulk-asert:
@blackdrag Did you want to update the PR with the error case when class loader resolving is turned off?

blackdrag:
> @blackdrag Did you want to update the PR with the error case when class loader resolving is turned off? If so, then +1 from me.

Yes, I will update it, but it may take a week

blackdrag:
> > @blackdrag Did you want to update the PR with the error case when class loader resolving is turned off? If so, then +1 from me.
> 
> Yes, I will update it, but it may take a week

@paulk-asert @eric-milles I updated the PR. instead of just catching everything I go specifically for the IllegalArgumentException. I think that is a better variant

eric-milles:
The changes all look fine except the `IllegalArgumentException` bit.  Can you change the name of the parameter to indicate fail on asm parsing or whatever you'd call the purpose of the `try` block?

Are you saying that if ASM and our ASM-plugins can't work out what the class file says, it throws `IllegalArgumentException`?  Is there a test or two that cause this to happen?

blackdrag:
> The changes all look fine except the `IllegalArgumentException` bit. Can you change the name of the parameter to indicate fail on asm parsing or whatever you'd call the purpose of the `try` block?

sure

> Are you saying that if ASM and our ASM-plugins can't work out what the class file says, it throws `IllegalArgumentException`? Is there a test or two that cause this to happen?

This is the code from ClassReader:
```
  @SuppressWarnings("PMD.ConstructorCallsOverridableMethod")
  ClassReader(
      final byte[] classFileBuffer, final int classFileOffset, final boolean checkClassVersion) {
    this.classFileBuffer = classFileBuffer;
    this.b = classFileBuffer;
    // Check the class' major_version. This field is after the magic and minor_version fields, which
    // use 4 and 2 bytes respectively.
    if (checkClassVersion && readShort(classFileOffset + 6) > Opcodes.V22) {
      throw new IllegalArgumentException(
          "Unsupported class file major version " + readShort(classFileOffset + 6));
    }
```
So it is not if the library fails to read the class version, but if the read class version is for above Java 22. Of course this changes with the next asm releases.

daniellansun (approved):

daniellansun:
Merged. Thanks.

PR #2020, opened 2023-12-22
Replacing custom hashmap MetaMethodIndex with standard maps

blackdrag:
This replaces the old custom hashmap variant in MetaMethodIndex with standard maps. Problem though is that this change is a breaking change. MetaMethodIndex itself looks very different now, but is not likely used. There is one protected method in MetaClassImpl which is also used for example by ExpandoMetaClass, which I consider slightly more as problem. Also I think the refactoring needs input on how to improve further

### DIFF
--- a/src/main/java/groovy/lang/MetaClassImpl.java
+++ b/src/main/java/groovy/lang/MetaClassImpl.java
@@ -18,6 +18,36 @@
  */
 package groovy.lang;
 
+import java.beans.BeanInfo;
+import java.beans.EventSetDescriptor;
+import java.beans.Introspector;
+import java.beans.MethodDescriptor;
+import java.beans.PropertyDescriptor;
+import java.lang.reflect.Array;
+import java.lang.reflect.Constructor;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
+import java.net.URL;
+import java.security.PrivilegedActionException;
+import java.security.PrivilegedExceptionAction;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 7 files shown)


daniellansun:
LGTM 🙂

eric-milles:
The changes in `MetaClassImpl` look fine.  `MetaMethodIndex` does not even load it is so different.

Why are the methods deprecated in `MetaMethod` and `ParameterTypes`?  Are they not useful for other purposes?

Besides replacing with standard types, does this improve performance or memory usage or some other characteristic?

blackdrag:
> Why are the methods deprecated in `MetaMethod` and `ParameterTypes`? Are they not useful for other purposes?

There are two cases, both of them are unused. 
First there is the ParameterTypes constructor, which loads types based on String. This is something we have to be extra careful with, because  this works only if the correct class loader is used. Which means there is only a limited use for this method in specific cases and in general we cannot use it. IMHO it makes no sense to keep this around. 
Next is the MetaMethod:
   ```
      public void checkParameters(Class[] arguments) {
        // let's check that the argument types are valid
        if (!isValidMethod(arguments)) {
            throw new IllegalArgumentException(
                    "Parameters to method: "
                    + getName()
                    + " do not match types: "
                    + FormatHelper.toString(getParameterTypes())
                    + " for arguments: "
                    + FormatHelper.toString(arguments));
        }
    }
```
for our runtime this method has no  use. We have isValidMethod instead, which is also used in this case. The method is just a wrapper around that for error reporting. This could be useful in theory, but I don't see why we should keep it. I mean what is the scenario for it?
 
> Besides replacing with standard types, does this improve performance or memory usage or some other characteristic?

It should be a bit more threadsafe since I use the concurrent hashmap. But I think in department we will have to do quite a bit more work. Otherwise the goal was no improving performance or memory usage but maintainability. It is now much more clear what this structure is compared to before.

daniellansun (approved):

daniellansun:
Merged. Thanks.

PR #2022, opened 2023-12-29
GROOVY-11260: ASTMatcher should support matching a var-arg placeholder

paulk-asert:


paulk-asert:
It currently doesn't check whether the placeholder names and varargPlaceholder names are distinct. It may not matter but code with such name clashes might become harder to understand, so maybe we should just rule it out.

Also, it currently just matches against all arguments - rather than allowing some normal args followed by the varargs pattern. I'd prefer the latter but it adds much more complexity and the examples where I am using this haven't required this functionality yet. It should be backwards compatible to add such functionality later.

PR #2024, opened 2024-01-01
GROOVY-11264: Bump jarjar to 1.13.1

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-11264


daniellansun:
Let's check if the latest jarjar still leaves empty META-INF/services files.

https://github.com/apache/groovy/blob/80e92ec31a6afe2d8b1260a1452b698fbec9c963/versions.properties#L34

daniellansun:
`com.eed3si9n.jarjar.JarJarTask` is missing in jarjar 1.13.1

PR #2037, opened 2024-01-20
GROOVY-11292: Class without sealed parent cannot be non-sealed

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-11292

### DIFF
--- a/.github/workflows/groovy-build-test-main.yml
+++ b/.github/workflows/groovy-build-test-main.yml
@@ -26,7 +26,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-22.04]
-        java: [11, 17]
+        java: [11, 17, 21]
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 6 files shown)


eric-milles:
Could you add the Java 19+ build stuff separately?

Is it possible to incorporate the non-sealed checking into `isSealed()` and then you would be checking `!isSealed()` instead of `isNonSealed()`?  Otherwise if we need both `isSealed()` and `isNonSealed()` can they be stood up in `ClassNodeUtils` instead?  These both came so quickly without much time to consider the impact to public API.

daniellansun:
> Could you add the Java 19+ build stuff separately?
Sure. I just want to see if the fix works for Java 19+.

> Is it possible to incorporate the non-sealed checking into isSealed() and then you would be checking !isSealed() instead of isNonSealed()?
`isNonSealed()` means the type is declared `non-sealed`, it is not `sealed` and its parent must be `sealed`, so `isNonSealed()` can not just be expressed by `!isSealed()`. So I would like to accept your suggestion putting `isNonSealed()` into `ClassNodeUtils`.

paulk-asert (approved):

PR #2044, opened 2024-01-26
GROOVY-11294: Make generated toList() and toMap() methods on records return immutable collections

PaintNinja:
https://issues.apache.org/jira/browse/GROOVY-11294

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/InvokerHelper.java
+++ b/src/main/java/org/codehaus/groovy/runtime/InvokerHelper.java
@@ -405,6 +405,10 @@ public class InvokerHelper {
         return answer;
     }
 
+    public static Map createImmutableMap(Object[] values) {
+        return VMPluginFactory.getPlugin().createImmutableMap(values);
+    }
+
     public static void assertFailed(final Object expression, final Object message) {
         throw createAssertError(expression, message);
     }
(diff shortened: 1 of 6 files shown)


eric-milles:
Adding reliance on the runtime library (`InvokerHelper` and `VMPlugin`) means a record can no longer be tagged `@POJO` and used in a vanilla java environment.  And though an immutable retval could be cached, it isn't here.  So what is the primary benefit?  It looks like a pound of cure for an ounce of problem.

PaintNinja:
Groovy records already rely on the runtime library - the generated toList() and toMap() use ScriptBytecodeAdapter#createList/createMap in Groovy 4 and Groovy 5 snapshots (pre this PR).

As explained on the mailing list, using the runtime library allows for using Java's faster immutable List.of()/Map.of() when possible without needing many changes to the AST transform or bytecode gen.

After reading some of the feedback here, on the mailing list and jira, I better understand the importance of preferring simpler approaches and will be replying to the comments and re-evaluating my PR accordingly when I have time.

paulk-asert:
As per comment in the issue, I'll close this PR. There is plenty of scope for further improvements - as noted, even with POJO, records still make calls into runtime. This is why @POJO is marked @Incubating and one of the things we still plan to improve. But it is the kind of thing we'd like to improve for more than just records if we can, so we'd target a broader change if we can make such a change.

PR #2047, opened 2024-01-29
GROOVY-11294: Make generated toList() and toMap() methods on records …

paulk-asert:
…return immutable collections


daniellansun:
+1

PR #2050, opened 2024-02-04
Remove redundant `gradle_version` property

Goooler:
It's easy to forget to modify this property, and we have no much necessary to input it for Wrapper tasks.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -107,8 +107,8 @@ As mentioned in the previous paragraph, if you download the source distribution
 you need to bootstrap Gradle. This isn't needed if you clone from the GitHub repo.
 
 Each version of Groovy is built and tested using a specific version of Gradle.
-That version is specified by the `gradle_version` property defined in the `gradle.properties`
-file within the root directory. Luckily you shouldn't need to know that version and,
+That version is defined in the `gradle/wrapper/gradle-wrapper.properties` file.
+Luckily you shouldn't need to know that version and,
 after bootstrapping, you should use the `gradlew` command which will ensure the
 correct version is always used.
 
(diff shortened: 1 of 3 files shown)


paulk-asert:
I don't think we can remove this. Current ASF policy prohibits binary files like the wrapper jar from being included in the src zip. So, to build from source, we need to have the bootstrap step. Here is where we exclude 'wrapper' (which contains version in properties file):
https://github.com/apache/groovy/blob/master/subprojects/groovy-binary/build.gradle#L184
Best bet is to only update version using bootstrap step - though I realise that isn't what dependabot et al will use or the standard instructions from Gradle.
There are other approaches to solve the problem, which I guess we could relook at, but it wouldn't be (just) the change suggested here.

Goooler:
If we can remove that line to prevent `gradle/wrapper` been deleted?

paulk-asert:
That is the spec for building the source zip. As per ASF requirements, it can't contain binary files, e.g. gradle-wrapper.jar. We have asked for an exemption in the past but haven't been successful. ASF values highly the "openness" of the source zip. The implication is that anyone should be able to look at the files used within the source zip when building from source and understand what everything does without having to try to work out what might be inside any binary object such as a jar.

Goooler:
Got it! Thanks for your explanation.

aalmiray:
FWIW I followed the trail of the Maven wrapper to bootstrap it from source. The initial commit is here https://github.com/jreleaser/jreleaser/commit/82f0706d98674f761d7253b81952b19756100ada with updates posted a few times later. In case you'd like to do something similar.

PR #2056, opened 2024-02-14
GROOVY-11308: Restore compatibility of Collection#unique

octylFractal:
The documentation for this method states that when `mutate` is false, a new collection is _always_ returned. The previous code only did so when `self.size()` was greater than 1, and otherwise returned null.


paulk-asert:
Merged, thanks! There is a small tweak we can do to get the minor performance gain that I think the original code was after, but I can do that after the merge.

PR #2063, opened 2024-03-07
GROOVY-11330:

paulk-asert:
Remove experimental AGM minBy/maxBy @Incubating methods now that selection by parameter count is improved


daniellansun:
+1

PR #2072, opened 2024-04-03
Test on JDK 22 and update JDK 23 EA

Goooler:


Goooler:
Ooops, this may need Gradle 8.8

daniellansun:
Gradle 8.7 does not support JDK 22 because it still uses Groovy 3.

paulk-asert:
> Gradle 8.7 does not support JDK 22 because it still uses Groovy 3.

Gradle 8.7 uses Groovy 3.0.17. Groovy 3.0.21 embeds ASM 9.6 and hence can read JDK 22 class files. We have additional support for recent JDKs in Groovy 4+. We haven't done any testing to try to determine whether Gradle needs the additional support or is unaffected.

Goooler:
@paulk-asert PTL

PR #2082, opened 2024-05-01
Fix description of Observable example code.

brettle:
The descriptions for 2 of the references were switched.


paulk-asert:
Merged, thanks! Nice catch.

PR #2084, opened 2024-05-10
Use Thread subclass instead of Closure in Console

pditommaso:
The use of closure to run the method `doRun` in the Groovy `Console` class makes impossible to extend the `Console` in custom code due to this [Groovy issue](https://issues.apache.org/jira/browse/GROOVY-2433). 

This PR solves the problem creating running thread using a `Thread` subclass instead of a Closure

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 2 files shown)


daniellansun:
Merged. Thanks!

https://github.com/apache/groovy/commit/ab3bc99c7b663fba284403b768b2e7179a298439

PR #2086, opened 2024-05-19
Make `Console.prefs` public accessible 

pditommaso:
The [Console.prefs](https://github.com/apache/groovy/blob/GROOVY_4_0_21/subprojects/groovy-console/src/main/groovy/groovy/console/ui/Console.groovy#L111) property is declared private, however, it's accessed by ConsolePreference object, for example [here](https://github.com/apache/groovy/blob/GROOVY_4_0_21/subprojects/groovy-console/src/main/groovy/groovy/console/ui/ConsolePreferences.groovy#L52-L52), causing an error to be reported when the Console class is extended by a subclass

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,9 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-groovyVersion=4.0.21-SNAPSHOT
+groovyVersion=4.0.21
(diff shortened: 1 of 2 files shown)


daniellansun:
Merged. Thanks.

See also: https://github.com/apache/groovy/commit/492b91aedce05e7b56ec99c8aa50e461f5fc5955

PR #2092, opened 2024-06-12
GROOVY-10993: Add CycloneDX SBOM files

paulk-asert:
This creates and publishes the cyclonedx sbom files but there are essentially empty ones for groovy-all, groovy-bom and groovy-binary. I don't know if that is a bug or feature. I.e. I don't know whether security scanning tools follow the transitive dependencies and merged to sbom data. If not we might have to do some aggregation like we do for groovy-all docs.


daniellansun:
Merged. Thanks.

PR #2097, opened 2024-07-04
GROOVY-11432: Support method references/method pointers in annotation…

paulk-asert:
… attributes


daniellansun:
+1

paulk-asert:
I'll write an email to the dev list shortly and await feedback before merging. As currently designed, it works for dynamic Groovy and fails when using static compilation. This is kind of as I'd expect since we have a "dangling" method reference not associated with a functional interface. But we could also make some assumptions and use a `Consumer` and make it work for statically compiled code.

paulk-asert:
**Update**: the current PR now uses `withMethodClosure` instead of the previous version which used `with`. This means that it works with static compilation and static type checking.

PR #2100, opened 2024-07-17
GROOVY-11443: Support multiple Requires/Ensures/Invariant annotations…

paulk-asert:
… in groovy-contracts


paulk-asert:
I still have some tests to write but manual testing seems to work as expected.

daniellansun:
+1

PR #2116, opened 2024-10-09
fix: slashy string doc

jlguenego:
escaping the slash.
"ends with slash ${'\\'}"


paulk-asert:
Merged, thanks! It will appear on the website when we do the next release (or if we do a manual website update in the meantime).

PR #2118, opened 2024-11-03
GROOVY-11516: Improve consistency of treatment for internal properties

paulk-asert:


paulk-asert:
Relooking at GROOVY-11928.

PR #2119, opened 2024-11-09
Update style-guide.adoc with 4 proper rendering of "`" character

fanofxiaofeng:
As mentioned in Example 6 in https://docs.asciidoctor.org/asciidoc/latest/syntax-quick-reference/#text-formatting, 
two consecutive "`" character can be used to render unconstrained italic (I think it means italic that are not constrained to boundary of words)
assert
Throwable

There are four places that "`" is not rendered as expected in this adoc file (as shown above), I tried to fix all of them.
Please help review. Thanks

### DIFF
--- a/src/spec/doc/style-guide.adoc
+++ b/src/spec/doc/style-guide.adoc
@@ -622,7 +622,7 @@ and the resulting value will be null if something's `null`.
 
 To check your parameters, your return values, and more, you can use the `assert` statement.
 
-Contrary to Java's `assert`, `assert`s don't need to be activated to be working, so `assert`s are always checked.
+Contrary to Java's `assert`, ``assert``s don't need to be activated to be working, so ``assert``s are always checked.
 
 [source,groovy]
 ----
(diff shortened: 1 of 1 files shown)


src/spec/doc/style-guide.adoc
```diff
@@ -622,7 +622,7 @@ and the resulting value will be null if something's `null`.
 
 To check your parameters, your return values, and more, you can use the `assert` statement.
 
-Contrary to Java's `assert`, `assert`s don't need to be activated to be working, so `assert`s are always checked.
```
paulk-asert, line 625: looks like two more at the start of this line?

paulk-asert: Actually, we probably want all those examples to be code escaping.

paulk-asert:
Thanks for spotting those. I suspect they might have worked as is on a very old version of Asciidoc. In any case, I think the sentence could be made clearer without the pluralization of those keywords. I might try to reword if I can find time.

fanofxiaofeng:
OK, thanks for the details.

PR #2139, opened 2025-01-03
GROOVY-11541: Sql wildcard params

seregamorph:
https://issues.apache.org/jira/browse/GROOVY-11541
`groovy.sql.Sql` declares several methods with SQL statement and parameters like this
```java
public boolean execute(String sql, List<Object> params) throws SLQException {
```

This declaration has one major drawback: it's not possible to pass as an argument the List which is not exactly `List<Object>`, e.g. `List<Integer>`.

This List is only iterated, hence using wildcard here is safe. Also changing the type to wildcard as receiving argument should be binary compatible.

Suggested method signature (example; address all such methods of Sql class):
```java
public boolean execute(String sql, List<?> params) throws SLQException {
```

Additional notes:
* there are few methods that return `List<Object>`, e.g. `SqlWithParams.getParams()` - in this case the Object generic should be preserved.
* For reference: pretty similar problem solved for cassandra-java-driver https://github.com/apache/cassandra-java-driver/commit/be07a77e4f80435a52b1347a07fd566448d9f4ef

### DIFF
--- a/subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
+++ b/subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
@@ -1048,7 +1048,7 @@ public class Sql implements AutoCloseable {
      * @param closure called for each row with a <code>ResultSet</code>
      * @throws SQLException if a database access error occurs
      */
-    public void query(String sql, List<Object> params, @ClosureParams(value=SimpleType.class, options="java.sql.ResultSet") Closure closure) throws SQLException {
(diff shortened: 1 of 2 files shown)


seregamorph:
The build is broken, but doesn't seem related as master branch is also broken: https://github.com/apache/groovy/actions/runs/12586627965/job/35088979045

daniellansun:
LGTM.

No worries. JDK 13 and 18 build fails due to timeout unaccountably someday, it is tricky for us too.

daniellansun:
LGTM.

No worries. JDK 13 and 18 builds fail due to timeout unaccountably someday, it is tricky for us too.

daniellansun:
Merged. Thanks!

PR #2146, opened 2025-01-25
GROOVY-11560: Invalid compiler error for class which overrides a meth…

paulk-asert:
…od having duplicate default definitions from interfaces


daniellansun:
+1

PR #2152, opened 2025-02-15
GROOVY-11569: Create a sprintf shortcut for 1-arg usage

paulk-asert:


eric-milles:
The other 2 `sprintf` extensions don't make use of the "self" parameter.  It is confusing to add one that passes self along as the arg.  It almost makes more sense to add a `toString(String format)` extension.

Your example would be `assert sprintf('%4.2f', Math.PI) == '3.14'` with no changes.

paulk-asert:
It is very useful to have a variant that can be used in conjunction with the spread-dot operator.

```
def nums = (2..7).collect{ 1/it }
assert nums*.sprintf('%.2f') == ['0.50', '0.33', '0.25', '0.20', '0.17', '0.14']
```

eric-milles:
I understand the use case.  But the existing `sprintf` extensions don't work this way.  Adding a variant with much different usage is a problem IMO.  We're only saving a few chars versus:

```groovy
def nums = (2..7).collect { 1 / it }
def strs = nums.collect { sprintf('%.2f', it) }
assert strs == ['0.50', '0.33', '0.25', '0.20', '0.17', '0.14']
```

I was trying to think if `import static java.lang.String.format as sprintf` could replace the existing extension methods.  However, I see that they were added in 1.5.0 so I'm not sure if dropping them is a good plan.

paulk-asert:
Actually, I just remembered String::formatted. It allows:

```
def nums = (2..7).collect { 1 / it }
assert nums.collect('%.2f'::formatted) == ['0.50', '0.33', '0.25', '0.20', '0.17', '0.14']
```

Only for JDK15+, but I can live with that. I'll close the PR/issue.

PR #2154, opened 2025-02-25
GROOVY-11582: remove excess manifest attributes

eric-milles:
I suspect not all of these are still required.  This is a draft to explore removing all or some.  And for any kept, adding some docs and tests.

* `Ant-Version`
* `Bundle-ClassPath`
* `Created-By`
* `DynamicImport-Package`
* `Eclipse-BuddyPolicy`
* `Eclipse-ExtensibleAPI`
* `Extension-Name`
* `Tool`

### DIFF
--- a/build-logic/src/main/groovy/org.apache.groovy-core.gradle
+++ b/build-logic/src/main/groovy/org.apache.groovy-core.gradle
@@ -119,7 +119,7 @@ tasks.named('jarjar') { JarJarTask jjt ->
     jjt.withManifest {
         instruction '-nouses', 'true'
         instruction 'Export-Package', "*;version=${groovyBundleVersion}"
-        instruction 'Eclipse-ExtensibleAPI', 'true'
+        instruction 'Eclipse-ExtensibleAPI', 'true' // GROOVY-8713, GROOVY-11582
     }
 }
 
--- a/build-logic/src/main/groovy/org/apache/groovy/gradle/GroovyLibraryExtension.groovy
+++ b/build-logic/src/main/groovy/org/apache/groovy/gradle/GroovyLibraryExtension.groovy
@@ -211,8 +211,6 @@ class GroovyLibraryExtension {
                 'Bundle-Vendor'         : 'The Apache Software Foundation',
                 'Bundle-Version'        : groovyBundleVersion,
                 'Bundle-License'        : 'Apache-2.0',
-                'DynamicImport-Package' : '*',
-                'Eclipse-BuddyPolicy'   : 'dependent',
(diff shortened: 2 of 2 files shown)


eric-milles:
This is also a work-in-progress for dropping `Ant-Version` and `Created-By` from the jarjar output, which retains those attributes due to use of ant jar task to repackage the jarjar archive.

build-logic/src/main/groovy/org/apache/groovy/gradle/JarJarTask.groovy
```diff
@@ -169,16 +168,9 @@ class JarJarTask extends DefaultTask {
                 it.into(outputFile.parentFile)
                 it.rename { outputFile.name }
             }
-            project.ant.jar(destfile: outputFile, update: true, index: true, manifest: manifestFile, modificationtime: tstamp) {
-                manifest {
-                    // because we don't want to use JDK 1.8.0_91, we don't care and it will
-                    // introduce cache misses
-                    attribute(name: 'Created-By', value: 'Gradle')
```
paulk-asert, line 176: -1 on deleting this
MANIFEST.MF now contains something like:

> Created-By: 17.0.13+11-LTS (Azul Systems, Inc.)


Instead of:

> Created-By: Gradle

Which, as well as impacting Gradle cache, breaks our reproducible build goal. "Created-By" is already in the `excludedFromManifest` list, so I guess can't really be excluded, hence its forced value.

eric-milles: This should be all set now.  I can promote it from draft or you can do some investigation.  Just let me know.

paulk-asert: +1 for all refactoring changes and removing Ant-Version, Created-By, Extension-Name, Tool. My preference is to keep OSGi changes in a separate PR, ideally with some testing, but at least to make it easier/cleaner for manual testing.

paulk-asert:
I merged in the "pre-work" and minor refactoring. Do you want to rebase this against master? Should reduce to just a few lines.

eric-milles:
I have rebased, added comments to the `Eclipse-ExtensibleAPI` line, and moved `DynamicImport-Package` and `Eclipse-BuddyPolicy` to main jar only.  I should be able to dig up a jira ticket number for those two from the comments you left on GROOVY-11582.

I still think all 3 can go, but the urgency is not that great.

Just a note, I tried `Main-Class` in the `org.apache.groovy-main` plugin, but it did not end up at the end of the manifest, where I wanted it to be.  That is why I added a project name check to `GroovyLibraryExtension`.

PR #2155, opened 2025-02-27
GROOVY-11577: Add withIndex/indexed methods for ArrayGroovyMethods

paulk-asert:


src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
```diff
@@ -4526,7 +4541,26 @@ public static Map<Integer, Long> indexed(long[] self, int offset) {
     }
 
     /**
-     * Zips a double[] with indices in (index, value) order.
+     * Zips a float[] with indices in (index, value) order starting from a given index.
+     * <p/>
+     * Example usage:
+     * <pre class="groovyTestCase">
+     * double[] nums = [10.0f, 20.0f, 30.0f]
```
eric-milles, line 4548: should be `float[]`

paulk-asert: Good catch, fixed.

PR #2156, opened 2025-03-16
GROOVY-11513: include `import java.time.*` by default

eric-milles:
For your consideration.  Here are the types in `java.time` package:

image

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java
+++ b/src/main/java/org/codehaus/groovy/control/ResolveVisitor.java
@@ -95,9 +95,9 @@ import static org.codehaus.groovy.ast.tools.ClosureUtils.getParametersSafe;
  */
 public class ResolveVisitor extends ClassCodeExpressionTransformer {
     // note: BigInteger and BigDecimal are also imported by default
-    // `java.util` is used much frequently than other two java packages(`java.io` and `java.net`), so place java.util before the two packages
-    public static final String[] DEFAULT_IMPORTS = {"java.lang.", "java.util.", "java.io.", "java.net.", "groovy.lang.", "groovy.util."};
-    public static final String[] EMPTY_STRING_ARRAY = new String[0];
+    // `java.util` is used much frequently, so place it before `java.io`, et al.
+    public static final String[] DEFAULT_IMPORTS = {"java.lang.", "java.util.", "java.io.", "java.net.", "java.time.", "groovy.lang.", "groovy.util."};
+    public static final String[] EMPTY_STRING_ARRAY = {};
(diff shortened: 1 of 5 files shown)


daniellansun:
It's better not to break existing users' code unless we have to do.

codeconsole:
@daniellansun what exactly is breaking?

This commit brings consistency to existing legacy imports

https://groovy.apache.org/blog/groovy-dates-and-times-cheat

paulk-asert:
In general, it is domain classes that have the same name as the java.time classes and no package that catch us out.

E.g. for a `Main.groovy` file containing `println Period.ZERO` and a project containing this top-level class:
```
class Period {
	public static final ZERO = 'Nothing to see here'
}
```
Running `Main` currently outputs:

> Nothing to see here

But after this change will output (toString for java.time.Period#ZERO):

> P0D

I don't believe there is a problem for classes with a package since classes in the same package and references via existing imports would come before the default imports.

The other potentially problematic cases would be existing classes that are currently picked up by e.g. the groovy.util.* default import that might now come after java.time.*, but groovy.time.Duration (for instance) isn't a default import, so I believe we are good there. If folks have not followed split packaging naming conventions and have gone and created their own groovy.util packaged classes, maybe wrapping joda time or some such, with names like Month, Year, etc., they would be impacted. We don't prohibit that in the compiler currently but I would think it rare and maybe something we wouldn't care strongly about.

codeconsole:
Well, I have found, as is, unintentionally encourages deprecated behavior in domain classes due to the added complexity of having to add imports in view layers. Even in standalone applications and scripts, I find a bias to utilizing Date over LocalDate to avoid having to add redundant imports.

It just seems logical to automatically support modern Data/Time usage by importing the java.time package.  Working with Date's is such a common task, not having comparable support for LocalDate and LocalDateTime can be very limiting.

paulk-asert:
@codeconsole I agree this would be nice to have. We just have to weigh up all pros and cons. We treat breaking existing code seriously and such changes proceed more slowly due to the extra caution. We sometimes do it anyway when it is a clear win. We know in such cases that we have to be extra clear in our release notes and messaging.

The particular aspect of breaking domain classes with no package is one that has bitten us before. I think it was in fact with Grails. Some folks had no package domain classes named Singleton and we introduced an annotation in groovy.lang which is auto-imported. (It was made worse by various other inject Singleton annotations too.)

@eric-milles Are you still in favor? Did you want to do a VOTE/DISCUSS thread on the mailing list?

paulk-asert:
I'll create a thread on the mailing lists. If we were going to make such a change, Groovy 5 would be an ideal time.

PR #2157, opened 2025-03-16
GROOVY-11486: rework `flatten` type parameters, similar to `collectMany`

eric-milles:
I tried to follow `collectMany` in terms of generics.  There is no connection that can be made between the in type and the out type, since aggregates are recursively processed.  And the closure accepting forms have weaker guarantees for the parameter type, so I added `@ClosureParams` just to indicate the number of parameters.

https://issues.apache.org/jira/browse/GROOVY-11486


daniellansun:
Merged. Thanks

PR #2158, opened 2025-03-21
GROOVY-11568: try `invokeMethod(Class,Object,Object[],boolean,boolean)`

eric-milles:
When MethodSelector#setMetaClassCallHandleIfNeeded links to MetaObjectProtocol#invokeMethod(Object,String,Object[]), the sender class is dropped.


daniellansun:
Merged. Thanks.

PR #2160, opened 2025-03-21
Gradle 8.13

eric-milles:
Need 8.14 for Java 24 compatibility


paulk-asert:
For 24 support, we could use toolchains for main source (would need some work in the build files). For the CI tests, we can use the `-Ptarget.java.home` property as per existing `ea` steps.

PR #2162, opened 2025-03-23
build(console): move `groovy-swing` to API scope

matrei:
`SwingBuilder` is part of `groovy-console`'s public API, so it must be in the `api` configuration scope. This prevents a downstream `NoClassDefFoundError: groovy.swing.SwingBuilder` at compile time.


matrei:
Will create PR against master branch instead

PR #2169, opened 2025-03-30
GROOVY-10683: support `for (index, item in list)` looping

eric-milles:
This is a total experiment and a work-in-progress.  I don't know quite how to represent the extra variable/parameter.  `ForStatement` is already a bit of a mess in how it supports both `for(int i = 0; i < n; ++i)`, `for (var item : list)` and `for(item in list)`.


daniellansun:
How about `for (int index, var item : items)`?
For example:
```groovy
// ① traverse a list
List<String> names = Arrays.asList("Alice", "Bob", "Charlie");
for (int i, String name : names) {
    System.out.println(i + ": " + name.toUpperCase());
}

// ② traverse a matrix
int[][] matrix = {{1,2}, {3,4}};
for (int row, int[] cols : matrix) {
    for (int col, int val : cols) {
        System.out.printf("[%d,%d]=%d\n", row, col, val);
    }
}

```

blackdrag:
Daniel's variant is very near to what I know from Python:

    for idx,data in enumerate(content_list):
        print(idx) # will print the index of the loop starting from 0
        print(data) # will print the actual data,
        print(content_list[idx]) # will manually access the list using the index. both this and data will  print the same thing

Which means "for (idx,data in list)" and "for (idx,data : list)" would also be needed besides the typed variants.

glaforge:
I like the Python approach suggested by @blackdrag, it's quite natural (and certainly familiar for those who have done some Python)

eric-milles:
Sorry, "I don't know quite how to represent the extra variable/parameter" was not intended as a discussion point for the syntax.  After the syntax is parsed, I have to solve the AST representation.

If you don't like the syntax, that's fine.  As I said, "[t]his is a total experiment".  The JIRA ticket sat for nearly 3 years without comment and I thought I had enough stuff in place to pull off a proof-of-concept.

eric-milles:
I'm going to push just the clean ups as a commit that can be reviewed and cherry-picked on its own.  The `for (int i, var e in x)` makes sense as simple and understandable.  I should be able to put together the parser changes for it and push that as well.

eric-milles:
This is functional enough for an initial review.

daniellansun:
LGTM

paulk-asert (approved):
I like the individual commits - much easier to see the "non-refactoring" changes.

Not critical, but would be nice to support also "long" as well as "int/var/def". Primitive int would make a good default if we did support long. I note we don't support "Integer". I think that is okay.

Not critical, but would be nice for a specific error message where an initValue is supplied (currently unsupported) for the index value:

```
for (int i = 3, j in 1..5) {
  println "$i:$j"
}
```

Unrelated, but I noticed we support this:

```
for (def ij : (1..5).withIndex()) {
  (i, j) = ij
  println "$j:$i"
}

```
but not this:
```
for (def (i, j) : (1..5).withIndex()) {
  println "$j:$i"
}

``` 
We could consider adding support (in a separate issue) but do you think the current changes would make supporting this any harder?

eric-milles:
Supporting `long` would not be a big step from what is there -- grammar support is done.  I can't envision the case where you would want linear iteration over that many elements.  Maybe a stream where you skipped a lot of indices?  You can open a new ticket that links back to GROOVY-10683 for this proposal.

There is no specific error on `for (int i = 0, var x : y)` or `for (i, x = 0 : y)`.  I tried to keep the grammar as short as possible since the classic for loop often starts with "int name".  In general we have a problem of "unexpected input" for out-of-bounds syntax.  [GROOVY-9192](https://issues.apache.org/jira/browse/GROOVY-9192) speaks to this issue.

I'm checking the subprojects for references to `ForStatement` in case there are any adjustments required.

eric-milles:
Spock's `spock.util.AstNodeToScriptVisitor#visitForLoop` will need a similar change to Groovy's `groovy.console.ui.AstNodeToScriptVisitor#visitForLoop`.

PR #2171, opened 2025-04-02
Update SDKMAN install URL

roulpriya:
Update SDKMAN URL to include `"https://get.sdkman.io"` as the old link does not work.


daniellansun:
Merged. Thanks!

PR #2172, opened 2025-04-02
GROOVY-4721: variable declared in try block is in scope in finally block

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-4721


eric-milles (approved):
Seems reasonable.  Nice compact change in `StatementWriter`.

subprojects/groovy-groovysh/src/test/groovy/org/apache/groovy/groovysh/GroovyshTest.groovy
```diff
@@ -392,17 +394,34 @@ class GroovyshInterpreterModeTest extends GroovyshTest {
...
     }
 
+    // collecting variables relies on GROOVY-4721, e.g. accessing variable declared in try block is in scope in finally block
+    // but after GROOVY-4721 is fixed, this test will fail, but it is aligned with the behavior of the groovySh:
+    // groovy:000> int x, y, z
+    // ===> 0
+    // groovy:000> y
+    // Unknown property: y
+    // groovy:000>
+    @NotYetImplemented
```
paulk-asert, line 424: This isn't the existing behavior of groovysh with interpreterMode set.

> groovy:000> :set interpreterMode
> groovy:000> int x = 3
> ===> 3
> groovy:000> x
> ===> 3
> groovy:000> int a, b, c
> ===> 0
> groovy:000> b
> ===> 0

daniellansun: https://github.com/apache/groovy/blob/87390561659f50e5a117bebb697731e8666f97b3/subprojects/groovy-groovysh/src/main/groovy/org/apache/groovy/groovysh/Groovysh.groovy#L294

Groovysh will try to collect variables in the `finally` block, but variables declared in `try` block can not be accessed in `finally` after GROOVY-4721 is fixed.

paulk-asert: I added the change to StatementWriter and changed groovysh to just have the "variableBlocks" execute directly after "current" (i.e. just removed the try...finally) then interpreterMode still seemed to work. The current test suite and the new tests all passed. There could be variations that we don't have coverage for in our test suite that may become broken with that change.

daniellansun: We can wrap the `current` with `try-finally` to guarantee the variables will be collected eventually.

PR #2174, opened 2025-04-04
GROOVY-11601: remove support for expression list in `for`-each statement

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-11601


daniellansun:
It's better not to break existing users' code unless we have to do.

eric-milles:
Do you have an example of something that requires this?  It is quite costly to support arbitrary expression list in for loop.

eric-milles:
sorry, I keep misinterpreting the icon in the branches list

PR #2175, opened 2025-04-04
GROOVY-11597: STC: loop over enum class

eric-milles:
`for` loop item type for `Class<EnumType>` is `EnumType`


daniellansun:
Merged. Thanks.

PR #2177, opened 2025-04-05
GROOVY-11229: limit scope of `instanceof` variable

eric-milles:
Support this common case in dynamic mode:
```groovy
def x = null
if (x instanceof String s) {
  // ...
} else {
  s // no such property
}
```


daniellansun:
LGTM

paulk-asert (approved):
LGTM

daniellansun:
Merged. Thanks.

PR #2182, opened 2025-04-09
GROOVY-11606: Lazy findAll, collect, collectMany

paulk-asert:


eric-milles:
I have started to look this over.  I am trying to frame up my understanding of the "collect" semantic.  And since there is a simple "collect(Iterator): List" that is basically the same as "toList(Iterator)", my understanding is that "collect" gathers up elements and produces a collection.  With that understanding, a "collectLazy" that produces an iterator does not fit IMO.

My initial take is that the Stream API offers all of this -- I think -- and the difference between ".collectLazy(" and ".stream().map(" is not that much in terms of character savings.  Is the idea here that an iterator chain is more lightweight compared to a stream pipeline?  If yes, I'm trying to frame up the possibility of adding "map", "filter" and one or more terminal operations (possibly existing) to support "aggregate.iterator().map(transform).filter(predicate).toList()" or just "iterator.map(transform).filter(predicate).toList()".

I will probably need to go through each example given in the [blog post](https://groovy.apache.org/blog/exploring-gatherers4j) to understand what is missing and what is proposed to fill the gap.

See also: https://groovy.apache.org/blog/groovy-gatherers

eric-milles:
I'm also trying to take in [GROOVY-11603](https://issues.apache.org/jira/browse/GROOVY-11603), which added some factory methods for an iterator.  I seem to recall Stream added such things post Java 8.  And I may have been expecting them as DSGM extensions rather than new `groovy.util.` classes, which are default imported.

paulk-asert:
The Smalltalk semantics for "collect" is transform. Think: I have an aggregate of People and I'll produce an aggregate of Names by "collect"ing each person's name. The "into a collection" is definitely true when we supply a collector but isn't a required interpretation when the incoming aggregate isn't a collection. Currently though, if the incoming aggregate is Set or array, we produce List as the output aggregate (though collectNested preserves the Set). The collect using the identity transform is a bit of an anomaly. Having some "map" variants certainly let's us not break compatibility if we don't mind the extra variants.

PR #2186, opened 2025-04-12
GROOVY-11609: Avoid finding same variable/class member repeatedly

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-11609

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/VariableScopeVisitor.java
+++ b/src/main/java/org/codehaus/groovy/classgen/VariableScopeVisitor.java
@@ -64,8 +64,11 @@ import org.codehaus.groovy.control.SourceUnit;
 import org.codehaus.groovy.syntax.Types;
 
 import java.util.Deque;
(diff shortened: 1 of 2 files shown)


eric-milles:
Do you have something to measure the performance difference?  We have nothing to weigh the complexity of the changes against.

Even if complex, it is compile-time behavior.  I would consider that lower priority when compared to reports of runtime performance differences between 2.5 and 3+ (indy and parser).

blackdrag:
@daniellansun   there are already  3 maps/caches in VariableScope. Why are those no good?

daniellansun:
> @daniellansun there are already 3 maps/caches in VariableScope. Why are those no good?

Finding logic will still be triggered for each same variables/class members in the VariableScopeVisitor. It's similar to difference between application cache and database cache.

blackdrag:
@daniellansun I think that if the logic tries to find duplicates, then because those three caches are not utilized properly.

PR #2187, opened 2025-04-12
GROOVY-11607: `groovyc`: `fileset` lists files

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-11607


daniellansun:
+1

PR #2189, opened 2025-04-13
GROOVY-11613: inner class of interface with default methods is static

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-11613
https://issues.apache.org/jira/browse/GROOVY-11600


daniellansun:
LGTM

PR #2190, opened 2025-04-13
GROOVY-10904, GROOVY-11618: reference to inner class property method

eric-milles:
https://issues.apache.org/jira/browse/GROOVY-11618
https://issues.apache.org/jira/browse/GROOVY-10904
https://issues.apache.org/jira/browse/GROOVY-10741


daniellansun:
LGTM

PR #2192, opened 2025-04-14
GROOVY-11611: outer class super method invocation

eric-milles:
```groovy
class A {
  String value() { 'A' }
}
class B extends A {
  String value() { 'B' }
  /*non-static*/ class C {
    def m() {
      B.super.value() // this$0 super call
    }
  }
}
```


daniellansun:
LGTM

PR #2194, opened 2025-04-15
use src/test/groovy instead of src/test

paulk-asert:
Moves our tests into the standard "src/test/groovy" directory.


daniellansun:
+1

daniellansun:
Merged. Thanks.

PR #2195, opened 2025-04-15
GROOVY-11637: Gradle 8.14

eric-milles:
supports Java 24

see #2160

https://issues.apache.org/jira/browse/GROOVY-11637


daniellansun:
+1

paulk-asert:
I am -0 for this change just now. I think it is good (so great to have this branch) to test that we can move to 8.14 when it comes out but I'd prefer to wait until we are on the final version not the RCs. Unless there is some compelling reason that 8.14 offers us. Each gradle change seems to bring additional gradle deprecations and it's work to hush them each time.

daniellansun:
FYI.

Gradle 8.14 is available:
https://docs.gradle.org/current/release-notes.html

daniellansun:
+1

PR #2197, opened 2025-04-16
GROOVY-11621: STC: fix for `list[idx] = null` and `map[key] = null`

eric-milles:
Special checks for `putAt` lack `null` handling.  Adding "isNullConstant(enclosingExpressionRHS) ? UNKNOWN_PARAMETER_TYPE :" to `getType` would expose to all STC users/scripts.


daniellansun:
+1

PR #2198, opened 2025-04-16
GROOVY-11623: STC: support return of list expression for array method

eric-milles:
```groovy
String[] foo() {
    ['a','b','c' ]
}
```
and
```groovy
@interface A {
  String[] value() default []
}
```


daniellansun:
Out of curiosity, will multiple dimension array be supported too? For example,
```groovy
String[][] foo() {
    [['a','b'], ['c', 'd']]
}
```

eric-milles:
It supports whatever the dynamic mode conversion is.  I have not tried with multi-dimensional, but I expect that it works.

daniellansun:
Great. Would you like to add some test cases for multi-dimensional array?

daniellansun:
Merged. Thanks.

PR #2199, opened 2025-04-17
GROOVY-11614: Enums in switch/case statements that are not fully qual…

paulk-asert:
…ified will cause a groovy compile error for switch expressions even with type checking enabled

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/CompilationUnit.java
+++ b/src/main/java/org/codehaus/groovy/control/CompilationUnit.java
@@ -30,7 +30,9 @@ import org.codehaus.groovy.ast.CompileUnit;
 import org.codehaus.groovy.ast.GroovyClassVisitor;
 import org.codehaus.groovy.ast.InnerClassNode;
 import org.codehaus.groovy.ast.ModuleNode;
+import org.codehaus.groovy.ast.expr.ClosureExpression;
 import org.codehaus.groovy.ast.expr.Expression;
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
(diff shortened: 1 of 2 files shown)


paulk-asert:
> Can you explain the design of this one? I don't quite follow. I think this plays into the reason we limited the previous fix to STC.

This just applies the previous fix (so only STC) to Switch Expressions as well as Switch Statements. The ClassCodeExpressionTransformer for the previous fix doesn't normally traverse deeply. We "create" a Switch Expression by wrapping a Switch Statement into a Closure. The change here just causes it to traverse deeper into the wrapped code. I did try adding some additional marker metadata to make traversal even more selective, but it was transformed away before it gets to where we'd look it up. We could try to track down where that happens and make it preserve the metadata, but looking inside for the Closure expression should eliminate many traversal paths.

daniellansun:
LGTM.

P.S. It's better to add some marker to limit visiting further.

eric-milles:
Is it possible for the switch expression closure to delegate to the enum class?  Would that be safe to do?

daniellansun:
If switch expression closure delegates to the enum class, switch expression and switch statement will be different. As a result, it's hard for users to understand if they do not know the delegating.

eric-milles:
They are already different.  Switch statement supports unqualified enum constants under STC.  Switch expression does not.  The expression case could be made to work with delegation, in my estimation.  It could be limited to STC as well for the initial cut.  Any other differences in behavior could be listed.

daniellansun:
If delegation applied, method calls within switch expression can also be delegated to enum class. But switch statement treats the method calls as invalid.

daniellansun:
Merged. Thanks.

PR #2203, opened 2025-04-21
GROOVY-11630: optimize `void` groovy method return in expression statement

eric-milles:
consolidate metadata keys to shared value


paulk-asert:
LGTM;

PR #2204, opened 2025-04-24
GROOVY-11636: Create a SGM#next method which also takes an integer in…

paulk-asert:
…crement value


eric-milles:
@blackdrag  raised concern that `charAt` should probably be code point at.  And is there a next code point method?  Any concern about invalid/unmapped code points?

paulk-asert:
I commented in the issue. I was following the code from next()/previous().

PR #2205, opened 2025-04-26
GROOVY-11639: include super interface fields in face's property index

eric-milles:
Support interface default method making reference to super interface constant.


daniellansun:
LGTM

PR #2206, opened 2025-04-27
GROOVY-11641: index trait field under original name

eric-milles:
It is possible this could be solved in the trait transformation or receiver transformer.  At the time of xform, the fields of super trait have already been moved to field helper, so it would be some work to locate and decode them.


daniellansun:
LGTM

daniellansun:
Merged. Thanks.

PR #2210, opened 2025-05-03
GROOVY-11649: add partitionPoint to Groovy Array and List(missed in java and groovy)

dongwq:
partition_point missed in java and groovy like in
* rust: https://doc.rust-lang.org/std/primitive.slice.html#method.partition_point
* c++:  https://en.cppreference.com/w/cpp/algorithm/ranges/partition_point

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
@@ -87,6 +87,9 @@ import java.util.function.IntConsumer;
 import java.util.function.IntUnaryOperator;
 import java.util.function.LongConsumer;
 import java.util.function.LongUnaryOperator;
+import java.util.function.IntPredicate;
+import java.util.function.LongPredicate;
+import java.util.function.DoublePredicate;
 
 /**
  * Defines new groovy methods which appear on arrays inside the Groovy environment.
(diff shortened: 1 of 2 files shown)


paulk-asert:
I created this issue to track this potential enhancement:
https://issues.apache.org/jira/browse/GROOVY-11649

paulk-asert:
This is an interesting proposal.

One immediate point of feedback is I would change the IntRange semantics to follow normal Groovy idioms. I know that means it might be slightly different to rust and other languages, but it follows what we do for all other Groovy methods. So, basically you'd get rid of the "- 1" when accessing getToInt() and change your examples from "0..arr.size()" to "0..<arr.size()" or "0..-1". Groovy let's you pick between (l, r), [l, r), [l, r] etc by using 0<..<4, 1..<4, 1..3 etc. I also haven't checked yet whether reverse ranges are handled properly, but perhaps you've already considered that.

Also, Java doesn't have a sorted list but offers Collections.sort(). Groovy doesn't have partition (currently) but does have groupBy. So, I think it makes sense to discuss in more detail how this would typically be used. Is it less useful than in other languages given the JDK collection semantics or is other functionality needed (like a partition extension method) to reap the full benefits. I think discussing such things is best on the mailing list or the issue.

src/main/java/org/codehaus/groovy/runtime/ArrayGroovyMethods.java
```diff
@@ -6576,6 +6579,528 @@ public static <T> T[] minus(T[] self, Object removeMe) {
...
 
+    //--------------------------------------------------------------------------
+    // partition_point
+
+    /**
+     * Returns the index of the partition point according to the given predicate
+     * (the index of the first element of the second partition).
+     * The arr is assumed to be partitioned according to the given predicate.
+     * <pre class="groovyTestCase">
+     * def arr = [7, 15, 3, 5, 4, 12, 6] as Integer[];
```
paulk-asert, line 6590: I'd probably also get rid of the unneeded semicolon here

dongwq: > I'd probably also get rid of the unneeded semicolon here (and other places)

thank you, done

dongwq:
> This is an interesting proposal.
> 
> One immediate point of feedback is I would change the IntRange semantics to follow normal Groovy idioms. I know that means it might be slightly different to rust and other languages, but it follows what we do for all other Groovy methods. So, basically you'd get rid of the "- 1" when accessing getToInt() and change your examples from "0..arr.size()" to "0..<arr.size()" or "0..-1". Groovy let's you pick between (l, r), [l, r), [l, r] etc by using 0<..<4, 1..<4, 1..3 etc. I also haven't checked yet whether reverse ranges are handled properly, but perhaps you've already considered that.
> 
> Also, Java doesn't have a sorted list but offers Collections.sort(). Groovy doesn't have partition (currently) but does have groupBy. So, I think it makes sense to discuss in more detail how this would typically be used. Is it less useful than in other languages given the JDK collection semantics or is other functionality needed (like a partition extension method) to reap the full benefits. I think discussing such things is best on the mailing list or the issue.

@paulk-asert  Thank you 
- the range arg: have adjusted to normal Groovy idioms.
- reverse ranges: with an assert check
- the usage: here partition_point can be view as a modern/general vision api of binary search with the help of closure/lambda. old binarySearch in java , lack of the ability to process these cases:
   1. directly provide the postion for insert like: lowerBound/uppperBound 
   2. the result of ordered arr/list with one val occured more than once is uncertain.
   
after all, so I change the method name to binaryPartitionPoint to be more precise.

paulk-asert:
I actually think i prefer just `partitionPoint` to match Rust and C++, but let me put my thoughts into GROOVY-11649 because I am not sure all of the variants are equally compelling.

paulk-asert:
Also, was there anything in particular you used from `AbstractList`. Wondering why not to use the more general `List` interface.

dongwq:
> Also, was there anything in particular you used from `AbstractList`. Wondering why not to use the more general `List` interface?

cause partition_point is O(logn),  suppose that element.get(i) is O(1).  so here we need a "random access" data store (such as an array).  

the java doc says for AbstractList:
 * This class provides a skeletal implementation of the {@link List}
 * interface to minimize the effort required to implement this interface
 * backed by a "random access" data store (such as an array)
 
though there no actual limit in java collection framework.

paulk-asert:
I was thinking a bit more about usage overnight.

For arrays, the "clean insertion" point argument is less compelling but somewhat useful none-the-less.

Given that it isn't easy to partition primitive arrays, I think partitionPoint is slightly less compelling. We could get by with both bisectLeft and bisectRight for instance. But even for natural ordering scenarios, the convenience of a single method is nicer than the two separate methods in my view.

I looked at the grades example from Python bisect: https://docs.python.org/3/library/bisect.html#examples

```
def scores = [33, 99, 77, 70, 89, 90, 100] // convert these to A grade (90 and above),
                                           // B grade (80 up to but not including 90), etc.
```

With binary search we have something like this:

```
def grade(score) {
  int[] breakpoints = [59, 69, 79, 89]
  def idx = Arrays.binarySearch(breakpoints, score)
  if (idx < 0) {
    idx = (idx + 1).abs()
  }
  'FDCBA'[idx]
}
assert scores.collect(this.&grade) == ['F', 'A', 'C', 'C', 'B', 'A', 'A']
```

We have to artificially subtract one from each of the breakpoints compared to the Python bisect example.
 
With partitionPoint it would become:

```
int[] breakpoints = [60, 70, 80, 90]
assert scores2.collect { sc ->
    'FDCBA'[breakpoints.partitionPoint{ it <= sc }]
} == ['F', 'A', 'C', 'C', 'B', 'A', 'A']

```
This is much cleaner.

paulk-asert:
> > Also, was there anything in particular you used from `AbstractList`. Wondering why not to use the more general `List` interface?
> 
> cause partition_point is O(logn), suppose that element.get(i) is O(1). so here we need a "random access" data store (such as an array).
> 
> the java doc says for AbstractList:
> 
> * This class provides a skeletal implementation of the {@link List}
> * interface to minimize the effort required to implement this interface
> * backed by a "random access" data store (such as an array)
> 
> though there no actual limit in java collection framework.

In general, we would use List here. That makes it work for all lists. The default is ArrayList which should have good performance but if the user wishes to supply some other kind of list, they will be responsible for ensuring they pick a list matching their performance requirements.

paulk-asert:
Can you rename back to partitionPoint, and use List instead of AbstractList. I am still pondering the benefits of Predicate vs Closure for the DGM List and AGM T[] variants. Did you make them different for a reason?

dongwq:
> Can you rename back to partitionPoint, and use List instead of AbstractList. I am still pondering the benefits of Predicate vs Closure for the DGM List and AGM T[] variants. Did you make them different for a reason?

@paulk-asert ok, 3 have done:
- revert method name to partitionPoint, 
- prefer Predicate<T> to Closure<?>, 
- DGM prefer List to AbstractList

paulk-asert:
I am not sure if "prefer Predicate to Closure<?>" is correct yet. I think that will be the case for primitive arrays, still thinking about List and T[]. It might be best to align those with how we'd sort those two things but I am just thinking through the cases whether it turns out to be the same. I'll writer another comment in a few hours.

paulk-asert:
Predicate looks fine in all cases.

paulk-asert:
Merged! Thanks!

paulk-asert:
Apologies, I meant to squash when merging.

PR #2211, opened 2025-05-03
GROOVY-11642: include annotations and modifiers in inner class offset

eric-milles:
inner class offset like outer class offset


eric-milles:
I checked field, method and compact constructor.  The first two already include annotations and modifiers in their source offset and the compact constructor is mapped to a closure in the `TupleConstructor` `pre` attribute.  So this is ready to merge.

PR #2215, opened 2025-05-06
GROOVY-11655: Create extension method to make StringBuilder like a stack

dongwq:
see: 
https://issues.apache.org/jira/browse/GROOVY-11655

### DIFF
--- a/src/test/groovy/org/codehaus/groovy/transform/DelegateTransformTest.groovy
+++ b/src/test/groovy/org/codehaus/groovy/transform/DelegateTransformTest.groovy
@@ -72,7 +72,7 @@ final class DelegateTransformTest {
             }
             new C(numbers:[1,2,3])
         '''
-        assert err =~ /The return type of java.lang.Number remove\(int\) in C is incompatible with java.lang.String in java.util.ArrayList/
+        assert err =~ /The return type of java.lang.Number get\(int\) in C is incompatible with java.lang.String in java.util.ArrayList/
 
         err = shouldFail '''
             class C extends ArrayList<String> {


paulk-asert:
Nice idea, but personally, I am unsure whether it is compelling enough to be added to Groovy. I will provide some comments anyway.

src/main/java/org/codehaus/groovy/runtime/StringGroovyMethods.java
```diff
@@ -1877,6 +1878,77 @@ public static StringBuilder leftShift(final StringBuilder self, final Object val
...
+        return self.charAt(self.length() - 1);
+    }
+
+    public static char pop(final StringBuilder self) {
+        char last = peek(self);
+        self.setLength(self.length() - 1);
+        return last;
+    }
+
+    public static int size(final StringBuilder self) {
```
paulk-asert, line 1947: This is provided by #size(CharSequence) already.

src/main/java/org/codehaus/groovy/runtime/StringGroovyMethods.java
```diff
@@ -1877,6 +1878,77 @@ public static StringBuilder leftShift(final StringBuilder self, final Object val
...
+     * @return the StringBuilder obj self
+     * @since 5.0.0
+     */
+    public static StringBuilder push(final StringBuilder self, final char ch) {
+        Objects.requireNonNull(self);
+        self.append(ch);
+        return self;
+    }
+
+    public static StringBuilder push(final StringBuilder self, final String ch) {
```
paulk-asert, line 1923: This and the next method could be replaced by a single CharSequence version.

src/main/java/org/codehaus/groovy/runtime/StringGroovyMethods.java
```diff
@@ -1877,6 +1878,77 @@ public static StringBuilder leftShift(final StringBuilder self, final Object val
...
+     * assert st.isEmpty() == false
+     * // stack out cases:
+     * assert st.peek() == 'r' as char
+     * assert st.pop() == 'r' as char
+     * assert st.toString() == 'Ba'
+     * assert st.peek() == 'a' as char
+     * assert st.pop() == 'a' as char
+     * assert st.toString() == 'B'
+     * assert st.size() == 1
+     * assert st.isEmpty() == false
```
paulk-asert, line 1907: StringBuilder#isEmpty comes with JDK15.

eric-milles:
I don't think there is much general value in providing "stack of chars" semantic.  You can always provide these for your own projects through extension or categroy class.

dongwq:
> I don't think there is much general value in providing "stack of chars" semantic. You can always provide these for your own projects through extension or category class.

@eric-milles ,
the issue try to solve the basic Use Case String:
-  Useful in algorithms where you need to modify strings dynamically, such as backtracking or parsing tasks.

given some examples code here: https://issues.apache.org/jira/browse/GROOVY-11655

src/main/java/org/codehaus/groovy/runtime/StringGroovyMethods.java
dongwq: > Since Groovy 5 is JDK11+, I would recommend removing isEmpty() calls from the javadoc test. They could be placed in separate test with an appropriate version guard.

@paulk-asert 
maybe this is off the issue.  
since JDK17 is  almost 4 years old.  
wish Groovy 5  comes out with JDK17+!!!

PR #2222, opened 2025-05-13
GROOVY-11658: `groovyc`: log the command-line arguments at verbose level

eric-milles:
verbose is between info and debug


perNyfelt (commented):
Nice solution!

paulk-asert:
+1

PR #2229, opened 2025-05-21
GROOVY-11674: Grails reproducible build

paulk-asert:
One way to force deterministic behavior is to track the order of definitions in trait source files and propogate those to the interface and helper classes.


eric-milles:
What is the source of the non-deterministic order?  Members should be ordered by declaration order from the source.  From there are the helper classes processing in that order?

paulk-asert:
As slightly covered in my email, the order of members using javap on the interface and helper classes is as expected, i.e. in declaration order. The order returned when calling `Class#getMethods()` in `TraitComposer` when processing pre-compiled trait files is non-deterministic which I find unexpected but as per the javadoc for that method is not guaranteed.

paulk-asert:
It may be possible to look at a different way to do the ordering. We have some assumptions about ordering in some places, e.g.:
```
int z = 0
int a = z
```
We don't want the order of those two to be swapped in generated initializer code.

paulk-asert:
Actually, I redid sorting based on just the descriptor and that seems to be enough. I must have had some other changes in play in my previous testing.

PR #2230, opened 2025-05-22
GROOVY-11664: STC: combine generics of same-type witnesses

eric-milles:
Merging of type witness data -- see GROOVY-5692 and GROOVY-10006 -- was not applied to equal/compatible type case.


daniellansun:
Merged. Thanks.

PR #2231, opened 2025-05-23
Groovy11674b

paulk-asert:


eric-milles:
The list returned by `ClassNode#getMethods` is an internal list not a copy.  It is probably not safe to sort this without making a copy.

paulk-asert:
@eric-milles Good catch. I am also going to check whether something in the grails codebase is doing such a sort.

PR #2232, opened 2025-05-24
GROOVY-11676: object expression type in `MethodClosure` error message

eric-milles:
This is for the dynamic path


sbglasius (approved):
+1

PR #2242, opened 2025-05-30
misc: minor documentation edits

jonnybot0:
Was reading through [the guide on Type Checking Extensions](https://docs.groovy-lang.org/docs/latest/html/documentation/#_type_checking_extensions) today and my picky brain tripped over what seemed like a few typos. Figured I'd put up a small PR to fix them while I was here.

There is one spot where I've inferred a meaning that may be incorrect. I rewrote

> but also all support for typos at compile time!

to be

> but also all support for catching typos at compile time!

I think that's right, but am sophomoric at best when it comes to understanding type checking extensions. Just @me if that's wrong and I'll put it back.


paulk-asert:
I like your suggestions but I think it could be further improved. I'll merge and tweak if I can think of better wording.

PR #2244, opened 2025-06-01
GROOVY-11688: Potential minor improvement to STC error message

paulk-asert:


paulk-asert:
I rejigged to carry a little more info in the `entries` data structure which meant less work to get to the correct ASTNode.

PR #2245, opened 2025-06-01
GROOVY-11689: Grails reproducible build (fix @Delegate method ordering)

jdaugherty:
[GROOVY-11689](https://issues.apache.org/jira/browse/GROOVY-11689) - Follow-up to https://github.com/apache/groovy/pull/2231 - fix `@Delegate` generating inconsistent ordering of added methods.

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/DelegateASTTransformation.java
+++ b/src/main/java/org/codehaus/groovy/transform/DelegateASTTransformation.java
@@ -21,6 +21,7 @@ package org.codehaus.groovy.transform;
 import groovy.lang.Delegate;
 import groovy.lang.Lazy;
 import groovy.lang.Reference;
+import org.apache.groovy.ast.tools.MethodNodeUtils;
 import org.codehaus.groovy.ast.ASTNode;
 import org.codehaus.groovy.ast.AnnotatedNode;
 import org.codehaus.groovy.ast.AnnotationNode;
@@ -40,6 +41,7 @@ import org.codehaus.groovy.control.SourceUnit;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.ListIterator;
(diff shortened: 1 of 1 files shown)


daniellansun:
@paulk-asert has fixed the ordering issue at the commit:
https://github.com/apache/groovy/commit/e42a2a7dc3973a9c4007c4247e58360e95d02e5c

Thanks for your contribution all the same.

src/main/java/org/codehaus/groovy/transform/DelegateASTTransformation.java
```diff
@@ -215,7 +220,11 @@ public void visit(final ASTNode[] nodes, final SourceUnit source) {
     }
 
     private static Collection<MethodNode> collectMethods(final ClassNode type) {
-        List<MethodNode> methods = new java.util.LinkedList<>(getAllMethods(type));
+        List<MethodNode> allMethods = new ArrayList<>(getAllMethods(type));
+        if (!allMethods.isEmpty()) {
+            allMethods.sort(Comparator.comparing(MethodNodeUtils::methodDescriptorWithoutReturnType));
+        }
+        List<MethodNode> methods = new java.util.LinkedList<>(allMethods);
```
daniellansun: Why should we create `java.util.LinkedList` instance here? We have a sorted `allMethods` already, maybe we could rename `allMethods` to `methods`.

jdaugherty: That's a great observation.  I'll update this to simplify it.  Thank you!

daniellansun:
> @daniellansun Hi Daniel, this is a separate issue that's fixing the `@Delegate` in the same way as that one.

OK. I see. 
I left a comment for the PR. Please take a look when you have some spare time.

src/main/java/org/codehaus/groovy/transform/DelegateASTTransformation.java
```diff
@@ -215,7 +220,10 @@ public void visit(final ASTNode[] nodes, final SourceUnit source) {
     }
 
     private static Collection<MethodNode> collectMethods(final ClassNode type) {
-        List<MethodNode> methods = new java.util.LinkedList<>(getAllMethods(type));
+        List<MethodNode> methods = new ArrayList<>(getAllMethods(type));
+        if (!methods.isEmpty()) {
+            methods.sort(Comparator.comparing(MethodNodeUtils::methodDescriptorWithoutReturnType));
+        }
```
eric-milles: probably want to sort at the end since more methods are added at the bottom

jdaugherty: I've moved the sort to the end of the method, great catch!

src/main/java/org/codehaus/groovy/transform/DelegateASTTransformation.java
```diff
@@ -165,7 +167,10 @@ public void visit(final ASTNode[] nodes, final SourceUnit source) {
             if (!checkPropertyOrMethodList(delegate.type, delegate.excludes, "excludes", node, MY_TYPE_NAME)) return;
             if (!checkPropertyOrMethodList(delegate.type, delegate.includes, "includes", node, MY_TYPE_NAME)) return;
 
-            final Iterable<MethodNode> ownerMethods = getAllMethods(delegate.owner);
+            final List<MethodNode> ownerMethods = new ArrayList<>(getAllMethods(delegate.owner));
```
paulk-asert: getAllMethods creates a new list, so you shouldn't need the outer ArrayList (here and below).

jdaugherty: new ArrayList<>() is removed.

eric-milles:
You only really need to sort if size > 1.  But otherwise looks fine.

Is there a way to test this?

PR #2247, opened 2025-06-04
GROOVY-11692: Add a Stream#size extension method that is an alias for…

paulk-asert:
… Stream#count

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/StreamGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/StreamGroovyMethods.java
@@ -568,6 +568,89 @@ public class StreamGroovyMethods {
         return Arrays.stream(self);
     }
 
+    /**
+     * An alias for {@code count}. Returns the count of elements for this stream.
+     * This is a terminal operator and, depending on the underlying stream, may invoke
+     * the stream pipeline, leaving it empty after this call.
+     * Care should be taken with stream pipelines that have side effects.
+     * This method should not be called on an infinite stream.
+     * <pre class="groovyTestCase">
+     * assert [1, 2, 3].stream().size() == 3
+     * </pre>
+     *
+     * @param self A Stream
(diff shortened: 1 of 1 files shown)


src/main/java/org/codehaus/groovy/runtime/StreamGroovyMethods.java
```diff
@@ -568,6 +568,24 @@ public static DoubleStream doubleStream(final double[] self) {
...
+     * assert [1, 2, 3].stream().size() == 3
+     * </pre>
+     *
+     * @param self A stream
+     * @return the count of elements in this stream
+     * @see Stream#count()
+     *
+     * @since 5.0.0
+     */
+    public static <T> long size(final Stream<T> self) {
```
daniellansun: How about `public static long size(final Stream<?> self)`?

eric-milles: `BaseStream` will let you size the primitive type streams as well.  And I agree, use `?` since the stream element type is irrelevant.

eric-milles (commented):
I agree with Christopher Smith from the mailing list that the semantic of count is different from size.   https://lists.apache.org/thread/8yknyvcxbksoojh6rz5shwv6jdjy1dnx

paulk-asert:
> I agree with Christopher Smith from the mailing list that the semantic of count is different from size. https://lists.apache.org/thread/8yknyvcxbksoojh6rz5shwv6jdjy1dnx

For those folks who are expecting every aggregate to have a `size()` method, do you have an alternative implementation for `size()` in mind that would better match their expectation? Or are you indicating that we'd better explain the unusual behavior (like we do for `Iterator#size`)? Folks who want to make the different semantics clearer would be free to use `count()` and wouldn't have to use `size()`.

daniellansun:
+1

PR #2250, opened 2025-06-06
GROOVY-11693: improve specificity of error message in extension module loading

dongwq:
https://issues.apache.org/jira/browse/GROOVY-11693


eric-milles:
Is there a test case that can check the message format?

PR #2251, opened 2025-06-06
GROOVY-11694: SC: implicit-this call to static method of super class

eric-milles:
`StaticMethodCallExpressionTransformer` creates "Type.staticMethod(arguments)" and later `StaticInvocationWriter` gets type of receiver expression as `Class<Type>`.  Use `Type` for access checks.


daniellansun:
Merged. Thanks.

PR #2253, opened 2025-06-10
GROOVY-11683: STC: transfer generics to extension method return types

eric-milles:
```groovy
@TypeChecked test() {
    def optional = Optional.empty().map(Iterable<String>::asCollection)
    //  ^^^^^^^^ Optional<Collection<String>>
}
```

https://github.com/apache/groovy/blob/55f31a3aec6cdab1f80aa76d27af08e2c621876d/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java#L922

see also GROOVY-10984


daniellansun:
Merged. Thanks.

PR #2255, opened 2025-06-18
GROOVY-11615: STC: receiver type args in param can help with argument

eric-milles:
For an expression like `list.stream().collect(Collectors.toSet())` the return type is dependent on the return type of `toSet`, which can be partially resolved from the receiver type (for example `Stream<SomeType>`).

`Collector<#T,?,Set<#T>>` --> `Collector<SomeType,?,Set<SomeType>>`


daniellansun:
Merged. Thanks.

PR #2256, opened 2025-06-18
Tuple changes related to its backing array

Choosechee:
I have made some changes to `Tuple` that are related to its backing array. One of them fixes a pretty major bug I found:

- Made `Tuple` implement `RandomAccess`.
- Currently, `Tuple`'s 'toArray()' returns its backing array. This is _extremely_ bad, because modifying this array affects the `Tuple`, which is supposed to be immutable. Additionally, it breaks the contract of `Collection.toArray()`: 
  > The returned array will be "safe" in that no references to it are maintained by this collection. (In other words, this method must allocate a new array even if this collection is backed by an array). The caller is thus free to modify the returned array.

  I have made it return a copy of the array instead.
- Overrode `toArray(T[])` with an implementation that uses `System.arraycopy`, for better performance.

### DIFF
--- a/src/main/java/groovy/lang/Tuple.java
+++ b/src/main/java/groovy/lang/Tuple.java
@@ -22,13 +22,15 @@ import org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation;
 
 import java.io.Serializable;
 import java.util.AbstractList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.RandomAccess;
 
 /**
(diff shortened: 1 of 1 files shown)


paulk-asert:
I created this issue to track this change:
https://issues.apache.org/jira/browse/GROOVY-11702

paulk-asert:
You might want to squash your PR.

Choosechee:
> You might want to squash your PR.

Done.

paulk-asert:
Merged, thanks!

PR #2259, opened 2025-06-29
Remove obsolete ManagedConcurrentMap stress tests

perNyfelt:
The ManagedConcurrentMap and its base class AbstractConcurrentMapBase were removed in Groovy 4.x as part of internal cleanup.

This patch removes the corresponding ManagedConcurrentMapStressTest, which is no longer relevant or compilable.


daniellansun:
Merged. Thanks.

PR #2262, opened 2025-07-10
GROOVY-11681: add `ExecutorService#submit(Closure)` extension

eric-milles:
`ExecutorService` declares `submit(Callable)` and `submit(Runnable)` so add `submit(Closure)` to prevent routing of closures to the `Runnable` variant, which drops the return value.

Add check in `MetaClassHelper#calculateParameterDistance` for `Closure` parameter and closure literal argument.

see also GROOVY-10714


daniellansun:
Merged. Thanks.

PR #2263, opened 2025-07-12
GROOVY-8162: Update Groovysh to JLine3

paulk-asert:


subprojects/groovy-groovysh/src/main/resources/nanorc/groovy.nanorc
```diff
@@ -0,0 +1,45 @@
...
+KEYWORD:     "\<(case|catch|default|do|else|finally|for|if|return|switch|throw|try|while)\>"
+KEYWORD:     "\<(abstract|class|extends|final|implements|import|instanceof|interface|native|non-sealed|package)\>"
+KEYWORD:     "\<(permits|private|protected|public|record|sealed|static|strictfp|super|synchronized|throws|trait|volatile)\>"
+PACKAGE:     "(([a-z0-9]{2,}[.]{1}){2,10}([a-z0-9]{2,}){0,1})"
+
+CLASS:       "\<[A-Z]{0,2}([A-Z]{1}[a-z0-9]+){1,}\>"
+# Annotation
+VARIABLE:    "@[A-Za-z]+"
+
+FUNCTION:    "\<(assert|print|println|sleep|from|where|select|as|in)\>"
```
daniellansun, line 28: `in` is Groovy keyword, so it's better to move it to `KEYWORD` list. 
Also, it would be great to add `on` to `FUNCTION` list, as `on` is keyword of GINQ.

paulk-asert:
> GQ {
> add: }>   from n in [1, 2, 3]
> add: }>   innerjoin m in [2, 3, 4] on m = n
> add: }>   select m, n
> add: }> }

You currently have to do:

println GQ {
add: }>   from n in [1, 2, 3]
add: }>   innerjoin m in [2, 3, 4] on m = n
add: }>   select m, n
add: }> }

The JLine3 "demo repl" which this groovysh replacement was inspired by, adopts "inspect last result" behavior instead of "print last result". I suspect we want to swap back to how we have always done it but there might be some classes where inspect behavior makes sense, so I haven't looked at that yet.

subprojects/groovy-groovysh/src/main/resources/nanorc/groovy.nanorc
paulk-asert: In some cases I moved certain words to the group which looked nicest when testing with a particular theme. I think a review of all words could be done. Also, I made the regex expressions in other places know about additional characters that can be in methods but haven't relooked at the how that change affects syntax files.

daniellansun:
> > GQ {
> > add: }>   from n in [1, 2, 3]
> > add: }>   innerjoin m in [2, 3, 4] on m = n
> > add: }>   select m, n
> > add: }> }
> 
> You currently have to do:
> 
> ```
> println GQ {
> add: }>   from n in [1, 2, 3]
> add: }>   innerjoin m in [2, 3, 4] on m = n
> add: }>   select m, n
> add: }> }
> ```
> 
> or:
> 
> `println _`
> 
> The JLine3 "demo repl" which this groovysh replacement was inspired by, adopts "inspect last result" behavior instead of "print last result". I suspect we want to swap back to how we have always done it but there might be some classes where inspect behavior makes sense, so I haven't looked at that yet.
> 
> So, my thinking is that we'd do the toString() but you could get the "inspect" behavior using "/inspect" or "/prnt".

`toString()` by default could help us to switch back.

paulk-asert:
> I prefer the original style of output, as it is compact and more readable
> 
> **current PR:** (a bit verbose, IMHO) image
> 
> **5.0.0-beta-1:** (compact and readable) image

I agree with you. Just haven't found time to turn that off yet.

daniellansun:
> > I prefer the original style of output, as it is compact and more readable
> > **current PR:** (a bit verbose, IMHO) image
> > **5.0.0-beta-1:** (compact and readable) image
> 
> I agree with you. Just haven't found time to turn that off yet.

It seem that we need to invoke `toString()` of the result.
As for "===>", it looks redundant, we could remove it in this PR by the way.

daniellansun:
image

Looks better now ;-)

PR #2267, opened 2025-07-17
GROOVY-11715: Grails reproducible build part 3: @DelegatesTo ordering

paulk-asert:


jdaugherty:
Thank you @paulk-asert

daniellansun:
Merged. Thanks.

src/main/java/org/codehaus/groovy/classgen/AsmClassGenerator.java
```diff
@@ -2221,10 +2222,11 @@ private AnnotationVisitor getAnnotationVisitor(final AnnotatedNode targetNode, f
      * @param av the visitor to use
      */
     public void visitAnnotationAttributes(final AnnotationNode an, final AnnotationVisitor av) {
-        Map<String, Object> constantAttrs = new HashMap<>();
-        Map<String, PropertyExpression> enumAttrs = new HashMap<>();
-        Map<String, Object> atAttrs = new HashMap<>();
-        Map<String, ListExpression> arrayAttrs = new HashMap<>();
+        // GROOVY-11715 TODO If we can determine what was causing the issues mentioned on that ticket, we should change these back to LinkedHashMap
```
matrei, line 2225: @paulk-asert Would a plausible explanation to why `LinkedHashMap` isn't working be that the items are inserted into the maps in different order across different runs? As `LinkedHashMap` maintains insertion order.

paulk-asert: @matrei "different order across different runs" is indeed what is being seen but we don't know how that could be the case.

PR #2268, opened 2025-07-19
Sort element value pairs in `AstBuilder` for reproducibility

matrei:
Sort element value pairs by key before collecting them into a `LinkedHashMap` to ensure consistent iteration order across builds.

This change addresses GROOVY-11715 without requiring `TreeMap` usage downstream, contributing to more reproducible build outputs.

I'm not sure if this is better than #2267, but perhaps it moves the solution one step closer to the source of the problem.

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -139,6 +139,7 @@ import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


daniellansun:
AST transformation can add annotations, so I do not think parser is an ideal place to fix the issue.

matrei:
> AST transformation can add annotations, so I do not think parser is an ideal place to fix the issue.

Aha, good point.
Is there also a reproducibility issue with annotations added by AST transformations?

eric-milles:
Don't the name-value pairs come in source order?  I would not want them sorted to another order besides that.

matrei:
Not the right solution, sorry for the noise

PR #2275, opened 2025-08-17
Try out ubuntu-24.04-arm runners

Goooler:
This should speed up builds significantly.

https://github.blog/changelog/2025-08-07-arm64-hosted-runners-for-public-repositories-are-now-generally-available/

### DIFF
--- a/.github/workflows/groovy-build-test.yml
+++ b/.github/workflows/groovy-build-test.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-24.04-arm]
         java: [11, 17, 21]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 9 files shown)


.github/workflows/groovy-build-test.yml
```diff
@@ -74,25 +74,18 @@ jobs:
...
       - uses: actions/setup-java@v4
         with:
           distribution: 'zulu'
-          java-version: 21
+          java-version: 25-ea
           check-latest: true
       - uses: gradle/actions/setup-gradle@v4
       - name: Test with Gradle
-        run: ./gradlew test -Ptarget.java.home=/home/runner/openjdk
+        run: ./gradlew test
```
Goooler, line 88: This seems to require Gradle 9, the previous test may not work as expected.

Goooler: I'm wrong, this could be replaced by toolchain.

PR #2276, opened 2025-08-17
Replace setup-java-25-ea steps with actions/setup-java

Goooler:


build-logic/src/main/groovy/org.apache.groovy-tested.gradle
```diff
@@ -70,11 +70,9 @@ tasks.withType(Test).configureEach {
...
     ))
 
-    if (rootProject.hasProperty('target.java.home')) {
-        String targetJavaHome = rootProject.property('target.java.home')?.trim()
-        if (targetJavaHome) {
-            executable = "${targetJavaHome}/bin/java"
-            println "Using ${executable} to run tests"
+    if (rootProject.hasProperty('test.java.ea')) {
+        javaLauncher = javaToolchains.launcherFor {
+            languageVersion = JavaLanguageVersion.of(25)
```
daniellansun, line 75: It's better to pass the java version via arguments to achieve better flexibility.

daniellansun:
Does toolchain support more ea versions present in the following link, e.g. JDK 26-ea?

https://jdk.java.net/

build-logic/src/main/groovy/org.apache.groovy-tested.gradle
Goooler: Yeah, we can. That's a good idea!

Goooler:
Yeah, the EA supports are due to https://github.com/foojayio/discoapi.

Goooler:
Blocker: https://github.com/gradle/foojay-toolchains/issues/109.

PR #2288, opened 2025-09-05
GROOVY-11745: check for null sender

eric-milles:
I looked into adding `if (senderClass == null) senderClass = metaClass.getTheClass();` to `ScriptBytecodeAdapter` `getProperty` and `setProperty`.  From what I can gather, the old behavior -- before any of the "propagate sender class" changes went in -- passed along the null sender.  So I opted for a null test within `isVisibleProperty`.


daniellansun:
Merged.

PR #2289, opened 2025-09-05
GROOVY-11601: restore support for expression list in `for` statement initializer

eric-milles:
revert commit 7f6720b17cd81ba62f82a71d6353963ac31fc7d4

see #2174


daniellansun:
Could you clarify the reason why we should restore the feature that was just removed in 5.0.0-beta-2?

daniellansun:
Merged.

PR #2295, opened 2025-09-16
GROOVY-11732: Gradle 9 and Java 25 is LTS now (not EA)

eric-milles:
Gradle 9 does not support Java 25 yet, so I changed the LTS test workflow to build with Java 17 and test with each LTS release.


eric-milles:
I'm having trouble working out the CycloneDX error.  It is described here: https://github.com/gradle/gradle/issues/33938

But the exact nature of how the root project ends up with `groovy-ant:runtimeClasspath` and so on is a bit confusing.  The root project includes `groovyAnt`, `groovyXml`, `groovyJson`, `groovyTest`, `groovyMacro` and `groovyDateutil` in the `testImplementation` configuration.  That should be excluded by `skipConfigs = ['compileClasspath', 'detached.*', 'test.*']`.

PR #2299, opened 2025-09-18
CycloneDX 2.4.0 (build dependency)

eric-milles:
prep for Gradle 9

see #2295


paulk-asert:
Merged, thanks! I scanned the generated SBOM files. They seem similar to what we produce now with some minor additional metadata and some slight ordering differences.

PR #2300, opened 2025-09-18
Gradle 9.1.0 (build dependency)

eric-milles:
see #2299 #2295

I'm having trouble working out the CycloneDX error.  It is described here: https://github.com/gradle/gradle/issues/33938

But the exact nature of how the root project ends up with `groovy-ant:runtimeClasspath` and so on is a bit confusing.  The root project includes `groovyAnt`, `groovyXml`, `groovyJson`, `groovyTest`, `groovyMacro` and `groovyDateutil` in the `testImplementation` configuration.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.14.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
```
nealeu, line 3: Please help us avoid MITM attacks and use the distribution sha
```suggestion
distributionUrl=https\://services.gradle.org/distributions/gradle-9.1.0-bin.zip
distributionSha256Sum=a17ddd85a26b6a7f5ddb71ff8b05fc5104c0202c6e64782429790c933686c806
```

paulk-asert: Is there any tooling to assist with this? No probs with adding as a one-off but it seems like it would be manual each time and already would be caught by gradle-wrapper-validation action?

paulk-asert:
Shouldn't be needed now. I haven't include the distributionSha256Sum yet.

PR #2311, opened 2025-10-03
GROOVY-11375, GROOVY-11769: SC: lookup type of variable before target

eric-milles:
There may be some side effects from changing `StaticTypesTypeChooser` in this manner.  I cannot recall exactly why this was not ported to 4.x back in June 2024.

see also GROOVY-9344, GROOVY-9607


eric-milles:
One of the related problems from Grails (see GormEntityTransformation) is this:
```groovy
if (classNode instanceof InnerClassNode || classNode.isEnum())
```

The second occurrence of `classNode` gets cast to `InnerClassNode` due to improper or handling (see GROOVY-7971).

paulk-asert:
I am happy for this to be merged but we still need to sort out GROOVY-7971. I might try to relook at that again soon.

PR #2374, opened 2026-01-28
GROOVY-10307: Improve invokedynamic performance with optimized caching

jamesfredley:
This PR significantly improves invokedynamic performance in Grails 7 / Groovy 4 by implementing several key optimizations:

https://issues.apache.org/jira/browse/GROOVY-10307
https://github.com/apache/grails-core/issues/15293

Using https://github.com/jamesfredley/grails7-performance-regression as a test application

**10K/10K = groovy.indy.optimize.threshold/groovy.indy.fallback.threshold**

### Key Results

| Configuration | Runtime (4GB) | Startup |
|--------------|---------------|---------|
| **No-Indy** | **60ms** | ~14s |
| **Optimized 10K/10K** | **139ms** |  ~11s |
| **Baseline Indy** | 267ms | ~14s |

### Root Cause Analysis

The performance in the Grails 7 Test Application was traced to the **global SwitchPoint guard** in Groovy's invokedynamic implementation:

1. **Global SwitchPoint Invalidation:** When ANY metaclass changes anywhere in the application, ALL call sites are invalidated
2. **Grails Framework Behavior:** Grails frequently modifies metaclasses during request processing (controllers, services, domain classes)
3. **Cascade Effect:** Each metaclass change triggers mass invalidation, forcing all call sites to re-resolve their targets
4. **Guard Failures:** The constant invalidation causes guards to fail repeatedly, preventing the JIT from optimizing method handle chains

---

## Optimizations Applied

1. **Disabled Global SwitchPoint Guard**
    - Removed the global SwitchPoint that caused ALL call sites to invalidate when ANY metaclass changed
    - Individual call sites now manage their own invalidation via cache clearing
    - Can be re-enabled with: `-Dgroovy.indy.switchpoint.guard=true`

2. **Call Site Cache Invalidation**
    - Implemented call site registration mechanism (`ALL_CALL_SITES` set with weak references)
    - When metaclass changes, all registered call sites have their caches cleared
    - Targets reset to default (fromCache) path, ensuring metaclass changes are visible
    - Garbage-collected call sites are automatically removed

### Optimizations Removed (No Measurable Benefit)

Testing showed these additional optimizations provide no measurable performance benefit while adding complexity:

3. ~~**Monomorphic Fast-Path**~~ (REMOVED)
    - ~~Added volatile fields (`latestClassName`/`latestMethodHandleWrapper`) to `CacheableCallSite`~~
    - ~~Skips synchronization and map lookups for call sites that consistently see the same types~~
    - **Removal Reason:** No measurable improvement in Grails 7 benchmark; added complexity

4. ~~**Optimized Cache Operations**~~ (REMOVED)
    - ~~Added `get()` method for cache lookup without automatic put~~
    - ~~Added `putIfAbsent()` for thread-safe cache population~~
    - **Removal Reason:** No measurable improvement; existing `getAndPut()` sufficient

5. ~~**Pre-Guard Method Handle Storage**~~ (REMOVED)
    - ~~Selector stores the method handle before argument type guards are applied (`handleBeforeArgGuards`)~~
    - ~~Enables potential future optimizations for direct invocation~~
    - **Removal Reason:** Future optimization not implemented; added unused code

### What Was NOT Changed

- **Default thresholds remain 10,000/10,000** (same as baseline GROOVY_4_0_X)
- **All guards remain intact** - guards are NOT bypassed, only the global SwitchPoint is disabled
- **Cache semantics unchanged** - LRU eviction, soft references, same default size (4)

---

### Test Environment

- **Java:** Amazon Corretto JDK 25.0.2
- **OS:** Windows 11
- **Test Application:** Grails 7 performance benchmark
- **Data Set:** 5 Companies, 29 Departments, 350 Employees, 102 Projects, 998 Tasks, 349 Milestones, 20 Skills
- **Test Method:** 5 warmup + 50 test iterations per run, 5 runs per configuration

### Runtime Performance Comparison (4GB Heap - Recommended)

| Configuration | Min | Max | Avg | Stability |
|--------------|-----|-----|-----|-----------|
| **No-Indy** | 55ms | ~85ms | **60ms** | Excellent |
| **Optimized 10K/10K** | 128ms | ~216ms | **139ms** | Good |
| **Baseline Indy** | 224ms | ~336ms | 267ms | Good |

### Startup Time Comparison

| Configuration | 4GB | Notes |
|--------------|-----|-------|
| No-Indy | ~14s | Consistent |
| Baseline Indy | ~14s | Similar to no-indy |
| Optimized 10K/10K | ~11s | Fastest startup |

### Visual Performance Summary (4GB Heap)

```
Runtime Performance - Lower is Better
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
No-Indy:        ███ 60ms
Opt 10K/10K:    ███████ 139ms         
Baseline Indy:  █████████████ 267ms
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Startup Time - Lower is Better
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Opt 10K/10K:    ███████ 11s           
No-Indy:        ████████████ 14s
Baseline Indy:  ████████████ 14s
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
```

### Configuration Test Matrix

All tests performed on January 30, 2026 with JDK 25.0.2 (Corretto), 4GB heap (-Xmx4g -Xms2g). Runtime values are stabilized averages (runs 3-5).

| Configuration | Startup | Runtime (Avg) | Min | Max |
|--------------|---------|---------------|-----|-----|
| **No-Indy** | ~14s | **60ms** | 55ms | 85ms |
| **Baseline Indy** | ~14s | 267ms | 224ms | 336ms |
| **Optimized 10K/10K** | ~11s | **139ms** | 128ms | 216ms |

---

## Failed Optimization Attempts

These approaches were tried but either caused test failures, provided insufficient benefit (<6%), or made performance worse. They are documented here to prevent repeating the same mistakes.

#### 1. Skip Metaclass Guards for Cache Hits

**What it tried:** Store method handle before guards, use unguarded handle for cache hits.

**Why it failed:** Call sites can be polymorphic (called with multiple receiver types). The unguarded handles contain `explicitCastArguments` with embedded type casts that throw `ClassCastException` when different types pass through.

#### 2. Version-Based Metaclass Guards

**What it tried:** Use metaclass version instead of identity for guards, skip guards for "cache hits."

**Why it failed:** Same polymorphic call site issue - skipping guards for cache hits is fundamentally unsafe.

#### 3. Skip Argument Guards After Warmup

**What it tried:** After warmup period, use handles without argument type guards.

**Why it failed:** Call sites can receive different argument types over their lifetime. Skipping argument guards causes `ClassCastException` or `IncompatibleClassChangeError`.

#### 4. Per-Class Metaclass Invalidation

**What it tried:** Only invalidate call sites for the specific class whose metaclass changed, not all call sites globally.

**Why it failed:** Broke mock framework tests (`MockFor`, `StrictExpectation`). The mock framework works by temporarily replacing metaclass and expects ALL call sites to see the change immediately - not just call sites for that class.

#### 5. Monomorphic Fast-Path with Identity Check

**What it tried:** Store latest receiver class name and method handle for quick identity check before hash lookup.

**Why it failed:** UNSAFE with per-instance metaclasses. Causes NPE when `NULL_METHOD_HANDLE_WRAPPER` gets cached for stale handles.

#### 6. Increase Cache Size (4 to 16)

**What it tried:** Increase the per-call-site LRU cache from 4 to 16 entries to improve polymorphic call site performance.

**Result:** Less than 6% improvement in Grails benchmark. The polymorphic benchmark uses way more than 16 types, so cache size doesn't help.

#### 7. Per-Class Generation Tokens (JRuby Pattern)

**What it tried:** Add `classInfoVersion` field to cache entries. On each call, compare cache token with receiver class generation for cheap integer validation instead of hash lookup.

**Result:** Less than 6% improvement. Category and mock tests still require global cache clearing anyway, so generation tokens provide little benefit.

#### 8. Optimized Cache Operations (get/putIfAbsent methods)

**What it tried:** Add separate `get()` and `putIfAbsent()` methods instead of combined `getAndPut()`.

**Result:** No measurable benefit. The overhead is elsewhere, not in cache operations.

#### 9. Pre-Guard Method Handle Storage

**What it tried:** Store method handle before argument type guards are applied (`handleBeforeArgGuards`, `directMethodHandle`) for potential future optimizations.

**Result:** No measurable benefit. Added unused code complexity.

#### 10. Bimorphic Inline Cache

**What it tried:** Add a 2-slot inline cache with identity comparison before HashMap lookup. Fast path for call sites that see 1-2 types.

**Results:**
- Polymorphic benchmark: 1.42s vs 1.60s (**11% improvement**)
- Grails app: 127ms vs 132ms (**4% SLOWER**)
- Metaclass ratio: 56.54x vs 66.76x (**18% worse**)

**Why it failed:** Adds overhead on every call site access. While it helps highly polymorphic sites, it hurts monomorphic-dominant workloads like typical Grails apps.

#### 11. Parallel Stream for Cache Invalidation (Approach 1)

**What it tried:** Use `parallelStream().filter().collect()` with atomic swap for call site invalidation.

**Result:** Made Grails app **18% SLOWER** (167ms vs 141ms baseline).

#### 12. Parallel Stream for Cache Invalidation (Approach 2)

**What it tried:** Use `parallelStream().forEach()` for invalidation + `removeIf()` for cleanup.

**Result:** Made Grails app **12% SLOWER** (158ms vs 141ms baseline).

**Why parallel streams failed:** Thread coordination overhead outweighs parallelization benefit for typical call site counts (~1000-5000 sites). Sequential `removeIf` has better cache locality and no synchronization overhead.

### Key Learnings

**Any optimization that bypasses guards based on cache state is unsafe because:**
1. The cache key (receiver Class) identifies which handle to use
2. But the call site itself can receive different types over its lifetime
3. Guards in the method handle chain ensure proper fallback when types don't match
4. The JVM's `explicitCastArguments` will throw if types don't match

**Safe optimizations must:**
1. Keep all guards intact
2. Focus on reducing guard overhead, not bypassing guards
3. Reduce cache misses rather than optimize the miss path unsafely

**Performance tradeoffs:**
- Optimizations that help polymorphic call sites (bimorphic cache, larger cache) often hurt monomorphic-dominant workloads
- Per-class invalidation breaks Groovy's mock framework which relies on immediate global metaclass visibility
- Parallel processing doesn't help for small collections (~1000-5000 call sites)

---

## Remaining Performance Gap

The optimized indy is still ~2.3x slower than non-indy (139ms vs 60ms). The remaining gap comes from:

1. **Guard evaluation overhead** - Every call must check type guards
2. **Method handle chain traversal** - Composed handles have inherent overhead
3. **Cache lookup cost** - Cache lookup overhead on each call
4. **Fallback path cost** - Cache misses trigger full method resolution

### Potential Future Optimization Areas

1. **Reduce guard evaluation cost** - Make guards cheaper (not skip them)
2. **Improve cache hit rate** - Larger cache, better eviction policy
3. **Reduce method handle chain depth** - Fewer composed handles
4. **JIT-friendly patterns** - Ensure method handle chains can be inlined
5. **Profile-guided optimization** - Use runtime profiles to specialize hot paths
6. **Megamorphic threshold detection** - Track relink count per call site, use simpler unoptimized path after threshold (Nashorn/Scala pattern)

---

### System Properties

| Property | Default | Description |
|----------|---------|-------------|
| `groovy.indy.optimize.threshold` | 10000 | Hit count before setting guarded target on call site |
| `groovy.indy.fallback.threshold` | 10000 | Fallback count before resetting to cache lookup path |
| `groovy.indy.callsite.cache.size` | 4 | Size of LRU cache per call site |
| `groovy.indy.switchpoint.guard` | false | Enable global SwitchPoint guard (NOT recommended) |

### Source Files 

| File | Lines | Changes |
|------|-------|---------|
| `Selector.java` | +16 | Disabled SwitchPoint guard by default |
| `IndyInterface.java` | +41 | Call site registration, cache invalidation on metaclass change |
| `CacheableCallSite.java` | +17 | Added `clearCache()` method |
| **Total** | **+69** | Minimal changes for maximum impact |

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
@@ -128,17 +128,22 @@ public class CacheableCallSite extends MutableCallSite {
     /**
      * Clear the cache entirely. Called when metaclass changes to ensure
      * stale method handles are discarded.
+     * <p>
+     * The volatile field {@code latestHitMethodHandleWrapperSoftReference} is cleared
+     * outside the synchronized block to be consistent with {@link #getAndPut}, which
+     * also accesses this field outside the lock. Volatile provides sufficient visibility
+     * guarantees for this field.
      */
     public void clearCache() {
-        // Clear the latest hit reference
+        // Clear the latest hit reference using volatile semantics
         latestHitMethodHandleWrapperSoftReference = null;
         
-        // Clear the LRU cache
+        // Clear the LRU cache under lock to synchronize with concurrent getAndPut operations
         synchronized (lruCache) {
             lruCache.clear();
         }
         
-        // Reset fallback count
+        // Reset fallback count (atomic operation, doesn't need to be in sync block)
         fallbackCount.set(0);
     }
 }
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
@@ -38,6 +38,10 @@ public class CacheableCallSite extends MutableCallSite {
     private static final int CACHE_SIZE = SystemUtil.getIntegerSafe("groovy.indy.callsite.cache.size", 4);
     private static final float LOAD_FACTOR = 0.75f;
     private static final int INITIAL_CAPACITY = (int) Math.ceil(CACHE_SIZE / LOAD_FACTOR) + 1;
+    
+    // Monomorphic fast-path: direct reference to most recently used entry (avoids sync + map lookup)
+    private volatile String latestClassName;
+    private volatile MethodHandleWrapper latestMethodHandleWrapper;
     private volatile SoftReference<MethodHandleWrapper> latestHitMethodHandleWrapperSoftReference = null;
     private final AtomicLong fallbackCount = new AtomicLong();
     private MethodHandle defaultTarget;
@@ -57,6 +61,14 @@ public class CacheableCallSite extends MutableCallSite {
     }
 
     public MethodHandleWrapper getAndPut(String className, MemoizeCache.ValueProvider<? super String, ? extends MethodHandleWrapper> valueProvider) {
+        // Fast path: check if this is the same class as the last call (monomorphic optimization)
+        final String cachedClassName = latestClassName;
+        final MethodHandleWrapper cachedMhw = latestMethodHandleWrapper;
+        if (cachedClassName != null && cachedClassName.equals(className) && cachedMhw != null) {
+            cachedMhw.incrementLatestHitCount();
+            return cachedMhw;
+        }
+
         MethodHandleWrapper result = null;
         SoftReference<MethodHandleWrapper> resultSoftReference;
         synchronized (lruCache) {
@@ -83,10 +95,78 @@ public class CacheableCallSite extends MutableCallSite {
             latestHitMethodHandleWrapperSoftReference = resultSoftReference;
         }
 
+        // Update fast path cache
+        latestClassName = className;
+        latestMethodHandleWrapper = result;
+
         return result;
     }
 
+    /**
+     * Get from cache. Returns null on cache miss (does NOT put).
+     * The monomorphic fast-path avoids synchronization and map lookup entirely.
+     * Use putIfAbsent() to add to cache after a miss.
+     */
+    public MethodHandleWrapper get(String cacheKey) {
+        // Fast path: check if this is the same key as the last call (monomorphic optimization)
+        final String cachedKey = latestClassName;
+        final MethodHandleWrapper cachedMhw = latestMethodHandleWrapper;
+        if (cachedKey != null && cachedKey.equals(cacheKey) && cachedMhw != null) {
+            return cachedMhw;
+        }
+
+        // Check the LRU cache
+        MethodHandleWrapper result = null;
+        synchronized (lruCache) {
+            SoftReference<MethodHandleWrapper> resultSoftReference = lruCache.get(cacheKey);
+            if (resultSoftReference != null) {
+                result = resultSoftReference.get();
+                if (result == null) {
+                    removeAllStaleEntriesOfLruCache();
+                }
+            }
+        }
+        
+        if (result != null) {
+            // Update fast path cache
+            latestClassName = cacheKey;
+            latestMethodHandleWrapper = result;
+        }
+
+        return result;
+    }
+
+    /**
+     * Put into cache. Returns existing value if present, otherwise puts and returns the new value.
+     */
+    public MethodHandleWrapper putIfAbsent(String name, MethodHandleWrapper mhw) {
+        synchronized (lruCache) {
+            // Check if already present
+            SoftReference<MethodHandleWrapper> existing = lruCache.get(name);
+            if (existing != null) {
+                MethodHandleWrapper existingMhw = existing.get();
+                if (existingMhw != null) {
+                    // Already present - update fast path and return existing
+                    latestClassName = name;
+                    latestMethodHandleWrapper = existingMhw;
+                    return existingMhw;
+                }
+                // Soft reference cleared - clean up
+                removeAllStaleEntriesOfLruCache();
+            }
+            // Put the new value
+            lruCache.put(name, new SoftReference<>(mhw));
+        }
+        // Update fast path cache
+        latestClassName = name;
+        latestMethodHandleWrapper = mhw;
+        return mhw;
+    }
+
     public MethodHandleWrapper put(String name, MethodHandleWrapper mhw) {
+        // Invalidate fast path cache since we're updating the LRU cache
+        latestClassName = null;
+        latestMethodHandleWrapper = null;
         synchronized (lruCache) {
             final SoftReference<MethodHandleWrapper> methodHandleWrapperSoftReference;
             methodHandleWrapperSoftReference = lruCache.put(name, new SoftReference<>(mhw));
@@ -123,5 +203,46 @@ public class CacheableCallSite extends MutableCallSite {
 
     public void setFallbackTarget(MethodHandle fallbackTarget) {
         this.fallbackTarget = fallbackTarget;
+    }
+    
+    /**
+     * Reset hit counts on all cached MethodHandleWrappers.
+     * Called after switchpoint invalidation to allow guards to be re-established.
+     */
+    public void resetAllHitCounts() {
+        // Reset the fast-path cache entry
+        MethodHandleWrapper fastPathMhw = latestMethodHandleWrapper;
+        if (fastPathMhw != null) {
+            fastPathMhw.resetLatestHitCount();
+        }
+        
+        // Reset all entries in the LRU cache
+        synchronized (lruCache) {
+            for (SoftReference<MethodHandleWrapper> ref : lruCache.values()) {
+                MethodHandleWrapper mhw = ref.get();
+                if (mhw != null) {
+                    mhw.resetLatestHitCount();
+                }
+            }
+        }
+    }
+    
+    /**
+     * Clear the cache entirely. Called when metaclass changes to ensure
+     * stale method handles are discarded.
+     */
+    public void clearCache() {
+        // Clear the fast-path cache
+        latestClassName = null;
+        latestMethodHandleWrapper = null;
+        latestHitMethodHandleWrapperSoftReference = null;
+        
+        // Clear the LRU cache
+        synchronized (lruCache) {
+            lruCache.clear();
+        }
+        
+        // Reset fallback count
+        fallbackCount.set(0);
     }
 }
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
@@ -38,10 +38,6 @@ public class CacheableCallSite extends MutableCallSite {
     private static final int CACHE_SIZE = SystemUtil.getIntegerSafe("groovy.indy.callsite.cache.size", 4);
     private static final float LOAD_FACTOR = 0.75f;
     private static final int INITIAL_CAPACITY = (int) Math.ceil(CACHE_SIZE / LOAD_FACTOR) + 1;
-    
-    // Monomorphic fast-path: direct reference to most recently used entry (avoids sync + map lookup)
-    private volatile String latestClassName;
-    private volatile MethodHandleWrapper latestMethodHandleWrapper;
     private volatile SoftReference<MethodHandleWrapper> latestHitMethodHandleWrapperSoftReference = null;
     private final AtomicLong fallbackCount = new AtomicLong();
     private MethodHandle defaultTarget;
@@ -61,14 +57,6 @@ public class CacheableCallSite extends MutableCallSite {
     }
 
     public MethodHandleWrapper getAndPut(String className, MemoizeCache.ValueProvider<? super String, ? extends MethodHandleWrapper> valueProvider) {
-        // Fast path: check if this is the same class as the last call (monomorphic optimization)
-        final String cachedClassName = latestClassName;
-        final MethodHandleWrapper cachedMhw = latestMethodHandleWrapper;
-        if (cachedClassName != null && cachedClassName.equals(className) && cachedMhw != null) {
-            cachedMhw.incrementLatestHitCount();
-            return cachedMhw;
-        }
-
         MethodHandleWrapper result = null;
         SoftReference<MethodHandleWrapper> resultSoftReference;
         synchronized (lruCache) {
@@ -95,78 +83,10 @@ public class CacheableCallSite extends MutableCallSite {
             latestHitMethodHandleWrapperSoftReference = resultSoftReference;
         }
 
-        // Update fast path cache
-        latestClassName = className;
-        latestMethodHandleWrapper = result;
-
         return result;
-    }
-
-    /**
-     * Get from cache. Returns null on cache miss (does NOT put).
-     * The monomorphic fast-path avoids synchronization and map lookup entirely.
-     * Use putIfAbsent() to add to cache after a miss.
-     */
-    public MethodHandleWrapper get(String cacheKey) {
-        // Fast path: check if this is the same key as the last call (monomorphic optimization)
-        final String cachedKey = latestClassName;
-        final MethodHandleWrapper cachedMhw = latestMethodHandleWrapper;
-        if (cachedKey != null && cachedKey.equals(cacheKey) && cachedMhw != null) {
-            return cachedMhw;
-        }
-
-        // Check the LRU cache
-        MethodHandleWrapper result = null;
-        synchronized (lruCache) {
-            SoftReference<MethodHandleWrapper> resultSoftReference = lruCache.get(cacheKey);
-            if (resultSoftReference != null) {
-                result = resultSoftReference.get();
-                if (result == null) {
-                    removeAllStaleEntriesOfLruCache();
-                }
-            }
-        }
-        
-        if (result != null) {
-            // Update fast path cache
-            latestClassName = cacheKey;
-            latestMethodHandleWrapper = result;
-        }
-
-        return result;
-    }
-
-    /**
-     * Put into cache. Returns existing value if present, otherwise puts and returns the new value.
-     */
-    public MethodHandleWrapper putIfAbsent(String name, MethodHandleWrapper mhw) {
-        synchronized (lruCache) {
-            // Check if already present
-            SoftReference<MethodHandleWrapper> existing = lruCache.get(name);
-            if (existing != null) {
-                MethodHandleWrapper existingMhw = existing.get();
-                if (existingMhw != null) {
-                    // Already present - update fast path and return existing
-                    latestClassName = name;
-                    latestMethodHandleWrapper = existingMhw;
-                    return existingMhw;
-                }
-                // Soft reference cleared - clean up
-                removeAllStaleEntriesOfLruCache();
-            }
-            // Put the new value
-            lruCache.put(name, new SoftReference<>(mhw));
-        }
-        // Update fast path cache
-        latestClassName = name;
-        latestMethodHandleWrapper = mhw;
-        return mhw;
     }
 
     public MethodHandleWrapper put(String name, MethodHandleWrapper mhw) {
-        // Invalidate fast path cache since we're updating the LRU cache
-        latestClassName = null;
-        latestMethodHandleWrapper = null;
         synchronized (lruCache) {
             final SoftReference<MethodHandleWrapper> methodHandleWrapperSoftReference;
             methodHandleWrapperSoftReference = lruCache.put(name, new SoftReference<>(mhw));
@@ -206,35 +126,11 @@ public class CacheableCallSite extends MutableCallSite {
     }
     
     /**
-     * Reset hit counts on all cached MethodHandleWrappers.
-     * Called after switchpoint invalidation to allow guards to be re-established.
-     */
-    public void resetAllHitCounts() {
-        // Reset the fast-path cache entry
-        MethodHandleWrapper fastPathMhw = latestMethodHandleWrapper;
-        if (fastPathMhw != null) {
-            fastPathMhw.resetLatestHitCount();
-        }
-        
-        // Reset all entries in the LRU cache
-        synchronized (lruCache) {
-            for (SoftReference<MethodHandleWrapper> ref : lruCache.values()) {
-                MethodHandleWrapper mhw = ref.get();
-                if (mhw != null) {
-                    mhw.resetLatestHitCount();
-                }
-            }
-        }
-    }
-    
-    /**
      * Clear the cache entirely. Called when metaclass changes to ensure
      * stale method handles are discarded.
      */
     public void clearCache() {
-        // Clear the fast-path cache
-        latestClassName = null;
-        latestMethodHandleWrapper = null;
+        // Clear the latest hit reference
         latestHitMethodHandleWrapperSoftReference = null;
         
         // Clear the LRU cache
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
@@ -53,6 +53,13 @@ import java.util.stream.Stream;
 public class IndyInterface {
     private static final long INDY_OPTIMIZE_THRESHOLD = SystemUtil.getLongSafe("groovy.indy.optimize.threshold", 10_000L);
     private static final long INDY_FALLBACK_THRESHOLD = SystemUtil.getLongSafe("groovy.indy.fallback.threshold", 10_000L);
+    /**
+     * Initial capacity for the call site registry used to track all active call sites
+     * for cache invalidation when metaclass changes occur. The default of 1024 balances
+     * memory usage against resize overhead. Tune via {@code groovy.indy.callsite.initial.capacity}
+     * system property for larger applications.
+     */
+    private static final int INDY_CALLSITE_INITIAL_CAPACITY = SystemUtil.getIntegerSafe("groovy.indy.callsite.initial.capacity", 1024);
 
     /**
      * flags for method and property calls
@@ -176,7 +183,7 @@ public class IndyInterface {
      * Weak set of all CacheableCallSites. Used to invalidate caches when metaclass changes.
      * Uses WeakReferences so call sites can be garbage collected when no longer referenced.
      */
-    private static final Set<WeakReference<CacheableCallSite>> ALL_CALL_SITES = ConcurrentHashMap.newKeySet();
+    private static final Set<WeakReference<CacheableCallSite>> ALL_CALL_SITES = ConcurrentHashMap.newKeySet(INDY_CALLSITE_INITIAL_CAPACITY);
 
     static {
         GroovySystem.getMetaClassRegistry().addMetaClassRegistryChangeEventListener(cmcu -> invalidateSwitchPoints());
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
@@ -18,7 +18,6 @@
  */
 package org.codehaus.groovy.vmplugin.v8;
 
-import groovy.lang.GroovyObject;
 import groovy.lang.GroovySystem;
 import org.apache.groovy.util.SystemUtil;
 import org.codehaus.groovy.GroovyBugError;
@@ -37,6 +36,7 @@ import java.lang.ref.WeakReference;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -51,18 +51,7 @@ import java.util.stream.Stream;
  * methods and classes.
  */
 public class IndyInterface {
-    /**
-     * Threshold for setting guarded target on call site (after this many cache hits).
-     * Lower values enable optimization sooner but may cause more fallbacks for polymorphic sites.
-     * Can be configured via system property groovy.indy.optimize.threshold.
-     */
     private static final long INDY_OPTIMIZE_THRESHOLD = SystemUtil.getLongSafe("groovy.indy.optimize.threshold", 10_000L);
-    
-    /**
-     * Threshold for fallback detection - after this many guard failures, the call site
-     * is reset to use the cache lookup path instead of the guarded path.
-     * Can be configured via system property groovy.indy.fallback.threshold.
-     */
     private static final long INDY_FALLBACK_THRESHOLD = SystemUtil.getLongSafe("groovy.indy.fallback.threshold", 10_000L);
 
     /**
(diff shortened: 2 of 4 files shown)


jamesfredley:
### Optimizations Retained

1. **Disabled Global SwitchPoint Guard** (Selector.java)
   - Removed the global SwitchPoint that caused ALL call sites to invalidate when ANY metaclass changed
   - Can be re-enabled with: `-Dgroovy.indy.switchpoint.guard=true`

2. **Call Site Cache Invalidation** (IndyInterface.java, CacheableCallSite.java)
   - Implemented call site registration mechanism (`ALL_CALL_SITES` set with weak references)
   - When metaclass changes, all registered call sites have their caches cleared
   - Targets reset to default (fromCache) path, ensuring metaclass changes are visible

### Optimizations Removed (No Measurable Benefit)

1. **Monomorphic Fast-Path** - `latestClassName`/`latestMethodHandleWrapper` volatile fields
2. **Optimized Cache Operations** - `get()`, `putIfAbsent()` methods
3. **Pre-Guard Method Handle Storage** - `handleBeforeArgGuards`, `directMethodHandle` fields

Testing showed these additional optimizations provide **no measurable performance benefit** in the Grails 7 benchmark while adding complexity.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -940,9 +941,18 @@ public void setGuards(Object receiver) {
...
-            if (LOG_ENABLED) LOG.info("added switch point guard");
+            // Skip the global switchpoint guard by default.
+            // The switchpoint causes ALL call sites to fail when ANY metaclass changes.
+            // In Grails and similar frameworks with frequent metaclass changes, this causes
+            // massive guard failures and performance degradation.
+            // The other guards (metaclass identity, class receiver, category) should be
+            // sufficient, combined with cache invalidation on metaclass changes.
+            // 
+            // If you need strict metaclass change detection, set groovy.indy.switchpoint.guard=true
+            if (SystemUtil.getBooleanSafe("groovy.indy.switchpoint.guard")) {
```
jonnybot0: The one question I have about this is whether this check ought to be inverted. That is, should the default behaviour be the old way, which we know is less performant but more battle-tested, or this way? Could Grails simply turn this flag on by default, thus solving their problem, without impacting other users?

Also, as a nitpicky thing, I reckon this string ought to be a `public static final` field so that we could add some tests around it.

jonnybot0:
I'm also well outside my expertise here, but so long as we're doing this, I'd be curious to see how this runs against the performance tests we have, compared to baseline.
https://ci-builds.apache.org/job/Groovy/job/Groovy%204%20performance%20tests/ has a Jenkins build that runs these. That's no good for running against a community repository, but perhaps we could push this up to a different branch and run it. Little outside my usual wheelhouse, so I'd like the nod from @paulk-asert, @blackdrag, or someone else on the PMC before I just run it willy-nilly.

I tried running `./gradlew :performance:performanceTests` on my local machine with Java 17 on this branch:
```
Groovy 4.0.7 Average 376.26ms ± 64.42ms 
Groovy current Average 376.87ms ± 63.75ms (0.16% slower)
Groovy 3.0.14 Average 443.74ms ± 131.81ms (17.93% slower)
Groovy 2.5.20 Average 451.40999999999997ms ± 138.82ms (19.97% slower)
```

That seems comparable, so my assumption is this hasn't hurt the most basic performance measures we do. Main reason I'd want CI confirmation is to make sure it wasn't a quirk of my machine.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
jamesfredley: If this PRs approach is viable, but does not prove to be generally valuable for all/most code run with Indy on, than I think the inverse would be fine for Grails and we could set that flag in default generated applications.

jonnybot0: This is part of why I'd like to run this branch against the current benchmarks on CI. That would likely provide some confidence as to whether this was "safe" to merge without impacting anyone's expectations about how the language behaves.

jamesfredley:
**Test Date:** January 30, 2026  
**Runs per Configuration:** 3  
**Environment:** Windows 11, Java 17.0.18, 4GB heap, 20 CPUs

---

## Master Comparison Table

All values are **averages across 3 runs**. Times in milliseconds unless noted. **Bold** indicates best performance.

### Loop Benchmarks (ComprehensiveBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Loop: each + toString | 33.09 ms | 33.23 ms | **32.70 ms** |
| Loop: collect | 52.62 ms | **52.22 ms** | 54.59 ms |
| Loop: findAll | **113.08 ms** | 116.15 ms | 114.58 ms |

---

### Method Invocation Benchmarks (ComprehensiveBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Method: simple instance | 8.77 ms | **8.55 ms** | 8.94 ms |
| Method: with params | **10.00 ms** | 10.08 ms | 10.52 ms |
| Method: static | 9.19 ms | **7.73 ms** | 8.25 ms |
| Method: polymorphic | **163.35 ms** | 1.70 s | 1.88 s |

---

### Closure Benchmarks (ComprehensiveBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Closure: creation + call | **21.45 ms** | 24.89 ms | 24.54 ms |
| Closure: reused | 19.44 ms | 19.86 ms | **18.59 ms** |
| Closure: nested | 39.25 ms | **37.51 ms** | 38.22 ms |
| Closure: curried | **154.37 ms** | 159.80 ms | 161.86 ms |

---

### Other Benchmarks (ComprehensiveBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Property: read/write | 20.80 ms | **18.70 ms** | 19.33 ms |
| Collection: each | 108.71 ms | **107.19 ms** | 110.03 ms |
| Collection: collect | 121.07 ms | **116.21 ms** | 121.74 ms |
| Collection: inject | **127.13 ms** | 132.85 ms | 139.18 ms |
| GString: simple | **101.22 ms** | 102.89 ms | 105.09 ms |
| GString: multi-value | **112.55 ms** | 116.85 ms | 117.61 ms |

---

### Method Invocation Benchmarks (MethodInvocationBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Simple instance method | 7.23 ms | **6.59 ms** | 8.22 ms |
| Method with parameters | **7.76 ms** | 8.08 ms | 8.31 ms |
| Method with object param | 11.41 ms | **10.74 ms** | 11.10 ms |
| Static method | **1.24 ms** | 3.47 ms | 3.40 ms |
| Static method with params | **1.17 ms** | 7.75 ms | 8.11 ms |
| Monomorphic call site | **46.89 ms** | 130.65 ms | 126.87 ms |
| Polymorphic call site | **382.26 ms** | 3.52 s | 3.85 s |
| Interface method | 6.34 ms | 3.53 ms | **3.42 ms** |
| Dynamic typed calls | 5.86 ms | **3.04 ms** | 3.18 ms |
| Property access | 27.52 ms | **21.49 ms** | 22.58 ms |
| Method reference | 34.40 ms | **31.01 ms** | 31.25 ms |
| GString method | **190.04 ms** | 193.28 ms | 194.07 ms |

---

### Closure Benchmarks (ClosureBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Simple closure creation | 32.52 ms | **30.21 ms** | 31.72 ms |
| Closure reuse | 22.79 ms | **19.86 ms** | 21.15 ms |
| Multi-param closure | 39.00 ms | **38.34 ms** | 39.22 ms |
| Closure with capture | 23.19 ms | **19.94 ms** | 20.34 ms |
| Closure modify capture | 14.30 ms | **12.55 ms** | 13.45 ms |
| Closure delegation | 40.33 ms | **27.24 ms** | 31.99 ms |
| Nested closures | 59.92 ms | **56.24 ms** | 59.72 ms |
| Curried closure | 291.59 ms | 296.21 ms | **289.39 ms** |
| Right curried closure | 188.59 ms | **187.98 ms** | 189.34 ms |
| Closure composition | 64.37 ms | **57.93 ms** | 59.37 ms |
| Closure as parameter | 42.64 ms | **40.58 ms** | 41.48 ms |
| Closure spread | **177.89 ms** | 2.10 s | 2.10 s |
| Closure.call() | 25.69 ms | **22.22 ms** | 25.49 ms |
| Closure trampoline | 55.88 ms | **50.72 ms** | 55.22 ms |
| each with closure | 28.55 ms | **26.44 ms** | 28.72 ms |
| collect with closure | 27.95 ms | **26.13 ms** | 28.00 ms |
| findAll with closure | **35.72 ms** | 37.14 ms | 42.34 ms |
| inject with closure | 33.69 ms | **33.41 ms** | 36.29 ms |

---

### Loop Benchmarks (LoopsBenchmark)

| Benchmark | NO-INDY | Optimized-INDY | Baseline-INDY |
|-----------|---------|----------------|---------------|
| Original: each + toString | **43.65 ms** | 45.55 ms | 45.90 ms |
| Simple: each only | **38.93 ms** | 42.03 ms | 42.67 ms |
| Closure call | 22.63 ms | **19.77 ms** | 21.99 ms |
| Method call | 8.14 ms | 6.22 ms | **5.39 ms** |
| Nested loops | **70.82 ms** | 72.94 ms | 77.47 ms |
| Loop with collect | **89.20 ms** | 90.26 ms | 90.67 ms |
| Loop with findAll | 209.40 ms | **204.75 ms** | 209.98 ms |

---

### Call Site Invalidation (Critical for GROOVY-10307)

| Metric | NO-INDY | Optimized-INDY | Baseline-INDY |
|--------|---------|----------------|---------------|
| Metaclass change ratio | **1.27x** | 69.4x | 102.6x |
| Baseline time (no changes) | 6.83 ms | 7.44 ms | **5.18 ms** |
| With changes time | **8.51 ms** | 439.54 ms | 531.72 ms |

---

## Summary Table: Key Metrics

| Metric | NO-INDY | Optimized-INDY | Baseline-INDY |
|--------|---------|----------------|---------------|
| Polymorphic call site | **382 ms** | 3.52 s | 3.85 s |
| Metaclass change ratio | **1.27x** | 69.4x | 102.6x |
| Static method | **1.24 ms** | 3.47 ms | 3.40 ms |
| Monomorphic call site | **46.89 ms** | 130.65 ms | 126.87 ms |
| Closure spread | **178 ms** | 2.10 s | 2.10 s |
| Simple instance method | 7.23 ms | **6.59 ms** | 8.22 ms |
| Closure delegation | 40.33 ms | **27.24 ms** | 31.99 ms |
| Property access | 27.52 ms | **21.49 ms** | 22.58 ms |
| Interface method | 6.34 ms | 3.53 ms | **3.42 ms** |
| Dynamic typed calls | 5.86 ms | **3.04 ms** | 3.18 ms |

---

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
```diff
@@ -168,24 +171,55 @@ public static CallType fromCallSiteName(String callSiteName) {
...
 
         synchronized (IndyInterface.class) {
             SwitchPoint old = switchPoint;
             switchPoint = new SwitchPoint();
             SwitchPoint.invalidateAll(new SwitchPoint[]{old});
         }
+        
+        // Invalidate all call site caches and reset targets to default (cache lookup)
+        // This ensures metaclass changes are visible without using expensive switchpoint guards
+        ALL_CALL_SITES.removeIf(ref -> {
```
daniellansun, line 231: The size of `ALL_CALL_SITES`  could be very big, so it's better to use parallel stream for better performance.

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
```diff
@@ -168,24 +171,55 @@ public static CallType fromCallSiteName(String callSiteName) {
     }
 
     protected static SwitchPoint switchPoint = new SwitchPoint();
+    
+    /**
+     * Weak set of all CacheableCallSites. Used to invalidate caches when metaclass changes.
+     * Uses WeakReferences so call sites can be garbage collected when no longer referenced.
+     */
+    private static final Set<WeakReference<CacheableCallSite>> ALL_CALL_SITES = ConcurrentHashMap.newKeySet();
```
daniellansun: It's better to set initial capacity.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#newKeySet-int-

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
daniellansun: How about extract `SystemUtil.getBooleanSafe("groovy.indy.switchpoint.guard")` to a constant to avoid repeated invocation?

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
daniellansun: > When metaclass changes, all registered call sites have their caches cleared

Current implementation will clear all cache no matter whether the changed metaclasses are releated. 
Groovy 5 has `lookup` instance in the `CacheableCallSite`, it can help us determine which class is using the cache site, maybe we can backport it to Groovy 4:
https://github.com/apache/groovy/blob/249491a631357e4a08a8a884d515fb6a9711d54a/src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java#L47

jamesfredley: I tested two versions of the parallel stream change with the two test application and saw a reduction in performance.

| Metric | Baseline | Approach 1 | Approach 2 |
|--------|----------|------------|------------|
| Grails App (steady state) | ~141 ms | ~167 ms (18% SLOWER) | ~158 ms (12% SLOWER) |

Approach 1: Parallel Collect + Atomic Swap - 18% SLOWER
```
// Invalidate all call site caches and reset targets to default (cache lookup)
// This ensures metaclass changes are visible without using expensive switchpoint guards
Set<WeakReference<CacheableCallSite>> liveReferences = ALL_CALL_SITES.parallelStream()
    .filter(ref -> {
        CacheableCallSite cs = ref.get();
        if (cs == null) {
            return false; // Don't keep garbage collected references
        }
        // Reset target to default (fromCache) so next call goes through cache lookup
        MethodHandle defaultTarget = cs.getDefaultTarget();
        if (defaultTarget != null && cs.getTarget() != defaultTarget) {
            cs.setTarget(defaultTarget);
        }
        // Clear the cache so stale method handles are discarded
        cs.clearCache();
        return true; // Keep live references
    })
    .collect(Collectors.toSet());
// Atomic swap - clear and replace with live references only
ALL_CALL_SITES.clear();
ALL_CALL_SITES.addAll(liveReferences);
```

Approach 2: Parallel ForEach + Sequential RemoveIf - 12% SLOWER
// Invalidate all call site caches in parallel
```
ALL_CALL_SITES.parallelStream().forEach(ref -> {
    CacheableCallSite cs = ref.get();
    if (cs != null) {
        // Reset target to default (fromCache) so next call goes through cache lookup
        MethodHandle defaultTarget = cs.getDefaultTarget();
        if (defaultTarget != null && cs.getTarget() != defaultTarget) {
            cs.setTarget(defaultTarget);
        }
        // Clear the cache so stale method handles are discarded
        cs.clearCache();
    }
});
// Remove garbage collected references (must be sequential for ConcurrentHashMap.KeySetView)
ALL_CALL_SITES.removeIf(ref -> ref.get() == null);
```

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
jamesfredley: Updating with this change.

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
jamesfredley: 1024 performed well for an initial capacity vs 16 (default), 128 and 2048

`private static final int INDY_CALLSITE_INITIAL_CAPACITY = SystemUtil.getIntegerSafe("groovy.indy.callsite.initial.capacity", 1024);`

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
jamesfredley: Testing with the Groovy 5 caller-specific Lookup pattern is generally positive on the [micro benchmarks](https://github.com/jamesfredley/groovy-indy-performance) but 11% slower in the [grails test application](https://github.com/jamesfredley/grails7-performance-regression): 

More details: https://github.com/jamesfredley/groovy/pull/1

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
```diff
@@ -168,24 +178,55 @@ public static CallType fromCallSiteName(String callSiteName) {
...
+            if (cs == null) {
+                return true; // Remove garbage collected references
+            }
+            // Reset target to default (fromCache) so next call goes through cache lookup
+            MethodHandle defaultTarget = cs.getDefaultTarget();
+            if (defaultTarget != null && cs.getTarget() != defaultTarget) {
+                cs.setTarget(defaultTarget);
+            }
+            // Clear the cache so stale method handles are discarded
+            cs.clearCache();
```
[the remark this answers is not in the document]
jamesfredley: Fixed in commit 6c6593e25d. Added `synchronized(cs)` block around the target check, target reset, and cache clear operations to ensure atomicity. This prevents race conditions where another thread could modify the call site between checking and setting the target.

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
```diff
@@ -168,24 +178,55 @@ public static CallType fromCallSiteName(String callSiteName) {
...
 
     static {
         GroovySystem.getMetaClassRegistry().addMetaClassRegistryChangeEventListener(cmcu -> invalidateSwitchPoints());
     }
+    
+    /**
+     * Register a call site for cache invalidation when metaclass changes.
+     */
+    static void registerCallSite(CacheableCallSite callSite) {
+        ALL_CALL_SITES.add(new WeakReference<>(callSite));
```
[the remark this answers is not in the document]
jamesfredley: Added comprehensive Javadoc in commit 6c6593e25d documenting the cleanup behavior. Stale WeakReferences are cleaned up during each call to `invalidateSwitchPoints()`. In applications with infrequent metaclass changes, some dead references may accumulate, but this is acceptable because: (1) memory overhead per dead reference is minimal (~16 bytes), and (2) frameworks like Grails that benefit most from this optimization have frequent metaclass changes that trigger regular cleanup.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -124,6 +125,13 @@ public abstract class Selector {
      */
     private static final CallType[] CALL_TYPE_VALUES = CallType.values();
 
+    /**
+     * Whether to use global SwitchPoint guard for metaclass changes.
+     * Default is false for better performance in frameworks with frequent metaclass changes.
+     * Set groovy.indy.switchpoint.guard=true to enable strict metaclass change detection.
```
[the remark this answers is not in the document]
jamesfredley: Added detailed Javadoc in commit 6c6593e25d explaining the performance implications. The documentation now explicitly warns that enabling the global SwitchPoint guard will significantly degrade performance in frameworks with frequent metaclass changes (like Grails), and recommends using it only for specific debugging or backward-compatibility scenarios.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -940,9 +948,18 @@ public void setGuards(Object receiver) {
...
-            handle = switchPoint.guardWithTest(handle, fallback);
-            if (LOG_ENABLED) LOG.info("added switch point guard");
+            // Skip the global switchpoint guard by default.
+            // The switchpoint causes ALL call sites to fail when ANY metaclass changes.
+            // In Grails and similar frameworks with frequent metaclass changes, this causes
+            // massive guard failures and performance degradation.
+            // The other guards (metaclass identity, class receiver, category) should be
+            // sufficient, combined with cache invalidation on metaclass changes.
+            // 
+            // If you need strict metaclass change detection, set groovy.indy.switchpoint.guard=true
```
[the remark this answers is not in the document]
jamesfredley: Simplified the inline comment in commit 6c6593e25d to a single line: "Apply global switchpoint guard if enabled (disabled by default for performance)". The detailed explanation of why the guard is disabled by default was moved to the `INDY_SWITCHPOINT_GUARD` field's Javadoc, which is the more appropriate location for this documentation.

src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
```diff
@@ -124,4 +124,21 @@ public MethodHandle getFallbackTarget() {
...
+    /**
+     * Clear the cache entirely. Called when metaclass changes to ensure
+     * stale method handles are discarded.
+     */
+    public void clearCache() {
+        // Clear the latest hit reference
+        latestHitMethodHandleWrapperSoftReference = null;
+        
+        // Clear the LRU cache
+        synchronized (lruCache) {
```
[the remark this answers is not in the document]
jamesfredley: Fixed in commit 6c6593e25d. Moved the `latestHitMethodHandleWrapperSoftReference = null` assignment inside the `synchronized(lruCache)` block to ensure atomicity with concurrent `getAndPut()` operations. The `fallbackCount.set(0)` remains outside the sync block since AtomicInteger operations are inherently thread-safe.

src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
```diff
@@ -124,4 +124,22 @@ public MethodHandle getFallbackTarget() {
...
+     * This method synchronizes on the lruCache to ensure atomicity with
+     * concurrent {@link #getAndPut} operations.
+     */
+    public void clearCache() {
+        // Clear the latest hit reference and the LRU cache atomically
+        synchronized (lruCache) {
+            latestHitMethodHandleWrapperSoftReference = null;
+            lruCache.clear();
+        }
+
```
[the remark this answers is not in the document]
jamesfredley: Going back to original code before last change - the synchronization was inconsistent. Updated in commit 868cfc5714 to keep the volatile field access outside the synchronized block, consistent with how `getAndPut()` accesses the same field (lines 75-84). Volatile provides sufficient visibility guarantees for this field. Added Javadoc explaining this design decision.

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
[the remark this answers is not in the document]
jamesfredley: **Update:** After further analysis with multiple reviewers, we decided NOT to add the `synchronized(cs)` block. Here's why:

1. **No other code synchronizes on `cs`** - The threads calling `setTarget()` in `fromCache()` and `selectMethod()` don't lock `cs`, so our lock doesn't make the operation atomic relative to them.

2. **The operation is idempotent** - Setting target to `defaultTarget` multiple times is harmless.

3. **Guards provide the real safety** - Even if a concurrent thread briefly reinstalls an optimized target after we reset it, the method handle guards (SAME_MC, SAME_CLASS) will fail on the next call if the metaclass changed, forcing a fallback to `selectMethod()`.

4. **Overhead without benefit** - The synchronized block adds monitor enter/exit overhead for no real safety gain.

Updated in commit 8c05bd5f46 with documentation explaining the best-effort invalidation strategy.

src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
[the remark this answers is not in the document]
jamesfredley: **Update:** After analysis, we're keeping the volatile field access **outside** the synchronized block (matching the original code pattern). 

The field `latestHitMethodHandleWrapperSoftReference` is a **best-effort optimization heuristic** for tracking hit counts, not a correctness mechanism. The race between `clearCache()` and `getAndPut()` is benign - worst case is slightly skewed hit count statistics.

Since `getAndPut()` already accesses this volatile field outside the lock (lines 75-84), putting the null assignment inside the lock in `clearCache()` doesn't provide atomicity anyway.

Updated in commit 8c05bd5f46 with documentation explaining this design choice.

src/main/java/org/codehaus/groovy/vmplugin/v8/CacheableCallSite.java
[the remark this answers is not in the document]
jamesfredley: **Update:** This contradicts your previous comment (#5) which suggested moving it **inside** the synchronized block. After analysis, we concluded that **neither position affects correctness** because:

1. The volatile field is a best-effort heuristic, not a correctness mechanism
2. `getAndPut()` accesses it outside the lock anyway, so no atomicity is achievable
3. The race is benign (worst case: slightly skewed hit count statistics)

We're keeping it **outside** the synchronized block (as you suggested here) for consistency with `getAndPut()`, with documentation explaining the design intent. See commit 8c05bd5f46.

paulk-asert (changes requested):
I am generally in favor of this change but not for GROOVY_4_0_X in the first instance. Safest would be master only but I understand folks are still by-and-large porting to GROOVY 5 rather than looking at 6, so we might need to get it in 5 for further feedback. I still have to think further about scenarios for when it is safe vs unsafe for the proposed default settings. I am not averse to potentially eventually porting to 4 but there are lots of folks on 4 and I am unsure we understand enough which ones might be adversely affected.

jamesfredley:
Closing this since Grails 7 EOL is June 30th, 2026

PR #2387, opened 2026-03-01
GROOVY-9381: Add native async/await support

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-9381

See also: 
- https://github.com/apache/groovy/blob/GROOVY-9381_3/src/spec/doc/core-async-await.adoc
- https://github.com/apache/groovy/blob/GROOVY-9381_3/subprojects/groovy-rxjava/src/spec/doc/rxjava-userguide.adoc
- https://github.com/apache/groovy/blob/GROOVY-9381_3/subprojects/groovy-reactor/src/spec/doc/reactor-userguide.adoc


eyforia (approved):
I was pleasantly surprised to learn about native async / await coming in Groovy 6. Way to go. Please keep it coming. 

As a: **developer who uses Groovy alongside with Java** (our codebase is 80% groovy / 20% java):
- i'm extremelly happy to see syntactic sugar features like  async/await to help with parallelization
- I would like this feature to leverage Java's Structured concurrency as much as possible (e.g. built on top of it rather than being a fork/alternative) - to be able to pick up future enhancements as Java's structured concurrency keeps improving
- I would like this feature to be as concise as possible rather than being "do-it-all" - e.g. parallelization is already making code pretty complex. Making it simple is a big win
- I don't care as much for Mono/Flux support - we are actively trying to get off WebFlux/Reactor where possible to reduce code complexity. It still has its uses (streams) but I would never use it where I can use structured concurrency or async/await


I hope this helps. @daniellansun  @paulk-asert  Great work!

daniellansun:
Thanks a lot for your review and suggestions.

@paulk-asert is trying to tweak the design and implementation of async/await( https://github.com/apache/groovy/commits/GROOVY_9381_5/ ) according to the feedbacks from dev mailing list.

PR #2453, opened 2026-04-10
GROOVY-11915: GINQ: Add groupby...into with first-class GroupResult type

paulk-asert:
Summary of changes                                                                                                                                                                                                
   
  New files (2):                                                                                                                                                                                                    
  - GroupResult.java — interface extending Queryable<T> with getKey() and key(String)
  - GroupResultImpl.java — implementation extending QueryableCollection<T>, materializes group elements as a list for reuse                                                                                         
                                                            
  Modified files (6):                                                                                                                                                                                               
  - GroupExpression.java — added intoAlias field with getter/setter, updated getText()
  - GinqAstBuilder.java — added KW_INTO keyword, dispatch block for into that sets alias on GroupExpression, hedge error for where after groupby...into, reordered conditions so hedge is checked before generic    
  DataSourceHolder check                                                                                                                                                                                        
  - GinqAstWalker.groovy — added __GROUPBY_INTO_ALIAS metadata key, getGroupByIntoAlias() helper, modified visitGroupExpression to generate groupByInto() call with simple having lambda, modified                  
  getLambdaParamName/correctVariablesOfLambdaExpression/validateGroupCols to bypass rewriting in into-mode                                                                                        
  - Queryable.java — added groupByInto() method signatures                                                                                                                                                          
  - QueryableCollection.java — added groupByInto() implementation
  - ginq-userguide.adoc — added "Group Into" subsection with examples                                                                                                                                               
                                                            
  Tests added:                                                                                                                                                                                                      
  - 3 backend API tests in QueryableCollectionTest.groovy   
  - 5 DSL-level tests in GinqTest.groovy (single-key, multi-key, sum, having, multi-key with as)                                                                                                                    
  - 2 error tests in GinqErrorTest.groovy (where-after-into hedge, into-without-groupby)        
                                                                                                                                                                                                                    
  Key design decisions in implementation:                                                                                                                                                                           
  - GroupResultImpl materializes elements via toList() to avoid stream-already-consumed errors when aggregates are called multiple times (having + select)                                                          
  - Single-key getKey() unwraps the NamedRecord to return the raw value                                                                                                                                             
  - into is parsed as a chained method call (groupby(n).into(g)), not as part of groupby's arguments — cleanly separates key aliasing (as) from group aliasing (into)

### DIFF
--- a/subprojects/groovy-ginq/src/main/groovy/org/apache/groovy/ginq/dsl/GinqAstBuilder.java
+++ b/subprojects/groovy-ginq/src/main/groovy/org/apache/groovy/ginq/dsl/GinqAstBuilder.java
@@ -231,6 +231,12 @@ public class GinqAstBuilder extends CodeVisitorSupport implements SyntaxErrorRep
 
             if (latestGinqExpressionClause instanceof JoinExpression && filterExpression instanceof OnExpression) {
                 ((JoinExpression) latestGinqExpressionClause).setOnExpression((OnExpression) filterExpression);
+            } else if (latestGinqExpressionClause instanceof GroupExpression && filterExpression instanceof WhereExpression
+                    && ((GroupExpression) latestGinqExpressionClause).getIntoAlias() != null) {
+                this.collectSyntaxError(new GinqSyntaxError(
+                        "`where` after `groupby...into` is not yet supported; use `having` instead",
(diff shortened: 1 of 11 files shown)


daniellansun:
Merged. Thanks.

PR #2461, opened 2026-04-11
GROOVY-11922: Jakarta EE 11 compatibility for groovy-servlet

paulk-asert:


subprojects/groovy-servlet/src/spec/doc/servlet-userguide.adoc
```diff
@@ -29,7 +29,7 @@ rather than Java code.
 * An `AbstractHttpServlet` class for defining traditional Servlets.
 * Helper classes like `ServletBinding` and `ServletCategory` which let you access some of the features offered by the `groovy-servlet` module in additional contexts.
 
-From Groovy 5, `groovy-servlet` supports Jakarta EE Servlet specifications. Javax EE servlet specifications are supported in version before Groovy 5 or with Groovy 5 by using the `javax` classifier when specifying your dependency.
+From Groovy 6, `groovy-servlet` targets Jakarta EE 11 (Servlet 6.1). It remains compatible with Jakarta EE 10 (Servlet 6.0) containers. The `javax` classifier is available for pre-Jakarta containers (e.g. Tomcat 9, Jetty 9).
```
utafrali: The claim "It remains compatible with Jakarta EE 10 (Servlet 6.0) containers" is asserted without qualification, but it depends entirely on whether any Servlet 6.1-exclusive API is called by groovy-servlet at runtime. Since the dependency is `compileOnlyApi`, the container provides the Servlet implementation at runtime. If any method or class added in 6.1 is invoked, users running on Tomcat 10.1 or Jetty 12.0.x (Servlet 6.0 containers) will get a `NoSuchMethodError` or `NoClassDefFoundError`.

Consider either: (a) auditing that only Servlet 6.0 APIs are used and adding a note about this constraint, or (b) changing the wording to something more conservative like "It may be compatible with Jakarta EE 10 (Servlet 6.0) containers provided no Servlet 6.1-exclusive APIs are used."

gradle/verification-metadata.xml
```diff
@@ -963,12 +963,22 @@
             <sha512 value="b810ce040ed0671a3161fb197845f3ae750cd38c474ef54c351b7b127ff3b4d624a26c33450c62474f42c1cde21c65324cda63b1b3ea0515a286e0bcb0c0c9d5" origin="Generated by Gradle" reason="A key couldn't be downloaded"/>
          </artifact>
       </component>
+      <component group="jakarta.servlet" name="jakarta.servlet-api" version="6.1.0">
```
utafrali, line 897: The new `jakarta.servlet-api-6.1.0.jar` entry relies only on SHA-512 for verification. Older entries (e.g. `jakarta.servlet.jsp-api-3.1.1.jar`) include both a `<pgp>` value and a SHA-512, even when the key couldn't be downloaded. It would be worth adding PGP verification here as well (even with `reason="A key couldn't be downloaded"` if necessary) to stay consistent with the project's existing artifact verification approach.

utafrali (approved):
The version bumps from Servlet 6.0 to 6.1 and JSP API 3.1.1 to 4.0.0 are correct, and the verification metadata SHA-512 hashes are the right length (128 hex chars). The main concern is the documentation's unqualified backward compatibility claim with Jakarta EE 10 containers, which could mislead users if any Servlet 6.1-specific API is called at runtime.

subprojects/groovy-servlet/src/spec/doc/servlet-userguide.adoc
paulk-asert: The Groovy functionality doesn't (currently) use any 6.1 exclusive functionality but you are correct that if user code calls such things, they will have to understand which versions they are running on. I was planning to add some notes about that in the release notes but you are correct that something should be here as well.

gradle/verification-metadata.xml
paulk-asert: I have just swapped the build to use an armor formatted keyring file in the last 30 mins. I plan to update the keys file and also add a pgp entry in one go. This doesn't affect any functionality of the groovy-servlet component, just ensures we are confident that we compiled and ran tests against trusted versions. It is more critical for dependencies that we include in our distributions, but I'll fix it here too. Thanks for taking the time to write those comments.

PR #2462, opened 2026-04-11
GROOVY-11923: Provide a groovy-csv optional module

paulk-asert:


daniellansun:
+1

glaforge:
I remember there was this GroovyCSV library as well by the way:
* https://github.com/xlson/groovycsv (github repo)
* http://xlson.com/groovycsv/ (documentation)

Maybe it's worth having a look at its feature-set.
It's using OpenCSV library.

Via your PR, I discovered that Jackson was also parsing CSV!
It's nice to be able to read CSV lines as typed objects!

paulk-asert:
My thinking with Jackson CSV processing was that if folks are already using TOML or YAML jackson libraries via groovy-toml or groovy-yaml, then they have the shared databind jars already in the classpath. And it means typed object support identical across all of those: CSV, TOML and YAML. But I am not averse to GroovyCSV, OpenCSV, or Apache Commons CSV.

glaforge:
This is fine to use Jackson! And the rationale makes sense.
I really like the fact we can extracted proper strongly typed objects out of the CSV, that's pretty neat.

PR #2473, opened 2026-04-18
GROOVY-11935: Set invokedynamic call site target immediately to enable earlier JIT inlining

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-11935

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
@@ -21,7 +21,6 @@ package org.codehaus.groovy.vmplugin.v8;
 import groovy.lang.GroovySystem;
 import org.apache.groovy.util.SystemUtil;
 import org.codehaus.groovy.GroovyBugError;
-import org.codehaus.groovy.reflection.ClassInfo;
 import org.codehaus.groovy.runtime.GeneratedClosure;
 
 import java.lang.invoke.CallSite;
@@ -31,6 +30,7 @@ import java.lang.invoke.MethodHandles;
 import java.lang.invoke.MethodType;
 import java.lang.invoke.MutableCallSite;
 import java.lang.invoke.SwitchPoint;
+import java.lang.reflect.Modifier;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.logging.Level;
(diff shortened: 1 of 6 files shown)


src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
```diff
@@ -320,48 +320,50 @@ public static Object fromCache(CacheableCallSite callSite, Class<?> sender, Stri
      */
     private static MethodHandle fromCacheHandle(CacheableCallSite callSite, Class<?> sender, String methodName, int callID, Boolean safeNavigation, Boolean thisCall, Boolean spreadCall, Object dummyReceiver, Object[] arguments) throws Throwable {
         FallbackSupplier fallbackSupplier = new FallbackSupplier(callSite, sender, methodName, callID, safeNavigation, thisCall, spreadCall, dummyReceiver, arguments);
+        final Object receiver = arguments[0];
```
blackdrag: So what will happen in this case?
```
def foo(receiver){receiver.bar()}
class A{ static bar(){1} }
class B{ static bar(){2} }
foo(A)
foo(B)
foo(new A())
foo(new B())
```
for the first two calls the class is the same, since receiver.getClass will return Class for both. But that is imho a bug, that existed already before `callSite.getAndPut(receiverClassName,...`. Also I think the later two cases are not considered here... why not?  If we really only want to optimize `A.bar()` style calls, then we should check the receiver type through the callsite type, not the runtime type.

daniellansun:
If no one is opposed to the PR, I will merge it at this weekend.

paulk-asert (approved):
Claude's assessment:

Assessment summary — GROOVY-11935

*Production code change*

Surgical and well-guarded. In IndyInterface.fromCacheHandle the early setTarget is gated on three conjoined conditions: the wrapper is cacheable (isCanSetTarget), the compile-time first-parameter of the call site is java.lang.Class, and the resolved MetaMethod is static. MethodHandleWrapper is extended to carry the already-resolved MetaMethod so the static check happens without re-resolution. The 15 new unit tests in IndyInterfaceCallSiteTargetTest cover the relevant edge cases — Class-typed vs Object-typed receivers, instance receivers that happen to dispatch to a static method, spread calls, safe-navigation, per-instance metaclass, fallback cutoff, uncacheable wrappers, multi-Class receivers — and all pass.

*Observed performance characteristics*

Two benchmark families on JDK 17.0.18 Zulu, with the PR enabled vs a local patch-off baseline:

Steady-state throughput (StaticMethodCallIndyBench, 2 forks × 6 s warmup × 5×2 s measurement):
* staticSum_groovy: 396 ops/ms vs 398 ops/ms — identical within noise.
* Same for staticChain_groovy, staticFib_groovy, all controls.

That was expected: the old hit-count relink (INDY_OPTIMIZE_THRESHOLD = 1000) crosses the threshold inside the first warmup iteration at these rates, so the two paths converge before measurement begins. No regression, but also nothing to prove the win.

Cold dispatch (new StaticMethodCallIndyColdBench, SingleShotTime, fresh JVM per fork × 80 forks, zero warmup):

|| variant || n || with PR (µs/op) || without PR (µs/op) ||
| staticSum_groovy |   500 |  97,958 ± 3490 | 100,835 ± 2689 |
| staticSum_groovy |  2000 |  98,340 ± 2509 | 103,700 ± 3001 |
| staticSum_groovy | 20000 | 105,597 ± 2754 | 110,887 ± 3011 |
| instanceSum_groovy (control) | 500 | 103,530 | 104,182 |
| instanceSum_groovy (control) | 2000 | 103,720 | 107,997 |
| instanceSum_groovy (control) | 20000 | 119,041 | 118,042 |

The cleanest read is the marginal cost per inner call between n=500 and n=2000 (the JVM-startup component cancels out):
* With PR: (98,340 − 97,958) / 1500 = *0.25 µs/call*
* Without PR: (103,700 − 100,835) / 1500 = *1.91 µs/call*

So in the pre-threshold invocation window the PR makes each cold static call *~7.6× cheaper*, exactly as the design predicts. Between n=2000 and n=20000 the two paths show the same marginal cost (~0.40 µs/call), confirming convergence once the old hit-count path has also relinked. The instance control shows no consistent direction across the three n values, so the signal on static calls is not environmental drift.

*Net*

Correctness-preserving, no steady-state regression, measurable cold-path win (~5–7× faster per call below the relink threshold; ~3–5% absolute on the n=500–2000 total-time measurements). The benefit should be most visible in short-lived scripts, groovysh/REPL startup, test bootstrap, and similar workloads where many call sites never reach 1000 hits. Larger absolute gains are plausible on JDKs with more aggressive MutableCallSite handling — matches the faster numbers the PR author reported.

PR #2490, opened 2026-04-25
GROOVY-11965: groovydoc incorrectly includes members declared inside …

daniellansun:
…anonymous Java class bodies in the generated documentation of the enclosing class

https://issues.apache.org/jira/browse/GROOVY-11965


paulk-asert (approved):
+1

PR #2495, opened 2026-04-27
GROOVY-11968: SC: trait static field access generates invalid bytecod…

paulk-asert:
…e under indy=false (GROOVY-11907 follow-up)

https://issues.apache.org/jira/browse/GROOVY-11968

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesWriterController.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesWriterController.java
@@ -19,9 +19,15 @@
 package org.codehaus.groovy.classgen.asm.sc;
 
 import org.codehaus.groovy.ast.ClassNode;
+import org.codehaus.groovy.ast.CodeVisitorSupport;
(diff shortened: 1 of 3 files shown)


src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesWriterController.java
```diff
@@ -87,15 +94,77 @@ public void init(final AsmClassGenerator asmClassGenerator, final GeneratorConte
...
+            super.visitAttributeExpression(expression);
+        }
+
+        @Override
+        public void visitVariableExpression(final VariableExpression expression) {
+            if (mark(expression)) return;
+            super.visitVariableExpression(expression);
+        }
+
+        private boolean mark(final Expression e) {
```
blackdrag: not wanting to nitpick, but mark is something you do. This is a check. isMakedDynamic or something like that would express the intend of the method better

paulk-asert: changed to isMarked

PR #2496, opened 2026-04-28
GROOVY-11022: StackOverflowError when having parameterized function w…

paulk-asert:
…ith recursive bounds

This is the fix recommended for GROOVY_5_0_X

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
+++ b/src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
@@ -55,11 +55,13 @@ import org.codehaus.groovy.tools.GroovyClass;
 import org.codehaus.groovy.transform.trait.Traits;
 import org.objectweb.asm.Opcodes;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1769,6 +1771,9 @@ static GenericsType[] getGenericsWithoutArray(final ClassNode type) {
         return type.getGenericsTypes();
     }
 
+    private static final ThreadLocal<Deque<GenericsTypeName>> EXPANDING_BOUNDS =
+            ThreadLocal.withInitial(ArrayDeque::new);
```
blackdrag, line 1775: I am very critical to using a ThreadLocal. At the very least there should be a description of why this has to be ThreadLocal

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingSupport.java
```diff
@@ -1789,9 +1794,22 @@ private static GenericsType applyGenericsContext(final Map<GenericsTypeName, Gen
...
+                // bound (e.g. `<K extends Comparable<? super K>>`)
+                Deque<GenericsTypeName> expanding = EXPANDING_BOUNDS.get();
+                if (expanding.contains(name)) return gt;
+                expanding.push(name);
+                try {
+                    GenericsType newGT = new GenericsType(type, applyGenericsContext(spec, gt.getUpperBounds()), applyGenericsContext(spec, gt.getLowerBound()));
+                    newGT.setPlaceholder(true);
+                    return newGT;
+                } finally {
+                    expanding.pop();
```
blackdrag, line 1808: so we push to expanding and then pop from expanding. That means we remove the entry we just added, even in the case of line 1806, with the return. Why is it then not always empty again? Ah, because of the recursive call applyGenericsContext(spec, gt.getLowerBound()). But why does it have to be a ThreadLocal then? I would prefere going for a breaking change here nstead of using ThreadLocal

paulk-asert:
Abandoned, instead the slightly larger solution for master will be back-ported.

PR #2499, opened 2026-04-28
GROOVY-11970: Provide support for compound assignment operator overlo…

paulk-asert:
…ading (GEP-15)

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java
@@ -998,4 +998,24 @@ public class ScriptBytecodeAdapter {
             throw unwrap(gre);
         }
     }
+
+    /**
+     * GEP-15: dispatcher for compound-assignment operators in dynamic Groovy.
+     * If {@code receiver} responds to {@code assignName} with the supplied argument,
+     * invoke it and return {@code receiver} (so the caller's STORE assigns the same
+     * reference back, leaving the in-place mutation visible). Otherwise fall back to
+     * {@code baseName} and return its result for the caller to assign.
+     */
+    public static Object compoundAssign(final Object receiver, final Object arg,
+                                        final String assignName, final String baseName) throws Throwable {
+        if (receiver != null) {
+            MetaClass mc = InvokerHelper.getMetaClass(receiver);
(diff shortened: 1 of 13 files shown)


src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java
```diff
@@ -998,4 +998,34 @@ public static Object bitwiseNegate(final Object value) throws Throwable {
...
+    public static Object compoundAssign(final Object receiver, final Object arg,
+                                        final String assignName, final String baseName) throws Throwable {
+        if (receiver != null) {
+            MetaClass mc = InvokerHelper.getMetaClass(receiver);
+            if (!mc.respondsTo(receiver, assignName, new Object[]{arg}).isEmpty()) {
+                mc.invokeMethod(receiver, assignName, arg);
+                return receiver;
+            }
+        }
+        return InvokerHelper.invokeMethod(receiver, baseName, arg);
```
blackdrag: SCB already has methods for invocation, why not use those?

src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java
```diff
@@ -998,4 +998,34 @@ public static Object bitwiseNegate(final Object value) throws Throwable {
...
+
+    /**
+     * GEP-15: helper for {@code @OperatorRename(plusAssign="...")} expression rewrites.
+     * The user's renamed name is authoritative (no fallback), and the expression value of
+     * {@code x op= y} is the (mutated) {@code x}, not the called method's return value.
+     */
+    public static Object invokeRenamedCompoundAssign(final Object receiver, final Object arg,
+                                                     final String name) throws Throwable {
+        InvokerHelper.invokeMethod(receiver, name, arg);
+        return receiver;
```
blackdrag: why not use a normal method call for this then? I don't see why we need an extra method in SCB for this

src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java
paulk-asert: fixed

src/main/java/org/codehaus/groovy/runtime/ScriptBytecodeAdapter.java
paulk-asert: reworked this

PR #2500, opened 2026-04-29
GEP-21: Broadening AST transform awareness under joint compilation (i…

paulk-asert:
…nitial spike for Shape C)

### DIFF
--- a/src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
+++ b/src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
@@ -93,6 +93,22 @@ public final class ASTTransformationVisitor extends ClassCodeVisitorSupport {
         this.context = context;
     }
 
+    /**
+     * Returns a primary class node operation that invokes any AST transformations
+     * registered for {@code phase} on each class. Intended for callers that need
+     * to drive transform invocation outside the standard phase wiring (for example,
+     * {@code JavaAwareCompilationUnit} runs CONVERSION-phase transforms before the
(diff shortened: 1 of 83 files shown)


src/main/java/org/codehaus/groovy/transform/ASTTransformationVisitor.java
```diff
@@ -93,6 +93,22 @@ private ASTTransformationVisitor(final CompilePhase phase, final ASTTransformati
...
 
+    /**
+     * Returns a primary class node operation that invokes any AST transformations
+     * registered for {@code phase} on each class. Intended for callers that need
+     * to drive transform invocation outside the standard phase wiring (for example,
+     * {@code JavaAwareCompilationUnit} runs CONVERSION-phase transforms before the
+     * stub generator emits stubs — see GEP-21).
+     */
+    public static org.codehaus.groovy.control.CompilationUnit.IPrimaryClassNodeOperation
+            invocationOperation(final CompilePhase phase, final ASTTransformationsContext context) {
```
blackdrag, line 104: concept wise I am not so happy with using IPrimaryClassNodeOperation here. It is more of an internal interface of CompilationUnit, it should not be used directly outside really.

paulk-asert: I have changed this - please check if the new approach seems okay to you. I renamed the branch and it looks like GitHub doesn't let you keep that in the same PR. So please check the other PR.

paulk-asert:
This is now in:
https://github.com/apache/groovy/pull/2503

PR #2503, opened 2026-04-30
GROOVY-11976: stub generation should better account for the shape tha…

paulk-asert:
…t will exist after later transforms run when known


daniellansun:
+1

PR #2511, opened 2026-05-04
GROOVY-11987: groovydoc CLI fails with "Unsupported Java Version: fal…

paulk-asert:
…se" when --javaVersion omitted


daniellansun:
Merged. Thanks.

PR #2513, opened 2026-05-04
Add `groovy.io.fileLegacyTruthy` system property to opt out of new File/Path truthy semantics

pditommaso:
## Summary

Adds an opt-out system property that restores the pre-Groovy-5 truthy semantics for `java.io.File` and `java.nio.file.Path`: when `-Dgroovy.io.fileLegacyTruthy=true` is set, any non-`null` reference is truthy, regardless of whether the underlying file exists.

The new default behaviour introduced in 5.0 (delegating to `File.exists()` / `Files.exists()`) is preserved when the property is absent or false.

## Why

The new truthy semantics for `File`/`Path` introduced in Groovy 5 are a useful upgrade for greenfield code, but they silently change the behaviour of any existing code that uses `if (path) { … }` or `path ? a : b` to mean **"the reference is set"**. The release notes call this out and recommend rewriting affected sites to use explicit `!= null`.

That recommendation is reasonable for code authors who control the codebase, but it does not generalise. In platforms that **execute user-supplied Groovy code** — Nextflow being one example, but the same applies to Gradle build scripts, Jenkins pipelines, and many other Groovy-based DSLs — the upgrade does not just affect the platform itself. Every existing user pipeline written against earlier Groovy versions can be silently affected, and there is no realistic path to audit and rewrite all of them. The end-user authoring those pipelines may have no idea Groovy was upgraded under them, and a previously-working `if(file)` may now silently skip a branch when the referenced file happens not to exist yet at evaluation time.

A toggle flag here lets such platforms upgrade to Groovy 5 without forcing a coordinated rewrite of an unbounded set of downstream user code. Greenfield Groovy 5 users continue to get the new default; platform integrators have an explicit, well-named knob to preserve compatibility for their users.

## Implementation

- New system property: `groovy.io.fileLegacyTruthy` (single shared flag covering both `File` and `Path`, named per Groovy's existing convention — see `groovy.compiler.strictNames`, `groovy.json.internKeys`, etc.).
- Two changes:
  - `ResourceGroovyMethods.asBoolean(File)` — opt-out branch returns `file != null`.
  - `NioExtensions.asBoolean(Path)` — opt-out branch returns `path != null`.
- Read once at class load via `SystemUtil.getBooleanSafe(...)` to match the pattern used elsewhere (`StaticTypeCheckingVisitor.DEBUG_GENERATED_CODE`, `CompilerConfiguration.parameters`, …).
- Javadoc updated on both methods.

The default path is unchanged (no behaviour change, no perf impact beyond a single boolean read at JVM start).

## Test plan

- [ ] Verify default behaviour (property absent) still calls `File.exists()` / `Files.exists()` — covered by the existing `ResourceGroovyMethodsTest.test_asBoolean`.
- [ ] Add a test exercising the opt-out path. Note: because the constant is initialised at class load, this test needs to set the system property before the class is loaded (forked JVM, or junit-pioneer `@SetSystemProperty`). Happy to add the test in this PR if a particular fixture style is preferred — let me know.

## Notes

- Single property covers both `File` and `Path` because the breaking change is conceptually the same; users are likely to want to flip both together.
- No CHANGELOG / release-note entry yet — happy to add one in the form preferred by the project.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
@@ -31,6 +31,7 @@ import groovy.transform.stc.FromString;
 import groovy.transform.stc.PickFirstResolver;
 import groovy.transform.stc.SimpleType;
 import groovy.util.CharsetToolkit;
+import org.apache.groovy.util.SystemUtil;
 import org.codehaus.groovy.runtime.typehandling.DefaultTypeTransformation;
 
 import java.io.BufferedInputStream;
@@ -153,14 +154,22 @@ public class ResourceGroovyMethods extends DefaultGroovyMethodsSupport {
 
     /**
      * Coerce the file to a {@code boolean} value.
+     * <p>
+     * By default this returns whether the file exists. Set the system property
+     * {@code groovy.io.fileLegacyTruthy} to {@code true} to restore the
+     * pre-Groovy-5 behavior where any non-{@code null} {@code File} is truthy
+     * regardless of whether the underlying file exists.
      *
      * @param file a {@code File}
      * @return {@code true} if the file exists, {@code false} otherwise
      * @since 5.0.0
      */
     public static boolean asBoolean(final File file) {
+        if (LEGACY_TRUTHY) return file != null;
         return file.exists();
     }
+
+    private static final boolean LEGACY_TRUTHY = SystemUtil.getBooleanSafe("groovy.io.fileLegacyTruthy");
 
     /**
      * Create an object output stream for this file.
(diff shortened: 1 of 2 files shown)


paulk-asert:
Are you up for creating a JIRA ticket for this? Or let me know and I can create one.
We tend to not have capital letters in system properties and often avoid things like legacy but state the opposite and let it be set to false to turn off, e.g. groovy.truth.file.exists.enabled (default to true).

pditommaso:
Thanks for the quick reply. I'm totally open to naming and code improvements — I've pushed a commit using the suggested name `groovy.truth.file.exists.enabled` (default: `true`) and reused the same constant to avoid duplicating the definition.

I'd appreciate it if you could open the Jira ticket.

One more question: do you think this change could reasonably be back-ported to the Groovy 5 branch?

paulk-asert:
I created this:
https://issues.apache.org/jira/browse/GROOVY-11996

paulk-asert:
Proposed PR merged. Thanks! Will be on the next 6 release and 5.0.7.

pditommaso:
Thanks a lot, this is great 🙏

PR #2529, opened 2026-05-11
GROOVY-11985: Static method override on trait implementer ignored whe…

paulk-asert:
…n called via this in trait body
https://issues.apache.org/jira/browse/GROOVY-11985


eric-milles:
This looks awfully familiar to the recent change to field reference.  Was the dynamic marker the missing detail?  Or was it more than that?

eric-milles:
I was able to try this out yesterday (internet outage) and the change is in the spot that I would expect.  The difficulty I have is reasoning about and explaining to others why "this.m(x)" can be rewritten to 3 -- or more depending on how you look at it -- different forms:
* `($static$self or $self.getClass()).m(x)`
* `(this or T$Trait$Helper).m($self or $static$self or $self.getClass(), x)`
* `(this or $self or $static$self).m(x)`

I did see that between Groovy 3 and Groovy 4, the behavior was change for private static methods.  Then between Groovy 4 and 5 there was a change for the "override" or "cannot be replaced by trait method" case.  I'm still on the fence about allowing dynamic dispatch so the seen trait method is not the one called in all cases.

paulk-asert:
@eric-milles I'll answer in the Jira issue to keep together with some other comments.

PR #2546, opened 2026-05-19
GROOVY-12022: replace indy code for array access in static compilation

blackdrag:
replace indy code for array access in static compilation mode with existing code in BytecodeInterface8 and deprecate now surplus IndyStaticTypesMultiTypeDispatcher. Performance wise there is no gain to use indy here and it simplifies code paths and JIT work


daniellansun:
Merged. Thanks.

PR #2567, opened 2026-05-28
use Locale.ROOT in StringGroovyMethods.containsIgnoreCase

netliomax25-code:
`containsIgnoreCase` lowercases both arguments with the JVM default locale, so
under `tr_TR` `"ADMIN".containsIgnoreCase("admin")` returns false — `I` folds
to dotless `ı`. Switch to `Locale.ROOT`, matching the sibling
`startsWithIgnoreCase`/`endsWithIgnoreCase` which already fold locale-independently via `equalsIgnoreCase`.


paulk-asert:
merged, thanks!

PR #2570, opened 2026-05-29
use Locale.ROOT in CharSequenceValue.toEnum

netliomax25-code:
Noticed the fallback in `CharSequenceValue.toEnum` uppercases the token text with the JVM default locale before matching the enum constant. Under tr_TR a value like `id` folds to `İD`, so `Enum.valueOf` throws and a valid JSON value never resolves. Fold with Locale.ROOT so the lookup stays locale-independent.


paulk-asert:
merged, thanks!

PR #2572, opened 2026-05-30
GROOVY-12046: MissingMethodException reports the metaclass theClass (…

daniellansun:
…a supertype) instead of the receiver's runtime class, breaking the GroovyObject.invokeMethod MOP fallback

### DIFF
--- a/src/main/java/groovy/lang/MetaClassImpl.java
+++ b/src/main/java/groovy/lang/MetaClassImpl.java
@@ -998,7 +998,23 @@ public class MetaClassImpl implements MetaClass, MutableMetaClass {
             }
         }
 
-        throw original != null ? original : new MissingMethodExceptionNoStack(methodName, theClass, arguments, false);
+
+        // GROOVY-12046: MissingMethodException reports the metaclass theClass (a supertype)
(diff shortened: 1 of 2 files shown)


src/main/java/groovy/lang/MetaClassImpl.java
```diff
@@ -998,7 +998,20 @@ private Object invokeMissingMethod(final Object instance, final String methodNam
...
+        // The MOP fallback guard in IndyGuardsFiltersAndSignatures.invokeGroovyObjectInvoker checks
+        // `receiver.getClass() == e.getType()` before delegating to GroovyObject.invokeMethod.
+        // A per-instance metaclass may have theClass set to a supertype of the actual receiver (a
+        // common pattern in mocking frameworks), so we must use the receiver's runtime class as the
+        // exception type in that case; otherwise the guard fails and the fallback is silently skipped.
+        Class<?> type = theClass;
+        if (!(instance instanceof Class) && type != instance.getClass() && type.isAssignableFrom(instance.getClass())
+                && lookupObjectMetaClass(instance) == this) {
+            type = instance.getClass();
+        }
```
eric-milles, line 1016: no need for all of this if original is non-null

PR #2580, opened 2026-05-31
GROOVY-12024: add `getAt(Map,String)` and `putAt(Map,String,V)`

eric-milles:
Should `map['key']` work like `map.key` or `map.get('key')`?  In older Groovy versions, `getAt(Map,Object)` was considered closer than `getAt(Object,String)`.  Adding `getAt(Map,String)` and `putAt(Map,String,V)` restores alignment between `getAt` and `get` at the expense of the public property handling for the subscript syntax.

STC of `map['key'] = value` would indicate "Cannot assign value of type ..." but now reports "Cannot call putAt ... with arguments ..."

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/DefaultGroovyMethods.java
@@ -7352,6 +7352,24 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
     }
 
     /**
+     * Support the subscript operator for a Map.
+     * <pre class="language-groovy groovyTestCase">def map = [foo:'bar', class:'baz', empty:'nope']
+     * assert map['foo'] == 'bar'
+     * assert map['class'] == 'baz'
+     * assert map['empty'] == 'nope'
+     * assert map['properties'] == null // GROOVY-12024
+     * </pre>
+     *
+     * @param self a Map
+     * @param key  a String as a key for the map
+     * @return the value corresponding to the given key
+     * @since 6.0.0
+     */
+    public static <V> V getAt(Map<? super String,V> self, String key) {
+        return self.get(key);
+    }
+
+    /**
      * Support the subscript operator for a Bitset
      *
      * @param self  a BitSet
@@ -12950,10 +12968,29 @@ public class DefaultGroovyMethods extends DefaultGroovyMethodsSupport {
      * @param self  a Map
      * @param key   an Object as a key for the map
      * @param value the value to put into the map
-     * @return the value corresponding to the given key
+     * @return the value
      * @since 1.0
      */
     public static <K,V> V putAt(Map<K,V> self, K key, V value) {
+        self.put(key, value);
+        return value;
+    }
+
+    /**
+     * A helper method to allow maps to work with subscript operators.
+     * <pre class="language-groovy groovyTestCase">def map = [:]
+     * assert map['properties'] == null
+     * map['properties'] = new Object() // GROOVY-12024
+     * assert map['properties'] != null
+     * </pre>
+     *
+     * @param self  a Map
+     * @param key   an Object as a key for the map
+     * @param value the value to put into the map
+     * @return the value
+     * @since 6.0.0
+     */
+    public static <V> V putAt(Map<? super String,V> self, String key, V value) {
         self.put(key, value);
         return value;
     }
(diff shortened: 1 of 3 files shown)


paulk-asert:
I think this is an improvement, I will try to do an assessment of all the cases tomorrow.

paulk-asert:
> The key reframing: under the **dynamic** runtime, `m['k']` already means `m.get('k')` for every key — `[:].class`, `[:].properties`, `[:].metaClass` all return `null` (the absent entry), not the bean. The surprising behavior was confined to **static** resolution, where subscript was binding to the property-access overload. So this PR isn't trading one consistency for another so much as **realigning STC subscript with the long-standing dynamic semantics** (and with `get`/`put`).
>
> Verified values (dynamic against the runtime; static column = intended 6.0 / the STC test assertions in this PR):
>
> **Plain map** — `[foo:'bar']` / `[class:'C']` / `[:]`
>
> | key | `m.k` dynamic | `m.k` `@CompileStatic` | `m['k']` (6.0) | `m.get('k')` |
> |---|---|---|---|---|
> | ordinary (`'foo'`) | `'bar'` | `'bar'` | `'bar'` | `'bar'` |
> | meta-name present (`'class'`→`'C'`) | `'C'` (entry, not the Class) | `'C'` | `'C'` | `'C'` |
> | meta-name absent (`'class'`) | `null` | `null` | `null` | `null` |
>
> **Map subclass** — `class C extends HashMap { def foo = 1; def bar = 2 }`, then `put('foo',11)`, `put('baz',33)`
>
> | key | `m.k` dynamic | `m.k` `@CompileStatic` | `m['k']` (6.0) | `m.get('k')` |
> |---|---|---|---|---|
> | declared prop **and** entry (`'foo'`: prop=1, entry=11) | **`11`** (entry wins) | **`1`** (property wins) | `11` | `11` |
> | declared prop, **no** entry (`'bar'`: prop=2) | **`null`** (property ignored) | **`2`** (property) | `null` | `null` |
> | entry, **no** declared prop (`'baz'`: entry=33) | `33` | `33` | `33` | `33` |
>
> Two takeaways:
>
> 1. **After this PR, `m['k']` ≡ `m.get('k')` in every row and both modes** — one clean column, no asterisks. That's the win worth stating explicitly as the invariant.
> 2. The genuinely treacherous cell is **`m.k` on a Map subclass**, which disagrees dynamic-vs-static (entry wins dynamically, declared property wins under STC). This PR doesn't touch dot access, so that divergence remains — subscript just sidesteps it now. Might be worth a follow-up issue, or at least a doc note steering entry access to `[]`/`get`.
>
> Suggestion for locking it in: pin the invariant as a paired test that runs the same asserts under dynamic **and** `@CompileStatic`, so the static/dynamic gap that caused this can't silently reopen via overload-resolution drift. Happy to put that together if helpful.

paulk-asert (approved):
I think we should take this but then work on the table to close anomolies

paulk-asert:
The above table was wrong. Here is the revised one:

For reference, here's the full `m.k` vs `m['k']` vs `m.get('k')` matrix on 6.0 after this change, across dynamic and `@CompileStatic` access. Every cell is now pinned by the live tables in `_working-with-collections.adoc` / `WorkingWithCollectionsTest.groovy`.

**Plain map** — `[foo:'bar']` / `[class:'C']` / `[:]`

| key                                 | `m.k`                        | `m.k` (`@CompileStatic`) | `m['k']` | `m.get('k')` |
| ----------------------------------- | ---------------------------- | ------------------------ | -------- | ------------ |
| ordinary (`'foo'`)                  | `'bar'`                      | `'bar'`                  | `'bar'`  | `'bar'`      |
| meta-name present (`'class'`→`'C'`) | `'C'` (entry, not the Class) | `'C'`                    | `'C'`    | `'C'`        |
| meta-name absent (`'class'`)        | `null`                       | `null`                   | `null`   | `null`       |

**Map subclass** — `class C extends HashMap { def foo = 1; def bar = 2 }`, then `put('foo',11)`, `put('baz',33)`

| key                             | `m.k` | `m.k` (`@CompileStatic`) | `m['k']` | `m.get('k')` |
| ------------------------------- | ----- | ------------------------ | -------- | ------------ |
| `'foo'` (property=1, entry=11)  | `1`   | `1`                      | `11`     | `11`         |
| `'bar'` (property=2, no entry)  | `2`   | `2`                      | `null`   | `null`       |
| `'baz'` (no property, entry=33) | `33`  | `33`                     | `33`     | `33`         |

Takeaways:

1. **`m['k']` ≡ `m.get('k')` in every row and in both modes** — that's the invariant this PR restores (subscript was previously diverging under static resolution).
2. **`m.k` and `m['k']` agree on a plain map** but legitimately differ on a subclass: `m.k` reads the declared property when one exists (falling back to the entry), while `m['k']`/`get` always read the entry. Dynamic and `@CompileStatic` columns now match throughout, so the only axis that matters is property-vs-entry, not dynamic-vs-static.
3. Practical guidance: use `['k']`/`get` whenever you mean "the entry"; reserve `m.k` for property semantics.

---

paulk-asert:
To keep this up -to-date, I added a "live table" of these results in:
src/spec/doc/_working-with-collections.adoc

PR #2581, opened 2026-05-31
reject backslash in grape ivy coordinate validation

netliomax25-code:
The coordinate check in createGrabRecord rejects / : " < > | in version and revision values, but the character class leaves out the backslash it claims to forbid, so a value such as ..\..\x slips through on Windows and is later interpolated into grape cache file paths. GrapeMaven already escapes the backslash in the same class; this brings GrapeIvy in line.


paulk-asert:
Merged, thanks! I think this is already protected at the Ivy level but doesn't harm to also include it here.

PR #2583, opened 2026-06-01
use Locale.ROOT in DataSet table name folding

netliomax25-code:
DataSet(Sql, Class) lowercases the table name from the type's simple name with the default locale on DataSet.java:107, so under a Turkish locale a class like Item produces the SQL `select * from ıtem`. Pass Locale.ROOT, matching the GROOVY-12055 fixes in Sql and ResultSetMetaDataWrapper.


paulk-asert:
merged, thanks!

PR #2584, opened 2026-06-01
GROOVY-12073: reject dot-dot segments in grape coordinate validation

netliomax25-code:
createGrabRecord rejects path separators and shell metacharacters in coordinate values, but a value made only of dot segments still passes both the version blacklist and the group/module whitelist, so a version or group of '..' survives and is later interpolated into the ivy/maven cache file paths as a parent-directory hop. This adds a contains('..') guard next to the existing checks, after the backslash fix, and applies it to GrapeMaven too since it shares the same validation.

### DIFF
--- a/subprojects/groovy-grape-ivy/src/main/groovy/groovy/grape/ivy/GrapeIvy.groovy
+++ b/subprojects/groovy-grape-ivy/src/main/groovy/groovy/grape/ivy/GrapeIvy.groovy
@@ -401,6 +401,9 @@ class GrapeIvy implements GrapeEngine {
                         throw new RuntimeException("Grab: invalid value of '$v' for $k: should only contain - . _ a-z A-Z 0-9")
                     }
                 }
+                if (v.toString().contains('..')) {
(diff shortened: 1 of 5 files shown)


paulk-asert:
Thanks for your on-going interest! Did you want to take a look at the Copilot suggestions?

paulk-asert:
I created this issue: https://issues.apache.org/jira/browse/GROOVY-12073

netliomax25-code:
Done. Both Copilot points are covered:

1. Added a `module: '..'` test to GrapeIvyTest so the check is pinned for all three coordinate keys, not just version and group.
2. Added matching dot-dot tests (version, group, module) to GrapeMavenTest for the Maven engine. That needed a `testImplementation projects.groovyTest` dependency there so `shouldFail` resolves.

Also reworded the commit to reference GROOVY-12073. Ran both grape suites with `-Djunit.network=true` and the new tests pass.

paulk-asert:
Merged, thanks!

PR #2585, opened 2026-06-02
decode yaml/toml/csv input streams as utf-8

netliomax25-code:
YamlSlurper, TomlSlurper and CsvSlurper wrap their stream and path inputs in `new InputStreamReader(stream)`, so bytes are decoded with the JVM default charset. Under a non-UTF-8 default (Windows-1252, Shift_JIS, or a JVM started with `-Dfile.encoding`), UTF-8 `0xC3 0xA9` reads back as `U+00C3 U+00A9` instead of `U+00E9`. YAML and TOML mandate UTF-8 and JsonSlurper already defaults byte/stream input to it, so pin these readers to UTF-8.


paulk-asert:
Issue created: https://issues.apache.org/jira/browse/GROOVY-12074

paulk-asert:
Merged, thanks!

PR #2588, opened 2026-06-03
Default MarkdownSlurper stream decoding to UTF-8

netliomax25-code:
MarkdownSlurper.parse decodes InputStream and Path input through a plain InputStreamReader, so it picks up Charset.defaultCharset() instead of UTF-8. On a JVM whose default isn't UTF-8 this looks like it would corrupt any multibyte content in an otherwise valid UTF-8 document. The sibling YamlSlurper, TomlSlurper, CsvSlurper and JsonSlurper already default stream input to UTF-8, so this passes StandardCharsets.UTF_8 at both reader sites to match.


paulk-asert:
Merged, thanks!

PR #2590, opened 2026-06-05
GROOVY-12062: OptimizingStatementWriter __$stMC slow branch resolves …

paulk-asert:
…a try-block local as getProperty when the method has a non-empty finally (classic non-indy codegen) - regression in 5.0.x


daniellansun:
Merged. Thanks.

PR #2591, opened 2026-06-06
GROOVY-12065: Implement peephole optimization for bytecode generation

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12065


src/main/java/org/codehaus/groovy/classgen/asm/OperandStack.java
```diff
@@ -585,42 +571,11 @@ private static void pushPrimitiveConstant(final MethodVisitor mv, final Object v
...
-            if (value.equals(0d)) {
-                mv.visitInsn(DCONST_0);
-            } else if ((Double) value == 1d) {
-                mv.visitInsn(DCONST_1);
-            } else {
-                mv.visitLdcInsn(value);
-            }
+            mv.visitLdcInsn(val);
+        } else if (isPrimitiveLong(type) || isPrimitiveFloat(type) || isPrimitiveDouble(type)) {
+            mv.visitLdcInsn(value);
```
eric-milles, line 576: Doesn't this require LCONST_0 and so on so the operand stack type is correct?

blackdrag: Maybe it was different when most of the bytecode code was written, but current version of asm actually handle the special case internally: https://asm.ow2.io/javadoc/org/objectweb/asm/MethodVisitor.html#visitLdcInsn(java.lang.Object) - AFAIK

paulk-asert (approved):

PR #2598, opened 2026-06-10
GROOVY-12069: introduce methodhandles in dgm wrappers

blackdrag:
This change adds a target method handle in the dgm helper classes. Invokedynamic then does not have to use the path of a generic meta method invocation, but can invoke the target of the dgm helper directly. For this bytecode generation changes are done in the dgm converter, as well as changes in indy method handling for that case. The change contains also a cleanup of some warnings, a test for the dgm converter changes and some cleanup/refactoring in the Selector


daniellansun:
Merged. Thanks.

PR #2624, opened 2026-06-24
escape namespace uri in XmlNodePrinter.printNamespace

netliomax25-code:
XmlNodePrinter escapes element text and ordinary attribute values, but printNamespace writes the namespace URI straight to the output. When a parsed node carries a namespace URI containing an XML metacharacter such as & or < (for example xmlns:foo="http://example.com/ns?a=1&amp;b=2"), the printer emits the literal character and the result is no longer well formed, so re-parsing the output fails. The URI is an attribute value, so route it through the existing printEscaped path used for other attribute values; a round-trip test covers & and < in a declared namespace.


paulk-asert:
Merged. Thanks!

PR #2625, opened 2026-06-24
reject ?> and -- in markup builder pi and comment output

netliomax25-code:
1. AbstractStreamingBuilder.toMapStringClosure builds processing-instruction pseudo-attributes by writing each name and value straight into the `<?...?>` body, so a value holding `?>` closes the PI early and injects sibling markup. This closure backs both StreamingMarkupBuilder and StreamingDOMBuilder.
2. MarkupBuilderHelper.comment wraps its text in `<!-- ... -->` through yieldUnescaped, so text holding `--` closes the comment early the same way.

Entity escaping is not legal inside a processing instruction or comment, so the terminating sequence is rejected (`?>` for the PI data, `--` for the comment text) rather than escaped. The builders already escape element text and attribute values, these two bodies were the gap. Regression test added for both, plus a check that ordinary PI/comment content still round-trips.


paulk-asert:
Merged as part of #2626. Thanks!

PR #2627, opened 2026-06-25
reject non-hex characters in decodeHex

netliomax25-code:
1. decodeHex decodes each two-character group with Integer.parseInt(value.substring(i, i + 2), 16), which accepts a leading + or - sign and non-ASCII Unicode digit characters.
2. so malformed input such as "-1", "+a" or "-f" decodes to a byte (0xff, 0x0a, 0xf1) instead of throwing NumberFormatException, which the method javadoc documents for characters that are not valid hexadecimal values.

Replaced the per-group parseInt with a strict per-nibble decode over the ASCII ranges 0-9/a-f/A-F. Valid hex decodes to the same bytes (checked against the old path for every byte value in both letter cases); the rejected inputs are covered by new cases in HexTest. Happy to file a Jira and link it.


paulk-asert:
Merged. Thanks!

PR #2633, opened 2026-06-27
GROOVY-12093 adjusted: out with @Anchored in with @Virtual

paulk-asert:
Backport of master's post-review adjustment to the 5.0.x line. Replaces the unreleased groovy.transform.Anchored marker (added in this branch's GROOVY-12093 backport, never shipped) with groovy.transform.Virtual, AND flips the default dispatch shape for public trait statics so the two release lines stay aligned at the source level:

  default          declarer-bound (trait body sees its own copy)
  @Virtual         per-implementer override visible to trait body

On 5.x the default returns to the pre-existing 4.x shape; @Virtual is the opt-in that restores the per-implementer override path used by Grails' Validateable.defaultNullable() and similar framework hooks. The annotated tests (Groovy11985 and matrix rows 1/10/12) add the marker so they exercise the per-implementer path under the new model.

Out of scope for this backport (master-only on the 6.x line): GROOVY-12104 (reject T.this.* qualifier inside trait code), GROOVY-12105 (reject unqualified super.m() from a static trait method), and the Groovy-6 default of promoting public trait statics to JVM-native interface statics.

The matrix test loses its four @Anchored-marker rows (1a/7a/8a/13) and its header references; rows 1/10/12 gain @Virtual on the relevant trait static so they continue to assert per-implementer dispatch. A new VirtualAnnotationTest covers both @Virtual's dispatch contract (with and without the marker, dynamic and @CompileStatic) and the validation that rejects misapplied placements at compile time.


src/main/java/groovy/transform/Virtual.java
```diff
@@ -0,0 +1,54 @@
...
+ * <p>Valid only on public, non-abstract <code>static</code> trait
+ * methods. Applying it to an instance method, a private method, an
+ * abstract method, or anything outside a trait is a compile-time error.
+ *
+ * @since 5.0.7
+ */
+@Documented
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface Virtual {
```
eric-milles, line 53: Since this applies to traits, does it make sense to have this as an inner class of `Trait`?

PR #2634, opened 2026-06-27
GROOVY-12109: Harden Closure deserialization against owner/delegate r…

paulk-asert:
…eference cycle


daniellansun:
Merged. Thanks.

PR #2644, opened 2026-06-29
GROOVY-12117: Resolution of inherited static trait method from sub-tr…

paulk-asert:
…ait body is transform-order dependent


eric-milles:
I was unable to recreate the order issue with the latest Groovy 6 snapshot.  Did this test trip it before the change?

paulk-asert:
With the fix backed out, the failing case is exactly the added test:

FAIL: testGrailsHelperShapeWithSubTraitDeclaredFirst
[Static type checking] - Cannot find matching method
    Arguable#withDelegate(groovy.lang.Closure, Field)

eric-milles:
Not sure what I missed.  I have read over my test case several times and it looks the same...

Anyways, the change itself looks good, so no worries there.

PR #2645, opened 2026-06-29
GROOVY-12115: Reify a receiver's type argument as a Class parameter u…

paulk-asert:
…nder @CompileStatic (@ClassTag)

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/CompilerConfiguration.java
+++ b/src/main/java/org/codehaus/groovy/control/CompilerConfiguration.java
@@ -226,6 +226,11 @@ public class CompilerConfiguration {
         }
 
         @Override
+        public void setClassTagPreemptionDisabled(final boolean classTagPreemptionDisabled) {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
         public void setBytecodePostprocessor(final BytecodeProcessor bytecodePostprocessor) {
             throw new UnsupportedOperationException();
         }
@@ -435,6 +440,17 @@ public class CompilerConfiguration {
     private Set<String> scriptExtensions = new LinkedHashSet<>();
 
     /**
+     * Global consumer opt-out for {@link groovy.transform.stc.ClassTag @ClassTag} preemption under
+     * static compilation (GROOVY-12115). Preemption is declared by the API author
+     * ({@code @ClassTag(preempt=true)}) and contained to the declaring class; setting this flag
+     * lets the consuming build veto every preemptive upgrade anyway. Additive injection (supplying
+     * an otherwise-mandatory token, such as for {@code asChecked}) is never gated by this flag.
+     * Seeded from the {@code groovy.classtag.preemption.disable} system property; false (the
+     * default) honours declared intent.
+     */
+    private boolean classTagPreemptionDisabled = getBooleanSafe("groovy.classtag.preemption.disable");
+
+    /**
      * If set to true recompilation is enabled.
      */
     private boolean recompileGroovySource;
@@ -613,6 +629,7 @@ public class CompilerConfiguration {
         setPluginFactory(configuration.getPluginFactory());
         setDisabledGlobalASTTransformations(configuration.getDisabledGlobalASTTransformations());
         setScriptExtensions(new LinkedHashSet<>(configuration.getScriptExtensions()));
+        setClassTagPreemptionDisabled(configuration.isClassTagPreemptionDisabled());
         setOptimizationOptions(new HashMap<>(configuration.getOptimizationOptions()));
         setForInPerIterationCaptureEnabled(configuration.isForInPerIterationCaptureEnabled());
         setBytecodePostprocessor(configuration.getBytecodePostprocessor());
@@ -1158,6 +1175,32 @@ public class CompilerConfiguration {
             scriptExtensions = SourceExtensionHandler.getRegisteredExtensions(getClass().getClassLoader());
         }
         return scriptExtensions;
+    }
+
+    /**
+     * Whether {@link groovy.transform.stc.ClassTag @ClassTag} preemption is disabled globally
+     * under static compilation (GROOVY-12115). Preemption requires the API author's declared
+     * intent ({@code @ClassTag(preempt=true)}) and is contained to the declaring class; this flag
+     * lets the consuming build veto every preemptive upgrade anyway. Additive injection, such as
+     * for {@code asChecked}, is never gated by this flag. False (the default) honours declared
+     * intent.
+     *
+     * @return whether preemption is disabled
+     * @since 6.0.0
+     */
+    public boolean isClassTagPreemptionDisabled() {
+        return classTagPreemptionDisabled;
+    }
+
+    /**
+     * Disables (or re-enables) {@link groovy.transform.stc.ClassTag @ClassTag} preemption
+     * globally under static compilation (GROOVY-12115).
+     *
+     * @param classTagPreemptionDisabled whether to disable preemption
+     * @since 6.0.0
+     */
+    public void setClassTagPreemptionDisabled(final boolean classTagPreemptionDisabled) {
+        this.classTagPreemptionDisabled = classTagPreemptionDisabled;
     }
 
     /**
--- a/src/test/groovy/groovy/transform/stc/ClassTagStaticTest.groovy
+++ b/src/test/groovy/groovy/transform/stc/ClassTagStaticTest.groovy
@@ -1,0 +1,551 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package groovy.transform.stc
+
+import groovy.transform.CompileStatic
+import org.codehaus.groovy.control.CompilerConfiguration
+import org.codehaus.groovy.control.customizers.ASTTransformationCustomizer
+import org.junit.jupiter.api.Test
+
+import static groovy.test.GroovyAssert.shouldFail
+
+/**
+ * Tests for {@code @ClassTag} (GROOVY-12115): under static type checking, a call may omit the
+ * compiler-supplied {@code Class<X>} token(s) and the type checker synthesises them, at their
+ * declared parameter positions, from the receiver's type argument(s). The first consumers are the
+ * {@code asChecked} extension methods.
+ */
+final class ClassTagStaticTest extends StaticTypeCheckingTestCase {
+
+    @Test
+    void testListTokenInjectedFromElementType() {
+        assertScript '''
+            List<String> base = []
+            List<String> checked = base.asChecked()       // compiler injects String.class
+            checked.add('ok')
+            assert base == ['ok']                          // checked view writes through to base
+            boolean threw = false
+            try {
+                ((List) checked).add(42)                   // wrong element type via raw view
+            } catch (ClassCastException e) {
+                threw = true
+            }
+            assert threw
+        '''
+    }
+
+    @Test
+    void testMapTokensInjectedForKeyAndValue() {
+        assertScript '''
+            Map<Number,String> base = [:]
+            Map<Number,String> checked = base.asChecked()  // compiler injects Number.class, String.class
+            checked.put(1, 'one')
+            assert base == [1: 'one']
+
+            boolean badKey = false
+            try { ((Map) checked).put('x', 'y') } catch (ClassCastException e) { badKey = true }
+            assert badKey
+
+            boolean badValue = false
+            try { ((Map) checked).put(2, 99) } catch (ClassCastException e) { badValue = true }
+            assert badValue
+        '''
+    }
+
+    @Test
+    void testInjectionWorksUnderCompileStaticDirectCall() {
+        assertScript '''
+            @groovy.transform.CompileStatic
+            class C {
+                static List<String> make() {
+                    List<String> base = []
+                    List<String> checked = base.asChecked()
+                    checked.add('ok')
+                    base
+                }
+            }
+            assert C.make() == ['ok']
+        '''
+    }
+
+    @Test
+    void testExplicitTokenStillResolves() {
+        assertScript '''
+            List<String> base = []
+            List<String> checked = base.asChecked(String)  // nothing injected; existing overload
+            checked.add('ok')
+            assert base == ['ok']
+        '''
+    }
+
+    @Test
+    void testSortedMapSubtypeReceiver() {
+        assertScript '''
+            TreeMap<Number,String> base = new TreeMap<>()
+            Map<Number,String> checked = base.asChecked()  // ConcreteMap subtype still resolves K,V
+            checked.put(1, 'one')
+            assert base == [1: 'one']
+        '''
+    }
+
+    @Test
+    void testRawReceiverDoesNotResolve() {
+        // a raw receiver has no statically-known type argument, so no token is synthesised
+        shouldFailWithMessages '''
+            void useRaw(Map base) {
+                base.asChecked()
+            }
+        ''', 'Cannot find matching method', 'asChecked()'
+    }
+
+    @Test
+    void testWithDefaultPreemptedToKeyAndValueChecked() {
+        assertScript '''
+            Map<Number,String> base = [:]
+            Map<Number,String> m = base.withDefault{ 'n/a' }   // preempted: key+value checked
+            assert m[1] == 'n/a'                                // compatible key auto-grows with String default
+
+            boolean badKey = false
+            try { ((Map) m).put('x', 'y') } catch (ClassCastException e) { badKey = true }
+            assert badKey
+
+            boolean badValue = false
+            try { ((Map) m).put(2, 99) } catch (ClassCastException e) { badValue = true }
+            assert badValue
+        '''
+    }
+
+    @Test
+    void testWithDefaultPreemptedKeyCheckedWhenValueUnconstrained() {
+        assertScript '''
+            Map<Number,?> base = [:]
+            Map<Number,?> m = base.withDefault{ null }          // preempted: at least key is checked
+            assert m[1] == null
+
+            boolean badKey = false
+            try { ((Map) m).put('x', 1) } catch (ClassCastException e) { badKey = true }
+            assert badKey
+        '''
+    }
+
+    @Test
+    void testWithDefaultStaysLenientWhenNothingToCheck() {
+        // an untyped map erases both tokens to Object, so there is nothing to gain and the lenient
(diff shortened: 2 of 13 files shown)


daniellansun:
# Request changes before merge:

## 1. Method selection: inject → re-select → rollback

In `visitMethodCallExpression`, the implementation mutates `argumentList` in place, re-runs a near-duplicate receiver/`findMethod` loop, then clears and restores the original arguments on failure. That couples AST surgery to selection success via an implicit rollback contract and duplicates control flow already present in the same method.

**Suggestion:** Treat `@ClassTag` matching as a pure computation (no AST mutation). Reuse a single `selectOnReceivers(...)` for both plain and expanded argument types. Commit the argument rewrite only after tagged selection succeeds. That removes the rollback path and the duplicated loop.

---

## 2. Preemption policy via method-name config is the wrong layer

`CompilerConfiguration.classTagPreemptionTargets` gates preemption by bare method name (default `"withDefault"`). Method names are not unique, so any type’s `withDefault` shares the policy. User APIs that want preemption need both parameter annotations and a compiler config change—the library cannot express intent alone.

**Suggestion:** Co-locate preemption with the API (e.g. an attribute on `@ClassTag`, or a method-level marker). Keep configuration, if needed, as a global kill-switch rather than a name allowlist.

---

## 3. ClassTag logic should not grow `StaticTypeCheckingVisitor`

Roughly 250 lines of cohesive feature logic were added to a visitor already past 7k lines. The package already factors helpers into types such as `StaticTypeCheckingSupport`.

**Suggestion:** Extract matching/resolution into a dedicated helper (or Support methods) and leave a thin call site in the visitor.

**Also:** The block was inserted between `findMethodsWithGenerated`’s Javadoc and the method body, which orphans that documentation. Please restore adjacency.

---

## 4. Selection model and contract gaps

- Arity-only matching with deferred applicability forces the mutate/rollback design; fold reifiability and remaining-parameter applicability into one pass so a failed match never touches the AST.
- When equal-token candidates disagree on erased types, injection is abandoned silently—consider a diagnostic or a clearer failure policy for API authors.
- Tokens may be resolved against one receiver while selection binds on another (multi-receiver / delegate-owner cases). Bind `(receiver, match)` together.
- `@ClassTag` documentation emphasises trailing `Class` arguments, but any parameter position is supported; preemption is undocumented on the annotation. Align the docs with behaviour.
- On source methods, `@ClassTag` on a non-`Class` parameter should error rather than silently disable injection (same class of bug as the override typo check).

src/test/groovy/groovy/transform/stc/ClassTagStaticTest.groovy
```diff
@@ -0,0 +1,364 @@
...
+                List<Class> captured = []
+                Map<K,V> record(@ClassTag Class<V> valueType, @ClassTag Class<K> keyType, Closure init) {
+                    captured = [valueType, keyType]
+                    [:]
+                }
+            }
+
+            Box<Number,String> b = new Box<>()
+            b.record{ }
+            assert b.captured == [String, Number]   // valueType <- V=String, keyType <- K=Number
```
blackdrag, line 185: I think I am missing some tests here:
(1) What happens if there is a record(Closure) method additionally to the one declared in Box?
(2) There should be a test where Class<V> and Class<K> are switched in positions.
(3) What about generic parameters added by the method?

paulk-asert:
Thanks @daniellansun and @blackdrag for the thorough reviews — nearly everything has been actioned, and the preemption design has been reworked substantially in response. Details below; the branch is updated.

## Preemption redesign: intent + containment + consent

The `classTagPreemptionTargets` name allowlist is gone. Daniel was right that a config-level name list was the wrong layer (names aren't unique, and a library couldn't express intent alone), and Jochen's mailing-list concern was that preemption changes the meaning of existing source at all. The replacement splits the decision across three parties, each owning the part that's theirs:

1. **Intent (API author)** — `@ClassTag` gains `preempt() default false`. Only an overload explicitly marked `@ClassTag(preempt=true)` can ever displace a token-less incumbent; unmarked tags are strictly additive. A library now self-declares — no consumer configuration needed. DGM's checked `withDefault` overloads declare it; `asChecked` does not (nothing to preempt).

2. **Containment (structural invariant, not configurable)** — a tagged overload may only preempt a token-less overload declared by the *same owner*: for an extension-method incumbent, tagged overloads of the same extension class; for an instance-method incumbent, the receiver's own class hierarchy. This closes what I think is the most important hole in author-declared preemption: without it, any jar on the compile classpath could ship e.g. `collect(Iterable, @ClassTag(preempt=true) Class, Closure)` and silently capture every statically-compiled `collect{}` call on the next rebuild. With the same-owner rule, a library can upgrade callers of *its own* lenient API but structurally cannot capture calls it doesn't own — the residual exposure reduces to "a library can change the behaviour of calls to itself", which it could already do by editing its method bodies.

3. **Consent (consumer)** — `CompilerConfiguration.classTagPreemptionDisabled`, a veto set using the same selector forms as `groovy.extension.disable`: a bare name (`withDefault`), an exact signature in self-first static form (`withDefault(Map,Class,Closure)`), or `*` to disable all preemption. Seeded from `-Dgroovy.classtag.preemption.disable`, so no config script is needed for the common cases. Default empty = honour declared intent.

For Jochen specifically: `*` (or simply never writing `preempt=true`) is now a first-class "no preemption ever" mode, and out of the box the only method that preempts anything is `withDefault` — whose lenient variant is the GROOVY-11807 bug this exists to route around.

## Daniel's review items

- **#1 inject → re-select → rollback**: reworked as suggested. Matching is now a pure computation (`ClassTagSupport.matchAdditive`/`matchPreemption`); tagged argument types are computed by splicing token types into the args array, and the AST is rewritten only after the retried selection binds. The rollback path and the duplicated receiver loop are gone.
- **#3 visitor growth + orphaned Javadoc**: all ClassTag logic extracted to a new package-private `ClassTagSupport` (~25-line call site remains in the visitor, net −230 lines); `findMethodsWithGenerated`'s Javadoc is reunited with its method.
- **#4a arity-only matching**: subsumed by the pure-match refactor.
- **#4b silent ambiguity**: still fail-soft by design (degrading to the call as written must not become an error), but I agree the additive-case "no matching method" error deserves a hint that injection was attempted and abandoned — follow-up.
- **#4c receiver binding**: fixed. Additive matching walks receivers in delegate/owner priority order and constrains the retry to the matched receiver; preemptive matching reifies from the receiver the incumbent bound on, so an upgrade can change overload but never binding target.
- **#4d docs**: annotation Javadoc no longer says "trailing" (tokens fill declared positions, anywhere in the list), and preemption + the veto config are now documented on the annotation itself.
- **#4e non-`Class` parameter**: now a compile-time error on source methods, parallel to the override-typo check.

## Jochen's test questions

1. *token-less overload alongside the tagged one* — covered by `testUserApiWithoutDeclaredIntentStaysLenient` / `testUserApiDeclaresPreemptionIntent` (and now also `testUserApiIntentVetoedByConsumer`).
2. *switched `Class<V>`/`Class<K>` positions* — `testTokensReorderedByTypeVariableNotPosition` covers the swap; new `testTokensInterleavedWithSuppliedArguments` additionally covers a token declared *after* a supplied argument.
3. *method-declared type parameters* — this was a genuinely good catch: chasing it surfaced that a method-declared type variable shadowing a class-level one (per Java scoping) could have reified the *wrong* class. Now: on instance methods compiled from source, tagging a method-declared variable is a compile-time error; a matching-side guard makes precompiled libraries degrade to no-injection rather than mis-inject; static self-style methods (the extension-authoring pattern) are exempt since there the method-level variable *is* the mechanism. Tests: `testMethodTypeParameterIsRejectedAtCompileTime`, `testShadowedTypeParameterIsRejectedAtCompileTime`, `testStaticSelfStyleMethodTypeParameterAccepted`.

## Status / open items

`ClassTagStaticTest` 32/32, `CompilerConfigurationTest` 15/15, full `groovy.transform.stc.*` (2012 tests) and `classgen.asm.sc.*` (2295 tests) sweeps green, `Groovy11807` regression green.

Two known follow-ups: (a) the preemption check now runs without the old O(1) name pre-gate — it's bounded (cached DGM lookup filtered by owner / receiver-hierarchy scan), but I'd like a compile-time benchmark on a large `@CompileStatic` codebase before merge, with an `ExtensionMethodCache`-level "names with preempt intent" cache as the optimization if needed; (b) the cross-module containment ("an extension module cannot capture DGM's methods") is implemented but wants an m12n-style fixture-jar integration test.

Comments on the revised model welcome — particularly whether `preempt=true` + same-owner containment + selector veto addresses the "why is one method better" and "wrong layer" concerns, respectively.

paulk-asert:
## Follow-up (a) in more detail: preemption-check cost — benchmark plan + proposed gate

Expanding on the compile-time concern noted above, since it's the one open item that could block merge.

### The cost, precisely

The earlier allowlist design had an accidental virtue: `classTagPreemptionTargets.contains(name)` was an O(1) pre-gate that was false for virtually every call, so resolved calls paid nothing. In the revised design, `isPreemptionPossible()` is still O(1) but almost always *true*, after which `matchPreemption` runs for every resolved method call under `@TypeChecked`/`@CompileStatic`:

- an incumbent-kind scan over the matched methods (cheap, usually one element);
- for extension-method incumbents — the dominant case (`each`, `collect`, …) — a `findDGMMethodsForClassNode` lookup, which is cache-backed but still walks the receiver hierarchy and filters by name per call;
- intent/veto checks that call `getAnnotations(...)` per parameter per candidate.

Almost always the outcome is "no preempt-intent overload exists here" — we pay enumeration cost to discover a negative that is knowable in advance. That's the target.

### Proposed gate: a per-classloader "preemptive names" set in `ExtensionMethodCache`

`ExtensionMethodCache` already solves the exact lifecycle problem we need: it caches the receiver→methods map per classloader and invalidates when extension modules change. The plan piggybacks on the same scan:

- Change the cached value to a small holder — `{byReceiverType map; preemptiveNames set}` — so the name set shares the map's lifecycle and can never go stale independently (no separate invalidation logic to get wrong).
- While `AbstractExtensionMethodCache` accumulates each `MethodNode`, additionally record the method name if any parameter carries `@ClassTag(preempt=true)`. One-time cost, amortized across the compilation.
- Fast path in `matchPreemption`: when all incumbents are extension methods and `!preemptiveNames.contains(name)`, return before any candidate enumeration. Out of the box that set contains exactly `{"withDefault"}`, so ~every DGM call short-circuits on one set lookup — restoring the allowlist's O(1) profile while keeping author self-declaration.

Two constraints worth stating: the gate must be *conservative* (false positives fall through to the full match, which fails softly; false negatives would be a correctness bug — hence deriving the set from the same scan that feeds method lookup, never a hand-maintained list), and it only covers the extension-incumbent path. Instance-method incumbents still scan the receiver's same-name overloads — bounded, and the rarer path. If measurement says that matters, the cheap fix is reordering `matchOverload` to compare arity (integer math) before scanning tag annotations (list allocation), so untagged overloads bail for free; I'd hold that until the data asks for it.

### Benchmark plan

Compile-time benchmarks are noisy, so: JMH, fresh `GroovyClassLoader`/`CompilationUnit` per invocation, compiling to `INSTRUCTION_SELECTION` (captures all STC work, skips classgen noise), average-time mode with proper warmup rather than single-shot.

**Corpus**: a generated synthetic source set (reproducible, tunable) — ~200 `@CompileStatic` classes × 50 calls, in three mixes as benchmark params:
- `dgmHeavy` — DGM calls on typed receivers (the path the gate targets);
- `instanceHeavy` — same-class/hierarchy instance calls with overloads (the path it doesn't);
- `withDefault` — calls that actually preempt, confirming the happy path doesn't regress.

**Variants**: (1) master baseline; (2) this branch as-is; (3) branch with `classTagPreemptionDisabled = ['*']` — the floor the gate could reach; (4) branch + name-set gate.

**Acceptance criteria**: if (2) is within ~1–2% of (1) on `dgmHeavy`, ship as-is and the gate becomes optional hygiene; otherwise land the gate and expect (4) ≈ (3) ≈ (1). A coarse cross-check against the existing `:performance` corpus is easy to add, but the variant comparison wants the JMH harness.

**Scope note**: I don't propose extending the gate to instance methods via per-`ClassNode` cached flags — the invalidation story during compilation (classes still being built, AST mutation) is genuinely tricky, and the bounded per-call scan is very likely in the noise. The benchmark exists to prove or refute that cheaply before adding machinery.

I'll post numbers here once the harness runs; if anyone has a preferred real-world `@CompileStatic`-heavy corpus to include alongside the synthetic one, suggestions welcome.

paulk-asert:
## Follow-up (a): done — preemption-check cost measured, gate landed

Resolving the compile-time question I flagged above. Short version: the concern was real, the gate fixes it, and it's now measured rather than asserted.

### Benchmark

Added `ClassTagResolutionBench` (JMH, in `:performance`) — a generated `@CompileStatic` corpus compiled to `INSTRUCTION_SELECTION`, fresh `CompilationUnit` per invocation. Three call mixes and a controlled `active` vs `vetoAll` comparison, where `vetoAll` sets `classTagPreemptionDisabled=['*']` so the preemption path is never entered — the pre-feature floor for a corpus whose calls all resolve. `active − vetoAll` is therefore the cost of active preemption matching. The corpus and the numbers below live in the benchmark's Javadoc so they travel with the code.

### The regression, confirmed

Without the gate, preemption matching runs on every resolved call once a token-less overload has matched:

```
mix            active   vetoAll(floor)   delta
dgmHeavy       121.8    109.7            +11.1%   <- ordinary each/collect/findAll code
instanceHeavy   52.8     51.0            + 3.6%
withDefault    263.1    228.7            +15.0%   <- preemption actually firing every call
```
*(avgt ms/op, 60 classes × 80 calls, 2 forks × 5+5 iterations.)*

The `dgmHeavy` +11% is the pathology: `each`/`collect`/`findAll` on typed receivers each pay a full DGM candidate scan + intent check, only to find nothing preemptive (no DGM method except `withDefault` declares intent). Pure overhead — well past the ~1–2% I'd have shipped as-is on.

### The gate

`ExtensionMethodCache` now exposes `getPreemptiveNames(loader)` — a per-classloader `Set<String>` of extension-method names that declare `@ClassTag(preempt=true)`, derived lazily from the already-cached method map and sharing its lifecycle (so it can never go stale independently). Out of the box it holds exactly `{"withDefault"}`. `matchPreemption` consults it before the extension candidate scan: if the call's name isn't in the set, one `HashSet.contains` replaces the whole scan+filter. Conservative by construction — a false positive falls through to the full match and fails softly; a false negative can't occur because any intent-declaring overload's name is in the set by derivation. The instance-incumbent path is untouched (it must still see subclass-preempts-superclass), and additive matching is untouched (it only runs on otherwise-unresolved calls).

### Post-gate

```
mix            active(gated)   vetoAll(floor)   pre-gate
dgmHeavy       108.1           110.3            121.8    <- regression eliminated (at the floor)
instanceHeavy   54.6            51.4             52.8
withDefault    269.7           227.0            263.1    <- unchanged; still does the real work
```

`dgmHeavy` gated `active` is back at the `vetoAll` floor, within noise — the +11% is gone. `withDefault` is deliberately unchanged: it's in the preemptive-names set, so it still does the full token synthesis + rewrite the feature exists to do (this number is a worst case — *every* call preempts, which no real code does).

### Diff and verification

Three files: the benchmark (new), `AbstractExtensionMethodCache` (+21, the name-set cache), `ClassTagSupport` (+7, the guard). `ClassTagStaticTest` 32/32, `CompilerConfigurationTest` 15/15, `Groovy11807` 5/5, and the full `groovy.transform.stc.*` (2012 tests) + `classgen.asm.sc.*` (2295 tests) sweeps green — the shared-cache change is safe.

### One documented residual

`instanceHeavy` still shows ~+6% (gated 54.6 vs floor 51.4) on a corpus where *every* call is to a same-class overloaded method. The gate deliberately doesn't cover the instance path — a per-`ClassNode` equivalent runs into invalidation-during-compilation complexity, and real instance calls are overwhelmingly non-overloaded (the scan returns one method, trivial). I've left it as a documented residual rather than adding speculative `ClassNode`-metadata caching; happy to revisit if anyone has a real-world `instanceHeavy` profile that justifies it.

src/test/groovy/groovy/transform/stc/ClassTagStaticTest.groovy
paulk-asert: See later comment.

src/main/java/org/codehaus/groovy/control/CompilerConfiguration.java
```diff
@@ -424,6 +434,20 @@ public void setLogClassgenStackTraceMaxDepth(int logClassgenStackTraceMaxDepth)
...
+     * ({@code @ClassTag(preempt=true)}) and contained to the declaring class; this set lets the
+     * consuming build veto it anyway. Entries use the same selector forms as
+     * {@code groovy.extension.disable}: a bare method name ({@code withDefault}) vetoes all
+     * overloads of that name, a signature form ({@code withDefault(Map,Class,Closure)}, simple or
+     * fully-qualified type names) vetoes the exact overload, and {@code *} vetoes all preemption.
+     * Additive injection (supplying an otherwise-mandatory token, such as for {@code asChecked})
+     * is never gated by this set. Seeded from the {@code groovy.classtag.preemption.disable}
+     * system property; empty (the default) honours declared intent. Never null.
+     */
+    private Set<String> classTagPreemptionDisabled = splitRespectingParens(getSystemPropertySafe("groovy.classtag.preemption.disable"));
```
blackdrag: I actually still do not get this part. Why do you want to override the authors intend? Assuming we compile against a library that is compiled with this enabled. Then code inside is already generated, the intend is kept. That leaves code the library user writes. Then why, if I want to prevent a problem, would I use the compiler configuration for this? And would I not maybe want to use it one time and at another callsite not? If you say this is needed (why though?) Then why for the whole codebase at once? How about an annotation to turn it off? For either a receiver (annotation on variable) or the whole method, maybe the whole class.

paulk-asert: It will be rare. It is about code that runs today, that might break, e.g.:

```
static void main() {
    Map<String, Integer> counts = [:]
    def m = counts.withDefault { 0 }
    assert m.get('a') == 0
    assert m.get(42) == 0 // was: success, now: ClassCastException
}
```

It only runs today because MapWithDefault has unsound code that grows the Map with the default when we call get. This is what GROOVY-11807 was pointing out, `get` takes Object which looks safe but has the side effect. The compiler config/flag is all about, I have code that runs today and I don't want to touch it and I want it to still run. Eventually, I might be able to change that code to not rely on the unsoundness and then I can remove the config/flag.

paulk-asert: The chain is:

- GROOVY-11807 = the bug: lenient withDefault auto-grows a wrong-typed key into a Map<Number,?> (or Map<String,Integer>), corrupting keySet() — a latent violation of the map's own declared generics. Fixed in 6.0.0-alpha-2 by the checked overload.
- The checked withDefault(Map, Class, Class, Closure) = the fix, opt-in by passing the tokens.
- GROOVY-12115 @ClassTag(preempt=true) = makes that fix reach existing static callers automatically, without them editing source.
- classTagPreemptionDisabled = a migration escape hatch to defer the fix, not a general "override the author" lever.

paulk-asert: Actually, given that it is rare, perhaps the global opt-out is enough. The thinking behind the list is that a Grails app using multiple plugins might want to upgrade some of those plugins and so be able to opt out just some of the changes plugins might make in newer versions. But if you can change some of the code, you can opt-out individual ones by swapping out of CompileStatic or whatever.

paulk-asert:
@blackdrag @daniellansun Thanks for your reviews. I reverted back to the simple global disable and believe all concerns are now addressed. That doesn't stop us from having a more elaborate disable in the future if there is demand. I plan to include it in beta-1. If we spot anything further during the release window that can't be fixed by a normal tweak in a later release, we can re-roll beta-1.

PR #2652, opened 2026-07-01
use Files.createTempFile for owner-only groovysh/groovyc temp files

netliomax25-code:
1. groovysh `/less` and `/nano` write the interactive session buffer to a temp file created with `java.io.File.createTempFile`, which under a typical umask (022) is created world-readable (`rw-r--r--`). The buffer often holds connection strings, tokens, or credentials the user has typed.
2. `Groovyc.addSourceFiles` writes the forked compilers source-file list to a temp file the same way when the command line exceeds the Windows length limit.
3. On a shared host any other local user can then read these artifacts. This is the same CWE-377/378 class already addressed for temp *directories* by moving to NIO (the CVE-2020-17521 fix, listed as property P4 in `THREAT_MODEL.md`), but these temp *files* were still using the `java.io` API.

Switched the three sites to `java.nio.file.Files.createTempFile(...).toFile()`, which yields owner-only (`rw-------`) permissions on POSIX. Behavior for the owning user is unchanged; the forked compiler and the jline pager run as the same user, so the tighter mode does not affect them.

Verified locally that `java.io.File.createTempFile` produces `rw-r--r--` under umask 022 while `Files.createTempFile` produces `rw-------`; the two modules compile cleanly.

### DIFF
--- a/subprojects/groovy-ant/src/main/java/org/codehaus/groovy/ant/Groovyc.java
+++ b/subprojects/groovy-ant/src/main/java/org/codehaus/groovy/ant/Groovyc.java
@@ -54,6 +54,7 @@ import java.io.PrintWriter;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


paulk-asert:
Merged. Thanks!

PR #2656, opened 2026-07-02
don't follow symlinks when recursively deleting a directory

netliomax25-code:
Repro: inside a directory, place a symlink pointing at another directory that lives outside the tree, then call `deleteDir()` on the outer directory. The linked-to directory's files are deleted along with the tree.

Cause: recursion is gated on `File.isDirectory()` / `Files.isDirectory(path)`, both of which follow symbolic links, so a link to a directory is treated as a subdirectory and walked into. The delete then reaches files outside the tree being removed (CWE-59).

Fix: check for a symbolic link first (`Files.isSymbolicLink` for the `File` paths, `LinkOption.NOFOLLOW_LINKS` for the `Path` path) and remove the link itself instead of following it. Applied at the three recursive-delete sites that share this behavior:

1. `ResourceGroovyMethods.deleteDir(File)`
2. `NioExtensions.deleteDir(Path)`
3. `FileSystemCompiler.deleteRecursive(File)`

Files outside the deleted tree now survive; the tree and the link entry are still fully removed, and the return value is unchanged for the ordinary (no-symlink) case. Added regression tests for the `File` and `Path` variants that skip where the platform has no symlink support.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
@@ -1743,7 +1743,14 @@ public class ResourceGroovyMethods extends DefaultGroovyMethodsSupport {
         // delete contained files
         boolean result = true;
         for (File file : files) {
-            if (file.isDirectory()) {
+            if (Files.isSymbolicLink(file.toPath())) {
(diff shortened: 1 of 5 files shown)


paulk-asert:
Merged as part of #2657. Thanks!

PR #2658, opened 2026-07-03
Reference implementation: groovy-ginq SQL/JOOQ support (GEP-26 Groovy 7)

paulk-asert:
**NOT for MERGING** just yet! This is a reference implementation for GEP-26. _I think we still want to target Groovy 7 and we're not quite ready to branch for that yet, but this gives us something to discuss about the feature boundary for what we deliver and for performance testing_.

## GEP-26: GINQ SQL backend — reference implementation (native-sql provider)

Implements the core of [GEP-26](https://groovy.apache.org/wiki/GEP-26.html): the same GINQ syntax, executed against a relational database instead of in-memory collections. See the GEP for the full design rationale, architecture, and phasing; this PR is a working subset.

```groovy
def threshold = 50000
GQL(provider: 'native-sql', dataSource: db) {
    from e in 'employees'
    join d in 'departments' on e.deptId == d.id
    where e.salary > threshold
    select e.name, d.name as deptName
}
// SELECT e.name, d.name AS deptName FROM employees e
//   INNER JOIN departments d ON e.deptId = d.id WHERE e.salary > ?
```

### What's included
- **New optional module `groovy-ginq-sql`** (excluded from groovy-all, like groovy-ginq):
  GINQ AST → SQL IR translator, ANSI renderer with a minimal `Dialect` SPI, and a thin
  runtime wrapper over `groovy.sql.Sql` returning `Queryable<GroovyRowResult>`.
- **Coverage**: GEP phases 1–2 (single-table queries, all join types, classic and
  `groupby ... into` aggregation, having, set operations, distinct, orderby/limit) plus
  phase 4 subqueries (derived tables, `exists`, `in (subquery)`, scalar subqueries,
  including correlated references) and a first tranche of function mapping
  (`UPPER`/`LOWER`/`TRIM`/`CHAR_LENGTH`/`LIKE` with escaping/`SUBSTRING`/`REPLACE`/`||`/`CASE WHEN`).
- **Parameter safety**: the SQL text is a compile-time constant; every literal and captured
  variable becomes a positional bound parameter (see the GEP's security section).
- **Strict mode**: untranslatable constructs are compile-time errors with source positions.
- **Small additive changes to groovy-ginq core**: `provider`/`dataSource` options, a
  ServiceLoader-based `GinqProvider` registry, raw option-expression plumbing on the
  (incubating) `GinqAstVisitor` SPI, optimizer gating for non-collection providers, and a
  `@GQ(provider=...)` member.

### Not in this PR (per GEP phasing)
jOOQ provider (phase 3), mixed-mode fallback (phase 5), window functions, and dialect breadth.

### Notes for reviewers
- Trailing `orderby`/`limit` on a set operation applies to the **combined result** (SQL
  semantics), unlike the collection provider which sorts the final operand before combining.
  Flagged in the userguide; happy to adjust if the divergence is unwanted.
- Tests: renderer unit tests, generated-SQL shape tests, and end-to-end tests against
  in-memory HSQLDB (already in the dependency-verification metadata — no new artifacts).
- Because GINQ macros expand at test-compile time, `groovy-ginq-sql/build.gradle` declares
  the provider classes as an input to `compileTestGroovy` so macro-using tests aren't served
  stale from the build cache.


daniellansun:
LGTM so far

PR #2666, opened 2026-07-05
GROOVY-12131: OptimizerVisitor: detect constant nodes shared across c…

paulk-asert:
…lasses

### DIFF
--- a/src/test/groovy/org/codehaus/groovy/control/OptimizerVisitorTest.groovy
+++ b/src/test/groovy/org/codehaus/groovy/control/OptimizerVisitorTest.groovy
@@ -20,9 +20,11 @@ package org.codehaus.groovy.control
 
 import org.codehaus.groovy.ast.ClassHelper
 import org.codehaus.groovy.ast.ClassNode
+import org.codehaus.groovy.ast.CodeVisitorSupport
 import org.codehaus.groovy.ast.Parameter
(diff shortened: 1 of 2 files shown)


src/test/groovy/org/codehaus/groovy/control/OptimizerVisitorTest.groovy
```diff
@@ -0,0 +1,86 @@
...
+import static org.codehaus.groovy.ast.tools.GeneralUtils.stmt
+
+/**
+ * GROOVY-12131: constant caching communicates the assigned {@code $const$} field name by
+ * mutating the {@link ConstantExpression}, which assumes each node is reachable from exactly
+ * one class. An AST transform that aliases a node into a second class makes the classes
+ * overwrite each other's stamp, so one class silently reads the wrong cached constant at
+ * runtime. The visitor now warns when it re-stamps a node with a different field name.
+ */
+class OptimizerVisitorTest {
```
blackdrag, line 41: I think there should be also a script based test of the fix. This is AST level only and may be subject to change, a simple script failing before and now being correct exhibits the problem and will be kept in case we solve this differently in the future

paulk-asert: added

PR #2669, opened 2026-07-05
GROOVY-12133: Implement findBalancedGroups utility to support Regex Balanced Groups

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12133

### DIFF
--- a/src/main/java/groovy/util/regex/BalancedGroup.java
+++ b/src/main/java/groovy/util/regex/BalancedGroup.java
@@ -1,0 +1,488 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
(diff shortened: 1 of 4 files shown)


daniellansun:

paulk-asert (approved):
LGTM. AI suggestions:

> Small nits on the PR itself (none blocking)
>
>  - Two stray trailing blank lines added at EOF of StringGroovyMethods.java.
>  - The two public BalancedGroup constructors mutate each child's parent field (one-shot "child already has a parent" contract). For a result type whose intended entry point is find(...), exposing constructors that carry hidden wiring side-effects is a mild API smell — could be package-private, though they are documented and tested.
>  - getDepth() is an O(depth) ancestor walk per call; fine at expected nesting.
>
> Bottom line: solid, well-tested user-facing GDK addition.

blackdrag (approved):

PR #2673, opened 2026-07-07
GROOVY-12137: experimental reflective cold tier for indy dispatch (sp…

paulk-asert:
…ike)

Behind the groovy.indy.cold.reflection flag (default: off), plain method calls dispatch reflectively while a call site is cold, deferring all MethodHandle chain construction — and its one-time LambdaForm creation cost — to hit-count promotion. The shared cold dispatcher has a single (Object[])Object shape for every call site, arity, and primitive pattern; promotion installs the unchanged full guarded chain, so the hot path is identical to master.

Plain calls are public, non-static, non-category CachedMethods on public classes with a real receiver and no spread/safe-null/interceptor semantics; caller-sensitive targets (including unannotated caller-context-sensitive cases like object serialization, and interface selections resolving to sensitive implementations) stay on the full path. Selection reuses Selector's existing logic via a new selection-only hook; validity is re-checked per call with plain-Java equivalents of the chain's guards. A per-wrapper cumulative counter promotes polymorphic receiver shapes that the consecutive-hit counter never catches.

Measured (deterministic class-load counters + fresh-JVM timing): ~2/3 of per-shape LambdaForm cost removed on cold dispatch workloads, 1.34-1.40x faster cold dispatch with many cold sites, steady state unchanged after promotion, MOP-heavy test subset (3437 tests) green with the flag on.


src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
blackdrag, line 1: this is marked as binary file somehow?

src/main/java/org/codehaus/groovy/vmplugin/v8/ColdReflectiveMethodHandleWrapper.java
```diff
@@ -0,0 +1,203 @@
...
+            if (hasCallerSensitiveAnnotation(method)) return true;
+            if (declaringClass.isInterface() || Modifier.isAbstract(method.getModifiers())) {
+                // virtual dispatch may land on a caller-sensitive implementation
+                Method implementation = receiverClass.getMethod(method.getName(), method.getParameterTypes());
+                return hasCallerSensitiveAnnotation(implementation);
+            }
+            return false;
+        } catch (Throwable ignore) {
+            return true;
+        }
```
blackdrag: I think this, if we keep it, is an information that should exist on Cached method and reused there to not to have to pay the information on a per callsite base

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -1226,6 +1251,26 @@ void buildInvokeHandle() {
                 addExceptionHandler();
             }
         }
+
+        @Override
+        MetaMethod selectForColdReflection() {
+            if (safeNavigation || spread || args[0] == null
+                    || args[0] instanceof Class || args[0] instanceof GroovyInterceptable) {
+                return null;
+            }
```
blackdrag, line 1273: spread and safeNavidation should be always false for CastSelector. Checking for Class looks strange here. Why not look for a asType method on Class added by the metaclass? Also why is null excluded?

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
paulk-asert: there was a \0000 char in the file - fixed now

src/main/java/org/codehaus/groovy/vmplugin/v8/ColdReflectiveMethodHandleWrapper.java
paulk-asert: changed

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: AI read: That hook is only reached for CallType.METHOD — the fallback gates on it, so CastSelector/InterfaceSelector are never asked; I've added a defensive check plus comments since that wasn't visible from the diff. The instanceof Class test isn't about asType/cast: a Class receiver means a static-method-style call, and those promote on their first hit anyway (GROOVY-11935), so the reflective tier would buy nothing. Null receivers mean NullObject MOP dispatch, which I kept out of the plain-call scope for the spike — supportable later if the numbers justify it.

blackdrag: CastSelector is not supposed to be used for CallType.METHOD, also there are no static method calls for CastSelector. CastSelector is for non-trivial casts that may result in asType calls. if the receiver is Class it means we are dealing with something like "(T) c" where c is an instance of Class.  It is not for a normal method call to asType(Class). Things left out of the spike are ok, if the PR is not intended for merge. Otherwise it should have at the very least a TODO.

blackdrag: sorry, I think I was too fast and mixed up things. The potential method call is not asType, but asBoolean.

paulk-asert:
@blackdrag updated - please reassess

src/main/java/org/codehaus/groovy/vmplugin/v8/ColdReflectiveMethodHandleWrapper.java
```diff
@@ -0,0 +1,196 @@
...
+            // virtual dispatch may land on a caller-sensitive implementation
+            try {
+                Method implementation = receiverClass.getMethod(method.getName(), method.getParameterTypes());
+                CachedMethod implCached = CachedMethod.find(implementation);
+                if (implCached != null) return implCached.isCallerSensitive();
+            } catch (Throwable ignore) {
+            }
+            return true; // could not resolve the implementation: stay safe
+        }
+        return false;
```
blackdrag, line 191: I don't understand why we are doing this like this. Why can all this information not be behind cm.isCallerSensitive()? At least the part that depends on the declaring class.  Then... the interface and abstract method case. We really end up with those? furthermore I think this does not handle default methods, well I should say it may marks them as callsersitive even if they are not.

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
blackdrag, line 1: There is still no diff on the file and it is marked as binary

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -1226,6 +1251,35 @@ void buildInvokeHandle() {
                 addExceptionHandler();
             }
         }
+
+        @Override
+        MetaMethod selectForColdReflection() {
+            // defensive: the cold tier is only for plain method calls; the
+            // fallback gates on CallType.METHOD, so subclasses (CastSelector,
+            // InterfaceSelector) are never asked, but make it explicit here
+            if (callType != CallType.METHOD) return null;
```
blackdrag: I would prefer a GroovyBugException. Like this it could hide a problem introduced by a change.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -1226,6 +1251,35 @@ void buildInvokeHandle() {
...
+            // these all need the full handle path: spread and safe-null have
+            // their own adaptation/constant handles; a null receiver means
+            // NullObject MOP dispatch; a Class receiver is a static-method
```
blackdrag: why do we need to dispatch through the normal MOP for NullObject? NullObject has methods like any other class, only the receiver is null.  Maybe it does not have to be part of this PR, but I would mark its investigation as TODO

src/main/java/org/codehaus/groovy/vmplugin/v8/IndyInterface.java
paulk-asert: (binary file): The head side is already clean — the NUL lives in the base blob on master (a literal \0 embedded in the GEP-15 COMPOUND_ASSIGN javadoc to illustrate NAME_SEPARATOR), so git treats the base as binary and no diff can render. Pushed the one-line javadoc fix to master separately; after the next rebase this file will diff normally.

src/main/java/org/codehaus/groovy/vmplugin/v8/ColdReflectiveMethodHandleWrapper.java
paulk-asert: (isCallerSensitive): Restructured as you suggest: everything method-determined — the annotation probe and the declaring-class cases like serialization — is now behind cm.isCallerSensitive(), cached per method. And yes, abstract selections really occur: the metaclass picks ObjectInput.readObject() for a call on an ObjectInputStream (that exact case was found by a failing serialization test), and the sensitivity then belongs to the receiver's implementation, which is the one receiver-dependent check left in the wrapper. Good catch on defaults: the old isInterface() clause could conservatively mark a default method when implementation lookup failed — now non-abstract interface methods are probed directly, and a default method verifiably takes the reflective tier.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: (GroovyBugError): Agreed and done — reaching this with a non-METHOD call type would mean the fallback's gate broke, so it now throws GroovyBugError.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: (NullObject): You're right — nothing fundamental prevents serving NullObject's methods reflectively; the exclusion was spike conservatism. Marked as TODO as suggested.

paulk-asert:
@blackdrag Updated - ready for another assessment if you have time.

blackdrag (approved):

PR #2674, opened 2026-07-08
GROOVY-12138: emit invokedynamic call sites for property writes (spike)

paulk-asert:
Property writes were the last un-cached dispatch surface: even in indy mode they compiled to static ScriptBytecodeAdapter.setProperty calls, paying getMetaClass() plus a full MetaClassImpl.setProperty resolution walk on every write, in both indy and classic modes.

Behind the compile-time flag groovy.indy.setproperty (default: off), AsmClassGenerator now diverts plain (non-safe, non-spread) property writes to a new CallSiteWriter.makeSetPropertySite hook. The default implementation is the classic adapter call, so classic mode and flag-off builds are unchanged; IndyCallSiteWriter overrides it to emit invokedynamic setProperty:(ReceiverType, ValueType)V with the receiver first, preserving the runtime's arguments[0]-is-receiver invariant.

The new SetPropertySelector (replacing the 'not supported' GroovyBugError for CallType.SET) fast-paths only shapes whose resolution is provably sender- and metaclass-independent: a public setter whose backing field is private or absent (GROOVY-8283 field-precedence is sender-dependent), or a public non-final field, each accepting the runtime value type directly. Everything else — Maps (GROOVY-8065/11367), Class receivers, categories, overridden/mixed-in/expando setProperty (detected by both a reflection probe and a metaclass pickMethod probe; ExpandoMetaClass is excluded wholesale, mirroring the GET selector), value coercion, listeners, propertyMissing — binds the sender-aware ScriptBytecodeAdapter.setProperty as the call-site handle, making non-fast-path behavior identical to the classic path by construction. Guards, PIC, and hit-count promotion come free from the existing call-site machinery; the value participates in the argument-class guards so type changes at a site re-select correctly.

Measured: ~42x on a hot property-write loop (6M mixed setter/field writes: ~435ms -> ~10ms, written state consumed and verified). Full root test suite (15678 tests) passes with the flag enabled for all test-code compilation; a 12-scenario semantic stress produces identical output compiled with and without the flag.

### DIFF
--- a/src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
+++ b/src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
@@ -85,7 +85,6 @@ import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.MET
 import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.META_METHOD_INVOKER;
 import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.META_PROPERTY_GETTER;
 import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.MOP_GET;
-import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.SBA_SET_PROPERTY;
 import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.MOP_INVOKE_CONSTRUCTOR;
 import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.MOP_INVOKE_METHOD;
 import static org.codehaus.groovy.vmplugin.v8.IndyGuardsFiltersAndSignatures.NON_NULL;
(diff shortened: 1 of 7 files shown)


src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -441,6 +444,157 @@ public void setMetaClassCallHandleIfNeeded(boolean standardMetaClass) {
...
+                if (field != null) {
+                    setFieldWriteHandle(field, value);
+                }
+            } else if (mp instanceof CachedField cf && !mp.isStatic()) {
+                setFieldWriteHandle(cf, value);
+            }
+            // otherwise (no meta property, listeners, expando, propertyMissing, ...): adapter path
+        }
+
+        private static final Class<?>[] SET_PROPERTY_PARAMS = {String.class, Object.class};
```
blackdrag: just for style, this should be before the methods. Nothing blocking in my eyes

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -441,6 +444,157 @@ public void setMetaClassCallHandleIfNeeded(boolean standardMetaClass) {
...
+        @Override
+        public void setMetaClassCallHandleIfNeeded(boolean standardMetaClass) {
+            if (handle != null) return;
+            useMetaClass = true;
+            if (LOG_ENABLED) LOG.info("set adapter invocation path for property set.");
+            // SBA_SET_PROPERTY: (Object value, Class sender, Object receiver, String name)void
+            MethodHandle sba = MethodHandles.insertArguments(SBA_SET_PROPERTY, 1, sender);
+            sba = MethodHandles.insertArguments(sba, 2, name); // -> (Object value, Object receiver)void
+            handle = MethodHandles.permuteArguments(sba,
+                    MethodType.methodType(void.class, Object.class, Object.class), 1, 0); // -> (receiver, value)void
```
blackdrag: I would consider not using SBA directly here. Instead I would make a method maybe even in the the selector, that has the arguments already in the right order and allows you to use a single insertArguments to inject sender and name and has the right remaining signature for the invocation. For example (name, receiverClass, receiver, value). Then one insertArguments call would be enough. You can still let that one call SBA... though I imagine we want to inline the remaining logic there in the long run.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: moved

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: changed if I understood you correctly

blackdrag (approved):

daniellansun:
Merged. Thanks.

PR #2679, opened 2026-07-09
GEP-27: compact non-capturing lambda compilation (opt-in)

paulk-asert:
A statically compiled lambda targeting a functional interface currently generates a per-lambda class extending groovy.lang.Closure, even though its runtime value is a LambdaMetafactory-produced functional-interface instance and (when non-capturing) a zero-allocation singleton. The generated class only holds a static doCall.

Like the JVM does for Java lambdas, hoist the body of a non-capturing, non-serializable SAM lambda into a private static method on the enclosing class and bootstrap LambdaMetafactory directly against it, so no per-lambda class is generated. Capturing, instance-accessing, serializable and nested cases are unchanged (they keep the generated lambda class).

Behaviour is preserved (verified: Predicate/BiFunction, lambdas in static methods, qualified outer static calls, serializable round-trips, singleton identity); stack traces now name a $lambda$ method on the enclosing class (Java-like).

Gated by an opt-in system property, groovy.target.lambda.hoist (default off), read per lambda so it can be toggled without a JVM restart, while IDE/tooling compatibility with the changed generated-class shape is verified. Default off keeps existing behaviour and tests unchanged.

See GEP-27 (Compact Closure and Lambda Compilation) for the broader design.

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesLambdaWriter.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesLambdaWriter.java
@@ -18,9 +18,11 @@
  */
 package org.codehaus.groovy.classgen.asm.sc;
 
+import org.apache.groovy.util.SystemUtil;
 import org.codehaus.groovy.GroovyBugError;
 import org.codehaus.groovy.ast.ClassHelper;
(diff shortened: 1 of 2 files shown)


paulk-asert:
superseded by #2680

PR #2681, opened 2026-07-10
fix out-of-bounds read on dangling escape in decodeJsonString

netliomax25-code:
1. In `CharBuf.decodeJsonString` the backslash branch is guarded by `if (index < to)`, a condition the enclosing for-loop already guarantees, so a backslash in the final position of the string slice is treated as the start of an escape and the following `chars[index]` read runs one element past the slice.
2. When a JSON string is unterminated and ends in a lone backslash, `findEndQuote` returns the buffer length, so the over-read is `chars[to]`. This routine is the single escape-decoding sink for every `JsonSlurper` parser type; on the eager default (`CHAR_BUFFER`) parser a two-character document (a quote then a backslash, `new JsonSlurper().parseText('"\\')`) raises a raw `ArrayIndexOutOfBoundsException` instead of the documented `JsonException`, so callers that guard untrusted input with `catch (JsonException)` miss it.

Changed the guard to require a character after the backslash and made a dangling trailing escape raise `JsonException`, matching how the switch already handles other invalid escapes. Added a regression test covering the parser types and confirming valid escapes still decode.

### DIFF
--- a/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/CharBuf.java
+++ b/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/CharBuf.java
@@ -868,7 +868,7 @@ public class CharBuf extends Writer implements CharSequence {
         for (int index = start; index < to; index++) {
             char c = chars[index];
             if (c == '\\') {
-                if (index < to) {
+                if (index + 1 < to) {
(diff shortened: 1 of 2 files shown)


paulk-asert:
Merged. Thanks!

PR #2687, opened 2026-07-10
GROOVY-12146: order reflection calls to ensure Annotation members are ordered

jdaugherty:
Fixes https://issues.apache.org/jira/browse/GROOVY-12146 by canonicalizing at the source of the nondeterminism instead of at emission (the reverted 4.x approach also reordered source-compiled annotations, which stay in deterministic source order and don't need touching):

// Java8#configureAnnotation
declaredMethods = type.getDeclaredMethods();
Arrays.sort(declaredMethods, Comparator.comparing(Method::getName));

Annotation member methods cannot overload, so name gives a total order. Verified locally against GROOVY_5_0_X: the woven class emits a stable sorted order independent of the JVM, source-compiled annotations are unchanged, and TypeAnnotationsTest plus the trait transform suites pass.


paulk-asert:
LGTM

PR #2689, opened 2026-07-11
decode process output streams using the native encoding

netliomax25-code:
1. getText(Process) and TextDumper (which backs consumeProcessOutput, waitForProcessOutput and the new waitForResult/ProcessResult) decode the process output stream through new InputStreamReader(stream) with no charset, so they use Charset.defaultCharset().
2. since JEP 400 that default is UTF-8 regardless of platform, while a child process still writes its output in the OS native encoding, so on a non-UTF-8 host (a Windows code page, or Linux under a C/legacy locale) non-ASCII output no longer round-trips and is silently corrupted (a native 0xE9 byte for é decodes to U+FFFD).

Decode with the native.encoding system property (JDK 17+, the documented way to obtain the OS native encoding), falling back to Charset.defaultCharset() when it is unavailable. UTF-8 hosts are unaffected.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ProcessGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ProcessGroovyMethods.java
@@ -33,6 +33,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Writer;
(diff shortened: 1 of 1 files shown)


paulk-asert:
Superseded by #2700 with credit given in that issue.

PR #2692, opened 2026-07-11
GROOVY-12151: hoist eligible closure bodies to a shared adapter @Pack…

paulk-asert:
…edClosures/flag off by default (GEP-27 S1)

Add closure packing on top of the SAM-lambda hoisting: an eligible closure literal's body is hoisted into a synthetic method on the enclosing class and the literal is replaced by a shared PackedClosure adapter, removing the per-closure generated class (and the nested $_closure1$_closure2 name explosion) while the value stays a real groovy.lang.Closure.

Capability analysis (PackStrategy, ClosureWriter.chooseStrategy):
- a single decision point selects FULL_CLASS (S0) vs PACKED_ADAPTER (S1);
- triggered by the @PackedClosures annotation (trust path, dynamic or static), or automatically under @CompileStatic when the type checker PROVES the closure delegate-independent -- every implicitly-received name resolved against the owner, read from STC's IMPLICIT_RECEIVER paths (a path ending in "owner" is safe; "delegate", from @DelegatesTo/with, is not). Behind groovy.target.closure.pack.

Typed static dispatch (phase 4): the hoisted body reuses the expressions StaticCompilationVisitor already annotated, so under @CompileStatic it compiles statically -- read-only captures pass as typed leading parameters, inferred types (@ClosureParams, implicit it) arrive via metadata, and the method takes the closure's inferred return type. Packed bodies are statically dispatched with zero invokedynamic, often fewer indy sites than a closure class (which Reference-boxes captures as Object).

Written captures ride the compiler's holder machinery: the shared Reference is passed as a parameter (originType/closure-shared/UseExistingReference), so the body is not rewritten, compiles statically, and supports every write form (compound ops beyond +=, postfix value semantics).

Soundness for the dynamic trust path:
- PackedClosure runtime guard fails fast if a delegate / delegate-consulting resolveStrategy is set on a trust-packed closure; a statically proven closure tolerates it as a no-op (as a @CompileStatic closure class does today);
- compile-time escape decline keeps closures that visibly escape (stored to a field/property, returned, appended, in a collection literal) as classes.

Mechanism fidelity so a packed closure behaves like a generated closure class: correct owner/static dispatch in static methods; getParameterTypes() carries the real declared types; arguments are adapted as reflective dispatch would (vararg collection, single List/Tuple destructuring, per-arg coercion Closure->SAM and GString->String); call() dispatches straight to doCall with invoker-exception unwrapping; dispatch via a cached reflective Method (metaclass coercion would auto-dereference Reference holder arguments).

ClassNode.visitMethods: dedup newly-added methods by identity via an IdentityHashMap set (O(1)) instead of ArrayList.removeAll (O(n^2)). Visiting a method during code generation can add more (constructor-ref synthetic, lambda $deserializeLambda$, a hoisted body that hoists a nested one); the O(1) fixpoint visits them all and fixes a compile-time freeze the O(n^2) scan caused against method-generating visitors (e.g. groovy-contracts).

Tests: ClosurePackCapabilityTest and PackedClosuresTransformTest (proof, declines, auto-pack, zero-indy typed bodies, write forms, destructuring). Flag-on stress net 0 failures across LambdaTest + all *Closure*/*DelegatesTo* static-compile suites + ReproducibleBytecodeBugs; flag-off byte-identical (778 green); groovy-contracts compiles and tests clean.

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
@@ -18,12 +18,15 @@
  */
 package org.codehaus.groovy.classgen.asm;
 
+import groovy.transform.PackedClosures.PackMode;
 import org.codehaus.groovy.GroovyBugError;
+import org.codehaus.groovy.ast.AnnotationNode;
 import org.codehaus.groovy.ast.ClassHelper;
 import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.CodeVisitorSupport;
 import org.codehaus.groovy.ast.ConstructorNode;
 import org.codehaus.groovy.ast.FieldNode;
+import org.codehaus.groovy.ast.PropertyNode;
 import org.codehaus.groovy.ast.InnerClassNode;
 import org.codehaus.groovy.ast.MethodNode;
 import org.codehaus.groovy.ast.Parameter;
@@ -32,39 +35,62 @@ import org.codehaus.groovy.ast.VariableScope;
 import org.codehaus.groovy.ast.expr.ClassExpression;
 import org.codehaus.groovy.ast.expr.ClosureExpression;
 import org.codehaus.groovy.ast.expr.ConstructorCallExpression;
+import org.codehaus.groovy.ast.expr.ArrayExpression;
+import org.codehaus.groovy.ast.expr.BinaryExpression;
 import org.codehaus.groovy.ast.expr.Expression;
 import org.codehaus.groovy.ast.expr.FieldExpression;
+import org.codehaus.groovy.ast.expr.ListExpression;
+import org.codehaus.groovy.ast.expr.MapEntryExpression;
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
+import org.codehaus.groovy.ast.expr.PropertyExpression;
+import org.codehaus.groovy.ast.expr.PostfixExpression;
+import org.codehaus.groovy.ast.expr.PrefixExpression;
 import org.codehaus.groovy.ast.expr.TupleExpression;
 import org.codehaus.groovy.ast.expr.VariableExpression;
 import org.codehaus.groovy.ast.stmt.BlockStatement;
+import org.codehaus.groovy.ast.stmt.ExpressionStatement;
+import org.codehaus.groovy.ast.stmt.ReturnStatement;
+import org.codehaus.groovy.ast.stmt.Statement;
 import org.codehaus.groovy.ast.tools.GenericsUtils;
 import org.codehaus.groovy.classgen.AsmClassGenerator;
+import org.codehaus.groovy.control.messages.WarningMessage;
+import org.codehaus.groovy.syntax.Types;
 import org.objectweb.asm.MethodVisitor;
 
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static org.apache.groovy.ast.tools.ClassNodeUtils.addGeneratedMethod;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.args;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.callThisX;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.callX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.constX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.ctorSuperX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.nullX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.returnS;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.stmt;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.varX;
+import static org.codehaus.groovy.transform.sc.StaticCompilationMetadataKeys.STATIC_COMPILE_NODE;
 import static org.codehaus.groovy.transform.stc.StaticTypesMarker.INFERRED_RETURN_TYPE;
 import static org.objectweb.asm.Opcodes.ACC_FINAL;
 import static org.objectweb.asm.Opcodes.ACC_PRIVATE;
 import static org.objectweb.asm.Opcodes.ACC_PUBLIC;
 import static org.objectweb.asm.Opcodes.ACC_STATIC;
+import static org.objectweb.asm.Opcodes.AASTORE;
 import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
+import static org.objectweb.asm.Opcodes.ACONST_NULL;
 import static org.objectweb.asm.Opcodes.ALOAD;
+import static org.objectweb.asm.Opcodes.ANEWARRAY;
 import static org.objectweb.asm.Opcodes.DUP;
 import static org.objectweb.asm.Opcodes.GETFIELD;
 import static org.objectweb.asm.Opcodes.INVOKESPECIAL;
@@ -94,6 +120,17 @@ public class ClosureWriter {
     /** The controller coordinating all bytecode writers for the current class. */
     protected final WriterController controller;
     private final Map<Expression, ClassNode> closureClasses = new HashMap<>();
+    private int packedCounter;
+    // Closures found in a syntactic position that lets them outlive the method are cached per method.
+    private final Map<MethodNode, Set<ClosureExpression>> escapingClosuresByMethod = new HashMap<>();
+    // Names assigned anywhere in a method (excluding declarations), cached per method: a capture must
+    // be Reference-threaded if it is written ANYWHERE in the enclosing method, not just inside the
+    // closure -- e.g. `def fib; fib = { n -> ... fib(n-1) ... }` writes fib after the closure is
+    // constructed, so a by-value capture would see the stale (null) value.
+    private final Map<MethodNode, Set<String>> writtenNamesByMethod = new HashMap<>();
+
+    /** Name prefix of the synthetic methods holding hoisted closure bodies. */
+    private static final String PACKED_METHOD_PREFIX = "$packed$closure$";
 
     /**
      * Creates a closure writer with the given controller.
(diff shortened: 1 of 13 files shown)


src/main/java/groovy/transform/PackMode.java
```diff
@@ -0,0 +1,47 @@
...
+ * {@link #STRICT} all pack, differing only in how they report a closure that could <em>not</em> be
+ * packed (declines are otherwise silent). {@link #DISABLED} is the opposite: it opts the scope out
+ * of packing entirely, so the most-specific declaration wins — a {@code DISABLED} method inside a
+ * packed class, or vice versa, and it overrides the automatic {@code groovy.target.closure.pack} flag
+ * as well.
+ *
+ * @see PackedClosures#mode()
+ */
+@Incubating
+public enum PackMode {
```
blackdrag: suggestion: make this part of PackedClosure

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +197,533 @@ public void writeClosure(final ClosureExpression expression) {
...
+     */
+    private PackStrategy chooseStrategy(final ClosureExpression expression) {
+        if (!isPackable(expression)) return PackStrategy.FULL_CLASS;
+        // The most-specific @PackedClosures mode in scope (method wins over class), or null if not
+        // annotated. DISABLED is a fine-grained opt-out that beats an enclosing opt-in AND the flag.
+        PackMode mode = annotatedMode();
+        if (mode == PackMode.DISABLED) return PackStrategy.FULL_CLASS;
+        boolean annotated = (mode != null); // LENIENT/WARN/STRICT all opt in (DISABLED handled above)
+        boolean triggered;
+        if (isStaticCompilation()) {
```
blackdrag: You have a class and method that is special only for the subclass, while basically encoding the name and resulting logic of that subclass in the parent class. I think the name should be more general and the logic should be moved into that method call. For example:  `triggered = annotated || triggerClosurePacking(expression)` This class would then return false and StaticTypesClosureWriter would return the SystemUtil and isDelegateIndependent call.  Also isDelegateIndependent exist only for this purpose here, so it should be moved as well

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +197,533 @@ public void writeClosure(final ClosureExpression expression) {
...
+        }
+        return captured;
+    }
+
+    /** Captured variables that are written (reassigned) in the body — these need Reference threading. */
+    // Names assigned anywhere in a method (excluding declarations), cached per method: a capture must
+    // be Reference-threaded if it is written ANYWHERE in the enclosing method, not just inside the
+    // closure -- e.g. `def fib; fib = { n -> ... fib(n-1) ... }` writes fib after the closure is
+    // constructed, so a by-value capture would see the stale (null) value.
+    private final Map<MethodNode, Set<String>> writtenNamesByMethod = new HashMap<>();
```
blackdrag: field in between code. You should maybe add to avoid this as rule to your AI or  it is going to continue doing this.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +197,533 @@ public void writeClosure(final ClosureExpression expression) {
...
+    // the shared adapter cannot reproduce, so it declines like the other closure pseudo-properties.
+    private static final Set<String> FORBIDDEN_CLOSURE_NAMES =
+            new HashSet<>(java.util.Arrays.asList("owner", "delegate", "thisObject", "directive",
+                    "resolveStrategy", "metaClass"));
+
+    /** Accessor/mutator forms of the same pseudo-properties, called with implicit this. */
+    private static final Set<String> FORBIDDEN_CLOSURE_CALLS =
+            new HashSet<>(java.util.Arrays.asList("getOwner", "getDelegate", "getThisObject", "getDirective",
+                    "getResolveStrategy", "setDelegate", "setDirective", "setResolveStrategy",
+                    "getMaximumNumberOfParameters", "getParameterTypes", "getMetaClass", "setMetaClass"));
```
blackdrag: how about invokeMethod, get/setProperty?

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +197,533 @@ public void writeClosure(final ClosureExpression expression) {
...
+                methodParams, ClassNode.EMPTY_ARRAY, body);
+        // Added after the ReturnAdder phase, so run it ourselves: it wires implicit returns through
+        // every statement shape (trailing expressions, if/else branches, try/catch, ...).
+        new org.codehaus.groovy.classgen.ReturnAdder().visitMethod(hoisted);
+        // Typed static dispatch (GEP-27 phase 4): the original body expressions were already visited
+        // by StaticCompilationVisitor as part of the enclosing method, so they carry the metadata the
+        // static writer needs (DIRECT_METHOD_CALL_TARGET, inferred types) -- including types that were
+        // only inferred (@ClosureParams, implicit it), which arrive for free via checkcasts. Written
+        // captures ride the holder machinery, so their bodies compile statically too.
+        hoisted.putNodeMetaData(STATIC_COMPILE_NODE, typedBody ? Boolean.TRUE : Boolean.FALSE);
```
blackdrag, line 636: This is a strong indicator for code that should be in StaticTypesClosureWriter. You should investigate if this method and actually all the methods added in this class really should be in this class or in StaticTypesClosureWriter

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +197,533 @@ public void writeClosure(final ClosureExpression expression) {
...
+            mv.visitInsn(AASTORE);
+            os.remove(3);
+        }
+        // strict guard only for the dynamic trust path; a statically PROVEN delegate-independent
+        // closure stores-and-ignores a caller-set delegate, like a @CompileStatic closure class.
+        mv.visitInsn(isStaticCompilation() ? org.objectweb.asm.Opcodes.ICONST_0 : org.objectweb.asm.Opcodes.ICONST_1);
+        os.push(ClassHelper.boolean_TYPE);
+        mv.visitMethodInsn(INVOKESPECIAL, pc, "<init>",
+                "(Ljava/lang/Object;Ljava/lang/String;[Ljava/lang/Object;[Ljava/lang/Class;Z)V", false);
+        os.replace(ClassHelper.CLOSURE_TYPE, 5); // owner, name, captured[], paramTypes[], strict -> Closure
```
blackdrag: OperandStack is not merely for book keeping. It's main purpose is to support casting, provide a more easy way to handleswap with different types, that you do not really know and with casts for types you do not know.  If all you do is creating some Object[] and load references into it then I doubt it is needed. So my question here is: If you remove the usages of os, what would break?

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +197,533 @@ public void writeClosure(final ClosureExpression expression) {
...
+            BytecodeHelper.pushConstant(mv, i);
+            os.push(ClassHelper.int_TYPE);
+            BytecodeHelper.visitClassLiteral(mv, erasedType(closureParams[i].getType()));
+            os.push(ClassHelper.CLASS_Type);
+            mv.visitInsn(AASTORE);
+            os.remove(3);
+        }
+        // strict guard only for the dynamic trust path; a statically PROVEN delegate-independent
+        // closure stores-and-ignores a caller-set delegate, like a @CompileStatic closure class.
+        mv.visitInsn(isStaticCompilation() ? org.objectweb.asm.Opcodes.ICONST_0 : org.objectweb.asm.Opcodes.ICONST_1);
```
blackdrag: Again, this looks like mixing up responsibilities. Plus, why is strict mode SC only and default for SC? Why is not default in general?

src/main/java/org/codehaus/groovy/classgen/asm/PackStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /** S1 — hoist the body to a method on the enclosing class behind a shared {@code PackedClosure} adapter. */
+    PACKED_ADAPTER,
+
+    /** S2 — hoist the body and bootstrap {@code LambdaMetafactory} against it; no generated class (SAM targets). */
+    METAFACTORY,
+
+    /** S3 — neither escapes nor needs an object; elide it entirely. */
+    ELIDE
+}
```
blackdrag, line 35: Are the last two even used?

src/test/groovy/org/codehaus/groovy/transform/PackedClosuresTransformTest.groovy
```diff
@@ -0,0 +1,306 @@
...
+    void compileStaticPacksResultPassedAsArgument() {
+        String passedAsArg = '''import groovy.transform.CompileStatic
+            @CompileStatic
+            @groovy.transform.PackedClosures
+            class Z {
+                int need(List<Integer> ys) { ys.sum() as int }
+                int m(List<Integer> xs)    { need(xs.collect { Integer it -> it * 2 }) }
+            }'''
+        assertEquals(['Z'], generatedClassNames(passedAsArg))   // closure packs, no class remains
+        assertEquals(12, instance(passedAsArg + '\n new Z()').m([1, 2, 3]))
```
blackdrag, line 211: why is it compilestatic only and if it is, then we should consider splitting the test in a static version and non-static version. Of course same for the other tests in here

src/main/java/org/codehaus/groovy/runtime/PackedClosure.java
```diff
@@ -0,0 +1,243 @@
...
+    // Dispatch call() straight to doCall: the generic varargs doCall is otherwise vararg-ambiguous
+    // through the metaclass when a single argument is itself an Object[] (the metaclass would treat
+    // the element as the whole argument list and spread it). Unwrap invoker exceptions exactly as
+    // Closure.call does, so user exceptions surface unwrapped.
+    @Override
+    public Object call(final Object... args) {
+        try {
+            return doCall(args);
+        } catch (InvokerInvocationException e) {
+            org.apache.groovy.internal.util.UncheckedThrow.rethrow(e.getCause());
```
blackdrag, line 140: Why can this Exception happen and if it does, why do we have to unpack it here? This looks kind of wrong to me

src/main/java/org/codehaus/groovy/runtime/PackedClosure.java
```diff
@@ -0,0 +1,243 @@
...
+        // Normalise the visible arguments to the closure's declared arity: pad missing with
+        // null (implicit-it / under-application) and drop extras (over-application), matching
+        // how a normal Groovy closure tolerates arity mismatches from callers such as each().
+        Object[] all = new Object[captured.length + arity];
+        System.arraycopy(captured, 0, all, 0, captured.length);
+        for (int i = 0; i < arity; i++) {
+            all[captured.length + i] = (i < provided.length) ? provided[i] : null;
+        }
+        return invokeHoisted(all);
+    }
```
blackdrag, line 208: This doCall method is an inlining nightmare. The idea of doCall was to provide a fast way for the MOP to shortcut the call to doCall without ballast in doCall itself. a dynamic call to call will not even dispatch to call and instead go directly got to doCall in case of a GeneratedClosure. There is even a special MetaClass for this with a simplified method selection. And all of this static available information in here is now done at runtime?

src/main/java/org/codehaus/groovy/runtime/PackedClosure.java
```diff
@@ -0,0 +1,243 @@
...
+
+    private transient java.lang.reflect.Method target;
```
blackdrag: again field/constant randomly in the class body ;)

paulk-asert:
superseded by #2709

PR #2698, opened 2026-07-12
fix concurrent modification of ExpandoMetaClass mixinClasses

netliomax25-code:
Repro: on one thread call metaClass.mixin(someCategory) while another thread dispatches a method on the same class, so findMixinMethod iterates mixinClasses while addMixinClass adds to it.
Cause: mixinClasses is a plain LinkedHashSet with no locking, while every other backing collection in ExpandoMetaClass is a ConcurrentHashMap held alongside the read/write lock. A structural add during iteration throws ConcurrentModificationException in the dispatching thread.
Fix: keep the mixins in a CopyOnWriteArraySet, which is safe for concurrent read-while-modify and preserves insertion order plus the boolean add() dedup that GROOVY-11775 depends on.

Verified with a new test that throws ConcurrentModificationException on the old set and passes with the change; the existing mixin and ExpandoMetaClass suites stay green.

### DIFF
--- a/src/main/java/groovy/lang/ExpandoMetaClass.java
+++ b/src/main/java/groovy/lang/ExpandoMetaClass.java
@@ -49,11 +49,11 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


paulk-asert:
I created the following issue:
https://issues.apache.org/jira/browse/GROOVY-12163

paulk-asert:
Superseded by #2708

paulk-asert:
The superseded PR is merged. Thanks!

PR #2706, opened 2026-07-13
GROOVY-12161: Omit identity catch-all for try/catch without finally and slim exception bytecode

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12161

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/CompileStack.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/CompileStack.java
@@ -931,6 +931,16 @@ public class CompileStack {
         applyBlockRecorder(blockRecorders);
     }
 
+    /**
+     * Closes protected ranges, inlines each finally/synchronized guard, then
+     * re-opens a range after the guards for any code that follows (e.g. the
+     * return-value reload).
+     * <p>
+     * A leading {@code NOP} keeps the closed range non-empty when the try body
+     * is only an abrupt exit (empty exception-table ranges are illegal). No
+     * trailing {@code NOP} is emitted; the restarted range binds to the
+     * following real instruction (load/return/goto).
+     */
     private void applyBlockRecorder(final Collection<BlockRecorder> blockRecorders) {
         if (blockRecorders.isEmpty() || blockRecorders.size() == visitedBlocks.size()) return;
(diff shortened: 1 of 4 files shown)


src/main/java/org/codehaus/groovy/classgen/asm/CompileStack.java
```diff
@@ -941,19 +951,19 @@ private void applyBlockRecorder(final Collection<BlockRecorder> blockRecorders)
...
+            // Exclude the finally body from the exception table so it is not
+            // re-entered if it throws (avoid double application of finally).
             recorder.excludedStatement.run();
 
             recorder.startRange(start);
         }
 
-        mv.visitInsn(NOP);
+        // start marks the first instruction after all inlined finally blocks.
         mv.visitLabel(start);
```
blackdrag, line 967: I vaguely remember I did need those NOP instructions or else I could not set the label. I think this was for cases in which this was the first instruction in the method... maybe for an empty try-finally? not sure anymore. Or maybe I added it here back then defensively because of a similar problem in combinations with returns? But thinking back I don't think this was because of the bytecode, I think it was because of ASM. So either a bug or me not using it correctly. Very possible this is no longer an issue and thus not required.

src/test/groovy/bugs/Groovy11362.groovy
```diff
@@ -33,9 +33,11 @@ final class Groovy11362 extends AbstractBytecodeTestCase {
                 }
             }
         '''
+        // Catch parameter must be typed Exception (GROOVY-11362), not Object.
+        // Label numbers are not stable across try/catch codegen improvements.
```
blackdrag, line 37: They had been stable before, why not anymore?

daniellansun: I hope I have understood what you mean.

In the past years, I find groovy generates more complex bytecode for try-catch statement than Java does, so I wish we could refine them. As we all know, simpler and less bytecode is also friendly to JIT.

See also:
https://issues.apache.org/jira/browse/GROOVY-12161

daniellansun:

paulk-asert:
Merged. Thanks! Plus I added an additional test.

PR #2707, opened 2026-07-13
GROOVY-12162: Harden and extend bytecode peephole optimization

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12162

### DIFF
--- a/src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
+++ b/src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
@@ -330,7 +330,7 @@ final class PeepholeOptimizingMethodVisitorTest {
     }
 
     @Test
-    void removesStandaloneCheckcastBeforePop() {
+    void removesLoadAndAttachedCheckcastBeforePop() {
         def bytecode = sequenceFor('(Ljava/lang/Object;)V') {
             visitVarInsn(Opcodes.ALOAD, 0)
             visitTypeInsn(Opcodes.CHECKCAST, 'java/lang/String')
@@ -338,7 +338,82 @@ final class PeepholeOptimizingMethodVisitorTest {
             visitInsn(RETURN)
         }
 
-        assert opcodeLines(bytecode) == ['ALOAD 0', 'POP', 'RETURN']
+        // ALOAD + CHECKCAST are both dead once the value is discarded.
+        assert opcodeLines(bytecode) == ['RETURN']
+    }
+
+    @Test
+    void removesStandaloneCheckcastBeforePop() {
+        def bytecode = sequenceFor('(Ljava/lang/Object;)V') {
+            visitVarInsn(Opcodes.ALOAD, 0)
+            visitInsn(Opcodes.NOP) // force the load to flush so CHECKCAST is standalone
+            visitTypeInsn(Opcodes.CHECKCAST, 'java/lang/String')
+            visitInsn(Opcodes.POP)
+            visitInsn(RETURN)
+        }
+
+        assert opcodeLines(bytecode) == ['ALOAD 0', 'NOP', 'POP', 'RETURN']
+    }
+
+    @Test
+    void rewritesNullComparisonsAgainstAconstNull() {
+        def nullComparisons = [
+                (Opcodes.IF_ACMPEQ): 'IFNULL',
+                (Opcodes.IF_ACMPNE): 'IFNONNULL',
+        ]
+
+        nullComparisons.each { opcode, expected ->
+            def label = new Label()
+            def bytecode = sequenceFor('(Ljava/lang/Object;)V') {
+                visitVarInsn(Opcodes.ALOAD, 0)
+                visitInsn(Opcodes.ACONST_NULL)
+                visitJumpInsn(opcode, label)
+                visitInsn(RETURN)
+                visitLabel(label)
+                visitInsn(RETURN)
+            }
+            def lines = opcodeLines(bytecode)
+
+            assert lines.size() == 4
+            assert lines[0] == 'ALOAD 0'
+            assert lines[1].startsWith(expected)
+            assert lines[2..3] == ['RETURN', 'RETURN']
+            assert !lines[1].startsWith(Printer.OPCODES[opcode])
+        }
+    }
+
+    @Test
+    void removesBareDupBeforeMatchingPop() {
+        def bytecode = sequenceFor {
+            visitInsn(Opcodes.ICONST_1)
+            visitInsn(Opcodes.DUP)
+            visitInsn(Opcodes.POP)
+            visitLdcInsn(2L)
+            visitInsn(Opcodes.DUP2)
+            visitInsn(Opcodes.POP2)
+            visitInsn(RETURN)
+        }
+
+        assert opcodeLines(bytecode) == [
+                'ICONST_1',
+                'LDC 2L',
+                'RETURN',
+        ]
+    }
+
+    @Test
+    void preservesAttachedCheckcastWhenTheValueIsUsed() {
+        def bytecode = sequenceFor('(Ljava/lang/Object;)Ljava/lang/String;') {
+            visitVarInsn(Opcodes.ALOAD, 0)
+            visitTypeInsn(Opcodes.CHECKCAST, 'java/lang/String')
+            visitInsn(Opcodes.ARETURN)
+        }
+
+        assert opcodeLines(bytecode) == [
+                'ALOAD 0',
+                'CHECKCAST java/lang/String',
+                'ARETURN',
+        ]
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


daniellansun:

paulk-asert:
AI read:

> Verdict: sound, well-engineered, and ship-able — with one design inconsistency worth reconciling before merge.
> 
> The one issue to reconcile — discarded-`CHECKCAST` CCE is treated inconsistently.
> 
> Within this same PR, a cast whose result is discarded is handled two opposite ways:
> * before `POP` (`tryRemovePendingLoad`): the cast is dropped — "drop the dead cast, keep the pop." The PR extends this so even `ALOAD; CHECKCAST; POP` collapses to nothing (test `removesLoadAndAttachedCheckcastBeforePop` asserts → `['RETURN']`). → the `ClassCastException` is eliminated.
> * before void `RETURN` (`tryDropPendingLoadOnReturn`): the cast is preserved — emits `CHECKCAST; POP; RETURN` "so a type-check side effect is preserved." → the CCE is **kept**.
> 
> Same situation (cast value thrown away), opposite decisions based only on `POP` vs `RETURN`. By the author's own stated principle (preserve the type-check side effect), the POP path is the inconsistent one. Impact is likely low — the standalone drop shipped back in GROOVY-12065, the full CI matrix is green, and Groovy rarely emits a bare `CHECKCAST; POP` for a discarded cast (dynamic casts go through `castToType`) — so I'd rate it PLAUSIBLE, not a confirmed regression. But it's a real semantic question worth settling: either a discarded cast's CCE matters (then preserve it on the POP path too) or it doesn't (then simplify the RETURN path to drop it). A behavior test pinning a discarded failing cast under @CompileStatic would lock the intended semantics.

paulk-asert (changes requested):
LGTM but change if possible assuming the Claude read holds per your analysis.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -329,16 +329,91 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+            visitLdcInsn(2L)
+            visitInsn(Opcodes.DUP2)
+            visitInsn(Opcodes.POP2)
+            visitInsn(RETURN)
+        }
+
+        assert opcodeLines(bytecode) == [
+                'ICONST_1',
+                'LDC 2L',
+                'RETURN',
```
blackdrag, line 458: in theory we could go further in the future and this would be reduced to just RETURN. But does not have to be part of this PR

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -329,16 +329,91 @@ final class PeepholeOptimizingMethodVisitorTest {
...
         def bytecode = sequenceFor('(Ljava/lang/Object;)V') {
             visitVarInsn(Opcodes.ALOAD, 0)
+            visitInsn(Opcodes.NOP) // force the load to flush so CHECKCAST is standalone
             visitTypeInsn(Opcodes.CHECKCAST, 'java/lang/String')
             visitInsn(Opcodes.POP)
             visitInsn(RETURN)
         }
 
-        assert opcodeLines(bytecode) == ['ALOAD 0', 'POP', 'RETURN']
+        assert opcodeLines(bytecode) == ['ALOAD 0', 'NOP', 'POP', 'RETURN']
```
blackdrag: why can it not reduce this to just RETURN? would that be a followup?

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+            visitInsn(Opcodes.POP)
+            visitInsn(RETURN)
+        }
+        assert opcodeLines(secondIincFlushes) == [
+                'ILOAD 0',
+                'IINC 0 1',
+                'IINC 0 2',
+                'POP',
+                'RETURN',
+        ]
```
blackdrag, line 771: in theory we could collapse that as well, if 0 is a temporary variable aka has no entry for the local variable names.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+            visitLdcInsn(8.0d)
+            visitInsn(Opcodes.DUP2)
+            visitVarInsn(Opcodes.DSTORE, 2)
+            visitInsn(Opcodes.POP2)
+            visitInsn(RETURN)
+        }
+
+        assert opcodeLines(bytecode) == [
+                'LDC 9L',
+                'LSTORE 0',
```
blackdrag, line 901: removing the dup2 in this case is very nice. Good job.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+                'RETURN',
+        ]
+    }
+
+    @Test
+    void preservesMismatchedDupAndPopSizes() {
+        def bytecode = sequenceFor {
+            visitInsn(Opcodes.ICONST_1)
+            visitInsn(Opcodes.DUP)
+            visitInsn(Opcodes.POP2) // size mismatch → keep DUP
```
blackdrag, line 930: would POP2 here not simply remove the two ints from the stack? It might be a size mismatch with DUP itself, but not with the stack in total. Again. Nothing wrong, just wondering if we should try to pick this up in a later version

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+            visitInsn(RETURN)
+        }
+        assert opcodeLines(withPop) == ['RETURN']
+
+        def standalonePop2 = sequenceFor('(Ljava/lang/Object;)V') {
+            visitVarInsn(Opcodes.ALOAD, 0)
+            visitInsn(Opcodes.NOP)
+            visitTypeInsn(Opcodes.CHECKCAST, 'java/lang/String')
+            visitInsn(Opcodes.POP2) // only POP removes standalone cast
+            visitInsn(RETURN)
```
blackdrag, line 972: this test is a bit irritating because of the NOP.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+    @Test
+    void dropsBoxedValueDiscardedByPop() {
+        def bytecode = sequenceFor('(I)V') {
+            visitVarInsn(Opcodes.ILOAD, 0)
+            visitMethodInsn(Opcodes.INVOKESTATIC, 'java/lang/Integer', 'valueOf', '(I)Ljava/lang/Integer;', false)
+            visitInsn(Opcodes.POP)
+            visitInsn(RETURN)
+        }
+
+        assert opcodeLines(bytecode) == ['ILOAD 0', 'POP', 'RETURN']
```
blackdrag: This sequence I do not understand. You load 0, box it with valueOf, POP the result and then return. But why does the POP stay? If the POP invalidates the valueOf call it should be removed. And then ILOAD 0; RETURN should collapse to just RETURN, since that ILOAD 0 is dead code. Similar for the other cases

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+            visitMethodInsn(Opcodes.INVOKESTATIC, 'java/lang/Integer', 'valueOf', '(I)Ljava/lang/Integer;', false)
+            visitInsn(Opcodes.POP2)
+            visitInsn(RETURN)
+        }
+
+        // POP2 does not match a one-slot boxed reference; flush the box and keep POP2.
+        assert opcodeLines(bytecode) == [
+                'ILOAD 0',
+                'INVOKESTATIC java/lang/Integer.valueOf (I)Ljava/lang/Integer;',
+                'POP2',
```
blackdrag: a POP2 should be seen as 2 slot operand if in the stack before we have a 2 slot value, otherwise it should be seen as POP;POP and remove accordingly

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+            visitMethodInsn(Opcodes.INVOKESTATIC, 'java/lang/Integer', 'valueOf', '(I)Ljava/lang/Integer;', false)
+            visitMethodInsn(Opcodes.INVOKEVIRTUAL, 'java/lang/Long', 'longValue', '()J', false)
+            visitInsn(Opcodes.LRETURN)
+        }
+
+        assert opcodeLines(bytecode) == [
+                'ILOAD 0',
+                'INVOKESTATIC java/lang/Integer.valueOf (I)Ljava/lang/Integer;',
+                'INVOKEVIRTUAL java/lang/Long.longValue ()J',
+                'LRETURN',
```
blackdrag: actually we could compress this to ILOAD0; I2L; LRETURN.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
```diff
@@ -366,6 +441,788 @@ final class PeepholeOptimizingMethodVisitorTest {
...
+                    'org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation',
+                    'box', '(I)Ljava/lang/Object;', false)
+            visitMethodInsn(Opcodes.INVOKEVIRTUAL, 'java/lang/Integer', 'intValue', '()I', false)
+            visitInsn(Opcodes.IRETURN)
+        }
+
+        assert opcodeLines(bytecode) == [
+                'ILOAD 0',
+                'INVOKESTATIC org/codehaus/groovy/runtime/typehandling/DefaultTypeTransformation.box (I)Ljava/lang/Object;',
+                'INVOKEVIRTUAL java/lang/Integer.intValue ()I',
```
blackdrag, line 1488: I think that if we have _primitive->DTT.box->unbox of same type as primitive_ that it is then save to remove the box/unbox call

daniellansun:
> AI read:
> 
> > Verdict: sound, well-engineered, and ship-able — with one design inconsistency worth reconciling before merge.
> > The one issue to reconcile — discarded-`CHECKCAST` CCE is treated inconsistently.
> > Within this same PR, a cast whose result is discarded is handled two opposite ways:
> > 
> > * before `POP` (`tryRemovePendingLoad`): the cast is dropped — "drop the dead cast, keep the pop." The PR extends this so even `ALOAD; CHECKCAST; POP` collapses to nothing (test `removesLoadAndAttachedCheckcastBeforePop` asserts → `['RETURN']`). → the `ClassCastException` is eliminated.
> > * before void `RETURN` (`tryDropPendingLoadOnReturn`): the cast is preserved — emits `CHECKCAST; POP; RETURN` "so a type-check side effect is preserved." → the CCE is **kept**.
> > 
> > Same situation (cast value thrown away), opposite decisions based only on `POP` vs `RETURN`. By the author's own stated principle (preserve the type-check side effect), the POP path is the inconsistent one. Impact is likely low — the standalone drop shipped back in GROOVY-12065, the full CI matrix is green, and Groovy rarely emits a bare `CHECKCAST; POP` for a discarded cast (dynamic casts go through `castToType`) — so I'd rate it PLAUSIBLE, not a confirmed regression. But it's a real semantic question worth settling: either a discarded cast's CCE matters (then preserve it on the POP path too) or it doesn't (then simplify the RETURN path to drop it). A behavior test pinning a discarded failing cast under @CompileStatic would lock the intended semantics.

Agreed — same situation, opposite decisions. We settled on **preserve CCE on both paths** (Java-like; matches the void-RETURN intent).

**Change:** `tryRemovePendingLoad` now emits attached/standalone `CHECKCAST` before `POP` instead of dropping it. Pure loads without a cast still collapse.

**Tests:**
- unit: POP vs RETURN both keep `CHECKCAST` (`preservesLoadAndAttachedCheckcastBeforePop`, `preservesStandaloneCheckcastBeforeReturn`, …)
- SC: `Groovy12162` — discarded Groovy cast keeps side effect; after `instanceof String`, discarded `(String) o` still emits `CHECKCAST; POP` (not erased)

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Agreed — once DUP/POP is gone the constants are already flushed out of the
single load window, so a second pass / wider window would be needed to drop
them before void RETURN. Happy as a follow-up; not this PR.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Two separate points:

1. **ALOAD → RETURN:** NOP flushes the load, so ALOAD is already committed; the
   single-window pass can’t reclaim it. Pulling that back needs multi-block DCE
   / a second pass — follow-up.

2. **CHECKCAST drop:** per Paul’s review we no longer drop a discarded
   CHECKCAST (POP and void-RETURN now both keep the cast for CCE). So this
   sequence becomes `ALOAD; NOP; CHECKCAST; POP; RETURN`, not bare RETURN.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Yes — if slot 0 were known to be a compiler temp (no LVT entry), ILOAD+POP after
the IINCs could go away. That needs LVT / “is this a temp?” info the current
single-window visitor doesn’t have. Follow-up.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Thanks ;-)

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Exactly — vs the real stack `ICONST; DUP; POP2` is a total no-op; we only refuse
because DUP’s *form* size (1) ≠ POP2 (2). Full stack-shape tracking would catch
it. Fine as a later enhancement; left as-is for this PR.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Fair — NOP is only a test harness trick to flush ALOAD so CHECKCAST is seen as
standalone (not attached). Comments now say that explicitly; happy to rename /
split the case if you’d prefer a clearer shape.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: You were right — that was incomplete. We used to flush the load when buffering
the box, so the producer was already committed and only the box could die.

**Fix:** keep a compatible pure primitive load under the pending box; on
discard, cancel both. Result is now `['RETURN']` (and IINC side effects are
still preserved when present). Same for the wide / void-RETURN siblings.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Done. POP2 on a narrow boxed value is treated as two one-slot discards
(box/primitive + the slot underneath). With a still-pending pure ILOAD that
becomes: cancel producer, emit one POP for the underneath value.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: I’d rather not. `Integer.valueOf` + `Long.longValue` must fail at runtime on the
wrong receiver (CCE / ICCE path). Rewriting to `I2L` would silently change
semantics. Mismatched unbox stays intact on purpose.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: Agreed and done for same primitive type across conventions:
- `DTT.box(P)` + `Wrapper.pValue()` cancels
- `Wrapper.valueOf(P)` + `DTT.pUnbox` cancels

Mismatched types (e.g. box int then `Long.longValue`) still do not cancel.

daniellansun:
@blackdrag @paulk-asert 
Thanks for your reviewing. 
The PR has been updated, please review it again.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
blackdrag: I was going by the test, and here no wrong receiver is possible. You cannot get a CCE from an int using valueOf and not from longValue from the resulting instance. Though I wonder in what situation we would even emit that kind of code. For long l = 1? I think that would directly use I2L.

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
blackdrag: agreed, the cast exception should be retained.

blackdrag (approved):

src/test/groovy/org/codehaus/groovy/classgen/asm/PeepholeOptimizingMethodVisitorTest.groovy
daniellansun: You’re right — thanks for pushing on this.

With a pending Integer.valueOf the receiver is always Integer, so there is
no data-dependent “wrong receiver” CCE path worth preserving. Real classgen
already uses I2L for long l = intExpr; the box+Long.longValue form in the
test was artificial.

Done: pending box(P) followed by unbox-as-Q now rewrites to the matching
JVM conversion (I2L / L2I / I2F / …), including cross-wrapper cases such as
Integer.valueOf + Long.longValue. Same-type cancel is unchanged; Boolean is
still excluded from numeric conversion.

daniellansun:
If no one objects to the refined PR, I'll merge it this weekend.

paulk-asert:
Merged. Thanks!

PR #2709, opened 2026-07-14
GROOVY-12151: closure packing with per-class dispatch tables and full MOP integration (GEP-27 S1)

paulk-asert:
## GROOVY-12151: closure packing (GEP-27 S1) with per-class dispatch tables and full MOP integration

Under `@PackedClosures` (dynamic trust path) or `-Dgroovy.target.closure.pack=true` under `@CompileStatic` (where the type checker PROVES delegate-independence), an eligible closure literal's body is hoisted into a synthetic private method on the enclosing class and the literal is replaced by a shared `PackedClosure` adapter — removing the per-closure generated class (and the nested `$_closure1$_closure2` name explosion) while the value stays a real `groovy.lang.Closure`. Ineligible closures (escaping, delegate-dependent, default-param, nested-in-generated-function) keep their class exactly as today; opt-out via `@PackedClosures(mode = DISABLED)`.

### Dispatch architecture

Each hosting class gets compiler-generated dispatch tables: a general `$packedDispatch$(int id, Object[])` covering every hoisted target, plus array-free per-arity tables (`$packedDispatch1$`/`$packedDispatch2$`) for targets taking one or two values beyond the receiver (captures count as values — the overwhelmingly common shapes), each case casting to the target's declared parameter types and invoking directly. Tables observe the JIT's inline budget via two-level chunking. One `invokedynamic` accessor per class links all three through `GeneratedDispatcher.bootstrap` into a constant `Bundle`, lazily on first adapter creation; each literal's parameter-type `Class[]` is a condy constant resolved once per site. The whole chain is ordinary bytecode — no reflection, and no per-instance `MethodHandle` (not constant-foldable, hence many times slower). The id binds the exact hoisted method, so an inherited packed closure can never misdispatch to a subclass's same-named method.

### MOP integration (from the dev@ closure-dispatch discussion)

- Packed dispatch is guarded by the same mechanism as `Closure.call`'s cache (GROOVY-12165): the direct path is taken only when the instance's metaclass is stock and no category is active; anything MOP-relevant routes through `invokeMethod`. Interception via a per-instance metaclass is honoured identically on the dynamic and Java/GDK paths (pinned by tests, matching generated closure classes).
- `PackedClosureMetaClass` is the registered stock metaclass for the adapter (the analog of `ClosureMetaClass` for generated closures): reflection-free `call`/`doCall` dispatch, `respondsTo` faithful to the instance's declared parameter types, and the same category stance as `ClosureMetaClass` (which also fixes a latent packed-vs-classed divergence under `use{}`).
- Runtime guard: the dynamic trust path fails fast if a delegate or delegate-consulting resolveStrategy is set on a packed closure; the statically proven path stores-and-ignores, like a `@CompileStatic` closure class.

Documented residual semantic differences (all packed closures share one adapter class): per-literal class identity is gone (`closure.getClass()` distinctness, serialization); class-level `PackedClosure.metaClass` changes are global to packed closures (per-instance `setMetaClass` is fully honoured via the guard).

### JPMS note

Packed dispatch uses the capability model Java lambdas use: the hosting class's own `invokedynamic` bootstrap captures its lookup, and handing the closure across a module boundary is the grant — so packed closures work under strict encapsulation with no `opens`, where reflective dispatch of closure classes does not. This moves enforcement from invocation time to creation time, deliberately.

### Performance (JMH `ClosurePackBench`, 2 forks, ops/ms, vs current master incl. GROOVY-12164/12165)

| benchmark | classes | packed | ratio |
| --- | --- | --- | --- |
| sum_mono (written-capture `each`) | 9,103 ± 362 | **12,347 ± 321** | **1.36x** |
| sum_mega (16 classes, megamorphic) | 6,353 ± 1,071 | **12,228 ± 408** | **1.92x** |
| squared_mono (non-capturing `collect`) | 19,123 ± 684 | 11,503 ± 117 | 0.60x |
| squared_mega | 18,127 ± 771 | 12,145 ± 602 | 0.67x |

Capturing shapes are faster packed — and the win holds under megamorphism; the tight non-capturing `collect` trails, the cost being the MOP-transparency guard (loop-invariant loads the JIT hoists when the chain inlines; the same checks the classic call-site caches make) plus the per-creation adapter instance (a per-literal singleton is unsound: `Closure` carries mutable per-instance state). Allocation is at or below the closure-class baseline on capturing shapes (e.g. `times`-accumulator 160 vs 368 bytes/call) and within one adapter instance of it on non-capturing ones. Earlier figures posted in this thread predate the MOP guard and the 12164/12165 baseline improvements; this table supersedes them.

### In-situ check: Grails-shaped workloads

To verify the micro table translates to realistic code, a Grails-patterned fixture (mirroring the `GrailsWorkloadBench` shapes: an employee `findAll`/`collect`/`groupBy`/`collectEntries` pipeline and a map-iteration validation cycle with a written-capture accumulator, `@PackedClosures` trust path — every literal packs) was measured classes-vs-packed on the same build, interleaved rounds:

| workload | classes | packed | delta |
| --- | --- | --- | --- |
| domain pipeline | 219-222 ops/ms | 272-273 ops/ms | **+23%** |
| validation cycle | 1,962-2,001 ops/ms | 2,298-2,428 ops/ms | **+17-21%** |
| classes emitted | 10 (8 closure classes) | **2** | |

(Both compilations produce identical outputs for these workloads, so the comparison is two implementations of the same computation; the broader behavioural-equivalence evidence is the corpus parity and test suite below.) The blend is the point: realistic pipelines mix capturing, multi-argument and non-capturing dispatches with real body work, and the mix nets strongly positive — the micro table's adverse non-capturing row is real but dilutes out in situ (further lane micro-tuning was assessed against these shapes and found to have no in-situ effect, so it is deliberately not pursued).

Code size on the 29-closure measurement corpus: 30 -> 2 classes, -73% bytecode bytes; corpus transcripts byte-identical with the flag on and off; full suite 15,780 green.

### Commits

Packing base (capability analysis, hoisting, escape/decline rules, `@PackedClosures` modes) -> per-class-table optimisation (general + per-arity tables, condy parameter types, `ClosurePackBench`) -> MOP integration slices (dispatch guard; doCall alignment; `PackedClosureMetaClass`). Rebased on master after GROOVY-12165 merged; the `Closure.java` delta of this PR is a single line (the stock-metaclass set gains `PackedClosureMetaClass`).

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
@@ -18,12 +18,18 @@
  */
 package org.codehaus.groovy.classgen.asm;
 
+import groovy.transform.PackedClosures.PackMode;
 import org.codehaus.groovy.GroovyBugError;
+import org.codehaus.groovy.ast.AnnotationNode;
+import org.apache.groovy.util.SystemUtil;
 import org.codehaus.groovy.ast.ClassHelper;
+import org.codehaus.groovy.control.CompilerConfiguration;
 import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.CodeVisitorSupport;
+import org.codehaus.groovy.ast.DynamicVariable;
 import org.codehaus.groovy.ast.ConstructorNode;
 import org.codehaus.groovy.ast.FieldNode;
+import org.codehaus.groovy.ast.PropertyNode;
 import org.codehaus.groovy.ast.InnerClassNode;
 import org.codehaus.groovy.ast.MethodNode;
 import org.codehaus.groovy.ast.Parameter;
@@ -32,42 +38,82 @@ import org.codehaus.groovy.ast.VariableScope;
 import org.codehaus.groovy.ast.expr.ClassExpression;
 import org.codehaus.groovy.ast.expr.ClosureExpression;
 import org.codehaus.groovy.ast.expr.ConstructorCallExpression;
+import org.codehaus.groovy.ast.expr.ArrayExpression;
+import org.codehaus.groovy.ast.expr.BinaryExpression;
+import org.codehaus.groovy.ast.expr.CastExpression;
+import org.codehaus.groovy.ast.expr.DeclarationExpression;
 import org.codehaus.groovy.ast.expr.Expression;
 import org.codehaus.groovy.ast.expr.FieldExpression;
+import org.codehaus.groovy.ast.expr.ListExpression;
+import org.codehaus.groovy.ast.expr.MapEntryExpression;
+import org.codehaus.groovy.ast.expr.MethodCallExpression;
+import org.codehaus.groovy.ast.expr.PropertyExpression;
+import org.codehaus.groovy.ast.expr.PostfixExpression;
+import org.codehaus.groovy.ast.expr.PrefixExpression;
 import org.codehaus.groovy.ast.expr.TupleExpression;
 import org.codehaus.groovy.ast.expr.VariableExpression;
 import org.codehaus.groovy.ast.stmt.BlockStatement;
+import org.codehaus.groovy.ast.stmt.ExpressionStatement;
+import org.codehaus.groovy.ast.stmt.ReturnStatement;
+import org.codehaus.groovy.ast.stmt.Statement;
 import org.codehaus.groovy.ast.tools.GenericsUtils;
 import org.codehaus.groovy.classgen.AsmClassGenerator;
+import org.codehaus.groovy.classgen.ReturnAdder;
+import org.codehaus.groovy.control.messages.WarningMessage;
+import org.codehaus.groovy.runtime.GeneratedDispatcher;
+import org.codehaus.groovy.syntax.Types;
+import org.codehaus.groovy.transform.stc.StaticTypesMarker;
+import org.codehaus.groovy.transform.trait.Traits;
 import org.objectweb.asm.MethodVisitor;
 
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import static org.apache.groovy.ast.tools.ClassNodeUtils.addGeneratedMethod;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.args;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.callThisX;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.callX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.constX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.ctorSuperX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.nullX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.returnS;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.stmt;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.varX;
+import static org.codehaus.groovy.transform.sc.StaticCompilationMetadataKeys.STATIC_COMPILE_NODE;
 import static org.codehaus.groovy.transform.stc.StaticTypesMarker.INFERRED_RETURN_TYPE;
+import org.objectweb.asm.ConstantDynamic;
+import org.objectweb.asm.Handle;
 import static org.objectweb.asm.Opcodes.ACC_FINAL;
 import static org.objectweb.asm.Opcodes.ACC_PRIVATE;
 import static org.objectweb.asm.Opcodes.ACC_PUBLIC;
 import static org.objectweb.asm.Opcodes.ACC_STATIC;
+import static org.objectweb.asm.Opcodes.AALOAD;
+import static org.objectweb.asm.Opcodes.AASTORE;
 import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
+import static org.objectweb.asm.Opcodes.ACONST_NULL;
 import static org.objectweb.asm.Opcodes.ALOAD;
+import static org.objectweb.asm.Opcodes.ANEWARRAY;
+import static org.objectweb.asm.Opcodes.ARETURN;
+import static org.objectweb.asm.Opcodes.ATHROW;
+import static org.objectweb.asm.Opcodes.CHECKCAST;
 import static org.objectweb.asm.Opcodes.DUP;
 import static org.objectweb.asm.Opcodes.GETFIELD;
+import static org.objectweb.asm.Opcodes.H_INVOKESTATIC;
+import static org.objectweb.asm.Opcodes.ICONST_0;
+import static org.objectweb.asm.Opcodes.ILOAD;
+import static org.objectweb.asm.Opcodes.ISHR;
 import static org.objectweb.asm.Opcodes.INVOKESPECIAL;
+import static org.objectweb.asm.Opcodes.INVOKESTATIC;
 import static org.objectweb.asm.Opcodes.INVOKEVIRTUAL;
 import static org.objectweb.asm.Opcodes.NEW;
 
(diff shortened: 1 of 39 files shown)


src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +253,1030 @@ public void writeClosure(final ClosureExpression expression) {
...
+                capturedTypes.put(v.getName(), erasedType(v.getOriginType()));
+                capturedVars.put(v.getName(), v);
+            }
+        }
+
+        // An EMPTY parameter array is the implicit-it literal ({ ... }); null is the explicit
+        // zero-parameter form ({ -> ... }), which must report arity 0 (GString's writer-vs-call
+        // branch and SAM matching key on it), exactly as its generated class would.
+        Parameter[] declared = expression.getParameters();
+        boolean implicitParam = (declared != null && declared.length == 0);
```
blackdrag, line 878: what is declared==null? Why not `boolean implicitParam = (declared == null || declared.length == 0);`

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +253,1030 @@ public void writeClosure(final ClosureExpression expression) {
...
+                // the loads match the STC metadata already on the body expressions (otherwise the
+                // static writer falls back to dynamic dispatch on the type mismatch).
+                // an untyped capture may carry a flow-inferred type STC recorded on the original
+                // variable (e.g. an enclosing closure's inferred parameter, captured by a nested
+                // literal whose method reference expects the narrowed type). DECLARE the hoisted
+                // parameter by it: the dispatch table's case checkcasts to declared parameter
+                // types, so the body's load then carries the type the use sites were compiled
+                // against — where a plain Object parameter loads unverifiably wide (the classed
+                // emission gets the equivalent cast on its shared-Reference dereference).
+                ClassNode capturedType = typedBody ? capturedTypes.get(cn) : ClassHelper.OBJECT_TYPE;
```
blackdrag: I am using this line to attach the comment, but it is not limited to this. I see here still a mix of static and dynamic compiler code. For me the dynamic case should be the general one and static compilation should be a specialization, separated from the dynamic code generator. Why like this? Because if we change something for static compilation it is in one class and if we change something for dynamic compilation it is in another class. Andy dynamic>static because that is how we do it in other places as well, just trying to keep the style.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +253,1030 @@ public void writeClosure(final ClosureExpression expression) {
...
+                Variable v = capturedVars.get(cn);
+                if (typedBody && v instanceof org.codehaus.groovy.ast.ASTNode) {
+                    Object inferred = ((org.codehaus.groovy.ast.ASTNode) v)
```
blackdrag: 2 Points here... Why the fully qualified name? (This is in other places as well) and can we not use `x instanceof T t` for this? It is a minor point. Just mentioning it because the long lines stick out.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +253,1030 @@ public void writeClosure(final ClosureExpression expression) {
...
+        // declares for it. A vararg-shaped literal (trailing array parameter, e.g. { ...z -> })
+        // also gets FixedN: its generated class would declare a VARARG doCall, whose selection
+        // flexibility (zero args collect to an empty array where a fixed one-param doCall would
+        // null-fill) only a varargs declaration reproduces.
+        boolean varargShape = !implicitParam && arity > 0 && closureParams[arity - 1].getType().isArray();
+        String pc = "org/codehaus/groovy/runtime/PackedClosure";
+        if (implicitParam) pc = pc + "$FixedIt";
+        else if (arity <= 4 && !varargShape) pc = pc + "$Fixed" + arity;
+        else pc = pc + "$FixedN";
+        mv.visitTypeInsn(NEW, pc);
```
blackdrag: should we start supporting default parameters, this kind of logic will no work. Not a blocker, just a note.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +253,1030 @@ public void writeClosure(final ClosureExpression expression) {
...
+                BytecodeHelper.pushConstant(mv, i);
+                os.push(ClassHelper.int_TYPE);
+                if (holderThreaded.contains(cn)) {
+                    loadReference(cn, controller);       // shared Reference holder
+                } else {
+                    new VariableExpression(cn).visit(acg); // value
+                    os.box();
+                }
+                mv.visitInsn(AASTORE);
+                os.remove(3);
```
blackdrag: I don`t think we need "os" here, except for loading the variable and boxing it. But you could directly remove 1 after that and otherwise remove the os-code. That is only 4 lines of code less of course. Just irritates me a bit the way it is here.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +253,1030 @@ public void writeClosure(final ClosureExpression expression) {
...
+        // building one per creation measurably taxes creation-heavy code. The types travel as a
+        // method descriptor because primitive parameter types have no Class constant-pool form.
+        StringBuilder typesDescriptor = new StringBuilder("(");
+        for (int i = 0; i < arity; i++) {
+            typesDescriptor.append(BytecodeHelper.getTypeDescription(erasedType(closureParams[i].getType())));
+        }
+        typesDescriptor.append(")V");
+        mv.visitLdcInsn(new org.objectweb.asm.ConstantDynamic("paramTypes", "[Ljava/lang/Class;",
+                new org.objectweb.asm.Handle(org.objectweb.asm.Opcodes.H_INVOKESTATIC, DISPATCHER_TYPE, "paramTypes",
+                        "(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/Class;Ljava/lang/String;)[Ljava/lang/Class;",
```
blackdrag, line 1040: We should consider the interface surface with bytecode. Right now we have SBC, DTT and IndyInterface mostly, but not only. I would prefer this in a method in IndyInterface, to have a central point of methods we have to keep caring about. The same about the classes we expose in bytecode... would it be a problem to have a closureFactory on one of the already exposed classes instead?  I mean to produce instances of Bundle, PackedClosure, PackedClosure.Fixed and so on. It is slightly less efficient I imagine, but I doubt the impact would be major compared to the object creation cost.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: `declared == null` is the explicit no-arg case, no `it`: `{ -> ... }`
`declared.length == 0` is `{ ... }`: implicit `it`, fuzzy 0/1 arity

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: agreed

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: should be fixed

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: noted

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: should be fixed

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
```diff
@@ -158,6 +262,1038 @@ public void writeClosure(final ClosureExpression expression) {
...
+     * The erasure of a captured variable's type, safe to declare on the hoisted method: generic
+     * placeholders are replaced by their bound (the hoisted method does not declare the enclosing
+     * method's type variables) and type arguments are dropped ({@code List<T>} → {@code List}).
+     */
+    protected static ClassNode erasedType(final ClassNode type) {
+        ClassNode t = type;
+        if (t == null) return ClassHelper.OBJECT_TYPE;
+        if (t.isGenericsPlaceHolder()) t = t.redirect();
+        return t.getPlainNodeReference();
+    }
```
blackdrag: Would this not be better in one of the generics helper classes? Not a must.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
blackdrag: this is still open and uncommented

src/test/groovy/org/codehaus/groovy/transform/PackedClosureBoundariesTest.groovy
```diff
@@ -0,0 +1,129 @@
...
+        ['bare field-bound name',        false, true,  'int field = 1\ndef m(List<Integer> xs) { xs.collect { it + field } }'],
+        ['explicit this-property',       false, true,  'int field = 1\ndef m(List<Integer> xs) { xs.collect { it + this.field } }'],
+        // ---- real-Closure semantics: declines everywhere ----------------------------------------
+        ['uses delegate',                false, false, 'def m() { def c = { delegate.toString() }; c }'],
+        ['uses owner',                   false, false, 'def m() { def c = { owner.toString() }; c }'],
+        ['default parameter values',     false, false, 'def m() { def c = { int x = 1 -> x }; c(2) }'],
+        // ---- escapes: declines everywhere --------------------------------------------------------
+        ['returned',                     false, false, 'Closure m() { return { it } }'],
+        ['stored to property',           false, false, 'def m(Map attrs) { attrs.handler = { it } }'],
+        ['in a collection literal',      false, false, 'def m() { [{ it }] }'],
```
blackdrag: I think the given reason in those 3 last tests for why this is not inlined is incorrect. `returned` should be hoistable. That it is not is due to the lack of default parameter support for it. `stored to property` then actually looks like the exact same case. It does not matter, that `attrs.handler` would be dynamic. And in the case `in a collection literal` you could see the return list as a special case of a returned closure, but again the real reason is the missing default parameter support. If you wanted to imply with `escapes: declines everywhere` that an escaping Closure is never hoisted, then you should actually add a case without it and without default parameters. I guess I am missing an explanation of why an escaping Closure is declined. Especially if I assume non of the other tests here is testing that already

blackdrag (changes requested):

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: Agreed — moved. It's now `GenericsUtils.erasure(ClassNode)`, sitting next to its closest sibling `nonGeneric` (which strips type arguments but doesn't resolve a placeholder to its bound — the part the hoisted signatures need, so it wasn't quite reusable as-is). Both writers now call the shared helper.

src/test/groovy/org/codehaus/groovy/transform/PackedClosureBoundariesTest.groovy
paulk-asert: Implicit `it` is actually supported by packing — it isn't compiled as a default parameter on the hoisted method; the adapter's `FixedIt` family member reproduces the fuzzy 0/1 arity (`doCall()` + `doCall(Object)`). The `'implicit it'` row near the top of the matrix pins that: `xs.collect { it * 2 }` packs on both paths. So in those last three rows the escape gate genuinely is the operative decline, not default-param support.

That said, your point stands that the rows didn't *prove* it — all three happened to use `{ it }`, so escape and implicit-`it` weren't isolated from each other. They now use explicit parameters (`{ x -> x }`), same for the escaping shapes in `PackedClosuresTransformTest`, so the only difference from the packing rows is the escaping position.

On why an escaping closure is declined at all: it's a policy gate, not a capability limit. A visibly escaping literal reaches consumers unknown to this compilation, which may set a delegate on it (which fail-fasts at runtime on the dynamic trust path), serialize it, or rely on per-literal class identity — so the gate conservatively keeps it a class rather than converting working code into runtime surprises. The section comment in the test now says exactly this. For the statically-proven path (where a caller-set delegate is stored-and-ignored) the gate could in principle be relaxed later; keeping it uniform seemed the safer S1 default.

src/main/java/org/codehaus/groovy/classgen/asm/ClosureWriter.java
paulk-asert: Sorry, this one slipped past my earlier replies — you're right, and it's now done, slightly beyond what you asked for:

- Both bootstraps are now on `IndyInterface` (`packedDispatchers` for the invokedynamic accessor, `packedParamTypes` for the condy parameter-types constant), as one-line delegations to `GeneratedDispatcher`. The JVM hands the bootstrap the *caller's* lookup regardless of which class hosts the method, so the JPMS/capability behaviour is unchanged.
- The adapter factory is now `ScriptBytecodeAdapter.packedClosure(...)`, taking the dispatcher bundle as `Object` and returning `Closure`; the `$getPackedDispatchers$` accessor descriptor is widened to `()Object` to match.

Net effect: generated class files no longer reference `GeneratedDispatcher`, `Bundle`, or `PackedClosure` at all — the emitted surface is exactly `IndyInterface` + `ScriptBytecodeAdapter` (verified on compiled output for all three trigger paths). Cost is one cast inside the factory per closure creation (noise against the object allocation, as you anticipated) and nothing per invocation; the bootstraps still link once per class.

PR #2711, opened 2026-07-14
cap nesting depth in MarkdownSlurper to avoid stack overflow

netliomax25-code:
1. MarkdownSlurper walks the parsed CommonMark tree recursively (blocksToList/nodeToMap, and textOf/appendText) with one stack frame per nesting level and no depth bound.
2. CommonMark builds that tree iteratively, so a small but deeply nested document (a single ~8 KB line of nested block quotes, or a deeply nested list) overflows the walk and raises a StackOverflowError inside parse()/parseText(). That raw Error escapes the documented MarkdownRuntimeException that callers guarding untrusted input catch.
3. Added a configurable cap (maxNestingDepth, default 1000, property groovy.markdown.maxNestingDepth) checked once before the walk, throwing MarkdownRuntimeException past the limit, matching the nesting cap JsonSlurper already applies.

Validation: a 5000-deep block-quote input now throws MarkdownRuntimeException instead of overflowing, documents within the limit parse unchanged, and the added regression tests plus the existing MarkdownSlurper suite pass.

### DIFF
--- a/subprojects/groovy-markdown/src/main/java/groovy/markdown/MarkdownSlurper.java
+++ b/subprojects/groovy-markdown/src/main/java/groovy/markdown/MarkdownSlurper.java
@@ -77,7 +77,40 @@ public class MarkdownSlurper {
 
     private static final String TABLES_EXT_CLASS = "org.commonmark.ext.gfm.tables.TablesExtension";
 
+    /**
+     * Default maximum nesting depth of block/inline elements accepted before a
+     * {@link MarkdownRuntimeException} is thrown. Matches the default nesting cap of the
+     * sibling {@code JsonSlurper}.
+     */
+    public static final int DEFAULT_MAX_NESTING_DEPTH = 1000;
+
     private boolean tablesEnabled;
+
+    /**
+     * Maximum nesting depth permitted while converting a parsed document. A value of {@code 0}
+     * or less disables the check. Defaults to {@link #DEFAULT_MAX_NESTING_DEPTH}, overridable
+     * globally via the {@code groovy.markdown.maxNestingDepth} system property.
+     */
+    private int maxNestingDepth = Integer.getInteger("groovy.markdown.maxNestingDepth", DEFAULT_MAX_NESTING_DEPTH);
+
+    /**
+     * Returns the maximum nesting depth of block/inline elements the parser will accept.
+     *
+     * @return the maximum nesting depth, or a value {@code <= 0} when the check is disabled
+     */
+    public int getMaxNestingDepth() {
+        return maxNestingDepth;
+    }
+
+    /**
+     * Sets the maximum nesting depth of block/inline elements the parser will accept before
+     * throwing a {@link MarkdownRuntimeException}. A value of {@code 0} or less disables the check.
+     *
+     * @param maxNestingDepth maximum number of nested elements to allow
+     */
+    public void setMaxNestingDepth(int maxNestingDepth) {
+        this.maxNestingDepth = maxNestingDepth;
+    }
 
     /**
      * Enable GFM-style tables. Requires {@code commonmark-ext-gfm-tables} on the classpath.
@@ -122,6 +155,7 @@ public class MarkdownSlurper {
     public MarkdownDocument parse(Reader reader) {
         try {
             Node doc = buildParser().parseReader(reader);
+            checkNestingDepth(doc);
             return new MarkdownDocument(blocksToList(doc));
         } catch (IOException e) {
             throw new MarkdownRuntimeException(e);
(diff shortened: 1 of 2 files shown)


paulk-asert:
AI read on your PR below:

> Thanks for this — the underlying concern is real: a small but deeply-nested document can drive `MarkdownSlurper` into a `StackOverflowError`, and an untrusted-input caller expecting `MarkdownRuntimeException` gets a raw `Error` instead. Before we settle on the cap approach, though, I dug into *where* the overflow actually happens, and it turns out there are **two distinct vectors** that land on opposite sides of the Groovy/CommonMark boundary — and the current fix only covers one of them.
> 
> ### Two vectors (verified against commonmark 0.29.0)
> 
> | Vector                          | Trigger                        | Where it overflows                                                                                  | Whose code                |
> | ------------------------------- | ------------------------------ | --------------------------------------------------------------------------------------------------- | ------------------------- |
> | **1 — block/container nesting** | `'>' * N`, deeply nested lists | Groovy's `blocksToList → nodeToMap → blocksToList` and `appendText` recursion                       | **MarkdownSlurper**       |
> | **2 — inline emphasis nesting** | `('*' * N) + 'a' + ('*' * N)`  | CommonMark `InlineParserImpl.mergeChildTextNodes` / `mergeTextNodesInclusive`, **inside `parse()`** | **CommonMark (upstream)** |
> 
> Probing the two directly:
> 
> - `'>' * 50000 + ' x'` → **parses fine**, then Groovy's own recursive walk overflows. CommonMark did nothing wrong here; it faithfully returns a 50000-deep tree and our recursive conversion blows the stack. This is the case this PR addresses. 👍
> - `('*' * 50000) + 'a' + ('*' * 50000)` → **overflows inside `parseReader` itself**, before control ever returns to `MarkdownSlurper`:
> 
>   ```
>   java.lang.StackOverflowError
>       at org.commonmark.internal.InlineParserImpl.mergeIfNeeded(InlineParserImpl.java:859)
>       at org.commonmark.internal.InlineParserImpl.mergeTextNodesInclusive(InlineParserImpl.java:842)
>       at org.commonmark.internal.InlineParserImpl.mergeChildTextNodes(InlineParserImpl.java:824)
>       ... (repeats)
>   ```
> 
> ### The gap
> 
> `checkNestingDepth(doc)` runs *after* `buildParser().parseReader(reader)` returns. For **Vector 2 the overflow happens during that parse call**, so the new check never executes and the raw `StackOverflowError` escapes exactly as before — the very thing the PR sets out to prevent. In other words the cap makes block-nesting safe but leaves an equivalent inline-nesting `StackOverflowError` DoS wide open, which is worse than the status quo because callers now believe they're covered.
> 
> A test that demonstrates it (fails on this branch — throws raw `StackOverflowError`, not `MarkdownRuntimeException`):
> 
> ```groovy
> @Test
> void testDeeplyNestedInlineEmphasisRejected() {
>     def md = ('*' * 50000) + 'a' + ('*' * 50000)
>     assertThrows(MarkdownRuntimeException) { new MarkdownSlurper().parseText(md) }
> }
> ```
> 
> ### Suggested direction
> 
> We generally prefer to push size/robustness limits onto the dependency and only harden locally what's genuinely ours and can't be deferred. That split maps neatly here:
> 
> 1. **Vector 2 is upstream and undeferrable.** commonmark-java has no CVE/advisory for this; it's an acknowledged limitation of the recursive inline/tree processing (see commonmark-spec#187), there's **no inline nesting cap**, and it's **not fixed in 0.29.0** (the current latest). Since we can't defer our untrusted-input obligation to a fix that doesn't exist, the honest fix is to **catch `StackOverflowError` around the parse+walk in `parse()` and rethrow `MarkdownRuntimeException`**. It's ~3 lines, covers *both* vectors' contract, and is safe here — the SO is cleanly catchable and the slurper builds a fresh parser/tree per call, so there's no retained state to corrupt. I'll also file an upstream issue for the `mergeChildTextNodes` recursion.
> 
> 2. **Vector 1 can be deferred to the dependency.** commonmark exposes `Parser.builder().maxOpenBlockParsers(int)` (since 0.28.0, default unlimited). Setting it in `buildParser()` caps block nesting *at parse time* — I confirmed `maxOpenBlockParsers(100)` yields a tree only ~103 deep for a 50000-deep block-quote input, so our recursive walk is inherently safe and beyond-limit nesting degrades to text rather than throwing. That's deterministic (unlike catching an Error at a stack-size-dependent depth) and uses the library's supported knob. (Note it does **not** help Vector 2 — inline emphasis isn't a block.)
> 
> So the combination I'd suggest is: `maxOpenBlockParsers(n)` for deterministic block handling + a `StackOverflowError` catch for the inline vector we can't defer, keeping your configurable knob as optional DoS defense-in-depth. Your test scaffolding is a great start — please keep it and add the inline-emphasis case above.
> 
> Happy to pair on the revised version if useful, or push a follow-up commit onto the branch. And thanks again for chasing these down — this is exactly the kind of untrusted-input hardening worth getting right.

paulk-asert:
Subsumed into #2729

paulk-asert:
Merged as part of #2729. Thanks!

PR #2713, opened 2026-07-16
wrap malformed json number decode in JsonException

netliomax25-code:
1. A value that starts with a digit or minus enters number decoding, but a malformed exponent or a lone sign/point such as `1e`, `1e+`, `1E`, or `-.` is only rejected when the collected token reaches `new BigDecimal(...)`.
2. The INDEX_OVERLAY and LAX overlay parsers already wrap that `NumberFormatException` in a `JsonException` (`NumberValue.bigDecimalValue`), but the default CHAR_BUFFER parser (`CharScanner.parseJsonNumber`) and the CHARACTER_SOURCE parser (`JsonParserUsingCharacterSource.decodeNumber`) let it propagate unwrapped, so an application guarding untrusted JSON with `catch (JsonException)` sees a raw `NumberFormatException` instead.
3. Wrapped the numeric construction at both sites so malformed numbers surface as `JsonException`, matching the overlay parsers. Added a regression test that checks all four parser types.

### DIFF
--- a/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/CharScanner.java
+++ b/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/CharScanner.java
@@ -499,16 +499,20 @@ public class CharScanner {
 
         final int length = index - from;
 
-        if (!foundDot && simple) {
-            if (isInteger(buffer, from, length)) {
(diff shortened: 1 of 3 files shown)


paulk-asert:
Thanks! Merged as part of PR#2715 (GROOVY-12168) including tweaks to fix the build breakage and some consistency improvements.

PR #2714, opened 2026-07-16
Reference implementation: next steps for compact closures (GEP-27 Groovy 7)

paulk-asert:
Groovy closures without generated classes: bodies become methods, and a single pluggable resolver replaces the owner/delegate/strategy machinery — behavior-identical to today.


blackdrag (commented):
my comments are actually in the other PR

PR #2717, opened 2026-07-17
reject lone low surrogate in StreamingMarkupWriter.write

netliomax25-code:
Repro: write a lone low surrogate (a `\uDC00`-`\uDFFF` char with no preceding high surrogate) through StreamingMarkupBuilder, e.g. `new StreamingMarkupBuilder().bind { mkp.yield '\uDC00' }`.
Cause: the low-surrogate branch in `write(int)` assumes a high surrogate was already buffered, so it reads `surrogatePair.charAt(0)` (the low surrogate itself) and emits a numeric character reference for an unrelated code point (`\uDC00` becomes `&#x10000;`) instead of failing.
Fix: reject a low surrogate that is not preceded by a high surrogate, matching the existing rejection of a high surrogate not followed by a low one.


paulk-asert:
Merged as GROOVY-12186. Thanks!

PR #2718, opened 2026-07-17
GROOVY-12169: Improve syntax error messages and caret positions for missing ')', ']', and '}'

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12169

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/internal/DescriptiveErrorStrategy.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/internal/DescriptiveErrorStrategy.java
@@ -34,9 +34,16 @@ import org.antlr.v4.runtime.misc.ParseCancellationException;
 
 /**
  * Provide friendly error messages when parsing errors occurred.
+ * <p>
+ * Extends {@link BailErrorStrategy} so the successful-parse path stays
+ * allocation-free and ATN-light (no grammar-level error alternatives —
+ * those were removed after GROOVY-9588).  Missing-delimiter diagnostics
+ * ({@code ')'}, {@code ']'}, {@code '}'}) run only after a recognition
+ * failure via {@link MissingDelimiterDiagnostic}.
+ * </p>
  */
 public class DescriptiveErrorStrategy extends BailErrorStrategy {
-    private CharStream charStream;
+    private final CharStream charStream;
 
     public DescriptiveErrorStrategy(CharStream charStream) {
         this.charStream = charStream;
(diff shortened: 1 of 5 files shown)


paulk-asert (approved):
Non-blocking nits (optional follow-ups, worth a PR comment)

1. insertionPointAfter adds the token's codepoint length to its start column — a multi-line anchor token (triple-quoted string/GString before a missing }) overshoots the caret. Message stays correct; caret only. Rare.
2. fromCastPattern reports the earliest high-confidence cast site independent of where ANTLR actually failed — intentional and well-guarded, but in a multi-error file the location could point at a different construct.
3. The broad catch (IndexOutOfBounds | IllegalArgumentException) in collectDefaultChannelTokens silently proceeds with partial tokens — reasonable for non-buffered streams (and there's a ThrowingTokenStream test), already commented.

daniellansun:
> Non-blocking nits (optional follow-ups, worth a PR comment)
> 
> 1. insertionPointAfter adds the token's codepoint length to its start column — a multi-line anchor token (triple-quoted string/GString before a missing }) overshoots the caret. Message stays correct; caret only. Rare.
> 2. fromCastPattern reports the earliest high-confidence cast site independent of where ANTLR actually failed — intentional and well-guarded, but in a multi-error file the location could point at a different construct.
> 3. The broad catch (IndexOutOfBounds | IllegalArgumentException) in collectDefaultChannelTokens silently proceeds with partial tokens — reasonable for non-buffered streams (and there's a ThrowingTokenStream test), already commented.

Thanks for the careful review and the approval — notes below on each optional nit.

1) Multi-line insertion point / caret overshoot
Agreed. `insertionPointAfter` now reuses `PositionConfigureUtils.endPosition`, so a multi-line anchor (e.g. triple-quoted string before a missing `}`) puts the caret on the last line of that token, after its final content, instead of overshooting on the start line. Covered with unit tests and end-to-end cases in `SyntaxErrorTest`. Related: `lastNonEof` also skips trailing `NL` so EOF-based carets stay on the last meaningful line.

2) `fromCastPattern` reports the earliest high-confidence cast site
Yes — intentional. With bail-on-error we stop at the first recognition failure, so the earliest incomplete cast is almost always the right one. Biasing toward ANTLR’s offending token was tried; it regressed good cases such as `println ((int 123)` (caret moved off the cast operand). Left as earliest-site, with the multi-error trade-off documented on the method.

3) Broad catch in `collectDefaultChannelTokens`
Agreed it’s the right shape for non-buffered / partial streams. Comment tightened slightly, and the `IllegalArgumentException` branch is now exercised alongside the existing `ThrowingTokenStream` (`IndexOutOfBoundsException`) coverage.

PR #2724, opened 2026-07-19
GROOVY-9192: Implement optional parser error recovery for Parrot parser

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-9192

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -188,7 +188,12 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
 
         this.lexer = new GroovyLangLexer(charStream);
         this.parser = new GroovyLangParser(new CommonTokenStream(this.lexer));
-        this.parser.setErrorHandler(new DescriptiveErrorStrategy(charStream));
+
+        // Opt-in recovery (CompilerConfiguration.ERROR_RECOVERY): resync after syntax
+        // errors so IDEs can collect multiple diagnostics in one pass. Default is fail-fast.
+        // Multi-error truth for hosts is ErrorCollector; recovery may still yield a partial tree.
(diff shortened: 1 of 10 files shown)


daniellansun:

paulk-asert (approved):
Approved with some minor potential cleanup possible. AI read:

In short: well-scoped, well-tested, default behavior preserved, opt-in semantics correct.

Minor notes (none blocking):
* Sonar reported 7 new minor issues (gate still passed) — worth a glance before or after merge, likely trivia.
* Theoretical edge: under recovery, a CompilationFailedException thrown without any error having been recorded would surface as silent success — every current throw path records first, so this is hypothetical, but a cheap belt-and-braces guard (if collector empty, rethrow) could be suggested in review comments if you want.

daniellansun:
> Approved with some minor potential cleanup possible. AI read:
> 
> In short: well-scoped, well-tested, default behavior preserved, opt-in semantics correct.
> 
> Minor notes (none blocking):
> 
> * Sonar reported 7 new minor issues (gate still passed) — worth a glance before or after merge, likely trivia.
> * Theoretical edge: under recovery, a CompilationFailedException thrown without any error having been recorded would surface as silent success — every current throw path records first, so this is hypothetical, but a cheap belt-and-braces guard (if collector empty, rethrow) could be suggested in review comments if you want.

Thanks for the review and the careful notes.

1) Sonar (7 minor):
Addressed in a follow-up: assertSame for reference equality, single-throw assertThrows
lambdas, assertion on recoveringStrategySyncDoesNotThrow, and the unused-recognizer
hook on createFailedPredicateErrorMessage (kept for parity with the other create*
hooks; documented + requireNonNull). The AstBuilder "brain method" INFO is pre-existing
size on an unrelated visitor method and left alone.

2) Empty-collector + recovery → silent success:
Agreed — cheap and worth doing. buildAST now returns the partial module only when
errorRecovery && errorCollector.hasErrors(); otherwise the CFE is rethrown. Covered by
regression tests in Groovy9192 (empty collector rethrows; recorded diagnostics still
return the partial module; fail-fast unchanged).

PR #2730, opened 2026-07-22
GROOVY-12185: Move classic call-site caching into optional groovy-callsite module

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12185


blackdrag (approved):
I am a bit worried about the metaclass changes since it kind of means changing public API. And I was wondering if can somehow provide an alternative. Maybe we don´t need that as most people probably do not use those callsite methods. But if we do need to compatibility I was wondering if we could solve it with a parallel meta class system. What I mean is that if callsites are active we use the callsite variants and otherwise we use the normal meta class. Maybe the callsite meta classes could just wrap the normal metaclass and delegate calls... the classic way of trying to solve a problem by adding a level of indirection. But the basic question is really: Do we need to keep those methods on the meta classes? @paulk-asert what do you think? If we say we do not need it, then I would agree to the change.

paulk-asert:
I am keen for this work to be eventually done. I need to think more about whether we are ready for this in Groovy 6 or whether this might be better in Groovy 7. I will need a few more days to absorb it.

paulk-asert:
I am +1 for 6.0.0-beta-1, if we remove the @Deprecated annotations on the moved classes, MetaClassCallSites, CachedMethodCallSites, and the package-info. I am happy for javadoc that says planned for deprecation/removal in a future Groovy version. If Grails get to test beta-1 and indy still doesn't have the "up to 4x performance hit" for their use cases, I would be happy to add back in the @Deprecated annotations before 6 GA since these are all classes we have earmarked for deprecation for some time.

PR #2732, opened 2026-07-23
GROOVY-11792: for-in loop variable captured by closure/AIC sees final value, not per-iteration value

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-11792

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/CompileStack.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/CompileStack.java
@@ -776,6 +776,10 @@ public class CompileStack {
 
     /**
      * Wraps the current stack value in a {@code groovy.lang.Reference} for the given variable slot.
+     * <p>
+     * Expects a boxed/object value already on the JVM operand stack. Prefer
+     * {@link #storeVar(BytecodeVariable, boolean)} when the value still needs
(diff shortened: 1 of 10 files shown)


blackdrag (changes requested):
Could you add a test for the bytecode as well? Best would be a test that checks the old bytecode pattern for not being produced and a test checking for the new bytecode pattern. That also clearly documents the change

daniellansun:
> Could you add a test for the bytecode as well? Best would be a test that checks the old bytecode pattern for not being produced and a test checking for the new bytecode pattern. That also clearly documents the change

Hi Jochen, does the commit to add the test looks OK to you?
https://github.com/apache/groovy/pull/2732/changes/1197010e93a0e9a7fe0f15bd8b0d8bb23e48f6fe

blackdrag (approved):

paulk-asert (changes requested):
Approved but consider the following naming suggestion:

> Flag naming, now rather than later. groovy.for.loop.capture / setForLoopCaptureEnabled doesn't say what it controls: classic for loops are not affected despite the name, and capture=false reads like "captures disabled" rather than "legacy shared binding." Something like groovy.forin.per.iteration.capture would be self-describing.

PR #2733, opened 2026-07-24
GROOVY-12188: accept covariant override of abstract method regardless…

paulk-asert:
… of classgen order

Since GROOVY-10687, a top-level class whose inner classes contain a closure is generated after its nest members, which can place it after its own subclass in the classgen order. The abstract-method check in ClassCompletionVerifier relied on the covariant bridge method that Verifier adds to the declaring class, so verifying a subclass first raised a spurious "same name but different return type" error. Recognise a covariant override directly instead; genuinely incompatible return types are still rejected by Verifier's covariant-method check.


paulk-asert:
Cherry-picked manually onto the GROOVY_5_0_X branch as the minimal fix. For master, see #2734 which should be a better long-term fix. If no impacts from the long-term fix are noticed on master, it would be a candidate for back-porting to GROOVY_5_0_X.

GROOVY_5_0_X: https://github.com/apache/groovy/commit/09a2d2f01015adc50708e366c6936164b462baab
Master: https://github.com/apache/groovy/commit/361a02e91dfb33d794cfaa33db82c39262046d6a

PR #2734, opened 2026-07-24
GROOVY-10687: complete NestMembers after class generation instead of …

paulk-asert:
…reordering classgen

Closure and anonymous inner classes materialize while bytecode is being generated, so a nest host emitted in normal order could not list them in its NestMembers attribute. The previous approach sorted nest hosts to the back of the classgen queue, which broke the supers-before-subclasses invariant relied on by ClassCompletionVerifier (GROOVY-11753, GROOVY-12188) and still could not see a host's own nested closures, requiring name prediction (GROOVY-11780). No ordering satisfies both constraints once a nest member extends its own host.

Instead, restore plain hierarchy-order sorting and complete each nest host's NestMembers in a post-generation pass: once all of a module's classes are generated, the host's inner class list is complete, so its bytes are rewritten with the exact member set. GroovyClassLoader defines classes from the post-processed bytes. This removes the reordering, the closure-detection walk, and the member-name prediction, and is unaffected by whether a closure ultimately produces a class (GEP-27 closure packing).


eric-milles (approved):
I don't check the nest member attributes that much, but all my tests seem to be holding up fine.

PR #2740, opened 2026-07-26
GROOVY-12199: Replace ant-jarjar with com.gradleup.shadow for package relocation

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12199


paulk-asert:
AI read of differences, others don't seem to matter:

>  Bucket 4a — Java9.class in the grooid jar · the one genuine semantic divergence
> 
> The grooid relocation maps java.beans.** → groovyjarjaropenbeans, and shadow rewrote a string literal "java.beans" → "groovyjarjaropenbeans" in Java9.JAVA8_PACKAGES(). jarjar left it intact. That array (source comment: "should NOT be changed!") holds real JDK package names looked up against ModuleFinder.ofSystem() to decide which packages Groovy opens reflectively — "groovyjarjaropenbeans" isn't a real JDK package, so that lookup would no-op.
>   - Scope: grooid (Android) jar only — the core jar's Java9.class is byte-identical.
>   - Real-world impact: very likely inert (the Java9 plugin's static init needs java.lang.module, which doesn't exist on Android, so it won't load there). But it's an unintended bytecode change and the clearest example of shadow's more aggressive string-constant rewriting. Worth flagging to Daniel — ideally the java.beans.** grooid rule should suppress string-literal relocation (or the AutoComplete case too), to keep grooid bytecode faithful.

daniellansun:
@paulk-asert Does the commit ( https://github.com/apache/groovy/pull/2740/changes/16f8e2305cbeffb9f75c600ef4d5e0fe7b9367a8 ) look OK to you?

paulk-asert (approved):
LGTM

PR #2748, opened 2026-07-29
avoid decoding the request URI in HttpBuilder.appendQuery

netliomax25-code:
appendQuery rebuilds the request URI with the multi-argument URI constructor, which takes decoded components, so passing uri.path and uri.query decodes every percent escape and re-emits it literally: a path segment of a%2Fb turns into two segments, %2e%2e turns into a real dot segment, and an existing query value of a%26admin%3D1 splits off a second parameter. That also weakens confineToBaseUri, since resolveUri runs enforceConfinement before appendQuery and that check deliberately compares raw path segments, so get('..%2Fadmin') { query page: 1 } passes confinement and is then sent as /api/v2/../admin. Reassembling from the raw components leaves everything except the query untouched; the three new tests fail on the current code, and the existing queryDslUsesRfc3986StyleEncoding was asserting against requestURI.query, which decoded once and masked the DSL escapes being encoded twice, so its fixture now reads rawQuery.

### DIFF
--- a/subprojects/groovy-http-builder/src/main/groovy/groovy/http/HttpBuilder.groovy
+++ b/subprojects/groovy-http-builder/src/main/groovy/groovy/http/HttpBuilder.groovy
@@ -624,8 +624,8 @@ final class HttpBuilder {
         }
 
         List<String> pairs = new ArrayList<>()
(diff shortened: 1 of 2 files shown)


paulk-asert:
merged manually as GROOVY-12213 - thanks!

PR #2755, opened 2026-08-02
GROOVY-12223: Introduce hidden class support

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12223

### DIFF
--- a/src/main/java/org/apache/groovy/util/HiddenClassDefiner.java
+++ b/src/main/java/org/apache/groovy/util/HiddenClassDefiner.java
@@ -35,67 +35,87 @@ import java.util.Collections;
  * <h2>Lookup ownership (read this first)</h2>
  * <p>{@link MethodHandles#lookup()} is <em>caller-sensitive</em>: it returns a
  * full-privilege lookup only for the class that literally contains the call.
- * A lookup captured in this utility therefore only has full privilege for
- * {@code HiddenClassDefiner} itself — never for arbitrary foreign classes
- * (for example {@code java.lang.String} in {@code java.base}).
- *
- * <p>Consequently the <strong>preferred</strong> entry point is
- * {@link #tryDefineNestmate(Lookup, byte[], boolean)}, where the caller
- * supplies a {@link Lookup} obtained inside the intended nest-host class:
+ * Production call sites therefore capture a lookup on the intended nest host
+ * itself, for example:
  * <pre>{@code
- * // Inside the class that should host the hidden nestmate:
+ * // ReflectorLoader / ProxyGeneratorAdapter (or any nest host):
  * private static final MethodHandles.Lookup LOOKUP = MethodHandles.lookup();
  *
  * Class<?> hidden = HiddenClassDefiner.tryDefineNestmate(LOOKUP, bytecode, false);
- * if (hidden == null) {
- *     // fall back to ClassLoader.defineClass(...)
- * }
  * }</pre>
+ * The {@link Lookup#lookupClass() lookup class} determines the hidden class's
+ * defining loader, run-time package, protection domain, and nest host. A lookup
+ * from {@code ReflectorLoader} is therefore <em>not</em> interchangeable with
+ * one from {@code ProxyGeneratorAdapter} or from this utility: each hosts
+ * nestmates in a different package / loader / nest. What is shared is only the
+ * define policy (package alignment, {@code NESTMATE}+weak options, soft-fail).
+ *
+ * <p>A lookup captured inside {@code HiddenClassDefiner} only has full privilege
+ * for {@code HiddenClassDefiner} itself. It is used solely as the <em>caller</em>
+ * argument to {@link MethodHandles#privateLookupIn(Class, Lookup)} in the
+ * foreign-host overload — never as the nest host of production generators.
  *
  * <p>The overload {@link #tryDefineNestmate(Class, byte[], boolean)} is a
  * <em>best-effort</em> helper for foreign hosts (user classes Groovy does not
- * control). It uses {@link MethodHandles#privateLookupIn(Class, Lookup)} from
- * this class and therefore succeeds only when the host's package is accessible
- * to Groovy's module — typically true for unnamed-module application classes,
- * and typically false for sealed / unopened packages such as those in
- * {@code java.base}. Callers must always handle a {@code null} result.
+ * control). It succeeds only when the host's package is accessible to Groovy's
+ * module (typical for unnamed-module application classes; not for sealed /
+ * unopened packages such as those in {@code java.base}). Callers must always
+ * handle a {@code null} result and fall back to {@link ClassLoader#defineClass}.
  *
- * <h2>What this utility centralises</h2>
+ * <h2>Soft-fail contract</h2>
+ * <p>{@code try*} methods return {@code null} on expected failure modes so call
+ * sites can fall back with one null check:
  * <ul>
- *   <li>{@code NESTMATE} + weak-lifecycle policy for dynamic Groovy classes;</li>
- *   <li>rewriting {@code this_class} into the lookup class's package (required
- *       by {@link Lookup#defineHiddenClass});</li>
- *   <li>a soft API that returns {@code null} on the <em>expected</em> failure
- *       modes ({@link IllegalAccessException}, {@link IllegalArgumentException},
- *       {@link SecurityException}, {@link LinkageError}) so call sites fall back
- *       to {@link ClassLoader#defineClass} with one null check. Unexpected
- *       failures (e.g. programming errors in callers) are not swallowed as a
- *       blanket {@link RuntimeException}.</li>
+ *   <li>{@link IllegalAccessException}, {@link SecurityException}</li>
+ *   <li>{@link LinkageError}</li>
+ *   <li>{@link IllegalArgumentException}, {@link IndexOutOfBoundsException}
+ *       (invalid class-file bytes / ASM)</li>
+ *   <li>GraalVM native-image "feature unsupported" errors, detected by class
+ *       name ({@code com.oracle.svm.core.jdk.UnsupportedFeatureError}) without
+ *       a build dependency on GraalVM</li>
  * </ul>
+ * Other {@link Error}s and unexpected failures are rethrown.
  *
- * <h2>Kill switch</h2>
- * <p>{@code -Dgroovy.hidden.classes.disable=true} forces every {@code try*}
- * method to return {@code null}.
+ * <h2>Enablement (kill switch and native image)</h2>
+ * <p>{@link #isEnabled()} is evaluated on each call (class definition is not a
+ * hot path). That keeps {@code -Dgroovy.hidden.classes.disable=true} effective
+ * under GraalVM native image, where a {@code static final} snapshot taken at
+ * <em>build-time</em> class-init would freeze the wrong answer. When running
+ * inside a native image ({@code org.graalvm.nativeimage.imagecode=runtime}),
+ * enablement is {@code false}: runtime class definition is unsupported there,
+ * so the soft path never attempts it.
  *
  * @since 6.0.0
  * @see Lookup#defineHiddenClass(byte[], boolean, Lookup.ClassOption...)
  */
 public final class HiddenClassDefiner {
 
-    /** System property that disables hidden-class definitions. */
+    /** System property that disables hidden-class definitions at run time. */
     public static final String PROPERTY_DISABLE = "groovy.hidden.classes.disable";
 
     /**
-     * {@code true} when hidden-class definitions are globally disabled.
-     * Evaluated once at class-init so hot paths pay no property-lookup cost.
+     * GraalVM property set in native images. Value {@code "runtime"} means the
+     * current process is executing an already-built native image (as opposed to
+     * {@code "buildtime"} during image generation). Package-private: not a
+     * public configuration surface — use {@link #isEnabled()}.
      */
-    public static final boolean HIDDEN_CLASSES_DISABLED =
-            SystemUtil.getBooleanSafe(PROPERTY_DISABLE, false);
+    static final String PROPERTY_NATIVE_IMAGE_CODE = "org.graalvm.nativeimage.imagecode";
+
+    /** Value of {@link #PROPERTY_NATIVE_IMAGE_CODE} while executing a native image. */
+    static final String NATIVE_IMAGE_CODE_RUNTIME = "runtime";
+
+    /**
+     * Fully-qualified name of GraalVM's "feature not supported at runtime" error.
+     * Matched by name to avoid a compile-time dependency on {@code org.graalvm.*}.
+     */
+    static final String UNSUPPORTED_FEATURE_ERROR =
+            "com.oracle.svm.core.jdk.UnsupportedFeatureError";
 
     /**
      * Lookup for <em>this</em> class only — used exclusively as the caller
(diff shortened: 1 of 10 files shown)


blackdrag (commented):
- HiddenClassDefiner contains method used only in tests
- ProxyGeneratorAdapter and ReflectorLoader seems to duplicate work already done in HiddenClassDefiner, while still using HiddenClassDefiner

But the big point for me is `MethodHandles.lookup()`. It has the javadoc:

> Returns a [lookup object](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html) with full capabilities to emulate all supported bytecode behaviors of the caller. These capabilities include [full privilege access](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html#hasFullPrivilegeAccess()) to the caller. Factory methods on the lookup object can create [direct method handles](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandleInfo.html#directmh) for any member that the caller has access to via bytecodes, including protected and private fields and methods. This lookup object is created by the original lookup class and has the [ORIGINAL](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html#ORIGINAL) bit set. This lookup object is a capability which may be delegated to trusted agents. Do not store it in place where untrusted code can access it.
>
> This method is caller sensitive, which means that it may return different values to different callers. In cases where MethodHandles.lookup is called from a context where there is no caller frame on the stack (e.g. when called directly from a JNI attached thread), IllegalCallerException is thrown. To obtain a [lookup object](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandles.Lookup.html) in such a context, use an auxiliary class that will implicitly be identified as the caller, or use [publicLookup()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/invoke/MethodHandles.html#publicLookup()) to obtain a low-privileged lookup instead.

So if I create a Lookup for HiddenClassDefinerTest in HiddenClassDefinerTest I will get full access. But if I create that in the same class for java.lang.String, I will not get full access. In practice this usually means that unless the class creates the lookup (for example during invokedynamic) it is useless for many things. So if you need private module access, then `MethodHandles.lookup()` is most likely not the way to go.
Which means I don`t think that the intend of HiddenClassDefiner.privateLookupIn and usuability go hand in hand here.

daniellansun:
@blackdrag  Thanks for the careful review — those points were well taken. Here is how the current design addresses them.

## Test-only API on `HiddenClassDefiner`

Agreed. The extra helpers that existed only for tests (`findConstructor`, strict multi-option `define*` overloads, etc.) have been removed. Production surface is now essentially:

- `isEnabled()` / the disable property
- `tryDefineNestmate(Lookup, bytes, initialize)` — preferred
- `tryDefineNestmate(Class, bytes, initialize)` — best-effort for foreign hosts only

## Duplication with `ProxyGeneratorAdapter` / `ReflectorLoader`

Also agreed that define policy should not be reimplemented at each call site. Class definition (package alignment, `NESTMATE` + weak lifecycle, soft-fail → `null`) lives only in `HiddenClassDefiner`.

- `ReflectorLoader` / `ClassLoaderForClassArtifacts` are thin: one `tryDefineNestmate` + visible `defineClass` fallback.
- Proxy-specific *policy* (when a proxy may be hidden, which host to try, MockFor-style “must stay nameable” aggregates) was pulled out of `ProxyGeneratorAdapter` into package-private `ProxyClassDefiner`, so the adapter stays focused on bytecode generation and no longer reimplements the define path.

## `MethodHandles.lookup()` and private module access

Your reading of the javadoc matches how we now structure the API.

A `Lookup` only carries full privilege for the class that actually called `MethodHandles.lookup()`. Capturing a lookup inside `HiddenClassDefiner` (or any other utility) cannot magically grant full access into arbitrary hosts such as `java.lang.String`. Relying on that for private/module access would be the wrong model.

So the intended split is:

1. **Preferred:** `tryDefineNestmate(Lookup, …)` where the **nest-host class** owns the lookup, e.g.  
   `private static final Lookup LOOKUP = MethodHandles.lookup();`  
   on `ReflectorLoader` / `ProxyGeneratorAdapter`. That is the correct caller-sensitive pattern.

2. **Best-effort only:** `tryDefineNestmate(Class, …)` for **foreign** hosts (user classes we do not control). It uses `privateLookupIn` from `HiddenClassDefiner` and is documented to succeed only when the host package is accessible to Groovy’s module (typical for unnamed application classes; not a general substitute for a host-created `Lookup`). Callers must always handle `null` and fall back to `ClassLoader.defineClass`.

We no longer expose a public `HiddenClassDefiner.privateLookupIn` that would suggest “full access for any class from a utility.” That intent and usability are no longer conflated.

Happy to adjust naming or docs further if anything still feels misleading.

src/main/java/org/apache/groovy/util/HiddenClassDefiner.java
```diff
@@ -0,0 +1,215 @@
...
+     * @param lookup     full-privilege lookup for the nest host
+     * @param bytes      class-file bytes
+     * @param initialize {@code true} to run {@code <clinit>} immediately
+     * @return the hidden class, or {@code null} if definition is not possible
+     */
+    public static Class<?> tryDefineNestmate(
+            final Lookup lookup,
+            final byte[] bytes,
+            final boolean initialize) {
+        if (HIDDEN_CLASSES_DISABLED || lookup == null || bytes == null) {
```
blackdrag: Did I see it wrong or is this effectively called only with MethodHandles.lookup()? Does it really matter if it is one time HiddenClassDefiner and another time ReflectorLoader?

paulk-asert:
I haven't done a proper review yet, but as part of some other work, I assessed whether the PR impacts potential GraalVM support if we try harder to support that in the future. It came back with below, I'm not sure we want to do what is says yet - but just wanted to capture it somewhere for now:

> Two native-image observations from exercising this branch alongside the packed-closure work (GROOVY-12227). Both are small; the first is a genuine easy win.
> 
> ### 1. The soft-fail contract doesn't hold under native image
> 
> `tryDefineNestmate` catches `IllegalAccessException | SecurityException | LinkageError` (plus `IllegalArgumentException` and `IndexOutOfBoundsException`), and the class javadoc documents the intent as returning `null` on the expected failure modes "so call sites fall back to `ClassLoader#defineClass` with one null check".
> 
> GraalVM signals "this runtime cannot define classes" with `com.oracle.svm.core.jdk.UnsupportedFeatureError`, which extends `java.lang.Error` **directly**:
> 
> ```
> $ javap com/oracle/svm/core/jdk/UnsupportedFeatureError.class   # from lib/svm/builder/svm.jar, GraalVM CE 25.2.4
> public class com.oracle.svm.core.jdk.UnsupportedFeatureError extends java.lang.Error {
> ```
> 
> It is not a `LinkageError`, so it escapes the catch and propagates out of the `try*` method — in precisely the environment where the fallback matters most.
> 
> A blanket `catch (Throwable)` would contradict the javadoc's "unexpected failures ... are not swallowed as a blanket `RuntimeException`", so something targeted is probably wanted, e.g.:
> 
> ```java
> } catch (Error e) {
>     // GraalVM native image: runtime class definition is unsupported. Name-checked
>     // to avoid a build-time dependency on org.graalvm.
>     if ("com.oracle.svm.core.jdk.UnsupportedFeatureError".equals(e.getClass().getName())) {
>         return null;
>     }
>     throw e;
> }
> ```
> 
> ### 2. The kill switch is baked in at build time
> 
> ```java
> public static final boolean HIDDEN_CLASSES_DISABLED =
>         SystemUtil.getBooleanSafe(PROPERTY_DISABLE, false);
> ```
> 
> `static final`, documented as "evaluated once at class-init so hot paths pay no property-lookup cost". Under native image this class is very likely initialized at *build* time, so the value captured is the build JVM's, and `-Dgroovy.hidden.classes.disable=true` at run time silently does nothing — the one escape hatch a native user would reach for.
> 
> This is the same trap I hit in GROOVY-12227: I ended up evaluating the equivalent check per link rather than caching it in a static, because a build-time-initialized class bakes in the wrong answer (the image-code property reports `buildtime` there, not `runtime`).
> 
> ### Possibly one fix for both
> 
> If `isEnabled()` did a per-call check that also returned `false` when
> `"runtime".equals(System.getProperty("org.graalvm.nativeimage.imagecode"))`, then the native path would never attempt the definition at all, the kill switch would work at run time, and (1) becomes belt-and-braces rather than load-bearing. If the hot-path cost of the property read is the concern, a non-final holder initialised on first *use* rather than at class-init keeps both properties.
> 
> ### Caveat on incidence
> 
> I have not observed (1) fire in practice — the coercion I tested (`[run: { ... }] as Runnable`) goes through `java.lang.reflect.Proxy` and works natively on both master and this branch, so it does not reach `HiddenClassDefiner`. This is from reading the code plus confirming the class hierarchy, not from a reproduced failure. Worth a targeted test if you think the proxy/reflector paths are reachable in a native image.

daniellansun:
@paulk-asert  Thanks for capturing this — both points are real and we’ve addressed them in the current code.

### 1. Soft-fail under native image

Agreed that `UnsupportedFeatureError` extends `Error` directly and was not covered by `LinkageError`.

`tryDefineNestmate` now soft-fails that case by **class name** (`com.oracle.svm.core.jdk.UnsupportedFeatureError`) with no GraalVM compile dependency, and rethrows other `Error`s. The shared path is `softFailOrRethrow(Error)`.

### 2. Kill switch baked in at build time

Agreed. The old `static final HIDDEN_CLASSES_DISABLED` snapshot was unsafe if the class is initialized at native-image build time.

We removed that field. **`isEnabled()` is evaluated on each call** (class definition is not a hot path):

- returns `false` when `org.graalvm.nativeimage.imagecode=runtime` (skip define entirely);
- otherwise honours `-Dgroovy.hidden.classes.disable=true` at **run** time.

So native users get a working escape hatch, and the soft-fail on `UnsupportedFeatureError` is belt-and-braces if something still attempts define.

Happy to adjust further if you hit a concrete native path that still reaches `HiddenClassDefiner` unexpectedly.

src/main/java/org/apache/groovy/util/HiddenClassDefiner.java
daniellansun: You are not wrong that every production `Lookup` we pass in is obtained via `MethodHandles.lookup()` — we do not use another factory for that path.

What still matters is **which class** executes that call.

`Lookup.lookupClass()` becomes the nest host and fixes the hidden class’s defining loader, run-time package, and protection domain. So:

- `ReflectorLoader`’s `LOOKUP` → nestmates of `ReflectorLoader` (its package / loader)
- `ProxyGeneratorAdapter`’s `LOOKUP` → nestmates of `ProxyGeneratorAdapter`
- foreign hosts via `tryDefineNestmate(Class, …)` → nestmates of that user class (`privateLookupIn`)

Those are not interchangeable. Sharing only the *define policy* (package alignment, `NESTMATE`+weak, soft-fail) in `HiddenClassDefiner` does not make the lookups equivalent.

`HiddenClassDefiner`’s own `LOOKUP` is used only as the **caller** argument to `privateLookupIn` for the foreign-host overload — never as the nest host for those generators.

So: same API (`lookup()`), different ownership — and yes, that ownership still matters.

blackdrag: My Point is that this code is always called from within the Groovy Runtime. Which means MethodHandles.lookup() will always return the same. It will be for the module the runtime is in and allow access to public methods in that module as well as accessible methods in other classes. And that later case is the potential problem, since if the runtime has no access to a different module the ProxyAdapter will not have that too. You can do that MethodHandles.lookup() anywhere in the Groovy runtime to get the same result. It also means the private lookup is restricted. What does it mean for nest mates? Since we need private access privileges for the target host class I have high doubts that this works all as intended in a module system. Just try to create a nest mate for String for example.

Looking at the usage I see ReflectorLoader. I found only a test, no actual code. Which actually means this class should be deprecated/removed. 
ProxyGeneratorAdapter is used by InlinedASTCustomizerFactory and ProxyGenerator. InlinedASTCustomizerFactory is probably fine, but for ProxyGenerator: what if the interface is not accessible by the runtime? And I mean groovy can inspect the class, the methods are public? Map for example is no problem. It is in a different and restricted module, but Map itself is public API. The scenario I always keep in mind for such cases is that of 2 modules. Module A in Java, Module B in Groovy and Module C, the Groovy runtime. And B uses A. B has to give C access, but why should A give C access? And then all kinds of things may start failing. So I try to improve the code base step-by-step in a way that can handle such cases for the time when Groovy becomes a real module and has to handle Groovy programs in real modules.

Now you probably say something like that this is why we have the fallback. And you are right about that. But assume we are in the modules case, how useful is the nestmate variant then in the end? What would it cover outside of module B and C? And is that enough for us? If you think it is good enough feel free to resolve this conversation.

daniellansun: Thanks — agreed on the module-level point, and it is a fair stress test for this design.

**`MethodHandles.lookup()` and private access**

You are right that every production call site sits in the Groovy runtime (module C). So `lookup()` always carries the same *module* rights: public API C can already see, plus reflective access only where packages are open to C. Capturing the lookup in `ProxyGeneratorAdapter` rather than elsewhere in the runtime does not open module A. `privateLookupIn` into an unopened package fails; a nestmate of `String` is the canonical example, and we treat that as expected.

What still differs between those lookups is only the *nest host* (package, defining loader, protection domain, nest membership). That still matters for unloadability and linkage, but it is not a privilege escalation into a foreign module.

We now pre-filter foreign hosts with `Module.isOpen` (`canAttemptPrivateLookup`, `@Internal`) so unopened platform types never pay for a guaranteed `IllegalAccessException`, and soft-fail to `null` so callers fall back to `ClassLoader.defineClass`.

**`ReflectorLoader`**

Agreed — no production callers. It is `@Deprecated(since = "6.0.0", forRemoval = true)`, the hidden-class path is gone (plain `defineClass` only), and tests are reduced to binary-compat smoke. Happy to remove the class entirely in a follow-up when we drop that surface.

**`ProxyGenerator` / accessibility of interfaces**

Public types such as `Map` are fine as *nominal* supers/interfaces: they are exported API, and C can already resolve them without `privateLookupIn`. The hard case is not “public method on a type in another module”, but private reflective entry into a package that module never opened to C. We do not claim nestmates solve that.

Under A (Java library) / B (Groovy program) / C (runtime):

1. **Caller-owned nestmate** (of a type in C) — when every type the bytecode names is resolvable from C’s loader.
2. **Foreign host** — best-effort `privateLookupIn`, only when the host package is open to C (typical for unnamed-module B; not for strongly encapsulated A or `java.base`).
3. **Visible `defineClass`** — intentional path when neither nestmate option applies.

**Is the nestmate path useful enough outside B and C?**

Honestly, outside open/unnamed B and C it mostly is not — and that is deliberate. Nestmates are for non-discoverable, eagerly unloadable *generated* classes where C (or B) can already host them, not a tunnel into A’s encapsulation. For restricted A we rely on the visible fallback, same as before this PR for those cases.

For the goals of this change (less class-space pollution, better unloading, clear soft-fail under modules and native image), I think that coverage is enough. Happy to resolve on that basis; please say if you still see a gap under the A/B/C layout.

blackdrag (approved):

paulk-asert (approved):

PR #2757, opened 2026-08-03
validate the four hex digits of a json \u escape

netliomax25-code:
Repro: `new JsonSlurper().parseText('{"k":"\u12"}').k` returns the two-character string `12` instead of failing, and `"\u+041"` decodes to `A`; both hold for every `JsonParserType`, while `JsonSlurperClassic` rejects the same documents because `JsonTokenType.STRING` already validates `\\u[0-9a-fA-F]{4}`.
Cause: `CharBuf.decodeJsonString` reads the digits with `Integer.parseInt(hex, 16)`, which accepts a leading sign and non-ASCII digits, and its `index + 4 < to` guard has no else branch, so a truncated escape is left unconsumed and its digits fall through the loop as text.
Fix: decode the four digits through a strict ASCII hex check and reject a short or non-hex escape the way the neighboring invalid-escape branch already does.

### DIFF
--- a/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/CharBuf.java
+++ b/subprojects/groovy-json/src/main/java/org/apache/groovy/json/internal/CharBuf.java
@@ -906,12 +906,14 @@ public class CharBuf extends Writer implements CharSequence {
                             break;
 
                         case 'u':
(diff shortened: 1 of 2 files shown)


paulk-asert:
Merged as part of PR#2760. Thanks!

PR #2771, opened 2026-08-07
GROOVY-12238: SecureASTCustomizer does not check constructors, initia…

paulk-asert:
…lizer blocks or field initializers

SecureASTCustomizer visited the script statement block and method bodies only, so code outside a method body escaped every configured restriction: disallowedReceivers, the statement and expression allow/deny lists, and any registered StatementChecker or ExpressionChecker. With disallowedReceivers = ['java.lang.System'], a call in a constructor, a static or instance initializer block, or a field initializer all compiled and ran, while the same call in the script body was correctly rejected.

The existing filters could not reach these. A static initializer ends up in <clinit>, which is synthetic and so excluded by filterMethods; instance initializers live in a separate getObjectInitializerStatements() list; and field initializers hang off FieldNode, whose property backing fields are themselves synthetic.

Add visitConstructorsAndInitializers(), applying the securing visitor to declared constructors, object initializer statements, the statements inside <clinit>, and field initial expressions.

Only nodes carrying a source position are visited. Constructors and initializers are not written solely by the author of the secured source: every script class has generated constructors, and AST transformations add their own. Visiting those rejects valid programs -- a first cut broke four existing tests on the script class's generated super(Binding) call, which is not marked synthetic and so cannot be excluded by any flag. Note the <clinit> wrapper block is synthetic even when its statements are not, so the check is applied per statement.

Tests cover each closed gap, keep the script-body control, and pin the exemption for generated constructors so a later simplification cannot drop the source-position check unnoticed. Both Limitations sections, in the user guide and the javadoc, are updated to match.

Constructors still do not count towards methodDefinitionAllowed, and annotation members remain unvisited; both are separable changes.

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
+++ b/src/main/java/org/codehaus/groovy/control/customizers/SecureASTCustomizer.java
@@ -18,8 +18,11 @@
  */
 package org.codehaus.groovy.control.customizers;
 
+import org.codehaus.groovy.ast.ASTNode;
 import org.codehaus.groovy.ast.ClassNode;
 import org.codehaus.groovy.ast.CodeVisitorSupport;
+import org.codehaus.groovy.ast.ConstructorNode;
+import org.codehaus.groovy.ast.FieldNode;
 import org.codehaus.groovy.ast.GroovyCodeVisitor;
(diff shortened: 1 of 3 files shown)


paulk-asert:
Included in #2779

PR #2783, opened 2026-08-13
GROOVY-12254: Make defer fully contextual: parse as a keyword only inside async closures

paulk-asert:
See: https://issues.apache.org/jira/browse/GROOVY-12254


blackdrag (approved):

daniellansun:
Merged. Thanks.

PR #2784, opened 2026-08-13
GROOVY-12255: Compile switch expressions as first-class AST (no closure desugar)

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12255

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -19,7 +19,6 @@
 package org.apache.groovy.parser.antlr4;
 
 import groovy.lang.Tuple2;
-import groovy.lang.Tuple3;
 import groovy.transform.CompileStatic;
 import groovy.transform.NonSealed;
 import groovy.transform.Sealed;
@@ -1031,38 +1030,37 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
         switchExpressionRuleContextStack.push(ctx);
         try {
             validateSwitchExpressionLabels(ctx);
-            List<Tuple3<List<Statement>, Boolean, Boolean>> statementsAndArrowAndYieldOrThrow =
-                    ctx.switchBlockStatementExpressionGroup().stream().map(this::visitSwitchBlockStatementExpressionGroup).toList();
-            if (statementsAndArrowAndYieldOrThrow.isEmpty()) {
+            List<Statement> statementList = ctx.switchBlockStatementExpressionGroup().stream()
+                    .map(this::visitSwitchBlockStatementExpressionGroup)
+                    .reduce(new LinkedList<>(), (r, e) -> {
+                        r.addAll(e);
+                        return r;
+                    });
+            if (statementList.isEmpty()) {
                 throw createParsingFailedException("`case` or `default` branches are expected", ctx.LBRACE());
-            } else {
-                // Last group must complete: every path yields or throws (JEP 361).
-                // Intermediate colon groups may still fall through.
-                var lastGroup = last(statementsAndArrowAndYieldOrThrow);
-                List<Statement> lastStmts = lastGroup.getV1();
-                Statement lastArm = lastStmts.get(lastStmts.size() - 1);
-                Statement lastCode = lastArm instanceof CaseStatement cs ? cs.getCode() : lastArm;
-                if (mayCompleteNormally(lastCode)) {
-                    throw createParsingFailedException("`yield` or `throw` is expected", lastArm);
-                }
+            }
+
+            // Last group must complete: every path yields or throws (JEP 361).
+            // Intermediate colon groups may still fall through.
+            Statement lastArm = last(statementList);
+            Statement lastCode = lastArm instanceof CaseStatement cs ? cs.getCode() : lastArm;
+            if (mayCompleteNormally(lastCode)) {
+                throw createParsingFailedException("`yield` or `throw` is expected", lastArm);
             }
 
             List<CaseStatement> caseStatements = new ArrayList<>();
             Statement defaultStatement = null;
-
-            for (var tuple : statementsAndArrowAndYieldOrThrow) {
-                for (Statement s : tuple.getV1()) {
-                    if (s instanceof CaseStatement c) {
-                        if (defaultStatement != null) {
-                            throw createParsingFailedException("default case should appear last", defaultStatement);
-                        }
-                        caseStatements.add(c);
-                    } else if (isTrue(s, IS_SWITCH_DEFAULT)) {
-                        if (defaultStatement != null) {
-                            throw createParsingFailedException("switch expression should have only one default case", s);
-                        }
-                        defaultStatement = s;
+            for (Statement statement : statementList) {
+                if (statement instanceof CaseStatement caseStatement) {
+                    if (defaultStatement != null) {
+                        throw createParsingFailedException("default case should appear last", defaultStatement);
                     }
+                    caseStatements.add(caseStatement);
+                } else if (isTrue(statement, IS_SWITCH_DEFAULT)) {
+                    if (defaultStatement != null) {
+                        throw createParsingFailedException("switch expression should have only one default case", statement);
+                    }
+                    defaultStatement = statement;
                 }
             }
 
@@ -1070,8 +1068,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
                     new SwitchExpression(
                             this.visitExpressionInPar(ctx.expressionInPar()),
                             caseStatements,
-                            defaultStatement != null ? defaultStatement : EmptyStatement.INSTANCE
-                    ),
+                            defaultStatement != null ? defaultStatement : EmptyStatement.INSTANCE),
                     ctx);
         } finally {
             switchExpressionRuleContextStack.pop();
@@ -1079,128 +1076,86 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     }
 
     @Override
-    public Tuple3<List<Statement>, Boolean, Boolean> visitSwitchBlockStatementExpressionGroup(SwitchBlockStatementExpressionGroupContext ctx) {
-        int labelCnt = ctx.switchExpressionLabel().size();
-        List<Token> firstLabelHolder = new ArrayList<>(1);
-        final int[] arrowCntHolder = new int[1];
+    public List<Statement> visitSwitchBlockStatementExpressionGroup(final SwitchBlockStatementExpressionGroupContext ctx) {
+        List<SwitchExpressionLabel> labels = ctx.switchExpressionLabel().stream()
+                .map(this::switchExpressionLabel)
+                .toList();
+        if (labels.isEmpty()) {
+            throw createParsingFailedException("`case` or `default` branches are expected", ctx);
+        }
 
-        boolean[] isArrowHolder = new boolean[1];
-        boolean[] hasResultStmtHolder = new boolean[1];
-        List<Statement> result = (List<Statement>) ctx.switchExpressionLabel().stream()
-                .map(e -> (Object) this.visitSwitchExpressionLabel(e))
-                .reduce(new ArrayList<Statement>(4), (r, e) -> {
-                    List<Statement> statementList = (List<Statement>) r;
-                    Tuple3<Token, List<Expression>, Integer> tuple = (Tuple3<Token, List<Expression>, Integer>) e;
+        if (labels.stream().anyMatch(SwitchExpressionLabel::isArrow) && labels.size() > 1) {
+            throw createParsingFailedException("`case ... ->` does not support falling through cases", labels.get(0).keyword());
+        }
 
-                    boolean isArrow = ARROW == tuple.getV3();
-                    isArrowHolder[0] = isArrow;
-                    if (isArrow) {
-                        if (++arrowCntHolder[0] > 1 && !firstLabelHolder.isEmpty()) {
-                            throw createParsingFailedException("`case ... ->` does not support falling through cases", firstLabelHolder.get(0));
-                        }
-                    }
+        BlockStatement arm = this.switchExpressionArm(ctx, labels.get(0).isArrow());
+        Token firstKeyword = labels.get(0).keyword();
+        SwitchExpressionLabel lastLabel = last(labels);
 
-                    boolean isLast = labelCnt - 1 == statementList.size();
+        return labels.stream().collect(
+                ArrayList::new,
+                (cases, label) -> this.addSwitchExpressionCases(cases, label, arm, firstKeyword, label == lastLabel),
+                ArrayList::addAll);
+    }
 
-                    BlockStatement codeBlock = this.visitBlockStatements(ctx.blockStatements());
-                    List<Statement> statements = codeBlock.getStatements();
-                    int statementsCnt = statements.size();
-                    if (0 == statementsCnt) {
-                        throw createParsingFailedException("`yield` is expected", ctx.blockStatements());
-                    }
+    private BlockStatement switchExpressionArm(final SwitchBlockStatementExpressionGroupContext ctx, final boolean arrow) {
+        BlockStatement block = this.visitBlockStatements(ctx.blockStatements());
+        List<Statement> statements = block.getStatements();
+        if (statements.isEmpty()) {
+            throw createParsingFailedException("`yield` is expected", ctx.blockStatements());
+        }
+        if (!arrow) {
+            return block;
+        }
+        if (statements.size() > 1) {
+            throw createParsingFailedException("Expect only 1 statement, but " + statements.size() + " statements found", ctx.blockStatements());
+        }
 
-                    if (isArrow && statementsCnt > 1) {
-                        throw createParsingFailedException("Expect only 1 statement, but " + statementsCnt + " statements found", ctx.blockStatements());
-                    }
+        Statement body = unwrapSingletonBlock(statements.get(0));
+        if (body instanceof YieldStatement || body instanceof ThrowStatement) {
+            return configureAST(createBlockStatement(body), body);
+        }
+        if (body instanceof ExpressionStatement expressionStatement) {
+            return configureAST(
+                    createBlockStatement(configureAST(yieldS(expressionStatement.getExpression()), body)),
+                    body);
+        }
+        if (mayCompleteNormally(body)) {
+            throw createParsingFailedException("`yield` or `throw` is expected", body);
+        }
+        return configureAST(createBlockStatement(body), body);
+    }
 
-                    if (!isArrow) {
-                        boolean[] hasYieldHolder = new boolean[1];
-                        boolean[] hasThrowHolder = new boolean[1];
-                        codeBlock.visit(new CodeVisitorSupport() {
-                            @Override
-                            public void visitYieldStatement(final YieldStatement statement) {
-                                hasYieldHolder[0] = true;
-                            }
+    private void addSwitchExpressionCases(
+            final List<Statement> cases,
+            final SwitchExpressionLabel label,
+            final BlockStatement arm,
+            final Token firstKeyword,
+            final boolean lastLabel) {
+        switch (label.keyword().getType()) {
+            case CASE -> {
+                List<Expression> values = label.values();
+                for (int i = 0, n = values.size(); i < n; i++) {
+                    CaseStatement caseStatement = new CaseStatement(
+                            values.get(i),
+                            lastLabel && i == n - 1 ? arm : EmptyStatement.INSTANCE);
+                    caseStatement.setArrow(label.isArrow());
+                    cases.add(configureAST(caseStatement, firstKeyword));
+                }
+            }
+            case DEFAULT -> {
+                arm.putNodeMetaData(IS_SWITCH_DEFAULT, Boolean.TRUE);
+                cases.add(arm);
+            }
+            default -> throw createParsingFailedException("Unsupported switch label: " + label.keyword().getText(), label.keyword());
+        }
+    }
 
-                            @Override
-                            public void visitThrowStatement(ThrowStatement statement) {
-                                hasThrowHolder[0] = true;
-                            }
-
-                            @Override
-                            public void visitClosureExpression(final ClosureExpression expression) {
-                                // yield inside a nested closure does not complete this arm
-                            }
-                        });
-
-                        if (hasYieldHolder[0] || hasThrowHolder[0]) {
-                            hasResultStmtHolder[0] = true;
-                        }
-                    }
-
-                    Statement exprOrBlockStatement = statements.get(0);
-                    if (exprOrBlockStatement instanceof BlockStatement blockStatement) {
-                        List<Statement> branchStatementList = blockStatement.getStatements();
-                        if (1 == branchStatementList.size()) {
-                            exprOrBlockStatement = branchStatementList.get(0);
-                        }
-                    }
-
-                    if (!(exprOrBlockStatement instanceof YieldStatement || exprOrBlockStatement instanceof ThrowStatement)) {
-                        if (isArrow) {
-                            if (exprOrBlockStatement instanceof ExpressionStatement expressionStatement) {
-                                codeBlock = configureAST(
-                                        createBlockStatement(configureAST(
-                                                yieldS(expressionStatement.getExpression()),
-                                                exprOrBlockStatement
-                                        )),
-                                        exprOrBlockStatement
-                                );
-                            } else if (!containsYieldOrThrow(exprOrBlockStatement)) {
-                                throw createParsingFailedException("`yield` or `throw` is expected", exprOrBlockStatement);
-                            } else {
-                                codeBlock = configureAST(
-                                        createBlockStatement(exprOrBlockStatement),
-                                        exprOrBlockStatement
-                                );
-                            }
-                        }
-                    } else if (exprOrBlockStatement instanceof YieldStatement || exprOrBlockStatement instanceof ThrowStatement) {
-                        codeBlock = configureAST(
-                                createBlockStatement(exprOrBlockStatement),
-                                exprOrBlockStatement
-                        );
-                    }
-
-                    if (isArrow && mayCompleteNormally(codeBlock)) {
-                        throw createParsingFailedException("`yield` or `throw` is expected", exprOrBlockStatement);
-                    }
-
-                    switch (tuple.getV1().getType()) {
-                        case CASE:
-                            if (!asBoolean(statementList)) {
-                                firstLabelHolder.add(tuple.getV1());
-                            }
-                            for (int i = 0, n = tuple.getV2().size(); i < n; i += 1) {
-                                Expression expr = tuple.getV2().get(i);
-                                CaseStatement caseStatement = new CaseStatement(
-                                        expr,
-                                        (isLast && i == n - 1) ? codeBlock : EmptyStatement.INSTANCE
-                                );
-                                caseStatement.setArrow(isArrow);
-                                statementList.add(configureAST(caseStatement, firstLabelHolder.get(0)));
-                            }
-                            break;
-                        case DEFAULT:
-                            codeBlock.putNodeMetaData(IS_SWITCH_DEFAULT, Boolean.TRUE);
-                            statementList.add(codeBlock);
-                            break;
-                    }
-
-                    return statementList;
-                });
-
-        return tuple(result, isArrowHolder[0], hasResultStmtHolder[0]);
+    private static Statement unwrapSingletonBlock(final Statement statement) {
+        if (statement instanceof BlockStatement block && block.getStatements().size() == 1) {
+            return block.getStatements().get(0);
+        }
+        return statement;
     }
 
     /**
@@ -1220,27 +1175,6 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
         return false;
     }
 
-    private static boolean containsYieldOrThrow(final Statement statement) {
-        boolean[] found = new boolean[1];
-        statement.visit(new CodeVisitorSupport() {
-            @Override
-            public void visitYieldStatement(final YieldStatement yieldStatement) {
-                found[0] = true;
-            }
-
-            @Override
-            public void visitThrowStatement(final ThrowStatement throwStatement) {
-                found[0] = true;
-            }
-
-            @Override
-            public void visitClosureExpression(final ClosureExpression expression) {
-                // do not look inside nested closures / lambdas
-            }
-        });
-        return found[0];
-    }
-
     private void validateSwitchExpressionLabels(SwitchExpressionContext ctx) {
         Map<String, List<SwitchExpressionLabelContext>> acMap =
                 ctx.switchBlockStatementExpressionGroup().stream()
@@ -1253,14 +1187,17 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     }
 
     @Override
-    public Tuple3<Token, List<Expression>, Integer> visitSwitchExpressionLabel(SwitchExpressionLabelContext ctx) {
-        final Integer acType = ctx.ac.getType();
-        if (asBoolean(ctx.CASE())) {
-            return tuple(ctx.CASE().getSymbol(), this.visitExpressionList(ctx.expressionList()), acType);
-        } else if (asBoolean(ctx.DEFAULT())) {
-            return tuple(ctx.DEFAULT().getSymbol(), Collections.singletonList(EmptyExpression.INSTANCE), acType);
-        }
+    public Object visitSwitchExpressionLabel(final SwitchExpressionLabelContext ctx) {
+        return this.switchExpressionLabel(ctx);
+    }
 
+    private SwitchExpressionLabel switchExpressionLabel(final SwitchExpressionLabelContext ctx) {
+        if (asBoolean(ctx.CASE())) {
+            return new SwitchExpressionLabel(ctx.CASE().getSymbol(), this.visitExpressionList(ctx.expressionList()), ctx.ac.getType());
+        }
+        if (asBoolean(ctx.DEFAULT())) {
+            return new SwitchExpressionLabel(ctx.DEFAULT().getSymbol(), Collections.singletonList(EmptyExpression.INSTANCE), ctx.ac.getType());
+        }
         throw createParsingFailedException("Unsupported switch expression label: " + ctx.getText(), ctx);
     }
 
@@ -4965,6 +4902,12 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     }
 
     //--------------------------------------------------------------------------
+
+    private record SwitchExpressionLabel(Token keyword, List<Expression> values, int separator) {
+        private boolean isArrow() {
+            return ARROW == separator;
+        }
+    }
 
     private static class DeclarationListStatement extends Statement {
 
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -171,6 +171,7 @@ import static org.codehaus.groovy.ast.tools.GeneralUtils.callX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.closureX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.declX;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.listX;
+import static org.codehaus.groovy.ast.tools.GeneralUtils.mayCompleteNormally;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.returnS;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.stmt;
 import static org.codehaus.groovy.ast.tools.GeneralUtils.yieldS;
@@ -1021,10 +1022,9 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
     }
 
     /**
-     * Builds a first-class {@link SwitchExpression} (GROOVY-12255 / JEP 361).
+     * Builds a {@link SwitchExpression} as specified by JEP 361.
      * Arrow arms that are a single expression become {@link YieldStatement}s;
-     * colon arms use explicit {@code yield}. The expression is compiled inline
-     * — it is not rewritten to a closure wrapping a {@link SwitchStatement}.
+     * colon arms use explicit {@code yield}.
      */
     @Override
     public SwitchExpression visitSwitchExpression(final SwitchExpressionContext ctx) {
@@ -1036,9 +1036,14 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
             if (statementsAndArrowAndYieldOrThrow.isEmpty()) {
                 throw createParsingFailedException("`case` or `default` branches are expected", ctx.LBRACE());
             } else {
-                var tuple = last(statementsAndArrowAndYieldOrThrow);
-                if (!tuple.getV2() && !tuple.getV3()) { // if no arrow, yield or throw is required
-                    throw createParsingFailedException("`yield` or `throw` is expected", tuple.getV1().get(0));
+                // Last group must complete: every path yields or throws (JEP 361).
+                // Intermediate colon groups may still fall through.
+                var lastGroup = last(statementsAndArrowAndYieldOrThrow);
+                List<Statement> lastStmts = lastGroup.getV1();
+                Statement lastArm = lastStmts.get(lastStmts.size() - 1);
+                Statement lastCode = lastArm instanceof CaseStatement cs ? cs.getCode() : lastArm;
+                if (mayCompleteNormally(lastCode)) {
+                    throw createParsingFailedException("`yield` or `throw` is expected", lastArm);
                 }
             }
 
@@ -1167,7 +1172,7 @@ public class AstBuilder extends GroovyParserBaseVisitor<Object> {
                         );
                     }
 
-                    if (isArrow && org.codehaus.groovy.ast.tools.GeneralUtils.maybeFallsThrough(codeBlock)) {
+                    if (isArrow && mayCompleteNormally(codeBlock)) {
                         throw createParsingFailedException("`yield` or `throw` is expected", exprOrBlockStatement);
                     }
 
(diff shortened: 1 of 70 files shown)


paulk-asert:
I am still reviewing but an initial AI assessment below:

> Gaps I'd close before merge:
> 
> - **Labeled break/continue escaping an arm**: `outer: while (...) { def r = switch (x) { case 1 -> { for (;;) { break outer } } } }`. The parser's peek-based check sees the loop frame and allows it, and the new `LabelVerifier` code fences closures for `yield` but doesn't fence *labels* at the expression boundary. Java rejects this ("attempt to break out of a switch expression"); here it likely compiles to a jump that abandons the expression mid-evaluation. Needs a check plus a fail test.
> - **Colon group that can complete normally despite containing a yield**: `case 'a': if (cond) yield 1` as the last group passes the parser's contains-yield check, then at runtime falls into the "does not cover" ISE even though the selector *did* match — Java makes this a compile error, and the runtime message is misleading. Verify and decide.
> - **Null selector under `@CS`** — no test for any of the three fast paths (see above).
> - **`@CS` fast paths + yield inside try/finally** — the finally-block stashing logic is only exercised dynamically.
> - **Switch expressions in field initializers / constructors / static initializers under `@CS`** — the writer leans on controller state; untested.
> - **GINQ**: the `LabelVerifier` javadoc explicitly calls out switch-expressions-inside-GINQ-queries as allowed, but no GINQ test was added.
> - **`SwitchExpression.transformExpression`** is a design smell: for a plain `ExpressionTransformer` it only rewrites top-level yield/throw/expression-statement expressions inside arms (if-conditions and loop conditions are missed); when the transformer is also a `GroovyCodeVisitor` it additionally re-visits the arm code, risking double transformation. It also mutates the original arm statements while claiming to produce a copy. Worth tightening before third-party transforms depend on it.

paulk-asert:
@daniellansun I added a commit that potentially addresses the issues raised above. We can delete if you have another preference.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -1020,27 +1021,13 @@ public Expression visitSwitchExprAlt(final SwitchExprAltContext ctx) {
...
-     *       case 1:  return 'a'
-     *       case 2:  return 'b'
-     *       default: return 'z'
-     *     }
-     * }.call()
-     * </pre>
+     * Builds a first-class {@link SwitchExpression} (GROOVY-12255 / JEP 361).
+     * Arrow arms that are a single expression become {@link YieldStatement}s;
+     * colon arms use explicit {@code yield}. The expression is compiled inline
+     * — it is not rewritten to a closure wrapping a {@link SwitchStatement}.
```
blackdrag: AI has the tendency to add comments like this, but I don't think they are good in this case. I suggest to remove "The expression is compiled inline — it is not rewritten to a closure wrapping a {@link SwitchStatement}." Referencing the JEP is good. Referencing GROOVY-12255 is also not something I would do here. It is the base for the change, yes. And the base for the comment part I just suggested to delete. But the information for the changes is in the git history. I mean it does also not reference GROOVY-9272 for example. And why should it? As a comment in the code, that this and that is special because of a certain issue is ok for me, just I find it misplaced in the javadoc in this case.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -1160,26 +1141,34 @@ public void visitThrowStatement(ThrowStatement statement) {
...
+                            }
                         }
+                    } else if (exprOrBlockStatement instanceof YieldStatement || exprOrBlockStatement instanceof ThrowStatement) {
+                        codeBlock = configureAST(
+                                createBlockStatement(exprOrBlockStatement),
+                                exprOrBlockStatement
+                        );
+                    }
+
+                    if (isArrow && org.codehaus.groovy.ast.tools.GeneralUtils.maybeFallsThrough(codeBlock)) {
```
blackdrag: why org.codehaus.groovy.ast.tools.GeneralUtils? Above I see plenty of static imports for parts of that class why not import the method here as well? If not, then why not import the class itself? Is there already a class of that name that conflicts here that I overlooked?

src/spec/test/SemanticsTest.groovy
```diff
@@ -184,11 +184,28 @@ final class SemanticsTest {
             case 'Adam'   -> 'Eve'
             case 'Antony' -> 'Cleopatra'
             case 'Bonnie' -> 'Clyde'
+            default       -> 'Unknown'
```
blackdrag: why did you add the "default"?

src/test/groovy/org/codehaus/groovy/classgen/asm/SwitchExpressionBytecodeTest.groovy
```diff
@@ -0,0 +1,143 @@
...
+    void noClosureAllocationForSwitchExpression() {
+        def bytecode = compile(method: 'run', '''\
+            def r = switch (1) {
+                case 1 -> 'a'
+                default -> 'z'
+            }
+        ''')
+        assert !bytecode.hasStrictSequence(['INVOKESPECIAL', 'org/codehaus/groovy/runtime/callsite'])
+        assert !bytecode.toString().contains('InnerClassNode')
+        assert !bytecode.toString().contains('$_run_closure')
```
blackdrag: I think this test only works if Closures stay the same in their bytecode usage.  Callsite may not be in use anymore because of Indy. Not sure where InnerClassNode is coming from and the $_run_closure method may also change. So this test is only reliable with you also test a Closure usage for those things to exist. And it should be in the same test to ensure the tests are not "fixed" independently.

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,670 @@
...
+        def err = shouldFail('''
+            def r = switch (1) {
+                case 1 -> {
+                    def c = { yield 1 }
+                    yield c()
+                }
+                default -> 0
+            }
+        ''')
+        assert err.message.contains('yield cannot jump through a closure or lambda')
```
blackdrag, line 780: I actually would like to see here two things additionally: (1) async usage where yield in the closure is then valid. (2) a switch statement in a switch expression that uses yield.

blackdrag (changes requested):
Additionally to what Paul said I added some review comments. Some of the comments annotate only one place, but actually mean many similar. The fully qualified class name usage is such a case.

daniellansun:
> Additionally to what Paul said I added some review comments. Some of the comments annotate only one place, but actually mean many similar. The fully qualified class name usage is such a case.

Thank you for the review. The inline notes are addressed below; the same FQCN cleanup was applied at the other sites introduced by this PR.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
daniellansun: Agreed. The method javadoc now only describes what the builder produces (JEP 361, arrow vs colon). The GROOVY-12255 key and the old closure-desugar narrative were removed from this javadoc (and from the `SwitchExpression` / `SwitchExpressionWriter` type docs that repeated the same history). Compatibility notes stay in `COMPATIBILITY.md`.

src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
daniellansun: Tweaked in this PR.

src/spec/test/SemanticsTest.groovy
daniellansun: No good reason for *this* snippet. The example always matches `'Romeo'`; exhaustiveness is already described in the surrounding prose and in the yield-block example. The added `default` has been reverted so the published example matches the 4.0 shape.

src/test/groovy/org/codehaus/groovy/classgen/asm/SwitchExpressionBytecodeTest.groovy
daniellansun: The test now compiles a real `{ -> … }` closure first and asserts that `$_run_closure` is present, then compiles the switch expression and asserts that the same marker is absent. The indy/callsite check is gone. `InnerClassNode` is only asserted if the closure probe actually emits it.

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
daniellansun: Added:

1. `asyncYieldReturnInsideNestedClosureIsAllowed` — `yield return` inside `async { }` nested in a switch arm (generator yield, not switch `yield`).
2. `switchStatementInsideSwitchExpressionCanYield` — a colon `switch` *statement* whose `yield` completes the enclosing switch *expression*.

daniellansun:
> I am still reviewing but an initial AI assessment below:
> 
> > Gaps I'd close before merge:
> > 
> > * **Labeled break/continue escaping an arm**: `outer: while (...) { def r = switch (x) { case 1 -> { for (;;) { break outer } } } }`. The parser's peek-based check sees the loop frame and allows it, and the new `LabelVerifier` code fences closures for `yield` but doesn't fence _labels_ at the expression boundary. Java rejects this ("attempt to break out of a switch expression"); here it likely compiles to a jump that abandons the expression mid-evaluation. Needs a check plus a fail test.
> > * **Colon group that can complete normally despite containing a yield**: `case 'a': if (cond) yield 1` as the last group passes the parser's contains-yield check, then at runtime falls into the "does not cover" ISE even though the selector _did_ match — Java makes this a compile error, and the runtime message is misleading. Verify and decide.
> > * **Null selector under `@CS`** — no test for any of the three fast paths (see above).
> > * **`@CS` fast paths + yield inside try/finally** — the finally-block stashing logic is only exercised dynamically.
> > * **Switch expressions in field initializers / constructors / static initializers under `@CS`** — the writer leans on controller state; untested.
> > * **GINQ**: the `LabelVerifier` javadoc explicitly calls out switch-expressions-inside-GINQ-queries as allowed, but no GINQ test was added.
> > * **`SwitchExpression.transformExpression`** is a design smell: for a plain `ExpressionTransformer` it only rewrites top-level yield/throw/expression-statement expressions inside arms (if-conditions and loop conditions are missed); when the transformer is also a `GroovyCodeVisitor` it additionally re-visits the arm code, risking double transformation. It also mutates the original arm statements while claiming to produce a copy. Worth tightening before third-party transforms depend on it.

Thank you — including for [`9a0c682`](https://github.com/apache/groovy/commit/9a0c682afdaaed2ae706f7ca52973cead801a8db), which already closed labeled break/continue, the `@CS` null-selector paths, and several bytecode cases. The remaining items are handled as follows.

| Gap | Response |
|---|---|
| Labeled `break` / `continue` escaping an arm | Kept your `LabelVerifier` isolation of loop/switch/label state. Added a `@CompileStatic` for-loop-in-arm case with an implicit-this call so the copied `ForStatement` keeps its `VariableScope` after `transformExpression`. |
| Colon group that contains a `yield` but can still complete | Last group is now rejected unless every path yields or throws (`GeneralUtils.mayCompleteNormally`, which uses the existing statement-flow analysis). Intermediate colon groups may still fall through. Covered by `lastColonArmIfWithoutElseIsError`, `fail/SwitchExpression_14x.groovy`, and `colonArmIfFallsThroughToCompletingDefault`. |
| Null selector under `@CS` | Your three fast-path tests remain. |
| `@CS` fast paths + `yield` in `try`/`finally` | Added int / `String` / enum variants (`compileStaticYieldInsideTryFinally*`). |
| Field / constructor / static initializer under `@CS` | Added `compileStaticSwitchExpressionInFieldInitializer`, `…InConstructor`, `…InStaticInitializer`. |
| GINQ | Arrow-form coverage already lived in `GinqTest` (`testGinq - switch - 1`…`6`). Added `testGinq - switch - yield block` so an explicit `yield` inside `GQ { }` is covered as well. |
| `transformExpression` | It now returns a structural copy: selector and case labels are transformed, arm statements are copied (not mutated), and the transformer is not also applied as a `GroovyCodeVisitor`. `AssertStatement` and `ForStatement.variableScope` are preserved. `transformExpressionCopiesArmsAndDoesNotMutateOriginal` checks that a constant inside an `if` / `yield` / `assert` is rewritten only on the copy. |

Happy to adjust further if any of these should take a different shape.

paulk-asert:
AI read:

> Verdict: technically ready, but blocked on process; what remains is re-review/sign-off, not code work.

paulk-asert (approved):

src/main/java/org/codehaus/groovy/ast/expr/SwitchExpression.java
```diff
@@ -0,0 +1,293 @@
...
+        copy.copyStatementLabels(caseStatement);
+        return copy;
+    }
+
+    /**
+     * Returns a structural copy of {@code statement} whose nested expressions
+     * have been passed through {@code transformer}. The original tree is not
+     * mutated. Unknown statement types are returned as-is.
+     */
+    private static Statement copyAndTransform(final Statement statement, final ExpressionTransformer transformer) {
```
blackdrag: actually I think this does not belong in here. I think this should be its own class and become something like the ClassCodeTransformer/ClassCodeVisitorSupport. I actually think the way you use this here violates the idea of ExpressionTransformer. transformExpression is not supposed to create new statements. Which actually shows here that using CaseStatement is a logical problem for the otherwise relatively strict difference between statements and expressions we have. in fact... the code makes copies of statements inside the cases, but it does not keep all the metadata, like the metadata map, source positions or for example variable scopes. We do not know when transformExpression is called.  I think for the normal switch-case statement we do not copy statements, especially not nested statements. Why should we here?  So my suggestion is if you need this transformExpression call, make a new copy-visitor, that also copies metadata and use it through a different method. If you only wanted to provide a transformExpression implementation, then I do not think it is right to transform nested statements. The part on if a CaseStatement should actually be a CaseExpression for a SwitchExpression I leave mainly to you, but I think I see strong idicators for it, as it would be strange to have an expression consisting of statements.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        for (CaseStatement caseStatement : caseStatements) {
+            Integer key = intConstant(caseStatement.getExpression());
+            if (key == null) return false;
+            if (!caseStatement.getCode().isEmpty() || caseStatement.isArrow()) {
+                currentGroupBody = new Label();
+            }
+            if (currentGroupBody == null) {
+                currentGroupBody = new Label();
+            }
+            if (keyToBody.put(key, currentGroupBody) != null) {
```
blackdrag: Is this really a case we want to handle here? This looks like a compilation error to me instead

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+                return false; // duplicate case value
+            }
+        }
+        if (keyToBody.isEmpty()) return false;
```
blackdrag: same here... why no compilation error?

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+
+        ClassNode resultType = resolveResultType(expression, effective);
+        Label endLabel = compileStack.pushSwitchExpression(resultType);
+
+        expression.getExpression().visit(acg);
+        ClassNode selectorType = operandStack.getTopOperand();
+        ClassNode storedSelectorType = ClassHelper.isPrimitiveType(selectorType)
+                ? ClassHelper.getWrapper(selectorType)
+                : selectorType;
+        operandStack.box();
```
blackdrag: (a) if you simply do operandStack.box(); ClassNode storedSelectorType = operandStack.getTopOperand(); you would not need the primitive type check here. You are not using selectorType past this point anyway.
(b) why are you actually boxing? For a tableswitch or a lookupswitch you have to unbox again.
(c) what if the variable is a reference? For example what if the same variable is used in a switch expression and also used in a Closure, where it is written even? like 
´´´
def x = 1
def cl = {x=2}
cl()
def y = switch(x)....
´´´
Then for x getTopOperand would return most likely int, But it actually is a Reference, potentially endangering all manipulation you do on that variable that does not use OperandStack. That wrong type extends to the temporary variable you define in the line after this.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        OperandStack operandStack = effective.getOperandStack();
+        AsmClassGenerator acg = effective.getAcg();
+
+        acg.onLineNumber(caseStatement, "visitCaseStatement");
+
+        mv.visitVarInsn(ALOAD, selectorIndex);
+        caseStatement.getExpression().visit(acg);
+        operandStack.box();
+        effective.getBinaryExpressionHelper().getIsCaseMethod().call(mv);
+        operandStack.replace(ClassHelper.boolean_TYPE);
```
blackdrag: why not emit a method call here and let others handle this properly? This looks like it enforces the dynamic isCase call even in static compilation.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        if (completeEnum) {
+            throwIncompatibleClassChangeError(effective.getMethodVisitor());
+        } else {
+            throwIllegalState(effective.getMethodVisitor(), selectorIndex);
+        }
+    }
+
+    private boolean writeIntSwitch(final SwitchExpression expression, final int selectorIndex,
+            final ClassNode storedSelectorType, final ClassNode resultType, final Label endLabel) {
+        if (!isStaticCompilation()) return false;
```
blackdrag: I think this indicates a problem. We have a split between static compilation and normal compilation. Any normal path that has to check if it is a static compilation path indicates a problem in the usage of the current architecture. Looking at this and the other specialized method for the non-isCase variants they all do that and should actually be part of the StaticTypesStatementWriter infrastructure.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        if (!isIntegralType(storedSelectorType) && !isIntegralWrapper(storedSelectorType)) return false;
```
blackdrag: don´t you unconditionally box for the storedSelectorType? I understand defensive programming is probably why you did that. I think I would instead split the check and do it before the write, and then box only if needed for the case. That way you need no null check for the primitive case and save on a potentially unused boxing. For this case here for example I think you need it only for the null-case which is only needed if the variable is no primitive in the first place.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+
+        Label defaultLabel = new Label();
+        // a null selector matches no constant label; it selects the default
+        // arm (or the unmatched-selector throw), as in 4.x/5.x and dynamic mode
+        mv.visitVarInsn(ALOAD, selectorIndex);
+        mv.visitJumpInsn(IFNULL, defaultLabel);
+
+        mv.visitVarInsn(ALOAD, selectorIndex);
+        operandStack.push(storedSelectorType);
+        operandStack.doGroovyCast(ClassHelper.int_TYPE);
```
blackdrag: I think this is a problem with references. Rule of thumb, use load/store only directly if you defined the value. As soon as it comes from an expression you cannot do that anymore for your assumptions may be wrong. Instead you should leave that to OperandStack, that is what it was made for.  For example I expect this case to fail, since you just force the normal type on the stack, while it might be a Reference. And then you basically do the cast on the Reference, instead of the value in the reference. Which is all handled by OperandStack.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        Set<Label> emitted = new HashSet<>();
+        for (CaseStatement caseStatement : caseStatements) {
+            Integer key = intConstant(caseStatement.getExpression());
+            Label body = keyToBody.get(key);
+            if (emitted.add(body)) {
+                mv.visitLabel(body);
+            }
+            if (!caseStatement.getCode().isEmpty()) {
+                caseStatement.getCode().visit(acg);
+            }
```
blackdrag: why do we need to check for EmptyStatement here? I think you should rename the label and not call it body. It is not for the body in the first place, it is a jump target. A label for body implies scopes for variables and such. Just a minor thing.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+                table[keys[i] - min] = labels[i];
+            }
+            mv.visitTableSwitchInsn(min, max, defaultLabel, table);
+        } else {
+            mv.visitLookupSwitchInsn(defaultLabel, keys, labels);
+        }
+    }
+
+    private boolean writeStringSwitch(final SwitchExpression expression, final int selectorIndex,
+            final ClassNode storedSelectorType, final ClassNode resultType, final Label endLabel) {
```
blackdrag: I think you should conform with Sonatype here and remove the unused variables.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+    private static void emitStringIndexDispatch(final MethodVisitor mv, final int stringLocal, final int caseIndexLocal,
+            final List<String> ordered, final Map<String, Label> stringToBody, final Label defaultLabel) {
+        Map<Integer, List<String>> byHash = new TreeMap<>();
+        for (String s : ordered) {
+            byHash.computeIfAbsent(s.hashCode(), h -> new ArrayList<>()).add(s);
+        }
+        Label secondSwitch = new Label();
+
+        mv.visitVarInsn(ALOAD, stringLocal);
+        mv.visitTypeInsn(CHECKCAST, "java/lang/String");
```
blackdrag: you already made sure it is a string, why force the checkcast here? Also, if you wold use OperandStack here you could write the cast and OperandStack would check for you if it is actually needed?

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+
+        Label defaultLabel = new Label();
+        // a null selector matches no constant label (as in 4.x/5.x and dynamic
+        // mode); with no default it must throw ISE, never the complete-enum ICCE
+        Label nullLabel = complete ? new Label() : defaultLabel;
+        mv.visitVarInsn(ALOAD, selectorIndex);
+        mv.visitJumpInsn(IFNULL, nullLabel);
+
+        mv.visitVarInsn(ALOAD, selectorIndex);
+        mv.visitTypeInsn(CHECKCAST, "java/lang/Enum");
```
blackdrag: Aren´t we going here only if selectorIndex stores an enum? Why is the cast needed?

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        mv.visitMethodInsn(INVOKEVIRTUAL, "java/lang/Enum", "name", "()Ljava/lang/String;", false);
+        effective.getOperandStack().push(ClassHelper.STRING_TYPE);
+        int nameLocal = compileStack.defineTemporaryVariable("$switchEnumName", ClassHelper.STRING_TYPE, true);
+        int caseIndexLocal = compileStack.defineTemporaryVariable("$switchCase", ClassHelper.int_TYPE, false);
+
+        List<String> ordered = new ArrayList<>(nameToBody.keySet());
```
blackdrag: Besides the cases that should probably be a compilation error you do not need a map. You could work with a List from the get-go.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+        Set<Label> emitted = new HashSet<>();
+        for (CaseStatement caseStatement : caseStatements) {
+            String name = enumConstantName(caseStatement.getExpression(), enumType);
+            Label body = nameToBody.get(name);
+            if (emitted.add(body)) {
+                mv.visitLabel(body);
+            }
+            if (!caseStatement.getCode().isEmpty()) {
+                caseStatement.getCode().visit(acg);
+            }
```
blackdrag: same question for EmptyStatement.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
```diff
@@ -0,0 +1,610 @@
...
+     *
+     * @param statement the yield statement to compile
+     */
+    public void writeYield(final YieldStatement statement) {
+        WriterController effective = effectiveController();
+        CompileStack compileStack = effective.getCompileStack();
+        CompileStack.SwitchExpressionContext context = compileStack.getSwitchExpressionContext();
+        if (context == null) {
+            throw new GroovyBugError("yield outside of a switch expression");
+        }
```
blackdrag: I think that check should be in CompileStack

src/test-resources/fail/SwitchExpression_12x.groovy
```diff
@@ -0,0 +1,28 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+
+def a = 1
+def r = switch (a) {
+    case 1 -> {
+        if (true) {
+            yield 1
```
blackdrag, line 25: It is actually difficult to see why 14 fails and also 12. Maybe add a comment to the program why it fails? For each of the switch expression files in this directory

src/test/groovy/org/codehaus/groovy/classgen/asm/SwitchExpressionBytecodeTest.groovy
```diff
@@ -0,0 +1,154 @@
...
+            int m(int n) {
+                switch (n) {
+                    case 1 -> 10
+                    case 2 -> 20
+                    case 3 -> 30
+                    default -> 0
+                }
+            }
+        ''')
+        assert bytecode.hasSequence(['TABLESWITCH']) || bytecode.hasSequence(['LOOKUPSWITCH'])
```
blackdrag: why "or"? It is either one or the other. Same for staticStringSwitchUsesLookupSwitch. Also this is a static compilation test so far, thus it should be under sc

blackdrag (changes requested):
A summary for the review would be the following:
* please use more of the provided infrastructure of OperandStack and CompileStack
* tests where the switch is using a Reference
* split between static compiler and non-static compiler and tests
* maybe we can reuse some code for the normal case statement as well?

daniellansun:
> A summary for the review would be the following:
> 
> * please use more of the provided infrastructure of OperandStack and CompileStack
> * tests where the switch is using a Reference
> * split between static compiler and non-static compiler and tests
> * maybe we can reuse some code for the normal case statement as well?

Agreed on all four.

- Selector load/store, yield stashing, string/enum casts, and the unmatched
  throw now go through `OperandStack` / `CompileStack` (`load`, `box`,
  `doGroovyCast`, `storeVar` via `defineTemporaryVariable`,
  `requireSwitchExpressionContext`). Raw `ALOAD` / `CHECKCAST` on those
  paths is gone.
- Reference-selector tests cover a closure-shared local that is written
  before the switch, both dynamically and under `@CompileStatic` (`int` and
  `Integer`), plus the same pattern on a switch *statement*.
- Sequential `isCase` lives on `SwitchExpressionWriter`. Tableswitch /
  lookupswitch and resolved `isCase` live on
  `StaticTypesSwitchExpressionWriter`, installed from
  `StaticTypesWriterController` only while
  `isInStaticallyCheckedMethod`. Dynamic bytecode tests stay in
  `SwitchExpressionBytecodeTest`; static ones moved to
  `asm/sc/SwitchExpressionStaticCompileTest`.
- `BinaryExpressionHelper.writeIsCase` is now shared with
  `StatementWriter.writeCaseStatement`.

src/main/java/org/codehaus/groovy/ast/expr/SwitchExpression.java
daniellansun: Agreed. `transformExpression` no longer copies arm statements. It rewrites
the selector and case-label expressions only and shares the arm bodies,
the same way `SwitchStatement` is rewritten by visit rather than by a
deep copy.

When the transformer is also a `GroovyCodeVisitor` (`ResolveVisitor`,
`ClassCodeExpressionTransformer`, …), those shared arms are then
*visited* so nested expressions still go through resolve and rewrite.
That is visit, not a new statement tree. A plain
`ExpressionTransformer` that is not a visitor leaves the arms untouched.

On `CaseExpression`: we kept `CaseStatement`. A `ClosureExpression`
already holds a statement body; a switch-expression arm is the same
shape (expression label, statement body). Introducing `CaseExpression`
would duplicate visitor surface without changing that. Happy to reopen
if you would rather split the node.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
daniellansun: Agreed. Empty colon arms are now grouped with the *following* completing
arm (walked backwards). A trailing empty group, or a duplicate constant
label, is reported as a compilation error rather than inventing a
standalone jump target or silently falling back to `isCase`.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
daniellansun: Agreed.

(a)(b) The dynamic writer boxes once, because it only does `isCase`. The
static writer keeps the type `OperandStack` reports after visiting the
selector: a primitive `int` is stored as `int` and fed to `tableswitch`
with no box/unbox; a wrapper is stored as a wrapper, null-checked, then
unboxed only on that path.

(c) The selector is evaluated once through the normal visit (which
unwraps a holder via `OperandStack`) and stored in a non-holder temp.
Later loads use `OperandStack.load` on that temp, never a raw `ALOAD` of
the original slot. Tests:
`referenceSelectorWrittenByClosure`,
`referenceSelectorWrittenByClosureUnderCompileStatic`,
`referenceWrapperSelectorWrittenByClosureUnderCompileStatic`,
`staticReferenceSelectorWrittenByClosure`,
`staticIntegerReferenceSelectorWrittenByClosure`.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
daniellansun: Agreed. The static writer emits `caseValue.isCase(selector)` as a
`MethodCallExpression` with a resolved target (instance `isCase`, or a
DGM overload more specific than `Object`). Only if that resolution is
ambiguous does it fall back to
`BinaryExpressionHelper.writeIsCase` (`ScriptBytecodeAdapter`). The
dynamic writer uses that shared helper from the start.

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
daniellansun: Agreed. Keys and jump targets are collected as parallel lists. A map is
kept only where the algorithm needs one (duplicate detection, hash
buckets for the Java 7 string dispatch, sorted int keys for
`tableswitch`).

src/main/java/org/codehaus/groovy/classgen/asm/SwitchExpressionWriter.java
daniellansun: Agreed. `CompileStack.requireSwitchExpressionContext()` now owns the
guard; `writeYield` just uses the returned context.

src/test-resources/fail/SwitchExpression_12x.groovy
daniellansun: Agreed. Each `fail/SwitchExpression_*x.groovy` now has a one-line
comment stating the reject reason.

src/test/groovy/org/codehaus/groovy/classgen/asm/SwitchExpressionBytecodeTest.groovy
daniellansun: Agreed. Dense `int` asserts `TABLESWITCH` only; sparse `int` asserts
`LOOKUPSWITCH` only; `String` asserts the Java 7 pair (`LOOKUPSWITCH` on
`hashCode`, then `TABLESWITCH` on the case index). Those tests live in
`src/test/groovy/org/codehaus/groovy/classgen/asm/sc/SwitchExpressionStaticCompileTest.groovy`.

daniellansun:
@blackdrag 

`yield` can be used with colon:
```
int result = switch (s) {
    case "Foo": 
        yield 1;
    case "Bar":
        yield 2;
    default:
        System.out.println("Neither Foo nor Bar, hmmm...");
        yield 0;
};
```

See also: https://openjdk.org/jeps/361

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,1073 @@
...
+                }
+                assert log == ['fin']
+                return r
+            }
+            assert m(1) == 10
+        '''
+    }
+
+    @Test
+    void compileStaticYieldInsideTryFinallyStringSwitch() {
```
blackdrag: why is this specific to static compilation? Also why does it matter if r is int or String?

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,1073 @@
...
+            def r = switch (1) {
+                case 1:
+                    if (true) yield 1
+            }
+        ''')
+        assert err.message.contains('yield') || err.message.contains('throw')
+    }
+
+    @Test
+    void compileStaticYieldInsideTryFinallyIntSwitch() {
```
blackdrag: why is this specific to static compilation?

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,1073 @@
...
+
+    @Test
+    void lastColonArmIfWithoutElseIsError() {
+        def err = shouldFail('''
+            def r = switch (1) {
+                case 1:
+                    if (true) yield 1
+            }
+        ''')
+        assert err.message.contains('yield') || err.message.contains('throw')
```
blackdrag: again the question as of why yield *or* throw?

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,1073 @@
...
+            assert m('Aa') == 'first'
+            assert m('BB') == 'second'
+            assert m('Cc') == 'none'
+        '''
+    }
+
+    @Test
+    void compileStaticSparseIntKeysStillDispatch() {
+        assertScript '''
+            @groovy.transform.CompileStatic
```
blackdrag: I am randomly using this line to anchor the comment, but it actually  is about several places with CompileStatic.  You already have tests that ensure the bytecode contains structures you expect from static compilation. Which means I assume the test here is to finalize that the behavior is consistent. But for this you need a base to compare to, which should be the same test without static compilation. So I suggest you do something like

```
def script = """..."""
assertScript script
assertScript "@groovy.transform.CompileStatic\n" + script
```

And that way you ensure baseline (dynamic Groovy) and static compiler align in behavior. You should look at each @CompileStatic using test in this class and ask yourself if it is really specific to the static compiler. If not you should change it like suggested, if it is special, then is should go into the static compilation test suite instead - or at least the test should have something explaining why this is only with static compilation.

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,1073 @@
...
+                def r = switch (1) {
+                    case 1 -> {
+                        for (;;) { continue outer }
+                        yield -1
+                    }
+                    default -> 0
+                }
+            }
+        ''')
+        assert err.message.contains("cannot continue to label 'outer'")
```
blackdrag, line 599: This is a compilation error, right? You should assert the exception type as well.

src/test/groovy/org/codehaus/groovy/classgen/Groovy12255.groovy
```diff
@@ -0,0 +1,1073 @@
...
+                def r = switch (1) {
+                    case 1 -> {
+                        for (;;) { break outer }
+                        yield -1
+                    }
+                    default -> 0
+                }
+            }
+        ''')
+        assert err.message.contains("cannot break to label 'outer'")
```
blackdrag, line 582: same case as for continue

src/main/java/org/codehaus/groovy/ast/expr/SwitchExpression.java
```diff
@@ -0,0 +1,193 @@
...
+     * pattern {@link ClosureExpression} uses so resolve, static-import and
+     * similar rewrites still see nested arm expressions. Any other transformer
+     * gets a copy of the selector and case labels; arm statements are shared.
+     */
+    @Override
+    public Expression transformExpression(final ExpressionTransformer transformer) {
+        if (transformer instanceof ClassCodeExpressionTransformer visitor) {
+            visitor.visitSwitchExpression(this);
+            return this;
+        }
```
blackdrag: I think this code should be in ClassCodeExpressionTransformer. The other part of the method looks good to me now

src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesSwitchExpressionWriter.java
```diff
@@ -0,0 +1,439 @@
...
+        call.putNodeMetaData(StaticTypesMarker.INFERRED_TYPE, ClassHelper.boolean_TYPE);
+        call.setSourcePosition(caseValue);
+        call.visit(controller.getAcg());
+        operandStack.doGroovyCast(ClassHelper.boolean_TYPE);
+    }
+
+    private MethodNode resolveIsCaseTarget(final Expression caseValue, final ClassNode selectorType) {
+        ClassNode caseType = controller.getTypeChooser().resolveType(caseValue, controller.getClassNode());
+        ClassNode switchArg = ClassHelper.isPrimitiveType(selectorType)
+                ? ClassHelper.getWrapper(selectorType) : selectorType;
```
blackdrag: minor: getWrapper is already doing this check for you.

src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesSwitchExpressionWriter.java
```diff
@@ -0,0 +1,439 @@
...
+                loader, caseType, "isCase", new ClassNode[]{switchArg});
+        if (methods.size() != 1) {
+            return null;
+        }
+        MethodNode dgm = methods.get(0);
+        // Object.equals-style isCase is not a substitute for runtime dispatch
+        if (isGenericObjectIsCase(dgm)) {
+            return null;
+        }
+        return dgm;
```
blackdrag: Do we not have infrastructure for this in the static compiler? If yes, then it should be used here, if no, well then it is beyond the scope of this issue, but then we should at one point create that.

src/main/java/org/codehaus/groovy/transform/stc/StaticTypeCheckingVisitor.java
```diff
@@ -4878,6 +4880,125 @@ public void visitSwitch(final SwitchStatement statement) {
...
 
+    /**
+     * Type-checks a switch expression: visits the selector and each arm, unifies
+     * the yielded types (JEP 361 poly expression), and reports a non-exhaustive
+     * switch when the selector type is statically known (GROOVY-12255).
+     *
+     * @since 6.0.0
+     */
+    @Override
+    public void visitSwitchExpression(final SwitchExpression expression) {
```
blackdrag, line 4894: I already mentioned it before but I think I was way to unspecific and then you misunderstood me. There are basically 2 cases for the switch expression: the generic isCase variant and the intrinsic variant. I think you need to check isCase here and if it is the isCase variant, you should actually go through method selection here to have a direct method call target be chosen for isCase. If there is no target this is a compilation error. This should then handle instance and DGM, other extensions, as well as making the typechecking extensions work for the isCase call. The isCase write is then actually  a direct method call write only, which is handled by writeDirectMethodCall in StaticInvocationWriter.  The intrinsic cases are to be handled by StaticTypesSwitchExpressionWriter directly, while for isCase should then go through the invocation writer mechanism.

src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesSwitchExpressionWriter.java
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    private static MethodNode chooseInstanceIsCase(final ClassNode caseType, final ClassNode switchArg) {
+        if (caseType == null) return null;
```
blackdrag: in combination with my comment on StaticTypeCheckingVisitor you would actually only have to check for the direct method call target here and then write it, if it exists.

blackdrag (changes requested):
I think we are almost there. The split looks promising, the type extension issue is in my opinion the only thing that really blocks this.

blackdrag (approved):

blackdrag:
@daniellansun I quite like the improvement the PR made. It was a lot of work, but I think it was worth it.

PR #2785, opened 2026-08-14
GROOVY-12258: category/bulk invalidation walks all loaded classes eve…

paulk-asert:
…n when no indy site was ever linked


paulk-asert:
held open as the minimal fallback for GROOVY-12258; will be closed unmerged if #2786 is approved

PR #2787, opened 2026-08-15
Build: Make JMH regression alerts direction-aware; retire github-acti…

paulk-asert:
…on-benchmark compare on PRs

The action hardcodes JMH as smaller-is-better, so throughput (ops/time) benchmarks alerted on speedups and never on slowdowns, and it compared against the single previous data point on noisy shared runners.

Per-benchmark alerts now live in the jmh-summary.py PR comment instead: direction comes from each benchmark's scoreUnit, ratios are measured against the trailing 90-day baseline, and runner-speed skew is divided out via the existing ruler calibration. Benchmarks >= 1.5x slower are listed explicitly; otherwise a quiet all-clear line is shown.

The daily/adhoc workflows still use github-action-benchmark to publish gh-pages data, so dashboards and graphs are unchanged.

### DIFF
--- a/.github/workflows/groovy-jmh-classic.yml
+++ b/.github/workflows/groovy-jmh-classic.yml
@@ -86,36 +86,6 @@ jobs:
           mv subprojects/performance/build/results/jmh/results.json \
              subprojects/performance/build/results/jmh/results-${{ matrix.suite }}.json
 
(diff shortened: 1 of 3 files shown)


daniellansun:
Merged. Thanks.

PR #2801, opened 2026-08-17
GROOVY-12232: Wrap AIC cast arguments only when the generated constru…

paulk-asert:
…ctor delegates to super dynamically

Since 5.0.0-alpha-4 (GROOVY-6285/GROOVY-9244), writeAICCall wraps every non-primitive cast argument in a PojoWrapper so the anonymous inner class's generated constructor can select among ambiguous super constructors by the declared cast type at runtime — the dynamic selectConstructorAndTransformArguments path unwraps it. But when the super constructor is arity-unique, the generated constructor binds it statically and consumes its Object parameter with a plain cast, which the wrapper reaches intact: new Test(['a'] as Set) {} failed with GroovyCastException 'Cannot cast PojoWrapper to Set' (reduced from Gradle's test suite).

writeAICCall now evaluates the same predicate the constructor's own compilation uses for its super call (arity-unique candidate, no spread arguments — mirroring makeDirectConstructorCall) against the generated constructor's super call, found by scanning its body (the this$0 assignment may precede it), and wraps only when delegation will be dynamic. Both consumers now see what they expect: the dynamic selector receives wrappers and keeps choosing by cast type; a statically bound super call receives the plain value.

### DIFF
--- a/src/main/java/org/codehaus/groovy/classgen/asm/InvocationWriter.java
+++ b/src/main/java/org/codehaus/groovy/classgen/asm/InvocationWriter.java
@@ -37,6 +37,9 @@ import org.codehaus.groovy.ast.expr.SpreadExpression;
 import org.codehaus.groovy.ast.expr.StaticMethodCallExpression;
 import org.codehaus.groovy.ast.expr.TupleExpression;
 import org.codehaus.groovy.ast.expr.VariableExpression;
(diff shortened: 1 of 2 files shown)


blackdrag (approved):

daniellansun:
Merged. Thanks.

PR #2817, opened 2026-08-18
GROOVY-10355: restore binary reading of (name) +/- x and (name) in/as…

paulk-asert:
… x (design sketch)

A parenthesized bare name whose final segment starts lowercase is by convention a value, not a class, so a cast mis-parse of the ambiguous shapes is rebuilt in AstBuilder as the binary expression the syntax visually suggests, preserving textual left-to-right grouping across precedence levels. The binary-only keywords in/as captured as cast operand identifiers are restored to their relational reading for any capitalization. Unresolvable bare-name cast types now carry a hint explaining the ambiguity and the ((name)) workaround.

The grammar is unchanged: a predicate-gated castExprAlt is not viable because adaptive prediction under the me.sunlan antlr4 fork only consults semantic predicates when a decision conflict is registered, which this decision never produces, so the predicate would only fire as a parse-time FailedPredicateException.

### DIFF
--- a/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
+++ b/src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
@@ -127,6 +127,7 @@ import org.codehaus.groovy.classgen.Verifier;
 import org.codehaus.groovy.control.CompilationFailedException;
 import org.codehaus.groovy.control.CompilePhase;
 import org.codehaus.groovy.control.ModuleImportHelper;
+import org.codehaus.groovy.control.ResolveVisitor;
 import org.codehaus.groovy.control.SourceUnit;
 import org.codehaus.groovy.control.messages.SyntaxErrorMessage;
 import org.codehaus.groovy.transform.AsyncTransformHelper;
@@ -154,6 +155,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


src/main/java/org/apache/groovy/parser/antlr4/AstBuilder.java
```diff
@@ -5034,6 +5256,22 @@ public List<DeclarationExpression> getDeclarationExpressions() {
     private static final String PACKAGE_INFO = "package-info";
     private static final String PACKAGE_INFO_FILE_NAME = PACKAGE_INFO + ".groovy";
 
+    /**
+     * Marks a cast whose type is a bare (possibly qualified) name and whose operand shape
+     * could equally be read as an operator continuation of a parenthesized value expression,
+     * so that later resolution failures can explain the ambiguity (GROOVY-10355).
+     */
+    public static final String AMBIGUOUS_BARE_NAME_CAST = "_AMBIGUOUS_BARE_NAME_CAST";
```
blackdrag: could we move that constant to a different place? I think ResolveVisitor should not reference AstBuilder or any antlr4 class if we can avoid that.

paulk-asert: Inverted the relationship.

daniellansun:
A couple of things I wasn’t sure about — I may well have missed the intended scope.

**`in` grouping.** `repairBinaryKeywordCast` uses the whole command argument as the RHS. With `x = 1`, `list = [1]`:

- `x in list && true` is `(x in list) && true` → `true`
- `def r = (x) in list && true` becomes `x in (list && true)` → `false`
- `if ((x) in list && true)` takes the same path

On master the parenthesized command forms don’t compile. After the rewrite they compile, but with the inverted meaning. `return (x) in list && true` is already grouped the other way — `return` parses a normal expression, not a command.

Would it be reasonable to peel `&&` / `==` / a second `in` off the RHS, the way `combineRebalancing` does for `+/-`? If the sketch is only meant to cover a simple name on the right, it might help to say so — the commit currently reads as the full relational meaning.

**`as`.** `(x) as Long` works; `(x) as List<String>`, `(x) as String[]`, and `(x) as Long ?: 0` still fail to resolve `x`. For `List<String>` the argument is already a `ClassExpression` — `right.getType()` might be enough. The other two look harder (expression vs `coercionType`).

**Source positions.** Inner nodes from `combineRebalancing` show up as `line=-1` (e.g. the inner `+` in `"A" + (b) + "C"`). A `configureAST` on each new `BinaryExpression` would probably sort that out.

**`CAST_RESOLVE_HINT`.** It’s a new public constant; the parser only needs the string. `@Internal` (or a shared internal key) might keep it off the public surface.

**Comments / tests.** The text says “lowercase-initial”; the check is `!isUpperCase` (`_foo` is treated as a value). And the `in` grouping above isn’t in the suite yet — something like `assert ((x) in list && true)` would have caught it. `((x) as Long)` already works on master because of the extra parens; `(x) as List<String>` as a statement would cover the `as` gap. `(p)++` isn’t a cast, so it passes without the rewrite.

paulk-asert:
@daniellansun Response to your questions in the Jira ticket.

daniellansun (approved):

src/main/java/org/codehaus/groovy/control/ResolveVisitor.java
```diff
@@ -110,6 +110,15 @@ public class ResolveVisitor extends ClassCodeExpressionTransformer {
...
     public static final String QUESTION_MARK = "?";
+    /**
+     * Node-metadata key under which a parser may store an explanatory suffix for a
+     * {@link CastExpression} whose type it suspects may fail to resolve — for example a
+     * cast produced by a grammar ambiguity. When the cast's type cannot be resolved, the
+     * stored text is appended to the {@code unable to resolve class} error message.
+     * The value is the complete hint text; this visitor attaches no meaning to it.
+     */
+    @Internal
+    public static final String CAST_RESOLVE_HINT = "_CAST_RESOLVE_HINT";
```
blackdrag, line 121: It looks better that way, but I am still not happy about this being such a specialized mechanism for just one thing, but spanning so much across packages and compilation phases. If we made this a more general mechanism and for example call it "_RESOLVE_HINT", then it would look better, but is it better? If it stays a single use case, then probably not. So I guess in combination with @Internal this is ok. We can still change it later if we really want to.

blackdrag (approved):

PR #2818, opened 2026-08-19
GROOVY-12280: hold cached awaitable adapters through SoftReferences

paulk-asert:
A ClassValue association lives as long as its key class, and the common keys here are platform classes such as CompletableFuture, so the cached Groovy-loaded adapter pinned Groovy's class loader for the lifetime of the JVM (JDK-8136353 behavior, working as intended). The value is now held through a SoftReference: the association strongly reaches only java.base objects. A cleared reference is removed and recomputed once; should the fresh reference already be cleared, the answer comes from an uncached scan, so lookups terminate under any memory pressure. A NO_ADAPTER sentinel keeps unsupported types distinguishable from cleared references.

The site deliberately stays on java.lang.ClassValue rather than routing through GroovyClassValueFactory: with soft values nothing strongly Groovy-loaded remains in the association, so there is nothing left for the groovy.use.classvalue escape hatch to release here.

Soft references are cleared by the collector under heap pressure, so the class loader becomes collectible rather than instantly collected — the designed semantics of the strategy agreed in the GROOVY-12142 review discussion, not a gap.

The change was also verified against the groovy-concurrent-java deliverable, whose jar is a filtered extraction of core promising Java users no Groovy-runtime dependency: jdeps over the registry classes with only that jar on the classpath resolves every reference from java.base or the jar itself (SoftReference and ClassValue are java.base). That check is a second, independent reason the factory-routing alternative was declined — GroovyClassValueFactory lives in org.codehaus.groovy.reflection, outside the extraction set, so routing through it would have shipped the extracted jar with an unresolvable reference.

### DIFF
--- a/src/main/java/groovy/concurrent/AwaitableAdapterRegistry.java
+++ b/src/main/java/groovy/concurrent/AwaitableAdapterRegistry.java
@@ -22,6 +22,7 @@ import org.apache.groovy.runtime.async.AsyncSupport;
 import org.apache.groovy.runtime.async.FlowPublisherAdapter;
 import org.apache.groovy.runtime.async.GroovyPromise;
 
+import java.lang.ref.SoftReference;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


src/main/java/groovy/concurrent/AwaitableAdapterRegistry.java
```diff
@@ -132,25 +189,11 @@ public static <T> Iterable<T> toIterable(Object source) {
                         + ". Register an AwaitableAdapter via ServiceLoader or AwaitableAdapterRegistry.register().");
     }
 
-    private static ClassValue<AwaitableAdapter> buildAwaitableCache() {
-        /**
-         * Cache of awaitable adapters by source type.
-         */
+    private static ClassValue<SoftReference<AwaitableAdapter>> buildAwaitableCache() {
         return new ClassValue<>() {
```
blackdrag, line 196: very important here is that the method is static, otherwise there would be a reference to the enclosing class instance, which would pin down the classes and prevent class loader unloading. I would add this as a comment to ensure a future change does not accidentally introduce this trap. But not blocker.

PR #2820, opened 2026-08-20
GROOVY-12281: opt-in soft `GroovyClassValue` mode — unpins class loaders while keeping the ClassValue fast path

paulk-asert:
### What this PR is

The master-bound payload of the GROOVY-12281 investigation, now exactly two commits:

- **soft `GroovyClassValue` mode — soft values with resurrection (opt-in)**: the reviewed
  change, rebuilt without the declined hybrid prototype (the factory has only
  `true`/`soft`/`false` arms);
- **review response — ephemeron pinning + store capability**: the delta responding to the
  review below, kept as a separate commit so it can be re-reviewed against the earlier
  state. **Intended to be squashed into the first commit at landing** — the earlier
  global-root-set design should not exist in master history (it has the reverse-scenario
  leak described below).

The investigation artifacts (hybrid prototype, escape-hatch cost quantification for the
guide, assessment document, loader/perf spike harnesses) now live on the companion branch
[`groovy12281-investigation`](https://github.com/paulk-asert/groovy/tree/groovy12281-investigation):
[GROOVY-12281-assessment.html](https://github.com/paulk-asert/groovy/blob/groovy12281-investigation/GROOVY-12281-assessment.html)
(v3; §12 records the review round and response).

### Problem

A `java.lang.ClassValue` association lives as long as its key class, so
`ClassInfo.globalClassValue` entries on immortal platform classes (String, Integer, …) hold a
strong chain to the ClassInfo and through it to Groovy's class loader — leaking every Groovy
copy a container deploys and undeploys (JDK-8136353 / GROOVY-12142). The existing remedy,
`-Dgroovy.use.classvalue=false`, gives up the per-Class fast path for every key
(~4× on the raw lookup, +4.8% geomean on classic dispatch, hot platform idioms to 1.38×).

### Change

`-Dgroovy.use.classvalue=soft` keeps `ClassValue` for all keys but stores each value in a
per-Class slot (a bootstrap `AtomicReference`) holding a bootstrap `SoftReference`, cutting
the only strong chain from immortal keys to the Groovy island. Three cooperating pieces make
that safe (details and file references in the assessment):

- **Resurrection** — a weak-key/weak-value side map is the identity authority: a value still
  reachable anywhere is re-associated, never replaced, so a fresh instance can only exist once
  no guard can still observe the old one. Classic call-site version guards therefore stay
  sound: legacy-compiled jars using `groovy-callsite` from the classpath are unaffected.
- **Ephemeron pinning** — ClassInfos carrying non-reconstructible state (installed MetaClass,
  per-instance MetaClasses, registry-written DGM/extension arrays) are pinned *inside their
  own association*: the slot flips from soft to strong, so a pinned instance lives exactly as
  long as its class — an immortal platform key retains it, a dropped script class releases it
  together with its loader (the "reverse" scenario from review). The store advertises this
  via a `GroovyClassValue` capability (`valuesReclaimable()` + default-no-op `pin`/`unpin`),
  so `ClassInfo` never consults the mode; `ClassInfo.remove()` keeps its hard-detach undeploy
  semantics with no store-specific handling — the pin travels with the association.
- **Per-Class indy domain continuity** — SwitchPoint domains are keyed by `Class` in soft
  mode, so a successor ClassInfo adopts its predecessor's domain and mutations
  deterministically retire guards that captured only the SwitchPoint (POJO direct dispatch).
  Indy-only; the default mode's behaviour is unchanged. The `IndyInvalidation` /
  `SwitchPointInvalidator` layering (policy vs mechanism, with `ClassInfo` the only other
  supported consumer) is now documented on both classes, which are `@Internal`.

Default-mode behaviour is byte-identical apart from no-op hook calls; all soft-mode structures
are unallocated unless the flag is set.

### Evidence

- **Acceptance** (the ticket's point; JDK 17/21/23): a dropped child-loader Groovy copy is
  pinned forever under the default and **collected under soft mode** once pressure clears
  soft references.
- **Reverse acceptance** (from review): the runtime stays alive while script loaders come and
  go, each installing an EMC on a class it created — under soft mode the dropped loaders are
  collected on memory pressure exactly as under the default (20/20; the earlier root-set
  design leaked 0/20). Now a probe scenario.
- **Correctness**: 9 unit tests + a deterministic child-JVM probe with 8 scenarios
  (resurrection identity/version, pinned-slot assertions for DGM/EMC/per-instance state,
  EMC removal unpins, collection + recreation dispatch, classic `CallSiteArray` soundness,
  predecessor-SwitchPoint retirement, the reverse scenario); a real-GC concurrency
  stress probe (128 MB heap, `SoftRefLRUPolicyMSPerMB=0`): 66M dispatches, ~130 EMC
  generations, 33/36 platform-receiver ClassInfos collected and recreated mid-run, zero
  invariant violations; full core suite passes under soft mode (16,800 tests).
- **Cost**: raw `getClassInfo` 1.2 → 1.7 ns; compiler harness +0.0%; classic-bytecode JMH gate
  0 of 37 significant (geomean +1.7%; the declined map default failed the same rule with 12,
  worst 1.38×); classic *polymorphic* miss traffic ≈ +3.7% pooled, flagged for the idiom-suite
  sweep before any default-flip discussion.

### Open questions for review

Assessment §9 consolidates the resulting option landscape (`true`/`soft`/`false` ×
static/dynamic × indy/classic, with guidance); §10 lists the open items: adoption path
(opt-in now — recommended — vs eventual default), whether domain re-homing should become
unconditional rather than soft-gated, and container soak experience. §12 records the PR
review round. One separate finding will get its own JIRA: `=false` on current master fails
the reverse scenario (0/20 collected vs 20/20 on released 5.0.6/6.0.0-beta-2) — the
escape-hatch rework's weak-key/strong-value map is not an ephemeron.

JIRA: https://issues.apache.org/jira/browse/GROOVY-12281

### DIFF
--- a/src/main/java/org/apache/groovy/runtime/indy/IndyInvalidation.java
+++ b/src/main/java/org/apache/groovy/runtime/indy/IndyInvalidation.java
@@ -24,6 +24,7 @@ import groovy.lang.MetaClass;
 import groovy.lang.MetaClassImpl;
 import groovy.lang.MetaClassRegistryChangeEvent;
 import org.apache.groovy.util.SystemUtil;
+import org.apache.groovy.util.concurrent.ManagedIdentityConcurrentMap;
 import org.codehaus.groovy.reflection.ClassInfo;
 import org.codehaus.groovy.runtime.NullObject;
 import org.codehaus.groovy.util.ManagedReference;
@@ -124,6 +125,76 @@ public final class IndyInvalidation {
         private final SwitchPointInvalidator domain;
 
         DomainReclaim(final ClassInfo owner, final SwitchPointInvalidator domain) {
+            super(ReferenceBundle.getWeakBundle(), owner);
+            this.domain = domain;
+        }
+
+        @Override
+        public void finalizeReference() {
+            domain.invalidate();
+            super.finalizeReference();
+        }
+    }
+
+    // -------------------------------------------------------------------------
+    // Soft ClassValue mode: per-Class domain continuity (GROOVY-12281)
+    // -------------------------------------------------------------------------
+
+    /**
+     * Per-Class domain map used only when {@code groovy.use.classvalue=soft}.
+     * In soft mode a ClassInfo instance is replaceable — it can be soft-collected
+     * and a successor created for the same class — while POJO direct-dispatch
+     * guards capture only the SwitchPoint's internal invoker, never the
+     * ClassInfo. Keying the domain by the {@code Class} (weak identity keys,
+     * strong domain values) makes domain identity survive instance turnover:
+     * the successor adopts its predecessor's domain, so an invalidation applied
+     * through the successor deterministically retires guards linked under the
+     * predecessor, with no dependency on lazy reference-queue processing.
+     * <p>
+     * The map holds no classes alive (weak keys) and no foreign loaders
+     * (values reference nothing owner-ward); entries for dead classes are
+     * purged by the map, and their domains retired by the Class-keyed reclaim
+     * ({@link #anchorClassDomainToClass}) so the live-SwitchPoint registry
+     * cannot accumulate zombies.
+     */
+    private static final ManagedIdentityConcurrentMap<Class<?>, SwitchPointInvalidator> CLASS_DOMAINS =
+            new ManagedIdentityConcurrentMap<>();
+
+    /**
+     * Returns the canonical domain for {@code type}, seeding the per-Class map
+     * with {@code candidate} on first touch. Soft-mode only (callers gate).
+     *
+     * @param type      the class whose domain is requested (must not be {@code null})
+     * @param candidate the caller-owned domain to install if none is mapped yet
+     * @return the canonical per-Class domain
+     */
+    public static SwitchPointInvalidator classDomainFor(final Class<?> type, final SwitchPointInvalidator candidate) {
+        return CLASS_DOMAINS.getOrPut(type, candidate);
+    }
+
+    /**
+     * Anchors a reclaim reference keyed to the {@code Class} rather than a
+     * ClassInfo instance: soft mode retires a domain only when the class dies
+     * (no receiver can reach an installed guard afterwards), because ClassInfo
+     * instances are replaceable and their successors adopt the same domain.
+     *
+     * @param owner  the class owning the domain (must not be {@code null})
+     * @param domain the class-level domain (must not be {@code null})
+     */
+    public static void anchorClassDomainToClass(final Class<?> owner, final SwitchPointInvalidator domain) {
+        domain.setReclaimAnchor(new ClassDomainReclaim(owner, domain));
+    }
+
+    /**
+     * Weak reference to a domain's owning {@code Class} whose collection
+     * retires the domain (soft ClassValue mode). Mirrors {@link DomainReclaim}
+     * with the owner lifetime moved from the replaceable ClassInfo instance to
+     * the stable per-Class identity.
+     */
+    private static final class ClassDomainReclaim extends ManagedReference<Class<?>> {
+        private final SwitchPointInvalidator domain;
+
+        ClassDomainReclaim(final Class<?> owner, final SwitchPointInvalidator domain) {
             super(ReferenceBundle.getWeakBundle(), owner);
             this.domain = domain;
         }
--- a/src/main/java/org/apache/groovy/runtime/indy/IndyInvalidation.java
+++ b/src/main/java/org/apache/groovy/runtime/indy/IndyInvalidation.java
@@ -23,6 +23,7 @@ import groovy.lang.DelegatingMetaClass;
 import groovy.lang.MetaClass;
 import groovy.lang.MetaClassImpl;
 import groovy.lang.MetaClassRegistryChangeEvent;
+import groovy.transform.Internal;
 import org.apache.groovy.util.SystemUtil;
 import org.apache.groovy.util.concurrent.ManagedIdentityConcurrentMap;
 import org.codehaus.groovy.reflection.ClassInfo;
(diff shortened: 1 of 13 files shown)


src/main/java/org/apache/groovy/runtime/indy/IndyInvalidation.java
blackdrag, line 77: The architectural relationship between `IndyInvalidation` and `SwitchPointInvalidator` makes me wonder. The issue is not only that `IndyInvalidation` is tightly coupled to `SwitchPointInvalidator`, but also which of the two is supposed to be the abstraction boundary?

* If `IndyInvalidation` is a wrapper/facade around `SwitchPointInvalidator`, then `SwitchPointInvalidator` is effectively an implementation detail of the higher-level invalidation mechanism. In that case, it would be surprising for `SwitchPointInvalidator` to remain independently exposed and used. Consumers should normally go through `IndyInvalidation`, otherwise the wrapper does not really define the abstraction boundary.

* If `SwitchPointInvalidator` is intended to be an independently usable abstraction, then the two classes should be considered independent components which happen to cooperate. In that case, their respective responsibilities, lifecycle guarantees and invariants need to be clearly defined. In particular, it should be clear which guarantees `SwitchPointInvalidator` provides on its own and which only exist when it is managed through `IndyInvalidation`.

At the moment, the implementation seems somewhat in between these two models. `IndyInvalidation` directly manages `SwitchPointInvalidator` instances and depends on its lifecycle and global state, which makes the relationship look more like a layered abstraction than two independent components. At the same time, `SwitchPointInvalidator` is still exposed and can be used independently.

This should be document, otherwise it is difficult to determine if the current coupling is intentional or simply an implementation dependency that has become externally visible.

If `SwitchPointInvalidator` can be used independently as well as through `IndyInvalidation`, then the correctness of the overall invalidation machinery cannot simply rely on invariants established by `IndyInvalidation`. The independent use cases need to obey the same assumptions, or the boundary between the two needs to enforce those assumptions.

If `SwitchPointInvalidator` is really only the low-level mechanism behind `IndyInvalidation`, hiding it behind that abstraction would make the ownership and lifecycle model much easier to reason about and would also leave room to change the underlying invalidation mechanism later.

I think the important point is not simply whether the classes are coupled, but whether the current API and usage actually match the intended abstraction boundary. If they are one layered subsystem, that relationship should be explicit; if they are independent abstractions, their independence and invariants need to be justified.

src/main/java/org/codehaus/groovy/reflection/ClassInfo.java
```diff
@@ -110,6 +112,29 @@ public class ClassInfo implements Finalizable {
...
             new ManagedConcurrentLinkedQueue<ClassInfo>(weakBundle);
 
+    /**
+     * Whether {@link #globalClassValue} stores its values softly with
+     * resurrection ({@code -Dgroovy.use.classvalue=soft}, GROOVY-12281
+     * investigation prototype). Soft mode needs two cooperating pieces here:
+     * strong roots for non-reconstructible state ({@link #nonReclaimableRoots})
+     * and per-Class indy domain continuity (see {@link #indyDomain()}).
+     */
+    private static final boolean SOFT_CLASS_VALUES = GroovyClassValueFactory.isSoftMode();
```
blackdrag: This isSoftMode() is a bit suspicious. Again it produces a tight coupling to a certain implementation. Why is a ClassValue with SoftReferences a soft mode, but a ManagedMap based version not?

src/main/java/org/codehaus/groovy/reflection/ClassInfo.java
```diff
@@ -110,6 +112,29 @@ public class ClassInfo implements Finalizable {
...
+     * an installed class-level MetaClass ({@link #setStrongMetaClass}),
+     * per-instance MetaClasses, or registry-written DGM/extension method
+     * arrays ({@link CachedClass#setNewMopMethods}/{@code addNewMopMethods}).
+     * The set itself is Groovy-loaded, so it contributes no foreign-loader
+     * pinning: everything in it dies with Groovy's own loader, which is the
+     * lifetime every ClassInfo has today under the default strong ClassValue.
+     * {@code null} unless soft mode is active.
+     */
+    private static final Set<ClassInfo> nonReclaimableRoots =
+            SOFT_CLASS_VALUES ? ConcurrentHashMap.newKeySet() : null;
```
blackdrag: What happens if we always do that and not only in soft mode? This is also not the only place. Either we have to do it, then this is missing an abstraction in my eyes. Or we do not, well, then we should simplify the code

src/main/java/org/codehaus/groovy/reflection/GroovyClassValueFactory.java
```diff
@@ -33,13 +33,30 @@ class GroovyClassValueFactory {
...
+	/**
+	 * GROOVY-12281 investigation prototype: whether values are stored behind
+	 * {@link java.lang.ref.SoftReference}s with resurrection semantics
+	 * ({@code -Dgroovy.use.classvalue=soft}). Soft mode needs cooperation from
+	 * {@link ClassInfo} (strong roots for non-reconstructible state, per-Class
+	 * indy domain continuity), which is why it is exposed package-wide rather
+	 * than kept local to {@link #createGroovyClassValue}.
+	 */
+	static boolean isSoftMode() {
+		return "soft".equalsIgnoreCase(CLASSVALUE_MODE);
```
blackdrag: why not store soft mode as boolean?

src/main/java/org/codehaus/groovy/reflection/GroovyClassValueFactory.java
```diff
@@ -33,13 +33,30 @@ class GroovyClassValueFactory {
...
+
 	public static <T> GroovyClassValue<T> createGroovyClassValue(ComputeValue<T> computeValue) {
-		// GROOVY-12281 investigation prototype: "hybrid" routes platform-loader keys to the
+		// GROOVY-12281 investigation prototypes: "hybrid" routes platform-loader keys to the
 		// weak-key map and everything else to ClassValue, so immortal platform keys never
-		// pin the value's loader while user classes keep the per-class fast path.
+		// pin the value's loader while user classes keep the per-class fast path (measured,
+		// declined); "soft" keeps ClassValue for all keys but holds values softly with
+		// resurrection, so immortal keys hold no strong chain to the value's loader.
 		if ("hybrid".equalsIgnoreCase(CLASSVALUE_MODE)) {
```
blackdrag: not part of the PR really, but if you do a boolean for soft mode, you could also do one for hybrid then. Or you switch to a model where you do not use the boolean, but a provider GroovyClassValueHybrid::new and so on. Just a thought

src/main/java/org/codehaus/groovy/reflection/GroovyClassValueSoft.java
```diff
@@ -0,0 +1,149 @@
...
+     * exists to provide. (The map's own {@code putIfAbsent} cannot be used
+     * for this: on an entry whose collected value has not yet been purged it
+     * reports success without storing the replacement.)
+     */
+    private final Object[] creationLocks = new Object[64];
+
+    private final ClassValue<SoftReference<T>> store = new ClassValue<SoftReference<T>>() {
+        @Override
+        protected SoftReference<T> computeValue(final Class<?> type) {
+            return new SoftReference<>(canonical(type));
```
blackdrag: I actually no wonder... does the ClassInfo case really need a SoftReference or could it be a WeakReference? The SoftReference is maybe the more "general", "cache friendly" variant, but I have now second thoughts of if it should be defined at this level, also considering the factory. But it could stay like this for now

blackdrag (commented):
I think besides the architecture question I have one more... What happens if you do the reverse memory pressure test? I mean spinning up many GroovyShells and execute some scripts in there, that for example set EMC on classes they create. The expectation would be that on memory pressure the old script instances and their meta classes can be collected. I called it reverse, because in this case the Groovy runtime stays alive.

paulk-asert:
@blackdrag Response to your questions in the issue.

PR #2821, opened 2026-08-21
Groovy12281 investigation

paulk-asert:
not for commit to master - just in case anyone wants to repeat the investigation


paulk-asert:
Not intended for merging.

PR #2834, opened 2026-08-26
GROOVY-12303: ClassNodeResolver: NoClassDefFoundError during class-loader lookup aborts resolution

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12303

### DIFF
--- a/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
+++ b/src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
@@ -49,17 +49,29 @@ import java.util.Map;
  * source to the compilation queue; a {@link ClassNode} result completes resolving
  * for that name. The outcome is wrapped in {@link LookupResult}.
  * <p>
- * The default lookup prefers ASM decompilation of a {@code .class} resource so a
- * type can be described without linking it in the JVM. Class-loader lookup is the
- * fallback when decompilation is disabled or the class exists only in memory.
- * {@link NoClassDefFoundError} during that fallback means the class was found but
- * could not be linked. Matching bytecode is decompiled once (ASM does not link)
- * so a {@link ClassNode} can still be produced. A groovy source replaces that
- * node only when the {@code .class} came from another class loader and is older
- * than the source. A {@code .class} resource whose bytecode name does not match
- * is not this class; script lookup then follows the same rule as
- * {@link ClassNotFoundException}. An unrecoverable error is rethrown with the
- * looked-up name in the message and is not cached as a miss.
+ * Lookup is two independent strategies selected by the compilation unit's
+ * optimization options {@code asmResolving} and {@code classLoaderResolving}
+ * (both default on). They are not each other's fallback:
+ * <blockquote>
+ * <table>
+ *   <caption>ClassNodeResolver lookup modes</caption>
+ *   <tr><th>Mode</th><th>asm</th><th>class loader</th><th>Lookup</th></tr>
+ *   <tr><td>default</td><td>on</td><td>on</td><td>ASM first; {@code loadClass} only if ASM has no match</td></tr>
+ *   <tr><td>ASM-only</td><td>on</td><td>off</td><td>ASM only; class-format errors thrown</td></tr>
+ *   <tr><td>loader-only</td><td>off</td><td>on</td><td>{@code loadClass} only; no decompile</td></tr>
+ *   <tr><td>neither</td><td>off</td><td>off</td><td>groovy source only</td></tr>
+ * </table>
+ * </blockquote>
+ * ASM describes a type from bytecode without linking it, so a missing superclass
+ * does not prevent a {@link ClassNode}. {@code loadClass} is for types that exist
+ * only in memory (or when ASM is off). A {@link ClassHelper} hit for an already
+ * resolved name is not {@code loadClass}.
+ * <p>
+ * {@link NoClassDefFoundError} means the class was found but could not be linked.
+ * It is wrapped with the looked-up name and rethrown, and is not cached as a miss.
+ * Script fallback on that error is only for an ASM bytecode-name mismatch (the
+ * requested name never existed). A groovy source replaces a found class only when
+ * that class came from another loader and the source is newer.
  * <p>
  * Lookups are cached. Override {@link #cacheClass(String, ClassNode)} and
  * {@link #getFromClassCache(String)} to disable or replace the cache. Custom
@@ -265,11 +277,10 @@ public class ClassNodeResolver {
      * (or if no class was found).
      * <p>
      * {@link NoClassDefFoundError} from class loading is not treated as a miss.
-     * The class under lookup is present; a referenced type is not. Matching
-     * bytecode is decompiled (if that has not already been tried) so a
-     * {@link ClassNode} can be produced without linking. A groovy source
-     * replaces an existing class only when it came from another class loader
-     * and is newer. An unrecoverable error is rethrown, not cached.
+     * It is wrapped and rethrown. Decompilation is not used as a fallback from
+     * that error; matching bytecode is the ASM strategy, which runs first when
+     * it is enabled. A groovy source replaces an existing class only when it
+     * came from another class loader and is newer.
      *
      * @param name the fully qualified class name
      * @param compilationUnit the current compilation unit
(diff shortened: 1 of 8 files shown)


src/main/java/org/codehaus/groovy/control/ClassNodeResolver.java
```diff
@@ -304,28 +324,91 @@ private static LookupResult findByClassLoading(final String name, final Compilat
...
+     *       fall through to script lookup. Detection uses the bytecode name, not
+     *       {@link NoClassDefFoundError} text.</li>
+     *   <li>A groovy source of the same name, added to the compilation queue.</li>
+     * </ol>
+     * If none of those succeed the error is rethrown with {@code name} in the
+     * message so callers see both the class under lookup and the missing type.
+     */
+    private LookupResult recoverFromNoClassDefFoundError(final String name, final CompilationUnit compilationUnit,
+            final GroovyClassLoader loader, final NoClassDefFoundError ncdfe) {
+        LookupResult decompiled = findDecompiled(name, compilationUnit, loader, false);
```
blackdrag: I do not understand this step. The name we use here is the one we got the class loading error for before, but to get here, we already tried decompilation I thought - Unless decompilation is not enabled, in which case I think calling this method should not happen (or I overlooked the guard in the method). So I assume decompiled is always giving null here. The other decompilation in classFileDeclaresDifferentName as imho one that should be tried initially... also it ignores the flag to not to decompile - without decompilation the method cannot do the check it tries to do right now. 

And then the change basically boils down to tryAsScript as it is done for a ClassNotFoundException. Which is not always legal. Here exactly we have to consider if the class is in the same class loader or not. We can define the ClassNode as script only if the failing class is from a parent. tryAsScript is not doing that check... the current check is in tryAsScript -> oldClass!=null -> isSourceNewer, which does not work if we do not have a class. Which means we cannot reliably know much about the class unless we look at the class file resource. If there is no such resource - which does not have to be the case - then this fails too.

daniellansun: Thank you — that reading of the recovery path was right, and the patch has been restructured around it.

### 1. The extra `findDecompiled` in recovery

When `asmResolving` is on we have already parsed the resource. Calling `findDecompiled` again on the same name was redundant; a hit would have returned before `loadClass`.

The guard you wondered about is in the caller (`tryAsLoaderClassOrScript`), not in `findDecompiled`. So when `asmResolving` is off, that first parse never ran — and that is the GROOVY-12303 case (unlinkable bytecode, class-loader lookup only). There we still decompile **once**, as a last resort, because `loadClass` cannot produce a `ClassNode` for a type that failed to link. That hatch is no longer an unmarked second copy of the happy-path decompile.

`recoverFromNoClassDefFoundError` is gone. The first pass stores a `ClassFile` (`MATCH` / `MISMATCH` / `ABSENT`, or `null` if ASM was not attempted). The `NoClassDefFoundError` arm uses that result; it parses only when the value is `null`.

### 2. `classFileDeclaresDifferentName` belongs on the first parse

Agreed. The bytecode-name check is now part of that single ASM read (`readClassFile`). A mismatch is no longer collapsed to “not found” and rediscovered with a second `DecompiledClassNode`.

The check still requires bytecode, so it does not run as a free-standing decompile when ASM was already skipped — except in the last-resort hatch above, which is the first parse of that lookup. We do not inspect `NoClassDefFoundError` text (HotSpot’s `"wrong name"` phrase is not a portable API).

A mismatch means this resource is not `name`. `loadClass` still runs afterwards: `getResource` and `defineClass` / `findLoadedClass` are independent (in-memory class of the right name, case-insensitive `getResource`). If `loadClass` then throws `NoClassDefFoundError` and we already know the resource was a mismatch, that is treated as “the requested name never existed” (`tryAsScript` with no `oldClass`), not as a linkage failure of `name`.

### 3. `tryAsScript(name, cu, null)` is not legal for `NoClassDefFoundError`

Agreed. That was the CNFE contract (`oldClass == null` ⇒ any groovy source wins, no `isSourceNewer`, no origin check). `NoClassDefFoundError` means some loader found a class (or bytes it tried to derive). Replacing it with a script is only legal when that class is from a parent — the same rule as the success path (`cls.getClassLoader() != loader`) and as the decompile path (`isFromAnotherClassLoader` then `tryAsScript(name, cu, decompiled)` so `isSourceNewer` applies).

That is now the only class→script path. Same-loader unlinkable bytecode keeps the `DecompiledClassNode`. If there is no matching class-file resource, we cannot get a timestamp or a defining loader, so we rethrow the wrapped `NoClassDefFoundError` and do not cache `NO_CLASS` — including when a groovy source of the same name happens to exist.

Thank you again for catching this.

blackdrag (changes requested):
I think I am seeing a conceptual problem in this code. And the problem may have existed already after the asm mode had been added. What I am missing is a clear definition.

|                        | ASM   |  ClassLoader |
|-------------------|--------- |------------------- |
| normal         | on       | on                   |
| ASM-only     | on       | off                   |
| loader-only | off       | on                    |
| neither         | off       | off                  |

For each row in that table we have to define what is allowed and what should be done in general, when to fall back to script and what to do in case of a class loading error or an asm format error. When is it allowed to load a class, even though we are in asm-only mode, or when it is allowed to decompile even though we are in loader-only mode. The current code is mixing everything together. And if we do our work here on differing assumptions we will both invest a lot of work without getting a result. If some of those are impossible, then the question is where that gets established.

But I have two points about the code...
1. it makes no sense adding a test for a simple constructor. We need tests on findClassNode instead of the structures it uses. Maybe tests that enter even higher. In other words there is too much whitebox testing. Plus, the tests that are there, are not testing the different modes, they test only the default mode.
2. about the script being more new than the class. If we assume that the loader can contain the class we are actually looking for then a check for if the parent knows that by loading or knows the class as resource is actually not good enough. In case of a NCDFE we no longer have the class, and therefore we cannot perform the existing cls.getClassLoader() != loader check to establish that the class is really from the parent. The fact that a class can be found through the parent is not quite equivalent to knowing that the failed class was defined by the parent. This is of course largely because GroovyClassLoader does not behave like a standard class loader.

daniellansun:
@blackdrag  
Thank you — the missing definition was the real problem, and the previous recovery path was mixing the two strategies.

The contract is now on `ClassNodeResolver` (the four-mode table in the class javadoc). `asmResolving` and `classLoaderResolving` are independent searches, not each other’s fallback:

| Mode | ASM | Class loader | Lookup | ASM format error | `loadClass` CNFE | `loadClass` NCDFE | Script |
|---|---|---|---|---|---|---|---|
| default | on | on | ASM first; `loadClass` only if ASM has no match | swallowed, then `loadClass` | `tryAsScript(null)` | wrap and rethrow, unless ASM already classified a bytecode-name mismatch | replace a found class only with `cls.getClassLoader() != loader` (or, on an ASM match, parent-visible `.class`) and a newer source |
| ASM-only | on | off | ASM only | thrown | n/a | n/a | no match → `tryAsScript(null)` |
| loader-only | off | on | `loadClass` only | n/a | `tryAsScript(null)` | wrap and rethrow; **no decompile** | success path only (`cls.getClassLoader() != loader`) |
| neither | off | off | groovy source only | n/a | n/a | n/a | `tryAsScript(null)` (`java.*` / `$` skipped) |

`ClassHelper.make(name).isResolved()` in the ASM branch is the interned-type table (`String`, primitives, …), not `loadClass`. ASM-only never calls `loadClass`. Loader-only never decompiles.

GROOVY-12303 for unlinkable bytecode is the **default** (and ASM-only) path: ASM runs first, so a missing superclass never reaches `loadClass`. Loader-only with unlinkable bytes is unrecoverable by design — that is where the mix used to live.

On your two code points:

1. The `LookupResult` constructor tests are gone. Coverage is `resolveName` / `findClassNode` across all four modes (`@MethodSource('modes')`), plus `CompilationUnit.compile` for unlinkable types (default succeeds; loader-only throws the wrapped `NoClassDefFoundError`).

2. Agreed: `parent.getResource` is not `cls.getClassLoader() != loader`, and on NCDFE there is no `Class`. We do not treat a parent resource as proof that the failed class was defined by the parent, and we do not add a `parent.loadClass` probe (`GroovyClassLoader`’s protected `loadClass` would compile scripts). NCDFE therefore does not replace a class with a script, except an ASM bytecode-name mismatch (the requested name never existed). The live-`Class` check is unchanged on the success path.

Thank you again for insisting on the matrix before more recovery patches.

blackdrag (approved):
so much cleaner. Nice!

paulk-asert (approved):
Non-blocking nits, can be tackled after merge:
1. Five test names still say findDecompiled… though the method was renamed to readClassFile.
2. resolvePackageReturnsNullWhenClassLoaderIsNull pokes unit.@classLoader - whitebox, but justified by the comment.
3. Pre-existing inconsistency worth a follow-up: the PR deliberately avoids inspecting NCDFE text, yet GroovyClassLoader.loadClass (line 706 on master) still does getMessage().indexOf("wrong name").
5. Sonar reports 5 new issues - check if any need work.
6. Seven incremental commits - squash.

PR #2837, opened 2026-08-27
GROOVY-12308: MetaClassImpl: reject in-place additions to an initiali…

paulk-asert:
…zed metaclass

addMetaBeanProperty, addNewInstanceMethod and addNewStaticMethod mutated an initialized MetaClassImpl in place with no version bump and no invalidation, so call-site caches kept serving the pre-change state while fresh lookups saw it — silently undefined behavior. Guard all three like addMetaMethod always was, enforcing the contract MutableMetaClass already documents: additions belong before initialize(); runtime changes belong on ExpandoMetaClass, whose mutation window lifts the guard. EMC's lazy memoization of an inherited expando property now runs inside that window (onSuperPropertyFoundInHierarchy mirrors addSuperMethodIfNotOverridden, whose addMetaMethod call was always guarded), keeping the two hierarchy-found hooks symmetric.


blackdrag (approved):

daniellansun:
Merged. Thanks.

PR #2839, opened 2026-08-27
GROOVY-12305: STC: record narrowed type of instanceof-guarded propert…

paulk-asert:
…y expression for static compilation

### DIFF
--- a/src/test/groovy/groovy/transform/stc/FieldsAndPropertiesSTCTest.groovy
+++ b/src/test/groovy/groovy/transform/stc/FieldsAndPropertiesSTCTest.groovy
@@ -1321,6 +1321,63 @@ class FieldsAndPropertiesSTCTest extends StaticTypeCheckingTestCase {
         '''
     }
 
+    // GROOVY-12305
+    @Test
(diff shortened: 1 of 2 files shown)


src/test/groovy/groovy/transform/stc/FieldsAndPropertiesSTCTest.groovy
```diff
@@ -1321,6 +1321,30 @@ class FieldsAndPropertiesSTCTest extends StaticTypeCheckingTestCase {
         '''
     }
 
+    // GROOVY-12305
+    @Test
+    void testMapPropertyAccess19() {
+        assertScript '''
+            def test(Map foo) {
+                foo.a instanceof Map ? foo.a.b : foo.a
```
blackdrag, line 1329: I think the test is ok, but does it works with generics too?

paulk-asert: I added some more cases

blackdrag (approved):

daniellansun:
Merged. Thanks.

PR #2841, opened 2026-08-28
GROOVY-12312: Add a compact, machine-readable diagnostic output format

paulk-asert:


src/main/java/org/codehaus/groovy/tools/FileSystemCompiler.java
```diff
@@ -212,6 +213,7 @@ public static void commandLineCompile(String[] args, boolean lookupUnnamedFiles)
      */
     public static CommandLine configureParser(CompilationOptions options) {
         CommandLine parser = new CommandLine(options);
+        parser.setCaseInsensitiveEnumValuesAllowed(true); // GROOVY-12312: --error-format=short
```
blackdrag: This looks like it does not belong in this PR really.

src/main/java/org/codehaus/groovy/control/messages/Message.java
```diff
@@ -58,10 +60,54 @@ public static Message create(final SyntaxException error, final SourceUnit owner
...
     }
+
+    //--------------------------------------------------------------------------
+
+    /**
+     * Renders one diagnostic as a single {@code file:line:column: severity: message} line.
+     * Absent parts are dropped rather than guessed: a message with no source unit omits the
+     * file, and one with no position omits the line and column.
+     */
+    static void writeShort(final PrintWriter writer, final String name, final int line,
```
blackdrag: Can we add the new API in a way that preserves the existing Message.write() contract but keeps output-format-specific logic outside the message hierarchy? Because next we may add a json output format to let AI read it better or a different format I can not think of right now.

paulk-asert: rewritten to try to achieve this.

src/main/java/org/codehaus/groovy/tools/FileSystemCompiler.java
paulk-asert: not sure if the line you are commenting on is the one you mean - I found another nearby that was a merge glitch that is now fixed and this line is gone with the refactor.

blackdrag (approved):

daniellansun:
Merged. Thanks.

PR #2842, opened 2026-08-28
GROOVY-12314: improve protected field semantics edge cases

paulk-asert:


src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -443,13 +443,37 @@ public void chooseMeta(MetaClassImpl mci) {
...
+                // GROOVY-12314: the effective lookup skips fields whose access reflection
+                // cannot force, but this sender's own lookup may still reach an inherited
+                // protected (or same-package) field, e.g. FilterReader#in from a subclass
+                MetaProperty rawMp = mci.getMetaProperty(name);
+                if (rawMp instanceof CachedField cf && !cf.isStatic()
+                        && !cf.isAccessEstablishable() && senderPassesJavaAccessRules(cf)) {
+                    try {
+                        @SuppressWarnings("removal")
+                        MethodHandles.Lookup lookup = ((Java8) VMPluginFactory.getPlugin()).newLookup(sender);
+                        handle = cf.asAccessMethod(lookup);
```
blackdrag: From the perspective of Selector I think this is actually using the wrong lookup, same for the code in 443. There is a lookup object from the callsite. This gives the access rights the caller has for the field. My favorite example is: assume we have a Groovy module G, a module written in Groovy GM and a module written in Java JM. GM uses JM and of course GM uses G. But why should G have rights to read JM? the result is that even if code from GM can read JM, making a field accessible from G can fail to do so. Access must be evaluated in the context of the call-site using Lookup, not by trying to establish reflective access from the Groovy module. That is why ReflectionUtils.makeAccessible should be avoided. That is why I think adding that field to CachedField is probably not a good idea. Lookup already handles all the access rules, checking them manually is surplus.  

The fix to replace the GroovyBugError is good I think. It also moves potential reflective access problems to the MetaClass, but here access can still fail.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
blackdrag: This is actually also wrong.

src/test/groovy/org/codehaus/groovy/transform/packageScope/DifferentPackageTest.groovy
```diff
@@ -367,7 +367,8 @@ final class DifferentPackageTest {
                 '''
             )
         }
-        assert err.message =~ /Access to p.One#value is forbidden/
+        // GROOVY-12314: rejected by the type checker, no longer during class generation
+        assert err.message =~ /No such property: value for class: p.One/
```
blackdrag: access forbidden and no such property are not the same.

src/main/java/groovy/lang/MetaClassImpl.java
```diff
@@ -3267,6 +3283,9 @@ public void setAttribute(final Class sender, final Object object, final String a
             if (mp instanceof MetaBeanProperty mbp) {
                 mp = mbp.getField();
             }
+            if (mp != null && !isAccessEstablishable(mp)) { // GROOVY-12314: unreachable field reports missing
```
blackdrag: you could consider moving the null check inside isAccessEstablishable, but that is minor

src/main/java/org/codehaus/groovy/classgen/asm/sc/StaticTypesCallSiteWriter.java
```diff
@@ -883,8 +877,14 @@ private boolean setField(final PropertyExpression expression, final Expression o
     }*/
 
     private void addPropertyAccessError(final Expression receiver, final String propertyName, final ClassNode receiverType) {
-        String receiverName = (receiver instanceof ClassExpression ? receiver.getType() : receiverType).toString(false);
-        String message = "Access to " + receiverName + "#" + propertyName + " is forbidden";
-        controller.getSourceUnit().addError(new SyntaxException(message, receiver));
+        ClassNode receiverNode = (receiver instanceof ClassExpression ? receiver.getType() : receiverType);
+        if (receiverNode.isGenericsPlaceHolder()) receiverNode = receiverNode.redirect(); // GROOVY-12314: report the erasure, not "E"
```
blackdrag: why not simply always do `receiverNode = receiverNode.redirect();`?

src/main/java/org/codehaus/groovy/reflection/CachedField.java
```diff
@@ -47,9 +47,29 @@ public CachedField(final Field field) {
...
+     * Determines if reflective access to the underlying field can actually be
+     * established: accessibility has already been forced, or forcing it can
+     * succeed (the declaring class lives in an open package or module). A
+     * non-public field of a strongly encapsulated class — e.g. a JDK class
+     * without {@code --add-opens} — reports {@code false}, which lets the MOP
+     * treat the field as absent instead of failing when it is read or written.
+     *
+     * @since 6.0.0
+     */
+    public boolean isAccessEstablishable() {
```
blackdrag: I think we should change the name as it gives a wrong impression. I suggest (be aware I am bad at naming) isReflectiveAccessEstablishable because this is really only and solely for reflective access and maybe made access by refection... but that again should maybe also be reflected in the name, so "isReflectiveAccessible". I am sure AI can suggest better names.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: Same fix: setFieldWriteHandle now uses the call-site lookup too (5624741568). The isPublic pre-check stays only as the cheap filter for the direct-handle fast path; anything else goes through the sender-aware adapter path as before.

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
paulk-asert: Agreed, and thanks for the module example — it made the problem concrete. newLookup(sender) is privateLookupIn(sender, Groovy's lookup), so access was being judged from Groovy's module: across modules the JDK drops MODULE access from that lookup and requires the target package to be open to Groovy, which is exactly your G/GM/JM case. The call-site lookup's lookup class is the sender, so it carries precisely the caller's rights.
Fixed in 5624741568: the property-get path (both the effective-field case and the retry for a field the metaclass skipped) now unreflects against callSite.getLookup(), and senderPassesJavaAccessRules is gone — the lookup decides, the reflective fallback still applies where it can, and the generic MetaProperty path remains for the rest.
On isAccessEstablishable in CachedField: I've kept it, but only for the classic Field.get path in MetaClassImpl, which has no lookup to ask. Without it a non-public field of a strongly encapsulated class is selected as a property and only fails on read/write (previously a GroovyBugError); with it the MOP reports the member as missing, which is what 4.x did. It no longer influences the indy path's decision. While doing this I found makeAccessible() recorded success unconditionally, so one failed setAccessible made the flag report the field reachable afterwards — also fixed in that commit.

src/test/groovy/org/codehaus/groovy/transform/packageScope/DifferentPackageTest.groovy
paulk-asert: You're right, and I checked what javac does to make sure I wasn't just moving the boundary arbitrarily. javac reports an access error (value is not public in One; cannot be accessed from outside package) for both the qualified and the inherited case — but its resolution still treats the inaccessible field as absent: an inner class extending p.One whose outer class also has a value compiles and picks the outer one (JLS 8.2, "not inherited"). So the two halves are: absent for resolution, access error for diagnosis. I'd only done the first half.
4e7b0d11ad adds the second: resolution still falls through (accessor, extension, map/list, enclosing scope), but when nothing else resolves the property and the field is declared by the receiver's own class, STC now reports Cannot access field: value of class: p.One from class: q.Two instead of "No such property". The two expectations in this test assert that message now.
Two boundaries I left where they were, so this PR doesn't relitigate older decisions — happy to widen them if you think we should: private fields stay hidden (GROOVY-12290), and an inaccessible inherited field still reports "No such property" for the subclass (GROOVY-9093, GROOVY-9293), on the JLS 8.2 ground that it isn't a member of that class. javac would say "not public / has private access" in those cases too, so the fully consistent version is a one-line condition plus updated expectations in those tests.

paulk-asert:
Two follow-up commits addressing the review (kept separate so the threads stay anchored; I can squash before merge):
- 5624741568 — indy uses the call-site lookup for field reads and writes; manual access rules removed from Selector; CachedField.makeAccessible records the real outcome.
- 4e7b0d11ad — STC reports an inaccessible package-private/protected field declared by the receiver's class as an access violation rather than a missing property; inherited and private cases unchanged for now (see the thread).

src/main/java/org/codehaus/groovy/vmplugin/v8/Selector.java
```diff
@@ -438,14 +438,29 @@ public void chooseMeta(MetaClassImpl mci) {
...
                 }
             } else {
+                // GROOVY-12314: the effective lookup skips fields whose access reflection
+                // cannot force, but that is a property of the reflective (Field.get) path;
+                // the call-site lookup carries the caller's own access rights, exactly as
+                // the bytecode a Java compiler would emit, so it may still reach e.g. an
+                // inherited protected field (FilterReader#in from a subclass). The lookup
+                // decides: no access rules are re-implemented here.
+                MetaProperty rawMp = mci.getMetaProperty(name);
+                if (rawMp instanceof CachedField cf && !cf.isStatic() && !cf.isAccessEstablishable()) {
```
blackdrag: you can access the field only with an `asAccessMethod`, if the lookup allows it. What `cf.isAccessEstablishable()` says to it is of no relevance. Also, when is mp no CachedField, but rawMp is? I guess that is what the comment is for and I guess this is now needed because of the changes to MetaClassImpl. But then maybe this here should have the inverse logic. first get the property via mci.getMetaProperty(name); handle the cases from that and then only get the effective property to do the generic fallback. Frankly I think we should get rid of the effective* stuff in indy, but that is beyond the scope of this PR.

paulk-asert:
@blackdrag Response below:

> Good questions — taking them in turn.
> 
> **When is `mp` not a CachedField but `rawMp` is?** Exactly when `getEffectiveGetMetaProperty` skips a field it would otherwise have chosen. Your guess is right that the GROOVY-12314 change to MetaClassImpl created the main path here (a non-public field whose access reflection cannot force now falls through to the generic property, while `getMetaProperty` still returns the raw CachedField) — but there is also a pre-existing one, which is what the condition is really guarding; see below.
> 
> **On `cf.isAccessEstablishable()` being of no relevance:** agreed the name was the problem — renamed to `isReflectiveAccessEstablishable` per your suggestion (along with the other two comments) in <commit>. But the check isn't deciding whether the lookup may access the field — the lookup alone decides that, in `asAccessMethod`. It narrows the retry to "the metaclass chose this field and dropped it *only* because reflection cannot force it". Without that narrowing, the retry also fires when the metaclass deliberately preferred something else over the field. Concrete case: a Map subclass with a package-private field. `getEffectiveGetMetaProperty` returns the map-get property *before* it ever considers the non-public field (the GROOVY-11367 ordering), which lands us in this else branch with `rawMp` the CachedField:
> 
> ```groovy
> class M extends HashMap<String, String> {
>     @PackageScope String value = 'field'
> }
> def m = new M()
> m.put('value', 'map')
> assert m.value == 'map'
> ```
> 
> Drop the condition and a same-package caller's lookup unreflects the field happily — field access silently wins over `map.get`. I verified both directions: the assert passes today and fails with the condition removed. Happy to add this to the PR as a test so the guard's purpose is pinned in code rather than in this thread.
> 
> **On the inverse logic:** I looked at it, but `getMetaProperty(name)` is a bare index lookup — no sender/selectionBase (the GROOVY-5438 private-field selection), no getter-over-field precedence, no visibility check, no map/category ordering. "Handling the cases from that" in the Selector would mean re-deriving policy the metaclass already encodes, which cuts against the direction of the rest of this review: the metaclass decides *selection*, the lookup decides *access*.

paulk-asert:
**Summary of what this PR now includes** (after the review rework; the branch is squashed to a single commit covering GROOVY-12314 plus a directly-related GROOVY-12310 fix that builds on the same machinery).

**Runtime / MOP (GROOVY-12314).** Property or attribute access to a field whose reflective access cannot be forced — e.g. a non-public field of a strongly encapsulated JDK class, absent `--add-opens` — now degrades to the normal missing-member handling (`MissingPropertyException` / `MissingFieldException`, and `ReadOnlyPropertyException` on the write side), as in 4.x, instead of `GroovyBugError` or a raw `IllegalAccessException`. Following the review discussion, this ended up with **no cached, caller-independent accessibility verdict anywhere**: selection in `MetaClassImpl` no longer predicts what reflection can do, and the reflective path reports its own failure at the point of use (conversions are gated on an `IllegalAccessException` cause, so genuine errors still propagate). The `isAccessEstablishable` API that earlier iterations added to `CachedField` is gone. `makeAccessible` is attempted at most once, race-safely (the attempt completes before being recorded — a stress-test-caught fix; the once-only latch briefly made a concurrent first access observable as a spurious missing property).

**Runtime / indy.** Per the `ARCHITECTURE.md` note in #2843, the call-site lookup is the sole access authority: field read and write handles unreflect against `callSite.getLookup()` (never a lookup fabricated from Groovy's own module), so e.g. a `FilterReader` subclass reaches the inherited protected `in` field through its own access rights — the case that motivated the ticket. On refusal, the handle is left unset and the full MOP adapter path applies; the earlier retry block that reached around the metaclass is gone. Map semantics are unaffected by construction: map-entry-over-non-public-field precedence (GROOVY-5001/5491/11367) is enforced solely by the metaclass ordering, and is now pinned by an additional `MapTest` case for `HashMap` subclass property *reads*.

**STC.** Two accessibility gaps behind property syntax are now compile-time errors, matching javac:
- *GROOVY-12314:* an inaccessible package-private or protected field declared by the receiver's own class reports `Cannot access field: ... of class: ... from class: ...` instead of "No such property" (resolution still falls through to accessor/extension/map/list first). A private field stays hidden (GROOVY-12290), and an inaccessible inherited field still reports missing (JLS 8.2; GROOVY-9093, GROOVY-9293).
- *GROOVY-12310:* a member reached through an inaccessible **qualifying type** — the reported shape is `UIInput.PropertyKeys.localValueSet`, a public enum constant of a package-private nested enum — previously compiled to a direct reference that failed at runtime with `IllegalAccessError`; it is now rejected by the type checker (JLS 6.6.1, via a new `hasAccessToClass` check). The qualifying type is checked rather than the declaring class, so an inherited public field remains reachable through an accessible subtype. Dynamic access, class references, and subscript access are unaffected, and a type-checking extension can still resolve such a reference dynamically via `unresolvedProperty`/`makeDynamic`. The `valueOf(...)` method-call variant needs the same check in method selection and is left for a follow-up.

**Classgen.** The property-access safety-net error is now well-formed: a synthetic receiver no longer yields `@ line -1, column -1`, and a type-parameter receiver reports its erasure rather than `E` (while a parameterized receiver keeps its type arguments — `List<String>`, not `List<E>`). Separately, the `size`/`length`-to-`size()` rewrite for `Collection` receivers is dropped: STC no longer admits either name unless a genuine member backs it, so `list.size` under `@CompileStatic` is now a compile error rather than a silently rewritten value — a deliberate source-breaking alignment with dynamic Groovy, which already throws for it (tests using a genuine `size` field member pin that real members still win).

**Tests.** New `Groovy12314` suite (the JDK-field cases guard themselves with an assumption and skip under `--add-opens`; permissive dynamic read *and* write to open classes pinned; public-field-of-package-private-class pinned through the forced-access retry), new `Groovy12310` suite, the `MapTest` read-precedence pin, an STC extension resource for the parameterized-receiver classgen message, and updated package-scope/STC expectations for the new compile-time errors.

**Behaviour changes to call out:** cross-package field misuse behind property syntax under STC is now a compile error; `list.size`/`list.length` under `@CompileStatic` no longer compile; strongly encapsulated JDK fields report missing instead of crashing. Dynamic Groovy's permissive access to open/class-path classes is unchanged throughout.

**Follow-ups deliberately out of scope:** the method-path twins of the field fix (`CachedMethod` forcing access before `Lookup.unreflect` can check it, and `Java9.checkAccessible` hand-rolling module rules during linkage), the `valueOf` case of 12310, and removing the `effective*` machinery from indy — the latter expected to be subsumed by the realm-aware invokedynamic work under GEP-31 in Groovy 7.

src/main/java/org/codehaus/groovy/reflection/CachedField.java
```diff
@@ -46,10 +46,19 @@ public CachedField(final Field field) {
...
+    private volatile boolean madeAccessible;
+    private boolean accessAttempted; // guarded by synchronization on this
     private void makeAccessible() {
-        ReflectionUtils.makeAccessibleInPrivilegedAction(field);
-        madeAccessible = true;
+        // at most one attempt, remembering either outcome: a failed attempt (strongly
+        // encapsulated declaring class) cannot succeed later. The attempt is recorded
+        // only once it has completed, under synchronization, so a concurrent caller
+        // waits instead of reading the field before setAccessible has taken effect.
+        synchronized (this) {
```
blackdrag: I wonder if we can simplify the accessibility/handle caching here and avoid the synchronization.

Since `setAccessible(true)` changes the accessibility of the `Field` itself, once Groovy has successfully established deep access, we can use Groovy's own lookup to create the getter/setter handle and cache that handle. A handle created with the caller's `Lookup` should not be cached, since it is caller-specific.

This would give us roughly:

```java
MethodHandle h = getter;
if (h != null) return h;

if (!accessAttempted) {
    try {
        if (ReflectionUtils.makeAccessibleInPrivilegedAction(field).isPresent()) {
            h = getter = MethodHandles.lookup().unreflectGetter(field);
            return h;
        }
    } finally {
        accessAttempted = true;
    }
}

return lookup.unreflectGetter(field);
```

and analogously for the setter.

`accessAttempted` would then mean only that we have already tried Groovy's deep-reflection path. It could be `volatile`, allowing the benign race where multiple threads perform the same `setAccessible`/handle creation concurrently. I don't think that race matters, since all threads operate on the same `Field` and the resulting handles are equivalent, whereas synchronizing the initialization adds blocking just to avoid that duplicate work.

The important part for me is that the successful deep-reflection handle can be cached, while a handle obtained from the caller's `Lookup` cannot. This also means that if `setAccessible` fails, we remember that attempt and fall back to the caller's `Lookup` without repeatedly trying the deep-reflection path.

Using this implementation also changes the need for the exception handling changes in MetaClassImpl

paulk-asert: Implemented as you sketched in ed4ec398f8: the deep-reflection getter/setter handles are created through Groovy's own lookup once `setAccessible` has succeeded and cached; a handle from the caller's lookup is returned but never cached. The synchronization is gone.

Three details on top of the sketch, for the record:

- **The one thing I kept from the synchronized version is the write ordering**: `accessAttempted` is recorded only *after* the attempt completes (both volatile). The benign race is exactly as you described for the handle paths — duplicate `setAccessible`/unreflect on the same `Field` produce equivalent handles — but the classic `Field.get`/`set` paths share the same latch and have no handle to hand them safety, so the ordering is what gives a thread that observes `accessAttempted` a happens-before edge to the outcome and to the field's accessibility. That closes, without blocking, the race the stress test caught (a concurrent first access could observe the latch before `setAccessible` had taken effect, and the resulting `IllegalAccessException` surfaced as a spurious missing property). The `ClassInfoSoftModeStressTest` loop that failed 2/6 against the synchronized-less latch is 10/10 on this.
- **No deep setter is created for a final field**: after `setAccessible`, `unreflectSetter` would happily hand out a working write handle for a non-static final, so that case deliberately falls through to the caller's lookup and keeps failing, preserving the `ReadOnlyPropertyException` semantics.
- **Ordering flip acknowledged**: this tries deep reflection before the caller's lookup, the reverse of the GROOVY-9144/9596 shape. That ordering existed to avoid needless `setAccessible` in the illegal-access-warning era; on the JDKs we support, `setAccessible` either succeeds silently or throws, and I checked the four cases (open/encapsulated × entitled/unentitled caller) all end in the same outcome — the encapsulated-but-entitled case (`FilterReader#in` from a subclass) pays one latched failed attempt and then goes through the caller lookup as before.

On the exception handling in MetaClassImpl: I've left it in place for now. After this change it is at least *deterministic* — the only `IllegalAccessException` left on the classic path is the "deep path unavailable and no lookup to ask" state, so converting it to missing-member handling no longer risks masking a transient failure. I agree the conversions could become plain control flow (the classic path can now ask whether deep access is available before attempting the read), but I'd want that check to stay at the use sites rather than migrate into `getEffectiveGetMetaProperty` — filtering during selection is what the earlier round of this review removed, and it's what kept the map-precedence and Selector entanglement alive. Happy to do that simplification here or as an immediate follow-up, whichever you prefer.

paulk-asert: The simplification mentioned above is added as another commit.

blackdrag (approved):

PR #2845, opened 2026-08-29
GROOVY-12319: Java compatibility: remaining generic type syntax

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12319

### DIFF
--- a/src/main/java/org/codehaus/groovy/ast/ClassNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ClassNode.java
@@ -118,6 +118,11 @@ import static org.objectweb.asm.Opcodes.ACC_SYNTHETIC;
  * they describe the type signature used at the point of declaration or the
  * type signatures provided by the class. If the type signatures provided
  * by the class are needed, then a call to {@link #redirect()} will help.
+ * {@link #getOuterClassType()} / {@link #setOuterClassType(ClassNode)} are
+ * likewise not proxied: they record the parameterized enclosing type of a
+ * JLS 4.5 rare type <em>use</em> ({@code Outer<T>.Inner}), distinct from
+ * {@link #getOuterClass()} which is the enclosing class of a nested class
+ * declaration.
  *
  * @see org.codehaus.groovy.ast.ClassHelper
  */
@@ -190,6 +195,8 @@ public class ClassNode extends AnnotatedNode {
     protected Class<?> clazz;
     // if not null then this instance is an array
     private ClassNode componentType;
+    // parameterized enclosing type of a JLS 4.5 rare type (Outer<T>.Inner); type-use, not declaration
+    private ClassNode outerClassType;
     // if not null then this instance is handled as proxy for the redirect
     private ClassNode redirect;
 
@@ -2055,24 +2062,27 @@ faces:  if (method == null && asBoolean(getInterfaces())) { // GROOVY-11323
 
     /**
      * For a JLS 4.5 rare type {@code Outer<T>.Inner}, the parameterized enclosing
-     * type. Distinct from {@link #getOuterClass()}, which is the enclosing class
-     * of a nested class <em>declaration</em>.
+     * type stored on this node (type-use, not declaration; not node metadata and
+     * not redirected). Distinct from {@link #getOuterClass()}, which is the
+     * enclosing class of a nested class <em>declaration</em>. Copied by
+     * {@link #getPlainNodeReference()}.
      *
      * @return the parameterized enclosing type, or {@code null} if this is not a rare type
      * @since 6.0.0
      */
     public ClassNode getOuterClassType() {
-        return getNodeMetaData("outer.class");
+        return outerClassType;
     }
 
     /**
-     * Records the parameterized enclosing type of a JLS 4.5 rare type.
+     * Records the parameterized enclosing type of a JLS 4.5 rare type on this
+     * node. Does not write through {@link #redirect()}.
      *
      * @param outer the parameterized {@code Outer<T>} node; {@code null} clears it
      * @since 6.0.0
      */
     public void setOuterClassType(final ClassNode outer) {
-        putNodeMetaData("outer.class", outer);
+        this.outerClassType = outer;
     }
 
     /**
--- a/src/main/java/org/codehaus/groovy/ast/ClassNode.java
+++ b/src/main/java/org/codehaus/groovy/ast/ClassNode.java
@@ -257,17 +257,15 @@ public class ClassNode extends AnnotatedNode {
     }
 
     /**
-     * Constructs a non-primary array {@code ClassNode} when no real array class is available.
-     * This internal constructor is used by {@code makeArray()} to create a synthetic array ClassNode.
-     * The resulting node represents an array type with appropriate interfaces and modifiers.
-     *
-     * @param componentType the {@link ClassNode} representing the element type of the array
+     * Synthetic array {@code ClassNode} used by {@link #makeArray()} when no real
+     * array class is available. {@code componentType} is the element type.
      */
-    private ClassNode(final ClassNode componentType) {
-        this(componentType.getName() + "[]", ACC_ABSTRACT | ACC_FINAL | ACC_PUBLIC, ClassHelper.OBJECT_TYPE,
+    private static ClassNode syntheticArray(final ClassNode componentType) {
+        ClassNode node = new ClassNode(componentType.getName() + "[]", ACC_ABSTRACT | ACC_FINAL | ACC_PUBLIC, ClassHelper.OBJECT_TYPE,
           new ClassNode[]{ClassHelper.CLONEABLE_TYPE, ClassHelper.SERIALIZABLE_TYPE}, MixinNode.EMPTY_ARRAY);
-        this.componentType = componentType.redirect();
-        this.isPrimaryNode = false;
+        node.componentType = componentType.redirect();
+        node.isPrimaryNode = false;
+        return node;
     }
 
     //--------------------------------------------------------------------------
@@ -338,6 +336,10 @@ public class ClassNode extends AnnotatedNode {
         n.setRedirect(redirect());
         if (isArray()) {
             n.componentType = redirect().getComponentType();
+        }
+        ClassNode outerType = getOuterClassType();
+        if (outerType != null) {
+            n.setOuterClassType(outerType);
         }
         return n;
     }
@@ -503,6 +505,7 @@ public class ClassNode extends AnnotatedNode {
      * @return the modifier flags as an integer bitmask
      * @see #setModifiers(int)
      * @see #isAbstract()
+     * @see #isStatic()
      * @see #isInterface()
      * @see #isEnum()
      */
@@ -1862,6 +1865,22 @@ faces:  if (method == null && asBoolean(getInterfaces())) { // GROOVY-11323
     }
 
     /**
+     * Returns whether this {@link ClassNode} represents a static class.
+     * A class is static if the {@code ACC_STATIC} modifier flag is set in its modifiers.
+     * Static nested classes have no enclosing instance; they can be used without an instance of the enclosing class.
+     * This method checks the actual modifier bits, independent of the source code syntax.
+     *
+     * @return {@code true} if this class has the static modifier set; {@code false} otherwise
+     * @see #getModifiers()
+     * @see #setModifiers(int)
+     *
+     * @since 6.0.0
+     */
+    public boolean isStatic() {
+        return (getModifiers() & ACC_STATIC) != 0;
+    }
+
+    /**
      * Returns whether this {@link ClassNode} represents an interface type.
      * A class is an interface if the {@code ACC_INTERFACE} modifier flag is set in its modifiers.
      * Interfaces define contracts for implementing classes without providing implementation details.
@@ -1980,7 +1999,7 @@ faces:  if (method == null && asBoolean(getInterfaces())) { // GROOVY-11323
             // don't use the ClassHelper here!
             node = new ClassNode(type, this);
         } else {
-            node = new ClassNode(this);
+            node = syntheticArray(this);
         }
         return node;
     }
@@ -2032,6 +2051,28 @@ faces:  if (method == null && asBoolean(getInterfaces())) { // GROOVY-11323
         } while ((outer = outer.getOuterClass()) != null);
 
         return result;
+    }
+
+    /**
+     * For a JLS 4.5 rare type {@code Outer<T>.Inner}, the parameterized enclosing
+     * type. Distinct from {@link #getOuterClass()}, which is the enclosing class
+     * of a nested class <em>declaration</em>.
+     *
+     * @return the parameterized enclosing type, or {@code null} if this is not a rare type
+     * @since 6.0.0
+     */
+    public ClassNode getOuterClassType() {
+        return getNodeMetaData("outer.class");
+    }
+
+    /**
+     * Records the parameterized enclosing type of a JLS 4.5 rare type.
+     *
+     * @param outer the parameterized {@code Outer<T>} node; {@code null} clears it
+     * @since 6.0.0
+     */
+    public void setOuterClassType(final ClassNode outer) {
+        putNodeMetaData("outer.class", outer);
     }
 
     /**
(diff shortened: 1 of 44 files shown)


src/main/java/org/codehaus/groovy/ast/ClassNode.java
```diff
@@ -2034,6 +2053,28 @@ public List<ClassNode> getOuterClasses() {
...
+    }
+
+    /**
+     * Records the parameterized enclosing type of a JLS 4.5 rare type.
+     *
+     * @param outer the parameterized {@code Outer<T>} node; {@code null} clears it
+     * @since 6.0.0
+     */
+    public void setOuterClassType(final ClassNode outer) {
+        putNodeMetaData("outer.class", outer);
```
blackdrag: if you add a getter and setter for it, then we not a field? node meta data is for data the node does not know about and is "imposed" by outer processing. This change makes it known, thus nodemetadata is the wrong place.

src/main/java/org/codehaus/groovy/classgen/asm/BytecodeHelper.java
```diff
@@ -466,12 +467,41 @@ private static void writeGenericsBoundType(StringBuilder ret, ClassNode printTyp
...
+        String innerName = inner.getName();
+        String ownerName = owner.getName();
+        if (innerName.startsWith(ownerName) && innerName.length() > ownerName.length()) {
+            char sep = innerName.charAt(ownerName.length());
+            if (sep == '.' || sep == '$') {
+                return innerName.substring(ownerName.length() + 1).replace('$', '.');
+            }
+        }
+        int dot = Math.max(innerName.lastIndexOf('.'), innerName.lastIndexOf('$'));
+        return dot < 0 ? innerName : innerName.substring(dot + 1);
```
blackdrag: So the assumption seems to be that if the innerName does not start with ownerName, then it is a "plain" name. For example `Foo` as owner and `Foo$1` as inner.  
1. why is it that one time we have innerName starting with owner and another time not? This feels like some kind of normalization is missing
2. What if ower is `Foo` and inner is `Bar.X`? This will write `Foo.Bar.X` I assume, but is it legal? Considering that this means owner should have `Foo.Bar` since we do only then the recursive call, it seems that `Bar.X` is illegal for inner. But why can this not happen?

src/main/java/org/codehaus/groovy/control/GenericsVisitor.java
```diff
@@ -159,11 +189,61 @@ public void visitDeclarationExpression(final DeclarationExpression expression) {
...
+
+    /**
+     * Groovy represents the type operand of {@code instanceof} as a
+     * {@link ClassExpression}, but it is not a class literal. Skip it so
+     * {@link InstanceOfVerifier} can diagnose JLS 15.20.2.
+     */
+    @Override
+    public void visitBinaryExpression(final BinaryExpression expression) {
+        if (expression.getOperation().isA(Types.INSTANCEOF_OPERATOR)
+                && expression.getRightExpression() instanceof ClassExpression) {
```
blackdrag, line 268: What happens if we do "x instanceof Map<String,Integer>"? Is this forbidden by the grammar?

src/main/java/org/codehaus/groovy/ast/ClassNode.java
daniellansun: Agreed. `outerClassType` is now a real field on `ClassNode`, next to
`componentType`. Getter/setter only touch that field and do not write through
`redirect()`, matching `getGenericsTypes()` / `setGenericsTypes()`. A leftover
`"outer.class"` metadata key is no longer read. `getPlainNodeReference()` still
copies the field.

src/main/java/org/codehaus/groovy/classgen/asm/BytecodeHelper.java
daniellansun: 1. The two spellings are parser vs resolver, not two different nesting rules.
   A rare type is first recorded as `Outer.Inner`; after resolve,
   `ClassNode.getName()` follows the redirect and is `Outer$Inner`. Anonymous
   classes (`Foo` / `Foo$1`) already match the owner prefix, so they take the
   same path as `Outer$Inner` — there is no separate “plain name” case for
   `Foo$1`. Both `.` and `$` after a matching owner prefix are treated as
   nesting separators. We did not fold the parser onto `$` in this follow-up:
   classgen already sees the binary name via `getName()`, and accepting both
   separators is cheaper than changing how the parser names unresolved types.

2. `Bar.X` with owner `Foo` is not a well-formed rare type (that would be
   `Foo$Bar$X`). The fallback is the last identifier only (`.X`), never extra
   qualification of the owner. It does **not** emit `Foo.Bar.X`.

   Tests assert the JVMS 4.7.9.1 strings: `Outer.Inner` and `Outer$Inner` both
   emit `L…Outer<…>.Inner<…>;`, `Foo$1` emits `L…Foo<…>.1;`, `Bar.X` emits
   `L…Foo<…>.X;`, and `Foo` / `FooBar` (prefix but not a nesting separator)
   emits `L…Foo<…>.FooBar;`. Nested form is used only when the enclosing type
   is parameterized; a raw enclosing type keeps the binary name
   (`Outer$Inner<…>`).

src/main/java/org/codehaus/groovy/control/GenericsVisitor.java
daniellansun: Not forbidden by the grammar. `matchingType` / `standardType` still take type
arguments so `instanceof List<?>` (and `instanceof Map<?,?>`) can parse
(JLS 15.20.2 / 4.7 — unbounded wildcards are reifiable).

`x instanceof Map<String,Integer>` is a parse error from
`AstBuilder.rejectParameterizedInstanceof`
(`Cannot perform instanceof check against parameterized type …`). The same
applies to `!instanceof` and to bounded wildcards (`List<? extends T>`,
`List<? super T>`).

`GenericsVisitor` does not visit the `instanceof` operand as a class
expression (that would treat it as `T.class`). It does run
`checkGenericsUsage` on the type, so nested-type well-formedness matches
javac / JDK `T6665356`:

- `instanceof Outer<?>.Inner` is illegal when `Inner` is generic (raw member
  of a parameterized enclosing type);
- `instanceof Outer.Inner<?>` is illegal (type arguments on a member of a
  raw type). The same form is rejected as a field type.

`InstanceOfVerifier` still only checks primitives and type-parameter targets.
We did **not** reject parameterized `instanceof` there: resolved `ClassNode`s
often still carry declaration placeholders (`List<E>`), and treating those as
a parameterized *use* breaks legitimate Groovy (reproduced on `groovy-macro`).

The Java-vs-Groovy matrix is in `GenericsJavaCompatibilityTest` (JLS 4.7 /
15.20.2). Reifiable cases include `list instanceof List<?>` with a
`List<String>` left operand, `ArrayList<?>`, raw `List`, `Map<?,?>`,
`Class<?>`, `List<?>[]`, `Outer<?>.Inner` (non-generic `Inner`),
`Outer.Inner`, and `Outer<?>.Inner<?>`. Non-reifiable cases include
`List<String>`, `Map<String,Integer>`, `!(o instanceof Map<String,Integer>)`,
bounded and mixed wildcards, `List<String>[]`, `T` / `T[]`, `Class<C>` /
`Class<? extends C>`, and `Outer<String>.Inner`.

blackdrag (approved):

paulk-asert:
I am generally in favor of this change but want to draw attention to one aspect before we merge as per this AI read.

> Alongside the three new syntax forms, the `GenericsVisitor` changes add JLS well-formedness checks that run on **every class — dynamic Groovy and scripts included, before AST transforms, with no opt-out**. I ran probes on 4.0.27, 5.0.8, 6.0.0-beta-3 and builds of both PR heads (`f0de882dd4`, `86f1ab3e3c`): **eleven forms that compile and run today (and have since at least 4.0) become compile errors**. Each is Java-consistent, and several expose silently-wrong code, so break-and-document is defensible for all of them — but we should decide that knowingly, per form:
> 
> | Form | Behaviour ≤ beta-3 (dynamic) | Suggest |
> |---|---|---|
> | `new T[n]`, `new T[]{t}` in `class G<T>` | `Object[]` typed as `T[]` — the Groovy shortcut for Java's `(T[]) new Object[n]` | **team call** — the one form with a real constituency; a small carve-out in `isReifiable` for bare type variables is possible |
> | `class E<T> extends Exception` | works; no JVM/Groovy reason to forbid, pure javac parity | **team call** — cheapest to leave if we want to minimise breakage |
> | `new List<String>[n]` | raw array, args ignored | break & document (`new List<?>[n]` still works) |
> | `List<int>` (and nested) | dynamic: treated as raw | break & document (STC already rejected it with a worse message) |
> | `new T()` / `T.class` | silently `Object` / `Object.class` | break & document — exposes wrong code; point to the `Class<T>` token idiom |
> | `implements X<? extends …>` | works (superclass form was already rejected) | break & document |
> | `new ArrayList<?>()` | works, harmless | break & document (parity) |
> | class type in additional bound `<T extends A & B>` | works | break & document |
> | `Foo<String>.class` | evaluates to `Foo.class` | break & document, but the message needs fixing (see below) |
> | `Outer.Inner<String>` with raw generic `Outer` (added in the follow-up) | works | break & document (fix: `Outer<?>.Inner<String>`, expressible thanks to this PR) |
> 
> Not breaking, for the record: `new Comparator<int>() { … }` already failed at runtime with a corrupt `Signature` (`GenericSignatureFormatError`) since 4.x — the new error is a fix; the stricter `instanceof` forms and the static-context `T` reference were already errors (clearer messages now); third-party AST transforms are safe (they run after the `GenericsVisitor` phase op — verified); and since the follow-up, inner-class signatures of existing code are unchanged.
> 
> Two message nits: `Foo<String>.class` still gets "Cannot refer to a static member of a generic type through a parameterization" (the class-literal path missed the message pass — javac says "cannot select from parameterized type"); and `new T[n]` reports "generic array creation **of Object**", which names the erasure rather than the type variable and reads like a different complaint.
> 
> One API note: the follow-up stores the enclosing type as a `ClassNode` field and no longer writes the `"outer.class"` node metadata — fine by me, just noting that any external code reading that key (GROOVY-10646 era) now sees `null`.

Most breaking changes seem fine - we need to add a table like above in release notes and possibly in `core-object-orientation` §Generics. I am happy to work on these separately. It is really only the first two rows which I wonder about - spelt out below.

> Both workarounds verified on the PR install (dynamic and `@CompileStatic`). Here are the two team-call rows as complete, paste-ready samples — each with the code that works today, the exact PR error, and the verified workaround.
> 
> ## Team call 1 — `new T[n]` / `new T[]{…}` (generic array creation)
> 
> **Works on ≤ 6.0.0-beta-3 (and 5.x, 4.x), dynamic and `@CS`; errors on the PR:**
> 
> ```groovy
> // a typed fixed-capacity stack — the kind of generic utility class where this idiom shows up
> class Stack1<T> {
>     private T[] items
>     private int size = 0
>     Stack1(int capacity) { items = new T[capacity] }       // PR: generic array creation of Object
>     void push(T t) { items[size++] = t }
>     T pop() { items[--size] }
>     T[] snapshot() { new T[] { items[0] } }                // PR: generic array creation of Object
> }
> def s = new Stack1<String>(4)
> s.push('a'); s.push('b')
> assert s.pop() == 'b'
> ```
> 
> On beta-3 both array creations quietly build an `Object[]` (erasure) and everything runs. On the PR each line fails with `generic array creation of Object` — note the message names the erasure, not `T`, which is the wording nit in the draft comment.
> 
> **Workaround (the Java idiom; verified OK on the PR build, both modes):**
> 
> ```groovy
> class Stack2<T> {
>     private T[] items
>     private int size = 0
>     Stack2(int capacity) { items = (T[]) new Object[capacity] }   // reifiable creation + unchecked cast
>     void push(T t) { items[size++] = t }
>     T pop() { items[--size] }
> }
> def s = new Stack2<String>(4)
> s.push('a'); s.push('b')
> assert s.pop() == 'b'
> ```
> 
> The team question: is forcing that rewrite worth the Java parity, or do we carve out bare type variables in `isReifiable` (keeping `new List<String>[n]` rejected while `new T[n]` stays legal)?
> 
> ## Team call 2 — generic class extending `Throwable`
> 
> **Works on ≤ beta-3, dynamic and `@CS`; errors on the PR:**
> 
> ```groovy
> // a DSL-style exception carrying a typed payload — the plausible real-world shape
> class ValidationException<T> extends RuntimeException {
>     final T details                                        // PR: A generic class may not extend java.lang.Throwable
>     ValidationException(String msg, T details) { super(msg); this.details = details }
> }
> 
> try {
>     throw new ValidationException<Map>('bad input', [field: 'name'])
> } catch (ValidationException<Map> ignored) {               // note: parameterized catch never made sense…
>     // …
> } catch (ValidationException e) {
>     assert e.details.field == 'name'
> }
> ```
> 
> Java forbids this (JLS 8.1.2) because `catch` clauses can't be reified — but Groovy's `catch` is erased anyway, so the class runs fine on the JVM; the rejection is pure javac parity. That's why this is the cheapest row to *leave* if the team wants to minimise breakage.
> 
> **Workaround (verified OK on the PR build, both modes):**
> 
> ```groovy
> class ValidationException extends RuntimeException {
>     private final Object payload
>     ValidationException(String msg, Object payload) { super(msg); this.payload = payload }
>     def <T> T payload() { (T) payload }        // generics move from the class to the accessor
> }
> 
> try {
>     throw new ValidationException('bad input', [field: 'name'])
> } catch (ValidationException e) {
>     Map m = e.<Map>payload()
>     assert m.field == 'name'
> }
> ```
> 
> (A variant that keeps static typing without the explicit type witness — implementing `Supplier<Map>` or exposing a concretely-typed field — also verified fine.)

daniellansun:
@paulk-asert  Thanks, Paul — this was the right list to walk through before merge. We have decided each row as below; the well-formedness checks stay on **every** compilation unit (dynamic Groovy and scripts included), with no opt-out.

### Team call 1 — `new T[n]` / `new T[]{…}`

We will **reject** it, with no carve-out in `isReifiable`.

A type variable is not reifiable (JLS 4.7 / 15.10.1). What Groovy 4/5 compiled was `Object[]` (or the bound) typed as `T[]` — Java’s `(T[]) new Object[n]`, without the cast in the source. Treating `T` as reifiable would also be the wrong predicate for `instanceof T`.

The Java idiom remains the workaround, in both dynamic and `@CompileStatic` code:

```groovy
items = (T[]) new Object[capacity]          // unbounded T
items = (T[]) new Bound[capacity]           // T extends Bound
items = (T[]) Array.newInstance(token, n)   // when a Class<T> is at hand
```

`new List<String>[n]` stays rejected; `new List<?>[n]` stays legal.

### Team call 2 — generic class extending `Throwable`

We will **reject** it as well, for JLS 8.1.2 / javac parity. Groovy’s `catch` is erased, so the class would still run on the JVM — but this PR’s charter is well-formedness that javac already enforces, and Groovy 6 is the window.

Workaround: keep the exception non-generic and move the type parameter onto an accessor (or a factory). Parameterized `catch` is already impossible in the grammar (`catchType` has no type arguments).

### Remaining rows

Agreed: **break and document** for all of them, including `List<int>`, `new T()` / `T.class`, `implements X<? extends …>`, `new ArrayList<?>()` , a class type as an additional bound, `Foo<String>.class`, and `Outer.Inner<String>` with a raw generic `Outer`.

A table in that shape is now in `core-object-orientation.adoc` §Generics (`[[generics-well-formedness]]`), with the two workarounds above as executable examples. I would still welcome a matching table in the 6.0 release notes if you would like to take that, as you offered.

### Message nits

Both fixed:

- `new T[n]` now reports `generic array creation of T` (the type variable, not the erasure).
- `Foo<String>.class` now reports `Cannot select from a parameterized type` (JLS 15.8.2). `Cell<String>.ID` / `.id()` keep the 4.5.2 “static member through a parameterization” wording. `value.class` / `value*.class` are unchanged — those are `getClass()`, not class literals.

### `"outer.class"` metadata

Noted. The enclosing type lives on the `ClassNode` field (`getOuterClassType()` / `@since 6.0.0`); we will not restore a metadata fallback. Transform authors who still read the GROOVY-10646-era key will see `null`.

PR #2852, opened 2026-09-01
GROOVY-12325: Speed up CachedMethod.invoke with a generated JIT-constant trampoline

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12325

See also: https://github.com/daniellansun/fast-reflection#benchmark-test

# JEP 416 vs GROOVY-12325

Working note. Not a JIRA comment and not a commit message.

- JDK: [openjdk/jdk#5027](https://github.com/openjdk/jdk/pull/5027) ([JEP 416](https://openjdk.org/jeps/416) / [JDK-8271820](https://bugs.openjdk.org/browse/JDK-8271820)), integrated 2021-10-28 as `c6339cb`, **JDK 18**.
- Groovy: `c17676b9e7e3625fdcf0bfdaa7bb1e993e641d47` ([GROOVY-12325](https://issues.apache.org/jira/browse/GROOVY-12325) / [apache/groovy#2852](https://github.com/apache/groovy/pull/2852)).

## Verdict

**The JDK PR does not implement Groovy’s `DirectInvoker`.** GROOVY-12325 is not a duplicate. Keep it.

JEP 416 rewrites **`Method.invoke` from the inside**. GROOVY-12325 **stops calling `Method.invoke`** on the hot MOP path. Same family of ideas (make the call target a JIT constant); different layer, different product, different call shape.

## What each change is

| | OpenJDK #5027 / JEP 416 | GROOVY-12325 |
|---|---|---|
| Layer | Inside `java.lang.reflect.Method.invoke` | Above it: `CachedMethod.invoke` |
| Public API | Unchanged | None (`org.apache.groovy.internal…`, `@Internal`) |
| First JDK | 18 | Groovy still targets **17** (`targetJavaVersion=17`) |
| Shipped encoding | One `DirectMethodHandleAccessor` class; per-method **instance** `MethodHandle`; `invokeExact` by arity | One **hidden class per method**: direct `INVOKE*` (steps 1, 2, 4) or classData `MethodHandle` + `invokeExact` (step 3) |
| Peak path | `Method` in a `static final` field so the JIT constant-folds through to the MH | Constant `Methodref` / classData in the trampoline bytecode |
| Non-constant `Method` | JEP’s own micros: **slower** than the old accessor | The MOP shape this change is written for |
| Extra policy | Receiver / argc checks, ITE wrapping, `@CallerSensitiveAdapter` | Groovy as caller; indy must **not** use the trampoline; CS / abstract stay reflective; `InvokerInvocationException` + `normalizeBoxedReturn` |

JDK 17 (Groovy’s floor) still inflates `Method.invoke` to `GeneratedMethodAccessor` (`INVOKE*`, ~15 hits). JEP 416 **removes** that generator and replaces it with the shared MH accessor.

The PR text described spinning a hidden class that loads the MH from **class data** as a dynamically computed constant. **That stub is not in the landed code** (JDK 18 through current `master`): `DirectMethodHandleAccessor` keeps `target` as an instance field. Mandy Chung’s working branch had `AdaptiveMethodHandleAccessor` / `spinMethodHandleInvoker`; it was dropped before integration.

## Why the MOP is JEP 416’s slow path

JEP 416 is explicit: the 43–57% win needs `Method` / `Constructor` / `Field` in **`static final`**. From the JEP micros (`instanceMethod*`, ns/op):

| Shape | Old reflection | JEP 416 |
|---|---|---|
| Const (`static final Method`) | 77.9 | **33.4** |
| Var (non-final field) | 80.0 | **90.4** |
| Poly | 96.1 | **109.1** |

`CachedMethod` holds

```java
private final Method cachedMethod;
```

That is an instance field. The `CachedMethod` itself is taken from a `CachedClass` method table (an array). That is JEP 416’s **Var / Poly** case — the one the JEP measured as a regression versus JDK 17’s per-method `INVOKE*` stub.

Groovy cannot put every cached member in a `static final Method`. Raising the floor to JDK 18+ would not make DirectInvoker redundant.

## Technique overlap (and where it stops)

- **Steps 1 / 2 / 4** (`INVOKE*` in a unique hidden nestmate) restore what JEP 416 **deleted** (`GeneratedMethodAccessor`), one layer above `Method.invoke`. Not present in the shipped JDK PR.
- **Step 3** (classData MH + `invokeExact`) is what #5027 **described** and **did not ship**. Closest overlap; still not a substitute, because Groovy still skips the `Method.invoke` wrapper (`@CallerSensitive`, access check, accessor dispatch, ITE).
- Groovy also has constraints JEP 416 does not: Lookup split (do not feed the trampoline to indy / `MethodHandleWrapper`), sticky skip for caller-sensitive and abstract methods.

A second hidden class on JDK 17 (JDK inflates at ~15 hits, Groovy at 101) is real metaspace cost. On JDK 18+ there is no JDK-generated `INVOKE*` class, so Groovy’s trampoline is the only one.

## What would make GROOVY-12325 unnecessary

Not “JEP 416 exists.” All of the following would have to be true:

1. Groovy’s minimum JDK ≥ 18 (so JEP 416 is always present).
2. The JIT can constant-fold `CachedMethod` → `Method` → accessor → MH on the real MOP shape (instance field / array). JEP 416’s own numbers say it cannot.
3. The remaining `Method.invoke` wrapper is cheaper than `DirectInvoker.invokeinterface` at the shared `CachedMethod.invokeGenerated` call site.

Until (2) is true, staying on `Method.invoke` on JDK 18+ is the path JEP 416 made **worse** for this shape.

## Measure, don’t assume

Shared `CachedMethod.invokeGenerated` is `invokeinterface DirectInvoker`. Many inflated methods → megamorphic at that one bytecode. JEP 416’s accessor class is the opposite: **one class, easy to inline**, but the MH field is not a constant.

That trade-off is empirical. The in-tree JMH is

`subprojects/performance/src/jmh/java/org/apache/groovy/bench/CachedMethodInvokerBench.java`

(`:perf:jmh -PbenchInclude=CachedMethodInvoker`; `threshold=0` vs `disable=true`; includes `mega` rows). Judge **MOP shape** (method taken from a table, not `static final Method`) on **JDK 17 and 21+**. The JEP Const numbers are not evidence for Groovy.

## Bottom line

| Question | Answer |
|---|---|
| Did #5027 already implement `c17676b`? | No. It reimplemented `Method.invoke`. |
| Is Groovy repeating the JDK? | No as a product. Only step 3 matches the PR **description**, which the JDK did not land. |
| Still needed in Groovy? | Yes. Especially on JDK 18+, where non-constant `Method` is JEP 416’s slow path, and on JDK 17, where JEP 416 does not exist. |
| Revert GROOVY-12325 because of JEP 416? | No. Revert only if that JMH loses on both 17 and 21+ against inflated `Method.invoke`. |

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -284,6 +284,16 @@ above. Each bites contributors quickly if missed:
   reproduce Java/module access rules manually when a `Lookup` can
   perform the check. The cached member representation describes
   the member; the call site supplies the access context.
+- **JIT-constant invoke for MOP internals.** After a call target
+  is selected and stable, `CachedMethod.invoke` may install a
+  generated `DirectInvoker` (direct `INVOKE*` or a classData
+  `MethodHandle`) so the invoke is a JIT-constant call rather than
+  `Method.invoke`. This is the MOP "Groovy as caller" path; indy
+  continues to use the call-site `Lookup` and must not be fed the
+  trampoline. Kill switch:
+  `groovy.cachedmethod.invoker.disable`. Hit threshold:
+  `groovy.cachedmethod.invoker.threshold` (default 100). See
+  `org.apache.groovy.internal.runtime.invoke`.
 
 ## Operator families
 
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -289,7 +289,7 @@ above. Each bites contributors quickly if missed:
   caller). Indy already uses a linked `CallSite` and must keep the
   call-site `Lookup`; DGM / `CallSiteGenerator` already emit
   `INVOKE*`. After
-  `groovy.cachedmethod.invoker.threshold` hits (default 100) a
+  `groovy.cachedmethod.invoker.threshold` hits (default 1000) a
   generated `DirectInvoker` (direct `INVOKE*` or a classData
   `MethodHandle`) is installed so the invoke is a JIT-constant
   call. Generation cost is paid once per inflated method; the
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -284,15 +284,18 @@ above. Each bites contributors quickly if missed:
   reproduce Java/module access rules manually when a `Lookup` can
   perform the check. The cached member representation describes
   the member; the call site supplies the access context.
-- **JIT-constant invoke for MOP internals.** After a call target
-  is selected and stable, `CachedMethod.invoke` may install a
+- **JIT-constant invoke for MOP internals.** `CachedMethod.invoke`
+  is the remaining `Method.invoke` hole on the MOP (Groovy as
+  caller). Indy already uses a linked `CallSite` and must keep the
+  call-site `Lookup`; DGM / `CallSiteGenerator` already emit
+  `INVOKE*`. After
+  `groovy.cachedmethod.invoker.threshold` hits (default 100) a
   generated `DirectInvoker` (direct `INVOKE*` or a classData
-  `MethodHandle`) so the invoke is a JIT-constant call rather than
-  `Method.invoke`. This is the MOP "Groovy as caller" path; indy
-  continues to use the call-site `Lookup` and must not be fed the
-  trampoline. Kill switch:
-  `groovy.cachedmethod.invoker.disable`. Hit threshold:
-  `groovy.cachedmethod.invoker.threshold` (default 100). See
+  `MethodHandle`) is installed so the invoke is a JIT-constant
+  call. Generation cost is paid once per inflated method; the
+  performance baseline for the change is reflective
+  `CachedMethod.invoke`, not a Java direct call. Kill switch:
+  `groovy.cachedmethod.invoker.disable`. See
   `org.apache.groovy.internal.runtime.invoke`.
 
 ## Operator families
--- a/src/main/java/org/apache/groovy/internal/runtime/invoke/DirectInvoker.java
+++ b/src/main/java/org/apache/groovy/internal/runtime/invoke/DirectInvoker.java
@@ -1,0 +1,56 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.groovy.internal.runtime.invoke;
+
+import groovy.transform.Internal;
+
+/**
+ * JIT-constant trampoline for a single {@link java.lang.reflect.Method}.
+ *
+ * <p>Not user API. Not an invokedynamic target. Valid for the {@code Method}
+ * lifetime, not across {@code MetaMethod} wrappers.
+ *
+ * <p>Public because generated nestmates of a foreign host (other package,
+ * other loader) must {@code implements} this interface. The containing
+ * {@code internal} package is japicmp-excluded.
+ *
+ * @since 6.0.0
+ */
+@Internal
+@FunctionalInterface
+public interface DirectInvoker {
+
+    /**
+     * Invokes the bound method on {@code receiver} with {@code arguments}.
+     *
+     * <p>Arguments are assumed already coerced (see
+     * {@code MetaMethod.doMethodInvoke}) except when a
+     * {@code TransformMetaMethod} skipped coercion. Generated trampolines
+     * require a non-null {@code arguments} array (use
+     * {@code MetaClassHelper.EMPTY_ARRAY} for no args);
+     * {@code CachedMethod.invoke} performs that substitution. Primitive
+     * returns are boxed; the caller runs {@code normalizeBoxedReturn}.
+     *
+     * @param receiver  the receiver, ignored for static methods
+     * @param arguments the positional arguments (non-null; empty if none)
+     * @return the boxed result, or {@code null} for {@code void}
+     * @throws Throwable the target's thrown exception, unwrapped
+     */
+    Object invoke(Object receiver, Object[] arguments) throws Throwable;
+}
--- a/src/main/java/org/apache/groovy/internal/runtime/invoke/DirectInvoker.java
+++ b/src/main/java/org/apache/groovy/internal/runtime/invoke/DirectInvoker.java
@@ -50,7 +50,9 @@ public interface DirectInvoker {
      * @param receiver  the receiver, ignored for static methods
      * @param arguments the positional arguments (non-null; empty if none)
      * @return the boxed result, or {@code null} for {@code void}
-     * @throws Throwable the target's thrown exception, unwrapped
+     * @throws Throwable the target's thrown exception, unwrapped. A narrower
+     *                   type would wrap {@code Error} and checked exceptions
+     *                   the reflective path leaves as {@code ITE} causes.
      */
-    Object invoke(Object receiver, Object[] arguments) throws Throwable;
+    Object invoke(Object receiver, Object[] arguments) throws Throwable; // NOSONAR java:S112 — must match Method.invoke / ITE unwrapping
 }
--- a/src/main/java/org/apache/groovy/internal/runtime/invoke/InvokerBytecode.java
+++ b/src/main/java/org/apache/groovy/internal/runtime/invoke/InvokerBytecode.java
@@ -1,0 +1,232 @@
+/*
+ *  Licensed to the Apache Software Foundation (ASF) under one
+ *  or more contributor license agreements.  See the NOTICE file
+ *  distributed with this work for additional information
+ *  regarding copyright ownership.  The ASF licenses this file
+ *  to you under the Apache License, Version 2.0 (the
+ *  "License"); you may not use this file except in compliance
+ *  with the License.  You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing,
+ *  software distributed under the License is distributed on an
+ *  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ */
+package org.apache.groovy.internal.runtime.invoke;
+
+import org.codehaus.groovy.classgen.asm.BytecodeHelper;
+import org.codehaus.groovy.classgen.asm.util.TypeUtil;
+import org.codehaus.groovy.control.CompilerConfiguration;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.ConstantDynamic;
+import org.objectweb.asm.Handle;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
+import org.objectweb.asm.Type;
+
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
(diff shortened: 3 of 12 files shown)


blackdrag (changes requested):
So far I have nothing against the technique used here for the hidden class itself, I worry about the added complexity vs the actual gain a bit

The benchmarks show the steady-state speedup once DirectInvoker exists, but they do not measure the cost of creating it or  the break-even point. Since the optimization is triggered by a usage threshold on a generic invocation path (potentially different to a hot path), a cold/warm-up benchmark would help to justify the added machinery.

The results would be more meaningful if the normal CachedMethod.invoke() path were explicitly established as the baseline and the optimized results reported relative to it. The direct Java invocation is a useful lower bound, but it isn't the baseline that the change is actually optimizing.

The other point I have is that CachedMethod is changed for something that feels like a special case. What did motivate you to this implementation?

daniellansun:
# GROOVY-12325 / PR 2852 — performance report

Working note. Not a JIRA comment.

## Setup

| Item | Value |
|---|---|
| Baseline | `03e303e9fdb2715c1be064849984c19a14cab4b8` (`GROOVY-12306: compatibility note`) |
| Head | working tree on `GROOVY-12325` after review follow-ups |
| JDK | 25.0.4 (Amazon Corretto), 64-bit Server VM |
| JMH | 1.37 |
| Protocol | AverageTime, 1 fork, 3×1 s warmup, 5×1 s measurement, 1 thread |
| Command (head) | `./gradlew :performance:jmh -PbenchInclude=CachedMethodInvoker` |
| Command (baseline) | worktree at `03e303e` + `CachedMethodBaselineBench` (`-PbenchInclude=CachedMethodBaseline`) |

Operands are `@Param` fields so HotSpot cannot constant-fold `startsWith` to `true`.

**Baseline for this change is reflective `CachedMethod.invoke`**, not a Java direct call. Java is a lower bound only.

## Headline

Steady-state generated `CachedMethod.invoke` is **faster than `03e303e`**. The kill-switch / pre-generation path is **statistically the same as `03e303e`** after one hot-path fix (sticky skip when generation is disallowed). Hidden-class definition is ~80 µs **once per inflated method**; it does not pay off on a fresh `CachedMethod` that is discarded after a few hundred calls.

## Steady-state (the interned-MOP case)

`String.startsWith`, one `CachedMethod` reused across the whole fork.

| Path | Score (ns/op) | vs `03e303e` |
|---|---|---|
| Java `startsWith` (lower bound) | 7.34 ± 0.85 | 7.64 ± 1.35 on `03e303e` |
| **`03e303e` `CachedMethod.invoke`** | **18.04 ± 2.66** | — |
| HEAD, `disable=true` (reflective) | 17.85 ± 1.31 | ≈ baseline (overlapping CI) |
| HEAD, `threshold=0` (generated, warmed) | **10.09 ± 2.52** | **~1.8× faster** |

Megamorphic row (five distinct `CachedMethod`s, one `invoke` site, after sticky opt):

| Path | Score (ns/op, 5 invokes) |
|---|---|
| `mega_reflective` | 77.0 ± 11.9 |
| `mega_generated` | 57.5 ± 4.6 |

## Generation cost and break-even

Fresh `CachedMethod` per invocation (`@Setup(Level.Invocation)`):

| Path | Score |
|---|---|
| `generate_reflective_fresh` | 0.088 ± 0.012 µs |
| `generate_then_invoke` (`threshold=0`) | 76.6 ± 5.4 µs |

Almost all of the 77 µs is `defineHiddenClass` + instantiate. Saving vs reflective is ~8 ns/op once the trampoline exists (18 ns → 10 ns). Naïve break-even:

```
77e3 ns / (18 − 10) ns/op  ≈  10⁴ invokes after generation
```

That is **once per interned `CachedMethod` per VM**, not per call site and not per request. A method that stays on the MOP for the process lifetime (typical `MetaClass` entry) recoups quickly. A throwaway `CachedMethod` used 250 times does not — see bursts.

Default threshold 100 is **below** `groovy.indy.optimize.threshold` (1000), so cold indy can still sit on `doMethodInvoke` when the trampoline appears. Methods that never reach 101 hits never pay `defineHiddenClass`.

## Bursts (fresh `CachedMethod`, hit counter starts at 0)

| Burst | Default threshold (100) | `disable=true` |
|---|---|---|
| 100 calls (never inflates) | 2.73 ± 0.47 µs | 1.71 ± 0.23 µs |
| 250 calls (inflates at hit 101) | 104 ± 16 µs | 4.06 ± 0.47 µs |

The 250-call default row **includes one hidden-class define**. It is the cost blackdrag asked for, not the interned-MOP steady state. It is slower than staying reflective for that short lifetime — expected, and why the threshold exists.

## Iteration log

The user-required comparison is **head vs `03e303e`**, not generated vs reflective only.

| Round | Change | `startsWith` generated | `startsWith` disable=true | vs `03e303e` 18.0 ns |
|---|---|---|---|---|
| 0 | As reviewed (property re-read every invoke when disabled) | 9.98 ns | **30.6 ns** | disable path regresses |
| 1 | Sticky `invokerAttempted` when `generationAllowed()` is false | 9.70 ns | **18.7 ns** | disable path matches baseline |
| 2 | Cache threshold on the `CachedMethod` (no `Long.getLong` per lukewarm invoke) | 10.09 ns | **17.85 ns** | generated **1.8×**; disable **no regression** |

No third round: the success criterion (generated faster than `03e303e`, disable path not slower within error) is met.

## What we did *not* treat as a regression

- Burst-250 default vs burst-250 disable: measures **define cost on a throwaway method**, not process-lifetime interned `CachedMethod`.
- `startsWith_generated` vs Java: Java is a lower bound; 10 ns vs 7 ns is the remaining trampoline / boxing / `invokeinterface` tax.

## Repro

```bash
# head
./gradlew :performance:jmh -PbenchInclude=CachedMethodInvoker

# baseline (separate worktree)
git worktree add /tmp/groovy-baseline 03e303e9fdb2715c1be064849984c19a14cab4b8
# add CachedMethodBaselineBench.java (same shape as startsWith_reflective)
cd /tmp/groovy-baseline && ./gradlew :performance:jmh -PbenchInclude=CachedMethodBaseline
```

daniellansun:
@blackdrag Thanks for the review — the hidden-class encoding staying acceptable is a relief. The complexity-vs-gain concern is fair, and the benches were incomplete in the way you described. I have pushed follow-ups and a short measurement note (`jira-drafts/pr-2852-performance-report.md` locally; I can paste numbers here).

**Baseline.** `startsWith_java` is now labelled a lower bound only. The baseline is reflective `CachedMethod.invoke` (`-Dgroovy.cachedmethod.invoker.disable=true`). Operands are `@Param` fields so the JIT cannot fold `startsWith` to a constant.

Against `03e303e` on JDK 25 (JMH 1.37, 1 fork, 3×1 s / 5×1 s):

| Path | ns/op |
|---|---|
| `03e303e` `CachedMethod.invoke` | 18.0 ± 2.7 |
| HEAD, generation disabled | 17.8 ± 1.3 |
| HEAD, generated (warmed, `threshold=0`) | 10.1 ± 2.5 |
| Java `startsWith` | 7.3 ± 0.8 |

So the interned-MOP case is about **1.8×** the pre-change path, and the kill-switch path is not a regression within error.

**Create cost / break-even.** New rows use a fresh `CachedMethod` per invocation:

- `generate_reflective_fresh` ≈ 0.09 µs
- `generate_then_invoke` (`threshold=0`) ≈ 77 µs

Almost all of that is `defineHiddenClass`. At ~8 ns saved per later invoke, naïve break-even is on the order of 10⁴ calls **after** generation. That cost is paid **once per interned `CachedMethod` per VM**, which is the real MOP (`CachedClass` method table). A throwaway `CachedMethod` used a few hundred times does not recoup — the `burst_*` rows show that on purpose (250 calls with default threshold ≈ 104 µs vs 4 µs staying reflective, because the burst includes one define). Default threshold 100 is unchanged: below `groovy.indy.optimize.threshold` (1000), so methods that never look hot never define a class.

The first disable-path run was ~31 ns vs 18 ns on `03e303e` because `generationAllowed()` re-read system properties on every invoke. That is now sticky on the instance when generation is off (kill switch / native-image / Android). The threshold is cached on the instance so lukewarm invokes do not call `Long.getLong` every time.

**Why `CachedMethod`.** This is not a third invoke engine beside indy and `CallSiteGenerator`. Those already have a JIT-constant target (linked `CallSite` / emitted `INVOKE*`). `CachedMethod.invoke` is the leftover `Method.invoke` on the MOP “Groovy as caller” path (including cold indy still on `doMethodInvoke`). The trampoline is bound to that `Method` and must not be fed to indy (wrong `Lookup`). Putting the hook on `CachedMethod` is the smallest place that sees every such invoke without widening public API. I kept the fast path to a volatile read + generate-or-reflect; factory/ASM live under `org.apache.groovy.internal.runtime.invoke`.

Happy to drop the default threshold, or the whole hook behind `disable=true` by default, if you would rather ship the machinery dark until we have a longer-lived application bench.

blackdrag:
The ~10k break-even point makes me wonder about the default threshold. Given that groovy.indy.optimize.threshold is already 1000, did you consider using 1k here for example? 100 seems quite aggressive given the the generation cost.

Overall we have now improved benchmarks, but I wonder about numbers in real applications.

daniellansun:
@blackdrag You are right that 100 was aggressive next to a ~10k naïve break-even. I had kept it below the indy threshold so a trampoline could appear while cold indy was still on `doMethodInvoke`; that undervalued `defineHiddenClass`.

We did consider 1k. The default is now **1000** (`InvokerFactory.DEFAULT_THRESHOLD`, same order as `groovy.indy.optimize.threshold`). Generation still happens only on interned `CachedMethod`s that stay on the MOP that long; anything colder never pays the define.

On real applications: the numbers we have are still micros (`CachedMethod.invoke` in isolation, plus a short megamorphic row). They do not stand in for Grails, Gradle, or a long-lived Groovy service. I would not claim a wall-clock win there without a workload that keeps the same `MetaClass` method table hot. If you have a candidate app bench, I am happy to run `disable=true` against the default on it. The kill switch stays for anyone who would rather not generate.

blackdrag:
> On real applications: the numbers we have are still micros (`CachedMethod.invoke` in isolation, plus a short megamorphic row). They do not stand in for Grails, Gradle, or a long-lived Groovy service. I would not claim a wall-clock win there without a workload that keeps the same `MetaClass` method table hot. If you have a candidate app bench, I am happy to run `disable=true` against the default on it. The kill switch stays for anyone who would rather not generate.

@daniellansun the problem I have is simple, what kind of program would use this path so extensively? If we talk about a long running program the threshold is reached, but it does not mean we are talking about a hot path. I guess that means we have to look for a tight loop, and the call must be dynamic from the Java side to an existing method, instead of calling the method directly. That means paths that go maybe through GroovyObject.invoke, InvokerHelper, MetaClass or CachedMethod itself. In Groovy itself we use most of these somewhere in extension methods usually. 

Maybe DynamicDispatchBench, especially the ones using MethodInterceptor could answer the question better. But my expectation is that the surrounding machinery shadows the actual call cost.  
We could also create another benchmark around ObjectRange, basically use the for-loop to iterate through it and cause a lot of dynamic "next" calls from Java world.

Of course those are not real applications and just give indicators

Actually, the failure in this PR made me aware of something I forgot about:

[71](https://github.com/apache/groovy/actions/runs/33690247923/job/100447108661?pr=2852#step:9:672)
        at java.base/java.lang.reflect.Method.invoke(Method.java:569)
        at org.codehaus.groovy.reflection.CachedMethod.invokeReflective(CachedMethod.java:505)
        at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:455)
        at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:298)
        at org.codehaus.groovy.vmplugin.v8.IndyInterface.invokeColdReflective(IndyInterface.java:648)
        at org.apache.groovy.groovysh.jline.GroovyPosixCommandsTest.lsStripsControlCharactersFromSymlinkTargets(GroovyPosixCommandsTest.groovy:108)
        
which means there is a path from indy to this PR and it was not made for this. And that is good and bad. Good, because it means the change is less niche than I initially thought. But now indy and this PR compete in optimization. But it also means that all our indy benchmarks may be influenced by this.

blackdrag:
> Maybe DynamicDispatchBench, especially the ones using MethodInterceptor could answer the question better. But my expectation is that the surrounding machinery shadows the actual call cost. 

Which actually ran already on the PR.

Benchmark | Base | PR | Improvement
-- | -- | -- | --
invokeMethodInterception | 3.476 ms/op | 2.467 ms/op | 29.0%
invokeMethodAlternating | 11.490 ms/op | 10.560 ms/op | 8.1%

Base: `03e303e9fdb2715c1be064849984c19a14cab4b8`
PR: `e2767e19554985c2e03a0c002e23c614f944c790`

With that I am no longer going to block the issue.

blackdrag (approved):

PR #2864, opened 2026-09-02
GROOVY-12339: opt the remaining serializable closures into the cycle check

paulk-asert:
The 6.0.0 deserialization cycle check is opt-in per class, deliberately: Closure.checkForReferenceCycle is a static helper rather than a hook, because a hook would have to be protected to reach generated subclasses and would force every subclass declaring the idiomatic private readResolve to widen it. CurriedClosure, ComposedClosure, TrampolineClosure and MethodClosure opted in. Six serializable closures did not, and a forged owner cycle in any of them deserializes unchecked and recurses on first use.

WritableClosure needs only the readResolve: the closure it writes through is its owner, which the default walk already reaches. The memoize wrappers and the functional hybrids also override additionalReferences, because each dispatches through a field - closure, delegate - that the owner/delegate/thisObject walk does not see. Only Closure-valued links are followed, so a hybrid wrapping a plain lambda contributes nothing.

SoftReferenceMemoizeFunction declares its own readResolve rather than relying on its superclass, since a private readResolve is not inherited and a subclass without one is read unchecked. It cannot itself be serialized, holding a non-transient ReferenceQueue, but that does not put it out of reach: a gadget stream is authored rather than produced by serializing a live object, so the test clears those fields to write the stream an attacker would simply compose.

Closures.java is the notable case: it shipped in 6.0.0, the same release as the check, without opting in. Its three hybrids are new API rather than pre-existing classes the rollout missed.

### DIFF
--- a/src/test/groovy/groovy/lang/ClosureSerializationCycleTest.groovy
+++ b/src/test/groovy/groovy/lang/ClosureSerializationCycleTest.groovy
@@ -18,10 +18,14 @@
  */
 package groovy.lang
 
+import org.apache.groovy.util.Closures
 import org.codehaus.groovy.runtime.CurriedClosure
 import org.junit.jupiter.api.Test
 
 import java.io.InvalidObjectException
+import java.util.function.Consumer
+import java.util.function.Function
+import java.util.function.Predicate
 
 import static groovy.test.GroovyAssert.shouldFail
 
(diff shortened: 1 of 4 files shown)


src/test/groovy/groovy/lang/ClosureSerializationCycleTest.groovy
```diff
@@ -215,6 +269,56 @@ final class ClosureSerializationCycleTest {
...
+            // are cleared here to produce the stream an attacker would simply author.
+            def memoized = { x -> x * 2 }.memoizeAtLeast(4)
+            setDeclaredField(memoized, 'queue', null)
+            setDeclaredField(memoized, 'lruProtectionStorage', null)
+            setDeclaredField(memoized, 'closure', memoized)
+            serialize(memoized)
+        }
+
+        static byte[] serializePredicateClosure() {
+            serialize(org.apache.groovy.util.Closures.from({ n -> n % 2 == 0 } as java.util.function.Predicate))
```
blackdrag: the FQNs again

paulk-asert: fixed

PR #2868, opened 2026-09-03
GROOVY-12344: TemplateServlet: hold the template cache in a ConcurrentHashMap with soft refs

paulk-asert:
The cache was a `WeakHashMap` keyed on `file.getAbsolutePath()` or `url.toString()`. Both are freshly built strings local to `getTemplate`, and `TemplateCacheEntry` retains only `lastModified`, `length`, a `Date` and the `Template`, never the `File`, so nothing outside the map referred to a key once its request returned. Weakly held keys made every entry collectible as soon as it was stored, and the cache emptied at each collection however hot a template was.

The map was also read and written from `service()` without synchronization. `WeakHashMap` did not receive the JDK 8 rewrite that ended the `HashMap` resize infinite-loop, so its `transfer()` still head-inserts and concurrent resizes can build a cycle that leaves `get()` spinning.

A `ConcurrentHashMap` addresses both, and is faster besides: `WeakHashMap.get` polls a `ReferenceQueue` on every read and dereferences a `WeakReference` per probe. Staleness handling is unchanged, as `validate()` still compares `lastModified` and `length`.

Entries are held through a `SoftReference`, which keeps the memory sensitivity the original was reaching for and never delivered. A soft reference is reclaimed under memory pressure but not by an ordinary collection, so the cache both caches and gives memory back. Measured on compiled templates, which is what it holds, under a 192MB heap and metaspace: holding them strongly raised `OutOfMemoryError` after 4143 templates, while holding them softly stored 20000, keeping 163 and shedding the rest.

A weak reference would not do: it is cleared by any collection, so nothing would survive between requests and the defect would return from the other side. Across an ordinary collection with no memory pressure, soft keeps five of five templates and weak keeps none.

A lookup that finds a reference the collector has emptied removes it, so keys do not accumulate pointing at nothing. The cache field is private and of declared type `Map` and `TemplateCacheEntry` is private, so the change is source and binary compatible.

`getTemplate` describes what a caller is promised rather than what holds the templates: that one already compiled is reused, that the file form checks length and timestamp while the URL form has nothing to check against, and that an instance is neither guaranteed to be the same one twice nor private to the request. Which map, and how it holds its values, is said only on the field that does it.

The regression test counts compilations through a `TemplateEngine` that wraps the real one, so it states the property the cache exists for — two requests either side of a collection compile the template once — rather than asserting an identity the documented contract disclaims. It was checked against both ways of getting this wrong: reverting the map to a `WeakHashMap`, and holding entries through a `WeakReference`, each make the count two.

### DIFF
--- a/subprojects/groovy-servlet/src/main/java/groovy/servlet/TemplateServlet.java
+++ b/subprojects/groovy-servlet/src/main/java/groovy/servlet/TemplateServlet.java
@@ -32,6 +32,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
 import java.io.Writer;
+import java.lang.ref.SoftReference;
 import java.lang.reflect.InvocationTargetException;
 import java.net.URL;
 import java.nio.file.Files;
@@ -165,9 +166,11 @@ public class TemplateServlet extends AbstractHttpServlet {
     }
 
     /**
-     * Simple file name to template cache map.
+     * File name to template cache, safe for concurrent use. Entries are held softly, so a
+     * template survives ordinary collection but can be reclaimed under memory pressure and
+     * compiled again on the next request.
      */
-    private final Map<String, TemplateCacheEntry> cache;
+    private final Map<String, SoftReference<TemplateCacheEntry>> cache;
 
     /**
      * Underlying template engine used to evaluate template source files.
@@ -187,7 +190,7 @@ public class TemplateServlet extends AbstractHttpServlet {
      * Create new TemplateServlet.
      */
     public TemplateServlet() {
-        this.cache = new ConcurrentHashMap<String, TemplateCacheEntry>();
+        this.cache = new ConcurrentHashMap<String, SoftReference<TemplateCacheEntry>>();
         this.engine = null; // assigned later by init()
         this.generateBy = true; // may be changed by init()
         this.fileEncodingParamVal = null; // may be changed by init()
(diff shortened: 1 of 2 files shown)


subprojects/groovy-servlet/src/main/java/groovy/servlet/TemplateServlet.java
```diff
@@ -187,7 +187,7 @@ public String toString() {
      * Create new TemplateServlet.
      */
     public TemplateServlet() {
-        this.cache = new WeakHashMap<String, TemplateCacheEntry>();
+        this.cache = new ConcurrentHashMap<String, TemplateCacheEntry>();
```
blackdrag: I agree that the concurrency issue is fixed very well with ConcurrentHashMap. But even if done wrong, the idea before was a memory sensitive cache. Your solution is removing that idea. Why not a Weak/SoftReference for the value? or a LRUCache or ConcurrentLinkedHashMap

paulk-asert: In the PDF, I sent out-of-band, you will see that the existing implementation never fulfilled the memory sensitive idea, but I'll do some further work.

paulk-asert: Updated to use `SoftReference`s. Also considered Weak and ConcurrentSoftCache but the soft references seemed best.

PR #2878, opened 2026-09-05
GROOVY-12355: STC: infer division and group arithmetic from the runti…

paulk-asert:
…me's number categories

The runtime divides with floating-point math when either operand is a Float or Double, wrapper or primitive, and with BigDecimal math for integral, BigInteger and BigDecimal operands. The type checker only recognised primitive float and double, so Double / Integer was inferred as BigDecimal, and an operand typed as Number, whose runtime category is unknown, was inferred as BigDecimal too. Statically compiled code then cast the Double the runtime produced and failed.

Unwrap before testing the floating category, and infer BigDecimal only when both operands belong to a category that divides that way; a Number operand yields Number. The group operations had the same blind spot from the other side: they took the category of whichever operand they recognised, so Number * int was int. An operand of unknown category now yields Number unless a floating-point partner decides the result, as it does at runtime.

An arithmetic result that could not be typed before may now be a Number where it was an int or BigDecimal, so code that assigned it to a narrower type needs an explicit conversion, as the GROOVY-5539 test now shows. StringUtil.bar with Double arguments works without change.

### DIFF
--- a/src/test/groovy/groovy/transform/stc/TypeInferenceSTCTest.groovy
+++ b/src/test/groovy/groovy/transform/stc/TypeInferenceSTCTest.groovy
@@ -1835,6 +1835,92 @@ class TypeInferenceSTCTest extends StaticTypeCheckingTestCase {
         '''
     }
 
+    // GROOVY-12355
+    @Test
+    void testDivisionInference() {
+        // the runtime divides with floating-point math when either operand is a
+        // Float or Double, and with BigDecimal math for the other known categories;
+        // an operand typed as Number may be either, so the result is only a Number
+        [['Number',     'Number',     'Number'],
+         ['Number',     'int',        'Number'],
+         ['Integer',    'Number',     'Number'],
+         ['Double',     'Integer',    'Double'],
+         ['Integer',    'Double',     'Double'],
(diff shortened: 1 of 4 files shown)


src/test/groovy/groovy/transform/stc/TypeInferenceSTCTest.groovy
```diff
@@ -1835,6 +1835,92 @@ class TypeInferenceSTCTest extends StaticTypeCheckingTestCase {
...
+         ['Float',      'Float',      'Double'],
+         ['double',     'Integer',    'Double'],
+         ['double',     'Number',     'Double'],
+         ['double',     'int',        'double'],
+         ['float',      'float',      'double'],
+         ['Integer',    'Integer',    'BigDecimal'],
+         ['int',        'int',        'BigDecimal'],
+         ['Long',       'BigInteger', 'BigDecimal'],
+         ['BigDecimal', 'Integer',    'BigDecimal'],
+         ['BigInteger', 'BigDecimal', 'BigDecimal']
```
blackdrag: you should add double/Double/BigDecimal combinations as well

blackdrag (approved):

PR #2880, opened 2026-09-05
GROOVY-12359: deleteDir: treat junctions and other special nodes as l…

paulk-asert:
…eaves

Since GROOVY-12125 a symbolic link inside a tree being deleted is removed as the link itself and never entered. The guard was Files.isSymbolicLink, which does not report a Windows directory junction, so a junction was still traversed and the contents of its target deleted. A junction is the link form that matters most there: mklink /J needs no privilege, where a Windows symbolic link needs one most users do not hold, and it is planted in advance rather than raced.

Both implementations, File.deleteDir and the groovy-nio Path.deleteDir, now read each node's attributes without following links and treat anything that is neither a regular file nor a directory as a leaf: a symbolic link is removed as before, and a junction or other reparse point, a pipe, a socket or a device is removed as the node it is, never entered or opened. Handed such a node directly, deleteDir returns false as it does for any non-directory, and does not touch what the node points at.

On POSIX the visible behaviour is unchanged — a pipe inside a tree was deleted before and still is, now through the leaf branch, which is covered by new mkfifo tests that run on the platforms the build runs on. The junction behaviour cannot be exercised on those platforms, so the Windows-gated tests are the verification rather than a formality: they create a junction with mklink /J, which succeeds unprivileged on the CI runners, delete the enclosing tree, and assert the target's contents survive. Until they have run on Windows CI, the junction half of this change is implemented to the documented attribute behaviour, not demonstrated.

### DIFF
--- a/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
+++ b/src/main/java/org/codehaus/groovy/runtime/ResourceGroovyMethods.java
@@ -64,6 +64,9 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.nio.charset.Charset;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.attribute.BasicFileAttributes;
+import java.nio.file.LinkOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


daniellansun:
Merged. Thanks

PR #2883, opened 2026-09-06
GROOVY-12354: Indy: reflective cold tier off by default on a JVM, on …

paulk-asert:
…for AOT-linked sites

While a call site is on the reflective cold tier, reflection and Groovy runtime frames sit between caller and target, so logging frameworks, StackWalker and stack traces report the wrong caller until the site promotes, or forever for rarely hit sites. Groovy 5 reported the caller correctly from the first call, so the on-by-default tier was a regression.

An explicit groovy.indy.cold.reflection setting still wins. When unset, only AOT-linked sites (GraalVM native image, or groovy.indy.aot.link) use the tier, where it is the steady state. The JMH A/B forks now enable the tier rather than disable it, and the invokedynamic docs describe the property, its defaults, and the logger-side workarounds for JVM and native-image users.


daniellansun:
Merged. Thanks.

PR #2887, opened 2026-09-06
GROOVY-12363: SC: compile anonymous inner classes of a @CompileStatic…

paulk-asert:
… method statically

An anonymous inner class declared in a method-level @CompileStatic method was marked as statically compiled only after the type checker had visited its body, so the body compiled dynamically while the class was generated with the static-compilation MOP writer, which emits no super$ bridges: any super.m() call failed with MissingMethodException. Mark the class from its enclosing method before the body is visited, and have the method-level transformation run StaticCompilationTransformer over such classes as the class-level path does, since operator expressions left unrewritten fail at class generation.


daniellansun:
Merged. Thanks.

PR #2888, opened 2026-09-06
GROOVY-12364: Indy: dispatch caller-sensitive targets reflectively on…

paulk-asert:
… AOT-linked sites

Lookup.unreflect returns a caller-bound handle for a @CallerSensitive method, which is exactly right on HotSpot but cannot be executed by GraalVM's MethodHandle interpreter: a native image aborts with the VM-fatal "Cannot invoke method that has a @CallerSensitiveAdapter without an explicit caller" as soon as dynamic code calls e.g. Logger.getLogger or MethodHandles.lookup. Reflection handles such methods fine there, so an AOT-linked site now builds the target handle around Method.invoke instead, shaped like the unreflected handle so the guarded chain is unchanged. The observed caller then becomes Groovy's runtime rather than the calling class, which is harmless for logger and bundle lookups in a single-module image; MethodHandles.lookup() should be static-compiled in native code.


daniellansun:
Merged. Thanks.

PR #2890, opened 2026-09-06
GROOVY-12360: GroovyDocWriter: keep generated pages inside the destin…

paulk-asert:
…ation directory

A class page was written to destdir + "/" + fullPathName + ".html", and a package page under destdir + "/" + packageDoc.name(), with the only guard being that the name is not absolute. The name is a package path, and it picks up a ".." segment when a source file is named through a source path with one — so a page could be written above the destination directory:

    fullPathName = ../srcroot/sub/Pwned
    -> <destdir>/../srcroot/sub/Pwned.html   written outside destdir

The absolute-only check is replaced by one that normalises the resolved path and confirms it stays within the destination directory, at both the class-page and package-page sites. A name that would escape is skipped with a warning rather than followed. The check is on the normalised paths, so it does not rely on names having been sanitised earlier, and a legitimate package like a/b/c still resolves within destdir and is unaffected.

The regression test builds a doc whose path carries "..", renders through a mock output tool, and asserts that nothing it wrote resolves outside the destination. It was confirmed to fail without the guard, writing to <destdir>/../.../Pwned.html.

MockOutputTool gains a getOutputs() accessor so the test can inspect every written path; it previously exposed only single-file lookup.

### DIFF
--- a/subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/GroovyDocWriter.java
+++ b/subprojects/groovy-groovydoc/src/main/java/org/codehaus/groovy/tools/groovydoc/GroovyDocWriter.java
@@ -92,10 +92,31 @@ public class GroovyDocWriter {
     /**
      * Writes the HTML page for a single class doc if its visibility matches the configured scope.
      */
+    /**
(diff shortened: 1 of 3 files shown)


daniellansun:
Merged. Thanks.

PR #2892, opened 2026-09-06
GROOVY-12368: groovy.ast=xml: give the AST dump the source file's per…

paulk-asert:
…missions

The groovy.ast=xml debug option writes a serialized AST to <source>.xml beside the source, through a FileWriter that creates the file at the process umask. A source the developer had kept to themselves produced a world-readable dump of its AST:

    -rw-------  Script.groovy
    -rw-r--r--  Script.groovy.xml

The dump is a serialization of that source's AST, so it is no less sensitive than the source. It is now created with the source file's own permissions, so it exposes no more than the source already does; a world-readable source still yields a world-readable dump, unchanged. Where there is no source file to match — a source compiled from a string — the dump defaults to owner-only rather than the umask; where the filesystem has no POSIX permissions, creation is left to the writer as before.

The file is created with the permissions rather than created and then adjusted, following the pattern used for in-place file rewrites, so its contents are never briefly readable by anyone those permissions exclude.

XStream is an optional feature, so a testRuntimeOnly dependency is added to let XStreamUtils be exercised; the test asserts the dump inherits a private source's permissions and was confirmed to fail without the change.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -93,6 +93,11 @@ afterEvaluate {
 dependencies {
     // Classic call-site runtime for tests that compile with indy=false (GROOVY-11158)
     testRuntimeOnly projects.groovyCallsite
+    // the groovy.ast=xml AST dump uses XStream, an optional feature; supply it to the tests so
+    // XStreamUtils can be exercised (its permission handling in particular) rather than skipped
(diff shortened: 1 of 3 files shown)


daniellansun:
Merged. Thanks

PR #2894, opened 2026-09-06
GROOVY-12370: GroovyDocWriter: do not copy a symbolic link out of a resource directory

paulk-asert:
A package's doc-files/ and snippet-files/ directories are mirrored into the output verbatim. The mirror walked the tree and copied each entry with Files.copy, which follows a symbolic link and copies the content of its target. A link pointing outside the source tree therefore placed that outside content into the published documentation:

    doc-files/leak.txt -> ../../../secret/passwd
    published as doc-files/leak.txt containing the secret

A link, or other reparse point such as a Windows junction, is now skipped with a warning rather than followed, so only the files actually inside the resource directory are published. An ordinary file beside the link is copied as before.

The regression test puts a symbolic link to a file outside the tree into a package's doc-files/, renders, and asserts the link was not copied while a real asset in the same directory was; it was confirmed to copy the link's target content without the guard.


daniellansun:
Merged. Thanks.

PR #2895, opened 2026-09-06
GROOVY-12371: bound the Sql statement cache and make it thread-safe

paulk-asert:
The prepared-statement cache was an unbounded HashMap keyed on the SQL text, mutated from getAbstractStatement without synchronization. Two problems followed. It grew without limit whenever the text varied rather than the parameters — an inList expands to a different placeholder count per list size, so a caller sizing the list mints a distinct cached statement each time, held open until the Sql is closed. And a shared Sql using the cache from more than one thread could corrupt the map.

The cache is now a synchronized, access-ordered LinkedHashMap that evicts the least recently used statement past a cap, closing it as it goes so a bounded cache does not leak the cursor it drops. The cap is statementCacheSize, default 256, settable and overridable with the groovy.sql.statement.cache.size system property; 0 or less keeps the old unbounded behaviour for anyone who relied on it.

Statement creation stays outside the lock — only the get-then-put is synchronized — so preparing a statement does not serialize other callers, and a create race keeps the first result and closes the loser rather than leaking it.

One consequence to note: with a shared Sql caching across threads, a bound below the working set can now evict and close a statement another thread is about to reuse. A single Sql is typically used serially or per connection; where it is shared, size the cache above the working set (or leave it unbounded). Default 256 is generous enough that ordinary use never evicts.

### DIFF
--- a/subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
+++ b/subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
@@ -44,6 +44,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


subprojects/groovy-sql/src/main/java/groovy/sql/Sql.java
```diff
@@ -5057,8 +5111,18 @@ private Statement getAbstractStatement(AbstractStatementCommand cmd, Connection
         if (cacheStatements) {
             stmt = statementCache.get(sql);
             if (stmt == null) {
-                stmt = cmd.execute(connection, sql);
-                statementCache.put(sql, stmt);
+                Statement created = cmd.execute(connection, sql);
+                synchronized (statementCache) {
+                    Statement raced = statementCache.get(sql);
```
daniellansun, line 5116: Why not get and check the `raced` outside of `syncronized` block first and use double-checked locking to avoid the statement with the same SQL is created repeatedly.

paulk-asert: I think either approach is fine in terms of pros/cons.

> DCL moves cmd.execute inside the lock. That JDBC prepare is a DB round-trip, and Collections.synchronizedMap uses one mutex for the whole map — so while one thread prepares a statement under the lock, every other thread's cache lookup and every other prepare — even of completely different SQL — blocks on that same mutex. DCL serializes all statement preparation across a shared Sql.
> 
> My version deliberately never holds the mutex across the DB call, so unrelated prepares proceed in parallel. The cost is exactly what Daniel spotted: a rare double-create.
> 
> So it's a genuine trade:
> - Mine: parallel prepares; a rare wasted prepare when two threads hit the same brand-new SQL simultaneously.
> - His DCL: never double-creates; but serializes all preparation behind one lock.

daniellansun: Understood

daniellansun:
LGTM

PR #2899, opened 2026-09-07
GROOVY-12376: LookAndFeelHelper: check the Metal theme type before in…

paulk-asert:
…stantiating

The Metal look-and-feel 'theme' attribute treated an unrecognized string as a class name and did Class.forName(name).getConstructor().newInstance() -- so it resolved, initialized and instantiated an arbitrary class named by the theme value before finding out whether it was even a MetalTheme (the cast failed only afterwards). It also resolved via this class's loader, so an application- or @Grab-supplied theme class was not found.

Resolve without initializing via the thread context classloader (falling back to this class's loader when none is set), confirm the class is a MetalTheme before constructing it, and throw IllegalArgumentException otherwise. A non-theme class is now never initialized or constructed, and custom or @Grab-supplied theme classes resolve.

### DIFF
--- a/subprojects/groovy-swing/src/main/groovy/groovy/swing/LookAndFeelHelper.groovy
+++ b/subprojects/groovy-swing/src/main/groovy/groovy/swing/LookAndFeelHelper.groovy
@@ -18,6 +18,7 @@
  */
 package groovy.swing
 
+import java.lang.reflect.Modifier
 import javax.swing.*
 import javax.swing.plaf.metal.DefaultMetalTheme
 import javax.swing.plaf.metal.MetalLookAndFeel
@@ -94,7 +95,23 @@ class LookAndFeelHelper {
                     } else if (theme == 'steel') {
                         theme = new DefaultMetalTheme();
                     } else {
-                        theme = Class.forName(theme as String).getConstructor().newInstance()
+                        // resolve without initializing, confirm it is a MetalTheme before
+                        // constructing it, and prefer the context classloader so application
+                        // and @Grab-supplied theme classes resolve; the context loader may be
+                        // null, so fall back to this class's loader (the pre-existing behaviour)
+                        def loader = Thread.currentThread().contextClassLoader ?: LookAndFeelHelper.classLoader
+                        def themeClass = Class.forName(theme as String, false, loader)
+                        if (!MetalTheme.isAssignableFrom(themeClass)) {
+                            throw new IllegalArgumentException("Metal theme class ${themeClass.name} is not a ${MetalTheme.name}")
+                        }
+                        if (Modifier.isAbstract(themeClass.modifiers)) {
+                            throw new IllegalArgumentException("Metal theme class ${themeClass.name} is abstract and cannot be instantiated")
+                        }
+                        try {
+                            theme = themeClass.getConstructor().newInstance()
+                        } catch (NoSuchMethodException e) {
+                            throw new IllegalArgumentException("Metal theme class ${themeClass.name} must provide a public no-argument constructor", e)
+                        }
                     }
                 };
                 MetalLookAndFeel.currentTheme = theme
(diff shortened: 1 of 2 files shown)


subprojects/groovy-swing/src/main/groovy/groovy/swing/LookAndFeelHelper.groovy
```diff
@@ -94,7 +94,16 @@ class LookAndFeelHelper {
...
                     } else {
-                        theme = Class.forName(theme as String).getConstructor().newInstance()
+                        // resolve without initializing, confirm it is a MetalTheme before
+                        // constructing it, and prefer the context classloader so application
+                        // and @Grab-supplied theme classes resolve; the context loader may be
+                        // null, so fall back to this class's loader (the pre-existing behaviour)
+                        def loader = Thread.currentThread().contextClassLoader ?: LookAndFeelHelper.classLoader
+                        def themeClass = Class.forName(theme as String, false, loader)
+                        if (!MetalTheme.isAssignableFrom(themeClass)) {
+                            throw new IllegalArgumentException("Metal theme class ${themeClass.name} is not a ${MetalTheme.name}")
```
utafrali, line 105: If `themeClass` is an abstract `MetalTheme` subtype, `isAssignableFrom` passes but `getConstructor().newInstance()` will throw `InstantiationException` with a confusing message. The fix already provides a clear `IllegalArgumentException` for wrong types; it should do the same for abstract classes. Consider adding before the `getConstructor()` call:
```groovy
if (java.lang.reflect.Modifier.isAbstract(themeClass.modifiers)) {
    throw new IllegalArgumentException("Metal theme class ${themeClass.name} is abstract and cannot be instantiated")
}
```

subprojects/groovy-swing/src/test/groovy/groovy/swing/LookAndFeelHelperThemeTest.groovy
```diff
@@ -0,0 +1,65 @@
...
+            handler(null, 'groovy.swing.NonMetalThemeProbe')
+        }
+        assert !ThemeProbeFlag.constructed: 'a non-MetalTheme class must not be constructed'
+    }
+
+    @Test
+    void acceptsMetalThemeSubclass() {
+        def handler = metalThemeHandler()
+        handler(null, 'groovy.swing.CustomMetalThemeProbe')
+        assert MetalLookAndFeel.currentTheme instanceof CustomMetalThemeProbe
```
utafrali: `handler(null, ...)` sets `MetalLookAndFeel.currentTheme` as a side effect and the test leaves it modified after completion. If other tests in the suite read `currentTheme`, this could introduce ordering-dependent failures. Consider saving and restoring:
```groovy
def previousTheme = MetalLookAndFeel.currentTheme
try {
    handler(null, 'groovy.swing.CustomMetalThemeProbe')
    assert MetalLookAndFeel.currentTheme instanceof CustomMetalThemeProbe
} finally {
    MetalLookAndFeel.currentTheme = previousTheme
}
```

utafrali (approved):
A well-targeted fix that correctly addresses both the security concern (arbitrary class initialization and construction via an unvalidated string) and the classloader resolution gap for @Grab-supplied themes. The approach is sound, the comments explain the reasoning clearly, and the new tests cover the key behaviors. Two minor points: abstract MetalTheme subclasses will produce a less descriptive InstantiationException instead of the clear IllegalArgumentException thrown for non-theme types, and the acceptance test leaves global MetalLookAndFeel state modified after it completes.

subprojects/groovy-swing/src/main/groovy/groovy/swing/LookAndFeelHelper.groovy
paulk-asert: Nice suggestion. Adopted plus I caught no-arg constructor as well. It would have a similar confusing message.

subprojects/groovy-swing/src/test/groovy/groovy/swing/LookAndFeelHelperThemeTest.groovy
paulk-asert: Good thinking. I did the same using @Before/AfterEach.

PR #2902, opened 2026-09-09
GROOVY-12379: native probe: cover groovy-nio, groovy-xml and groovy-json

paulk-asert:
Put the three module jars on the probe classpath, union their shipped metadata in the drift check, and exercise Path extensions, the XML parsers and builder, and JSON output and parsing. The strict-mode image found three gaps: groovy-xml's FactorySupport needs its JAXP ServiceLoader lookups and default implementations registered (generator gains a resources section), groovy-json needs its ServiceLoader-loaded string service (hand-written module metadata), and the Introspector's by-name resolution of Path's parameter types needs a hand-listed entry the JVM agent cannot record, now preserved across updateProbeMetadata.


paulk-asert:
Merged as part of #2903

PR #2915, opened 2026-09-09
GROOVY-12392: NativeImageMetadataGenerator references deprecated-for-removal Java9

daniellansun:
https://issues.apache.org/jira/browse/GROOVY-12392


paulk-asert (approved):
It needs a lintGate tweak but I'll do that after merge.

PR #2932, opened 2026-09-13
GROOVY-12408: a switch that is the last statement of a body is a stat…

paulk-asert:
…ement

GROOVY-12399 restored Java's rule that position, not arm shape, decides whether a switch produces a value, but carved out a third case: a switch in implicit-return position stayed an expression that yielded null when unmatched, so that a method whose body is a bare switch kept its 5.x value. That carve-out cost more than it bought.

It made one piece of syntax mean two things. Under @CompileStatic a `case label` with a declared-Object label resolved statically in that position and dynamically once any statement followed the switch, so adding a trailing line changed how labels matched. It also required about 110 lines in AstBuilder to recognise implicit-return position, duplicating what ReturnAdder already knows, on a different tree.

The carve-out existed because a SwitchStatement was a second-class citizen in static compilation, not because the semantics needed it. GROOVY-12405 gave statements the same jump-table dispatch and GROOVY-12406 the same duplicate-label diagnostic, so nothing is left to protect. A switch whose value is not used is now a statement wherever it appears, and Groovy's implicit return does to it what it does to any other statement, via ReturnAdder.

Three behaviours move, each toward consistency:

- A primitive-returning method whose bare switch matches nothing now yields the type's default, as `int f(int i) { if (i == 1) return 1 }` and the colon form always have. The arrow form was the only shape in the language that threw there, GroovyCastException dynamically and NullPointerException when statically compiled.
- `case label` in that position matches as it does when the value is discarded, and as it did in 5.x.
- The node is a SwitchStatement, which is visible to AST transforms. COMPATIBILITY.md already warns that audience about GROOVY-12255.

Deleted: the UNMATCHED_YIELDS_NULL marker and its three readers, and isImplicitReturnPosition, returnsValue, isLastStatement and isLenientPosition from AstBuilder. isSwitchValueUsed keeps the recursion added for nested arrow arms, since an arm of a switch that is itself a statement still produces nothing.


paulk-asert:
See the Jira issue for the complete description.

